PR #288, opened 2020-05-22
Implement GCPressure

### LINKED ISSUE #282: Implement something like Windows.Foundation.Metadata.GCPressure
See related dotnet issue for runtime support in net5:
https://github.com/dotnet/runtime/issues/36762

Once the runtime has settled on a new name and functionality is in master, cswinrt can provide a custom projection to the new type(s), possibly maintaining the old as well for netstandard2 compat.

/cc @MikeHillberg

Scottj1s:
fixes #282


Scottj1s:
@AaronRobinsonMSFT - for some reason, adding a finalizer to projected classes is causing random unit test failures, only with net5 (not netstandard2).  Tests are failing on object identity checks, and my spidey sense says its in the runtime.  Discovered it with this PR, but the repro is just to add an empty finalizer to a projected class.  Any ideas?

AaronRobinsonMSFT:
@Scottj1s Nothing obvious jumps out. The finalizer will extend the lifetime of the object slightly since it must be placed in the finalizer queue instead of immediately collected. Is there some way to give me an assembly I can run that will exhibit the behavior? Can you point at the object identity failure test? I don't see it here.

jkoritzinsky:
The object identity tests currently have just enough GC calls to pass. With the addition of a finalizer, we may need to add an extra call to GC.Collect after WaitForPendingFinalizers so we can deterministically collect the runtime class objects.

Scottj1s:
@jkoritzinsky - thanks, but the failing tests are TestClassGeneric and TestInterfaceGeneric (and once, TestStrings), with no explicit GC calls (nor should there be any).  Aaron fiddled with a test last week, which seems to suggest that the presence of a finalizer is causing the lifespans of projected class and nested object reference to diverge (is that possible?)

```
        // rewritten ...
        [Fact]
        public void TestClassGeneric()
        {
            var objs = TestObject.GetClassVector();
            Assert.Equal(3, objs.Count);
            bool a, b, c;
            {
                var obj = objs[0];
                a = object.Equals(obj, TestObject);
                Assert.Equal(TestObject.ThisPtr, objs[0].ThisPtr);
            }
            {
                var obj = objs[1];
                b = object.Equals(obj, TestObject);
                Assert.Equal(TestObject.ThisPtr, objs[1].ThisPtr);
            }
            {
                var obj = objs[2];
                c = object.Equals(obj, TestObject);
                Assert.Equal(TestObject.ThisPtr, objs[2].ThisPtr);
            }
            Console.WriteLine($"{a} {b} {c}");
            Assert.True(a && b && c);
        }  
```

  Message: 
    System.ObjectDisposedException : Cannot access a disposed object.
    Object name: 'ObjectReference'.
  Stack Trace: 
    IObjectReference.ThrowIfDisposed() line 97
    IObjectReference.get_ThisPtr() line 23
    IClass.get_ThisPtr() line 5371
    Class.get_ThisPtr() line 46
    Class.op_Equality(Class x, Class y) line 207
    Class.Equals(Object obj) line 210
    Object.Equals(Object objA, Object objB)
    TestComp.TestClassGeneric() line 754

jkoritzinsky:
Yes I believe that's possible. Here's what I think the series of events is that's causing the problem:

- TestObject.ThisPtr called
- GC triggers
- TestObject seen as no longer referenced, gets GCd
- ObjectReference moved to finalizer queue
- TestObject moved to finalizer queue <--- this step is new
- ObjectReference gets finalized
- objs[i] accessed, returns TestObject since the object is still technically alive (it's been GC'd but not finalized yet).
- ThisPtr throws because the ObjectReference has been disposed
- TestObject finalized (if we hadn't thrown)

In the current code, once TestObject is GCd, it's no longer alive so a new wrapper would be created instead of using the partial-state of the wrapper.

jkoritzinsky:
I think the merge for this PR got messed up. It looks like it's now including all of the commits on master since this branch was created (and is now +17k/-5k lines)

Scottj1s:
> I think the merge for this PR got messed up. It looks like it's now including all of the commits on master since this branch was created (and is now +17k/-5k lines)

hmm, so much for my rebase - thx

Scottj1s:
rebase confused github - [recreated the PR](https://github.com/microsoft/CsWinRT/pull/446)

PR #291, opened 2020-05-28
Add RID-specific runtime folder load fallback

### LINKED ISSUE #266: Type activation requires assemblies to be next to the exe
This breaks users who use a `DllImport` to load a WinUI assembly, as this will cause them to be loaded from the Runtime Identifier specific directory. What this means is that we'll have two different versions of the WinUI assemblies loaded in the process.

We should fix type activation so that it can look in the RID-specific folders.

This also breaks Runtime-Independent apps (AnyCPU) that need to bundle all their native dependencies with them. While this scenario may be considered out-of-scope for WinUI3, the fact that `DllImport` causes different assemblies to be loaded seems like an issue we need to address by WinUI3 RTM.

I think the current behavior is ok as-is for Preview1.

Scottj1s:
fixes #266

/cc: @stevenbrix


stevenbrix:
Thanks Scott! Could this just replace the call to `Platform.LoadLibraryEx`?

Scottj1s:
> Thanks Scott! Could this just replace the call to `Platform.LoadLibraryEx`?

NativeLibrary is only available on net5, so still need LoadLibrary at least for netstandard2, and it doesn't hurt as a fallback for net5

Issue #280, opened 2020-05-18
Evaluate pros and cons of switching geometry types from double to float


Scottj1s:
Having switched Point, Rect, Size from double to float (see #198), we can assess cost-benefit.

dotMorten:
I'm seeing 100s of build errors as well because of this change. Rect and Point is heavily used for performing layout, but almost everything else is done in doubles, so you're forced to making the casts almost everywhere in your code. IMHO if this change really is needed, it needs to be everywhere, including Width/Height/ActualWidth/ActualHeight/Thickness/GridLength etc etc.
And I'd have to if-def all those cast just for WinUI

michael-hawker:
Yup, changing this now seems like it'll cause a lot of friction for those migrating their existing WUX projects, and part of the selling point is it's supposed to be an easy transition. 

As we've been moving forward with the Toolkit it's been looking more and more like just the namespace changes to update (and the re-target to .NET 5), but this was definitely a  step backwards.

sylveon:
Width/Height and friends are doubles in the Windows Runtime types, not just the projections, changing this would affect all consumers, no matter the language. Code using the Python, C++ and Rust projections would all have to be updated as well.

michael-hawker:
Tagging the right Toolkit Alex as well @azchohfi. FYI @Scottj1s

yoshiask:
@sylveon Why would this affect other languages? I thought this was an issue with the C# projection?

sylveon:
IFrameworkElement uses doubles. It's only Point/Size/Rect that uses floats but where incorrectly projected as doubles.

@dotMorten's suggestion would require changing IFrameworkElement to use floats, which would break every consumer. (I guess that adjusting the projection of those to use floats as well could work, rather than changing the API surface)
![image](https://user-images.githubusercontent.com/6440374/82492057-9520ce00-9ab3-11ea-8866-a1fa31e715d6.png)

hshristov:
I hit the same issue. Migrating library to WinUI Win32 led to 198 errors due to a need to explicit cast.

Do all types need to be projected? I mean can't we have structs like Point, Size, Rect that are only in the C# world and only when they are sent to C++ they are converted to float?

MikeHillberg:
[...]

PR #295, opened 2020-06-03
Revert "flip geometry types from double to float, per community feedback"

Scottj1s:
Based on community feedback, this change represents too much of a breaking change and compatibility issue.


Scottj1s:
fixes #280

PR #303, opened 2020-06-09
Adding support for agile references

manodasanW:
- Added an AsAgile extension method which can be used on WinRT types to get an agile reference which can then be used to marshal an object across apartments in an agile way even if the original object wasn't agile.  Note an extension method was chosen to avoid filling up all the projected interfaces and classes with an AsAgile function.
- Making using of RoGetAgileReference when available (Win 8.1+) to accomplish this, otherwise falling back to using an git table
- Making delegates registered via the native ABI agile when they are aren't.
- Added tests to exercise these scenarios.

### DIFF
--- a/TestComponentCSharp/NonAgileClass.cpp
+++ b/TestComponentCSharp/NonAgileClass.cpp
@@ -1,0 +1,125 @@
+#include "pch.h"
+#include "NonAgileClass.h"
+#include "NonAgileClass.g.cpp"
+
+using namespace winrt;
+
+namespace
+{
+    struct __declspec(uuid("624cd4e1-d007-43b1-9c03-af4d3e6258c4")) __declspec(novtable)
+        INonAgileBindableVectorChangedEventHandler : ::IUnknown
+    {
+        virtual int32_t __stdcall Invoke(void*, void*) noexcept = 0;
+    };
+
+    struct NonAgileDelegate : implements<NonAgileDelegate, non_agile, INonAgileBindableVectorChangedEventHandler, IMarshal>
+    {
+        NonAgileDelegate()
+        {
+        }
+        
+        int32_t __stdcall Invoke(void* p1, void* p2) noexcept override
+        {
+            VectorChanged(*reinterpret_cast<Microsoft::UI::Xaml::Interop::IBindableObservableVector const*>(&p1),
+                          *reinterpret_cast<Windows::Foundation::IInspectable const*>(&p2));
+            return S_OK;
+        }
+
+        HRESULT __stdcall GetUnmarshalClass(REFIID riid, void* pv, DWORD dwDestContext, void* pvDestContext, DWORD mshlflags, CLSID* pCid) noexcept final
+        {
+            if (m_marshaler)
+            {
+                return m_marshaler->GetUnmarshalClass(riid, pv, dwDestContext, pvDestContext, mshlflags, pCid);
+            }
+
+            return E_OUTOFMEMORY;
+        }
+
+        HRESULT __stdcall GetMarshalSizeMax(REFIID riid, void* pv, DWORD dwDestContext, void* pvDestContext, DWORD mshlflags, DWORD* pSize) noexcept final
+        {
+            if (m_marshaler)
+            {
+                return m_marshaler->GetMarshalSizeMax(riid, pv, dwDestContext, pvDestContext, mshlflags, pSize);
+            }
+
+            return E_OUTOFMEMORY;
+        }
+
+        HRESULT __stdcall MarshalInterface(IStream* pStm, REFIID riid, void* pv, DWORD dwDestContext, void* pvDestContext, DWORD mshlflags) noexcept final
+        {
+            if (m_marshaler)
+            {
+                return m_marshaler->MarshalInterface(pStm, riid, pv, dwDestContext, pvDestContext, mshlflags);
+            }
+
+            return E_OUTOFMEMORY;
+        }
+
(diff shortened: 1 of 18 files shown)


Scottj1s:
The NonAgileClass test harness seems fairly generic (language-agnostic), so should probably live in TestWinRT.  I wouldn't let it hold up the PR, but maybe [open an issue](https://github.com/microsoft/TestWinRT/issues) to move this at some point.

Scottj1s:
Can you open an issue to document the new AsAgile extension method, and maybe discuss agile delegates, on http://aka.ms/cswinrt?  Similar to what's been done for cppwinrt?

TestComponentCSharp/NonAgileClass.cpp
```diff
@@ -0,0 +1,125 @@
...
+        int32_t __stdcall Invoke(void* p1, void* p2) noexcept override
+        {
+            VectorChanged(*reinterpret_cast<Microsoft::UI::Xaml::Interop::IBindableObservableVector const*>(&p1),
+                          *reinterpret_cast<Windows::Foundation::IInspectable const*>(&p2));
+            return S_OK;
+        }
+
+        HRESULT __stdcall GetUnmarshalClass(REFIID riid, void* pv, DWORD dwDestContext, void* pvDestContext, DWORD mshlflags, CLSID* pCid) noexcept final
+        {
+            if (m_marshaler)
```
Scottj1s, line 30: probably don't need all these if checks, since you're aggregating the FTM unconditionally (and WINRT_VERIFY-ing it)

WinRT.Runtime/AgileReference.cs
```diff
@@ -0,0 +1,97 @@
...
+        {
+            IntPtr agileReference = default;
+            IntPtr gitPtr = default;
+            Guid iid = typeof(IUnknownVftbl).GUID;
+            try
+            {
+                Marshal.ThrowExceptionForHR(Platform.RoGetAgileReference(
+                    Platform.AgileReferenceOptions.AGILEREFERENCE_DEFAULT,
+                    ref iid,
+                    instance.ThisPtr,
```
Scottj1s, line 25: check for null ThisPtr and short circuit out with null _agileReference?

WinRT.Runtime/AgileReference.cs
```diff
@@ -0,0 +1,97 @@
...
+                try
+                {
+                    if (_git != null && _cookie != IntPtr.Zero)
+                    {
+                        _git.RevokeInterfaceFromGlobal(_cookie);
+                    }
+                }
+                catch(ObjectDisposedException)
+                {
+                    // TODO: How to handle removing from git when it has already been disposed.
```
Scottj1s, line 65: were you actually seeing this? your dispose pattern is properly implemented.

manodasanW: In the delegate scenarios I was seeing the finalizer already disposed those instance variables causing me to be unable to remove from the git.

WinRT.Runtime/Interop/IAgileReference.cs
```diff
@@ -0,0 +1,203 @@
...
+    [WindowsRuntimeType]
+    [Guid("C03F6A43-65A4-9818-987E-E0B810D2A6F2")]
+    internal interface IAgileReference
+    {
+        IObjectReference Resolve(Guid riid);
+    }
+
+    [WindowsRuntimeType]
+    [Guid("94ea2b94-e9cc-49e0-c0ff-ee64ca8f5b90")]
+    public interface IAgileObject
```
Scottj1s, line 17: nit: fold https://github.com/microsoft/CsWinRT/blob/1dc85f4f14deb3d85f83c3fc60b6684d25ed733b/WinRT.Runtime/ComWrappersSupport.cs#L31 in with this definition

WinRT.Runtime/AgileReference.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        public unsafe AgileReference(IObjectReference instance)
+        {
+            IntPtr agileReference = default;
+            IntPtr gitPtr = default;
+            Guid iid = typeof(IUnknownVftbl).GUID;
+            try
+            {
+                Marshal.ThrowExceptionForHR(Platform.RoGetAgileReference(
+                    Platform.AgileReferenceOptions.AGILEREFERENCE_DEFAULT,
```
Scottj1s, line 23: not a big deal, but I'd probably just drop in 0 and a comment so enum isn't needed (like CLSCTX_INPROC_SERVER)

WinRT.Runtime/AgileReference.cs
Scottj1s: ok, I don't think it's an issue to leak a GIT cookie, but it so, we could add a wrapper at a later time (maybe change this comment to suggest that)

weltkante: > I don't think it's an issue to leak a GIT cookie

are you sure this won't pollute the GIT (i.e. actually leak memory) or cause access violations if someone tries to access the already-released COM pointer? Documentation says registered interfaces **_must_** stay alive until revoked.

WinRT.Runtime/AgileReference.cs
manodasanW: Addressed and handled other issues hit with the null test.

WinRT.Runtime/AgileReference.cs
manodasanW: Ended up making the git table a static lazy reference which avoids this issue and avoids allocating it each time given it is a per process instance.

PR #305, opened 2020-06-10
Fix error in README MSBuild sample

wjk:
5aef58f is optional, and I am fully open to reverting it if desired. I just thought it would be better to submit it here instead of in a separate PR. Thanks!


Scottj1s:
thanks William!

PR #321, opened 2020-06-16
Address ComWrappers breaking change and add prerelease policy

### LINKED ISSUE #316: Update to new ComSupport api surface
.Net 5 SDK Preview 5 introduced a breaking change, causing C#/WinRT-based apps to fail at runtime on finding System.Runtime.InteropServices.ComWrappers.RegisterAsGlobalInstance.  Need to update the projection and add user protection from future prerelease breaking changes via msbuild assertions.

Scottj1s:
fixes #316

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ _build/
 global.json
 .nuget
 .nupkg
+nuget/Microsoft.Windows.CsWinRT.Prerelease.targets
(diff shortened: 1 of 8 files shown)


Scottj1s:
@jkoritzinsky - note the workaround for the winmd reference enforcement, which applies to ReferencePath when it should be ReferencePathWithRefAssemblies:

C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin\Roslyn\Microsoft.CSharp.Core.targets(111):
         References="@(ReferencePathWithRefAssemblies)"

jkoritzinsky:
Can you file an issue on dotnet/sdk and tag me?

Issue #326, opened 2020-06-23
WinRT.Runtime should not register its ComWrappers instance for Marshaling


jkoritzinsky:
It should only register for Tracker Support since it handles objects of any type but does not handle classic COM semantics.

cc: @Scottj1s

PR #327, opened 2020-06-23
do not register for ComWrappers marshaling support, but only reference tracking

Scottj1s:
and some more nuget behavior tweaking


Scottj1s:
fixes #326

PR #333, opened 2020-06-26
Add ComImport casting support, subject to some constraints

### LINKED ISSUE #302: Proposal: Address gap in ComImport support of WinRT interop
**Summary**
There are a number of COM interfaces that are not modeled in Windows metadata (the format defined in ECMA-335), and yet are required to make full use of many WinRT APIs.  Historically, these interfaces have been supported by the ComImport attribute, which the C# compiler and runtime use to generate projected interfaces on demand.  With .Net5.0, while ComImport support remains, it can no longer be used to interop with WinRT interfaces, and thus needs a replacement.  

**Background**
Most non-metadata interfaces exist to provide interop functionality.  Currently, the SDK “um” folder has 28 interop style headers, the majority for explicit HWND hosting, to support Desktop and multi-window UWP apps - the so-called [interop pattern](https://devblogs.microsoft.com/oldnewthing/20170315-00/?p=95735).  There are also interop interfaces that don't follow this pattern, and of course there are many user-defined interop interfaces.  So while it's possible to hand roll each interop interface definition, the work is non-trivial and doesn't scale to the need.
 
With C++/WinRT, the common practice is to excerpt an interop interface definition from its header and include it near the call site.  This ignores any type mappings that would otherwise be provided by the projection (e.g., HSTRING <--> std::wstring).  The same could be done with C#/WinRT, only mapping return HRESULT to exceptions.  This would be much simpler than providing a complete projection of an interface, and could be implemented with a source generator.  But many ComImport definitions also include MarshalAs attributes to provide natural type mappings on params and returns.  Here, the caller would be programming to the ABI, making it much less useable.

Alternately, a source generator could duplicate much of the type-mapping (projection) logic of cswinrt.exe, but that's a poor design decision.  It's not clear whether a source generator could exec cswinrt.exe to provide the projection, given that a fully built assembly would be required to provide interface definitions and ignore everything else.  
 
**Proposal**
Given the constraints, the simplest and best solution is to model the interop interface in WinRT IDL so that a winmd can be produced from it, and consumed by cswinrt.exe.  In many cases, this would require modest edits to COM IDL to make it WinRT IDL compatible (adding namespaces, removing HRESULT returns, etc).  To support IUnknown-based interfaces, which are prohibited by ECMA 335, an InterfaceIsIUnknown attribute (defined in any namespace) could be attached to the interface definition and used by cswinrt modified to provide the proper projection.  The attribute name is borrowed from the [ComInterfaceType enum](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.cominterfacetype?view=netcore-3.1) used with ComImport.

This system is definitely constrained.  It does not provide exhaustive coverage of all ComImport capabilities (marshaling, proxying, BSTRs, SafeArrays, etc).  Instead, it provides the subset of functionality typically required of interop interfaces - access to HWNDs, buffers, etc.

Based on this scheme, these interop scenarios can be supported:
1.	IInspectable-based interfaces defined in WinRT IDL: currently supported
2.	IInspectable-based interfaces defined in COM IDL: migrate the definition to WinRT IDL
3.	IUnknown-based interfaces defined in COM IDL: migrate the definition to WinRT and mark it with an InterfaceIsIUnknown attribute
4.	Any interface not defined in IDL (e.g., hand-rolled): model the interface in WinRT IDL, marking it with an InterfaceIsIUnknown attribute as needed

**Sample Interop IDL**
```
/*  Interop.idl
 *
 *	Experiments in generating interop interface projections for C#/WinRT
 */

/* a namespace is required for all WinRT interfaces */
namespace Interop
{

/* For any IInspectable-based interop currently defined in  COM IDL, modify it to produce metadata */
[uuid(D3EE12AD-3865-4362-9746-B75A682DF0E6)]
interface IAccountsSettingsPaneInterop
{
    UInt64 GetForWindow(UInt64 appWindow, GUID riid);
    UInt64 ShowManageAccountsForWindowAsync(UInt64 appWindow, GUID riid);
    UInt64 ShowAddAccountForWindowAsync(UInt64 appWindow, GUID riid);
}

/* Define attribute that cswinrt can look for to provide IUnknown-based interface projection */
[attributeusage(target_interface)]
[attributename("InterfaceIsIUnknown")]
attribute InterfaceIsIUnknownAttribute
{
}

/* For IUnknown-based interops (including those cswinrt currently hand rolls), attribute accordingly */ 
[uuid(905a0fef-bc53-11df-8c49-001e4fc686da)]
[InterfaceIsIUnknown]
interface IBufferByteAccess
{
    UInt64 Buffer();    
}

}

```

Scottj1s:
fixes #302 

This is somewhat experimental.  The hope is that most ComImport interfaces that would be associated with WinRT are interop-related and do not require integration of COM/WinRT type systems.  The related issue describes these, and excludes outliers like IGraphicsEffectD2D1Interop.


Scottj1s:
@jkoritzinsky - There is a reference leak in the TestComImports test case in this PR - ComImport objects are not being collected properly.  Is this a bug in the runtime - maybe some fallout of winrt removal?

UnitTest/TestComponentCSharp_Tests.cs
```diff
@@ -1594,5 +1585,66 @@ public void TestNonAgileDelegateCall()
...
+            TestObject();
+            GC.Collect();
+            GC.WaitForPendingFinalizers();
+            Assert.Equal(0, ComImports.NumObjects);
+
+            TestImports();
+            GC.Collect();
+            GC.WaitForPendingFinalizers();
+            // TODO: refcount leaks for ComImports
+            // Assert.Equal(0, ComImports.NumObjects);
```
Scottj1s, line 1647: here's the leak - NumObjects should be 0 here

AaronRobinsonMSFT: A single `GC.Collect()` may not be enough. This kind of validation is incredibly hard when dealing with the built in system. I would try calling `GC.Collect()` 5 or more times and then `GC.WaitForPendingFinalizers()`.

Scottj1s: I'll experiment with that, but didn't think it was that nondeterministic as:
a) we have a number of unit tests for CCWs that only do a single collect
b) I have a baseline for just the IInspectable object that demos a single collect

jkoritzinsky: I believe Aaron is correct here. The lifetimes all look reasonable to me, so it looks like we may have to run multiple GC collects to ensure we get the objects collected.

jkoritzinsky: The lifetimes of the C#/WinRT generated types are more easily handled since there's no extra magic keeping things alive a little longer. The built-in system lacks the simplicity of the C#/WinRT system, so objects may have their lifetime slightly extended.

AaronRobinsonMSFT: >but didn't think it was that nondeterministic

It is incredibly nondeterministic. It is quite possible that the unit test framework is helping out a lot here, but I would be surprised if a small console would give you the desired reliability. The `GC.Collect` is only going to work if there is needed pressure. The fact that a few lines up a `GC.Collect` was done indicates to me that the system has already cleaned up so it is good. This also means `GC.WaitForPendingFinalizers()` is probably not doing anything because the queue is empty.

Scottj1s: Thanks guys - I had a missing Release

Projections/Test/Test.csproj
```diff
@@ -8,7 +8,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <GenerateTestProjection Condition="'$(GenerateTestProjection)$(Configuration)' == 'Release'">true</GenerateTestProjection>
+    <GenerateTestProjection>true</GenerateTestProjection>
```
jkoritzinsky, line 11: I think committing this change may have been unintentional?

Scottj1s: the test projection executes quickly, even with the debug cswinrt.exe

jkoritzinsky: Ok! Just wanted to make sure.

PR #350, opened 2020-07-16
Emit function pointers for non-generic vtable members

### LINKED ISSUE #90: Enabling UnmanagedCallersOnly, calli perf improvements

jkoritzinsky:
Emit function pointers instead of delegates when possible. Will significantly reduce the memory footprint of projection on both .NET Standard 2.0 and .NET 5, but will more significantly reduce the .NET 5 footprint since it can avoid allocations in both directions.

Fixes #90


cswinrt/main.cpp
```diff
@@ -31,6 +31,7 @@ namespace cswinrt
         { "output", 0, 1, "<path>", "Location of generated projection" },
         { "include", 0, option::no_max, "<prefix>", "One or more prefixes to include in projection" },
         { "exclude", 0, option::no_max, "<prefix>", "One or more prefixes to exclude from projection" },
+        { "target", 0, 1, "<tfm>", "Target TFM for projection. Leave empty for code compatible with newest supported TFM. Currently supported: netstandard2.0, net5.0." },
```
Scottj1s, line 34: how about:
        { "target", 0, 1, "<netstandard2.0 | net5.0>", "Target for projection, defaulting to the latest available TFM" },

WinRT.Runtime/ComWrappersSupport.netstandard2.0.cs
```diff
@@ -120,25 +120,30 @@ public static T FindObject<T>(IntPtr thisPtr)
...
+            QueryInterface = (delegate* stdcall<IntPtr, ref Guid, out IntPtr, int>)Marshal.GetFunctionPointerForDelegate(Abi_QueryInterface),
+            AddRef = (delegate* stdcall<IntPtr, uint>)Marshal.GetFunctionPointerForDelegate(Abi_AddRef),
+            Release = (delegate* stdcall<IntPtr, uint>)Marshal.GetFunctionPointerForDelegate(Abi_Release),
         };
         
         public static IntPtr AllocateVtableMemory(Type vtableType, int size) => Marshal.AllocCoTaskMem(size);
 
+        private delegate int QueryInterface(IntPtr pThis, ref Guid iid, out IntPtr ptr);
+        private static QueryInterface Abi_QueryInterface = Do_Abi_QueryInterface; 
         private static int Do_Abi_QueryInterface(IntPtr pThis, ref Guid iid, out IntPtr ptr)
```
Scottj1s, line 134: does this need:
#if !NETSTANDARD2_0
            [UnmanagedCallersOnly]
#endif

same for AddRef & Release?

WinRT.Runtime/ObjectReference.cs
```diff
@@ -113,44 +113,37 @@ protected virtual void Dispose(bool disposing)
...
-                {
-                    // If we're in the finalizer, then releaseDelegate might already be null.
-                    // In this case, we need to re-fetch the delegate from the vtable.
-                    // TODO: Fix this once we have function pointers.
-                    releaseDelegate = Marshal.PtrToStructure<IUnknownVftbl>(Unsafe.AsRef<VftblPtr>(ThisPtr.ToPointer()).Vftbl).Release;
-                }
-            }
-
-            releaseDelegate(ThisPtr);
+            VftblIUnknown.Release(ThisPtr);
```
Scottj1s, line 123: wow - one of the benefits of not using a GC'd delegate. nice!

WinRT.Runtime/ObjectReference.cs
```diff
@@ -197,25 +184,34 @@ public static ObjectReference<T> FromAbi(IntPtr thisPtr)
...
+#else
+            if (RuntimeHelpers.IsReferenceOrContainsReferences<T>())
+#endif
             {
                 vftblT = (T)typeof(T).GetConstructor(BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.CreateInstance, null, new[] { typeof(IntPtr) }, null).Invoke(new object[] { thisPtr });
             }
             else
             {
-                vftblT = Marshal.PtrToStructure<T>(vftblPtr.Vftbl);
+                vftblT = Unsafe.AsRef<T>(vftblPtr.Vftbl.ToPointer());
```
Scottj1s, line 212: is AsRef more performant that PtrToStructure?

Scottj1s:
>             private static unsafe int Do_Abi_add_CollectionChanged_0(IntPtr thisPtr, IntPtr handler, out global::WinRT.EventRegistrationToken token)

if !NETSTANDARD2_0
            [UnmanagedCallersOnly]
endif

for these too?
Do_Abi_add_CollectionChanged_0
Do_Abi_remove_CollectionChanged_1


---
Refers to: WinRT.Runtime/Projections/INotifyCollectionChanged.cs:55 in ec55b3c. [](commit_id = ec55b3c2bae5b12a15d39651ed7e38007cab479f, deletion_comment = False)

Scottj1s:
>             private static unsafe int Do_Abi_add_PropertyChanged_0(IntPtr thisPtr, IntPtr handler, out global::WinRT.EventRegistrationToken token)

if !NETSTANDARD2_0
            [UnmanagedCallersOnly]
endif

for these?
Do_Abi_add_PropertyChanged_0
Do_Abi_remove_PropertyChanged_1


---
Refers to: WinRT.Runtime/Projections/INotifyPropertyChanged.cs:56 in ec55b3c. [](commit_id = ec55b3c2bae5b12a15d39651ed7e38007cab479f, deletion_comment = False)

WinUIDesktopSample/App.xaml.cs
```diff
@@ -40,9 +41,18 @@ protected override void OnLaunched(LaunchActivatedEventArgs args)
             myWindow = window;
         }
 
-        private void Button_Click(object sender, RoutedEventArgs e)
+        private async void Button_Click(object sender, RoutedEventArgs e)
```
Scottj1s, line 44: just for testing? not intended for checkin?

UnitTest/Properties/launchSettings.json
```diff
@@ -2,7 +2,7 @@
   "profiles": {
     "UnitTest": {
       "commandName": "Project",
-      "nativeDebugging": false
+      "nativeDebugging": true
```
Scottj1s, line 5: not intended for checkin?

WinUIDesktopSample/Properties/launchSettings.json
```diff
@@ -0,0 +1,8 @@
+{
```
Scottj1s, line 1: intended for checkin?

WinRT.Runtime/Projections/INotifyPropertyChanged.cs
```diff
@@ -6,28 +6,45 @@
...
+            private delegate*<IntPtr, IntPtr, out global::WinRT.EventRegistrationToken, int> _add_PropertyChanged_0;
+            public delegate* stdcall<IntPtr, IntPtr, out global::WinRT.EventRegistrationToken, int> add_PropertyChanged_0 { get => (delegate* stdcall<IntPtr, IntPtr, out global::WinRT.EventRegistrationToken, int>)_add_PropertyChanged_0; set => _add_PropertyChanged_0=(delegate*<IntPtr, IntPtr, out global::WinRT.EventRegistrationToken, int>)value; }
+            private delegate*<IntPtr, global::WinRT.EventRegistrationToken, int> _remove_PropertyChanged_1;
+            public delegate* stdcall<IntPtr, global::WinRT.EventRegistrationToken, int> remove_PropertyChanged_1 { get => (delegate* stdcall<IntPtr, global::WinRT.EventRegistrationToken, int>)_remove_PropertyChanged_1; set => _remove_PropertyChanged_1=(delegate*<IntPtr, global::WinRT.EventRegistrationToken, int>)value; }
+#endif
 
             private static readonly Vftbl AbiToProjectionVftable;
             public static readonly IntPtr AbiToProjectionVftablePtr;
+#if NETSTANDARD2_0
+            private static Delegate[] DelegateCache = new Delegate[2];
```
Scottj1s, line 34: is this to prevent GC?

WinRT.Runtime/ComWrappersSupport.netstandard2.0.cs
jkoritzinsky: This file is .NET Standard 2.0 only, so not in this case.

WinRT.Runtime/ObjectReference.cs
jkoritzinsky: Significantly. The JIT can see that `Unsafe.AsRef` is a pointer dereference instead of having to get the type and calling into the runtime and doing a copy.

WinUIDesktopSample/App.xaml.cs
jkoritzinsky: Probably a bad merge from when I initially started this work.

WinRT.Runtime/Projections/INotifyPropertyChanged.cs
jkoritzinsky: Yep.

Scottj1s:
super exciting.  can't wait to see the perf impact!

jkoritzinsky:
I've updated all of the non-generic manual projections in WinRT.Runtime to use function pointers. Currently have all tests passing on Windows x64 for .NET 5.0. Still working on figuring out what's going on with the .NET Standard 2.0 test failures I'm seeing with this PR.

jkoritzinsky:
Fixed the .NET Standard 2.0 implementation. Last step is to update any non-generic interfaces in the `additions` folder, which I'll probably do just enough to make them work for simplicity since those files aren't really TFM-specific easily right now.

UnitTest/Properties/launchSettings.json
jkoritzinsky: I feel like every time I debug C#/WinRT tests I need to turn this on. I can remove this if you'd like though.

jkoritzinsky:
`additions` folder has been updated.

jkoritzinsky:
This PR is ready for review.

WinRT.Runtime/Projections/INotifyCollectionChanged.cs
```diff
@@ -7,27 +7,44 @@
...
-                    remove_CollectionChanged_1 = Do_Abi_remove_CollectionChanged_1
+#if NETSTANDARD2_0
+                    _add_CollectionChanged_0 = (void*)Marshal.GetFunctionPointerForDelegate(DelegateCache[0] = new _add_EventHandler(Do_Abi_add_CollectionChanged_0)),
+                    _remove_CollectionChanged_1 = (void*)Marshal.GetFunctionPointerForDelegate(DelegateCache[1] = new _remove_EventHandler(Do_Abi_remove_CollectionChanged_1)),
+#else
+                    _add_CollectionChanged_0 = &Do_Abi_add_CollectionChanged_0,
+                    _remove_CollectionChanged_1 = &Do_Abi_remove_CollectionChanged_1,
+#endif
                 };
                 var nativeVftbl = (IntPtr*)ComWrappersSupport.AllocateVtableMemory(typeof(Vftbl), Marshal.SizeOf<global::WinRT.IInspectable.Vftbl>() + sizeof(IntPtr) * 2);
```
jkotas, line 49: Using `Marshal.SizeOf` and `Marshal.StructureToPtr` is orders of magnitude more expensive than `sizeof(...)` and simple memory copy that all this does for the `!NETSTANDARD2_0` case. Can this be leaner?

jkoritzinsky:
cc: @AaronRobinsonMSFT

WinRT.Runtime/Projections/ICommand.cs
```diff
@@ -182,6 +203,10 @@ private static unsafe int Do_Abi_CanExecute_2(IntPtr thisPtr, IntPtr parameter,
                 }
                 return 0;
             }
+
+#if !NETSTANDARD2_0
+            [UnmanagedCallersOnly]
```
AaronRobinsonMSFT, line 208: Can we change this pattern and instead define a dummy/placeholder type? Then this attribute can always be added and substantially reduce these preprocessor locations.

```CSharp
#if NETSTANDARD2_0
internal UnmanagedCallersOnlyAttribute : Attribute
{
}
#endif
```

UnitTest/Properties/launchSettings.json
Scottj1s: funny - me too.  I once checked it in (inadvertently) and you commented on that.  I'm fine with it.

WinRT.Runtime/Projections/ICommand.cs
Scottj1s: Unless the preprocessor is super slow, I'd stick with status quo - the fake attribute is a bit misleading

WinRT.Runtime/Projections/INotifyCollectionChanged.cs
jkoritzinsky: I've slimmed it down for the generated code, but I'd rather merge this in and clean up the manual projections later since this PR already has a lot of changes and a large impact on perf.

jkoritzinsky:
@Scottj1s can you take one more look and merge?

Projections/WinUI/WinUI.csproj
```diff
@@ -12,9 +12,9 @@
     <NoWarn>8305;0618</NoWarn>
   </PropertyGroup>
 
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
-    <DebugType>full</DebugType>
-    <DebugSymbols>true</DebugSymbols>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
+    <DebugType>full</DebugType>
```
AaronRobinsonMSFT, line 16: Is there a compelling reason to use Full PDBs rather than Portable?

cswinrt/strings/WinRT_Interop.cs
```diff
@@ -17,12 +17,11 @@ namespace WinRT.Interop
 {
     // IActivationFactory
     [Guid("00000035-0000-0000-C000-000000000046")]
-    internal struct IActivationFactoryVftbl
+    internal unsafe struct IActivationFactoryVftbl
     {
-        public unsafe delegate int _ActivateInstance(IntPtr pThis, out IntPtr instance);
-
         public IInspectable.Vftbl IInspectableVftbl;
```
AaronRobinsonMSFT, line 22: I am still seeing this pattern. As previously mentioned the fully types vtable does introduce a fair bit of additional metadata overhead with little real value. The previous suggestion was to use a `void*[]` and index/cast at the dispatch site. The size reduction would be substantial in some instances. Is this something we just don't want to do in this PR?

jkoritzinsky: We can't just use `void*[]` for the vtable unless we want to re-allocate a delegate every time we have to call a generic method. We would need to cache those somewhere. I'm fine with moving to a caching mechanism for those and using pointer indirection for non-generic methods in the future, but I'd rather not move to that in this PR given its size.

Projections/WinUI/WinUI.csproj
jkoritzinsky: This change was done by VS, not by myself. I don't know of any particular reason to use full PDBs.

AaronRobinsonMSFT: I would highly recommend portable PDBs if possible.

cswinrt/strings/WinRT_Interop.cs
AaronRobinsonMSFT: We shouldn't require a delegate allocation for this with function pointers right?

Regardless of the answer I completely agree this PR is big enough. Let's continue this conversation during the perf/optimization pass.

jkoritzinsky: We don't need a delegate allocation for IActivationFactory, that is true. We only need delegate allocations for methods with generic parameters or return values on WinRT interfaces.

Issue #352, opened 2020-07-20
'MarshalInterface<IStringable>' does not contain a definition for 'CopyAbiArray' 


ujjwalchadha:
Looks like this PR introduced the scenario we should probably be handling but aren't: https://github.com/microsoft/TestWinRT/pull/29/files specifically 

delegate Windows.Foundation.IStringable[] Array16Handler(Windows.Foundation.IStringable[] a, ref Windows.Foundation.IStringable[] b, out Windows.Foundation.IStringable[] c);

Trying to build the project gives the following error(s):

```
"C:\Users\ujchadha\source\repos\CsWinRT\cswinrt.sln" (default target) (1) ->
"C:\Users\ujchadha\source\repos\CsWinRT\UnitTest\UnitTest.csproj" (default target) (11) ->
"C:\Users\ujchadha\source\repos\CsWinRT\UnitTest\UnitTest.csproj" (Build target) (11:2) ->
"C:\Users\ujchadha\source\repos\CsWinRT\Projections\Test\Test.csproj" (default target) (14:2) ->
(CoreCompile target) ->
  Generated Files\TestComponent.cs(1380,78): error CS0117: 'MarshalInterface<IStringable>' does not contain a definition for 'CopyAbiArray' [C:\Users\
ujchadha\source\repos\CsWinRT\Projections\Test\Test.csproj]
  Generated Files\TestComponent.cs(7109,74): error CS0117: 'MarshalInterface<IStringable>' does not contain a definition for 'CopyAbiArray' [C:\Users\
ujchadha\source\repos\CsWinRT\Projections\Test\Test.csproj]


"C:\Users\ujchadha\source\repos\CsWinRT\cswinrt.sln" (default target) (1) ->
"C:\Users\ujchadha\source\repos\CsWinRT\UnitTest\UnitTest.csproj" (default target) (11) ->
"C:\Users\ujchadha\source\repos\CsWinRT\UnitTest\UnitTest.csproj" (Build target) (11:3) ->
"C:\Users\ujchadha\source\repos\CsWinRT\Projections\Test\Test.csproj" (default target) (14:4) ->
  Generated Files\TestComponent.cs(1380,78): error CS0117: 'MarshalInterface<IStringable>' does not contain a definition for 'CopyAbiArray' [C:\Users\
ujchadha\source\repos\CsWinRT\Projections\Test\Test.csproj]
[...]

PR #353, opened 2020-07-20
Added Reflection Performance Benchmarks

ujjwalchadha:


Benchmarks/ReflectionPerf.cs
```diff
@@ -0,0 +1,35 @@
...
+        [GlobalSetup]
+        public void Setup()
+        {
+            instance = new ClassWithMarshalingRoutines();
+        }
+
+        [Benchmark]
+        public void ExecuteComWrappersCreateKeyValuePairsFactory()
+        {
+            var value = instance.DefaultKeyValuePairProperty;
```
manodasanW, line 20: We should have all these functions return the value to prevent the variable from being optimized.

Benchmarks/ReflectionPerf.cs
```diff
@@ -0,0 +1,65 @@
...
+        ClassWithMarshalingRoutines instance;
+
+        [GlobalSetup]
+        public void Setup()
+        {
+            instance = new ClassWithMarshalingRoutines();
+        }
+
+        [Benchmark]
+        public object ExecuteComWrappersCreateKeyValuePairsFactoryWithNewObject()
```
manodasanW, line 18: I would recommend renaming these functions to reflect user scenario and putting a comment if you want to indicate the exact function we are trying to benchmark in cswinrt which is what the function name reflects today.

manodasanW: Something like ExecuteMarshalingForNewKeyValuePair or so.

manodasanW:
After the function rename, I think this is good to merge, so I will sign off.  But you should block merging until #352 is resolved.

PR #363, opened 2020-08-07
Updating compiler toolset to support function ptrs without VS preview

manodasanW:


Directory.Build.targets
```diff
@@ -22,5 +22,8 @@
     </RestoreSources>
   </PropertyGroup>
 
+ <ItemGroup Condition="'$(MSBuildProjectExtension)' == '.csproj'">
+   <PackageReference Include="Microsoft.Net.Compilers.Toolset" Version="3.7.0" />
```
jkoritzinsky, line 26: Does version 3.7.0 have what we need? I thought we needed to use a pre-release version of the package.

nuget/Microsoft.Windows.CsWinRT.nuspec
```diff
@@ -14,8 +14,12 @@
     <license type="file">LICENSE</license>
     <projectUrl>https://github.com/microsoft/cswinrt/tree/master/</projectUrl>
     <dependencies>
-      <group targetFramework=".NETStandard2.0" />
-      <group targetFramework=".NET5.0" />
+      <group targetFramework=".NETStandard2.0" >
+        <dependency id="Microsoft.Net.Compilers.Toolset" version="3.7.0" />
```
jkoritzinsky, line 18: I don't think we should add the compiler toolset package as a dependency. We don't want to force a downgrade for anyone building with a more recent preview (this package doesn't do downgrades well). And we're going to remove this dependency once .NET 5.0 releases.

Directory.Build.targets
jkoritzinsky: Synced offline that this package version works.

nuget/Microsoft.Windows.CsWinRT.nuspec
manodasanW: My main goal was to avoid users getting confused why it isn't building, but I agree this is a dependency that comes from VS and if they were using a later VS version, they would have a newer version of the package.  I updated the readme to communicate it.

PR #365, opened 2020-08-10
Don't directly use the underlying ObjectReference as the marshaler when marshalling ABI interface types.

jkoritzinsky:
We need to create a new ObjectReference, otherwise we'll dispose the ObjectReference instance that the ABI interface object contains, making all future usages fail with an ObjectDisposedException.

I discovered this bug with @stevenbrix when working on our prototype XAML compiler that emits IL. I've validated that it fixes the issue.

### DIFF
--- a/TestComponentCSharp/Class.cpp
+++ b/TestComponentCSharp/Class.cpp
@@ -1176,4 +1176,17 @@ namespace winrt::TestComponentCSharp::implementation
         // Compile-only test for keyword escaping
         throw hresult_not_implemented();
     }
+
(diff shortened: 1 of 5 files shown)


manodasanW:
@jkoritzinsky Has none our tests caught this because it is essentially a race for when it gets disposed or are we missing some coverage?

jkoritzinsky:
I think we're missing test coverage. I'll try to add a test that reproduces this that doesn't use XamlDirect.

jkoritzinsky:
@manodasanW I've added a test.

PR #369, opened 2020-08-14
Consume IDynamicInterfaceCastable in the .NET 5 projection

jkoritzinsky:
Use IDynamicInterfaceCastable to drive the managed->native calls in the .NET 5 projection.

Stop shipping the .NET Standard 2.0 projection.

TODO:

- [x] Validate that interface that inherit from collection types correctly implement all of the interface members from direct and transitive parents.
- [x] Move manually written type mappings over to IDynamicInterfaceCastable for .NET 5.
- [x] Update the RCW creation (not activation) to instantiate WinRT.IInspectable for all interface types on .NET 5.


jkoritzinsky:
There are a few perf losses in this PR that we should work on before merging this in:

- Calls are slightly slower since they need to get the object reference from the cache instead of it just being a field on the object.
- A cast that is not cached is significantly slower due to the reflection used to get a correctly typed vtable.

Scottj1s:
>       <group targetFramework=".NETStandard2.0" />

let's not remove just yet - there are non-Unity projects (WinML) that are evaluating cswinrt netstandard2.0

---
Refers to: nuget/Microsoft.Windows.CsWinRT.nuspec:17 in 471979e. [](commit_id = 471979e418209e825d8298354cadf4b916dc007f, deletion_comment = True)

cswinrt/code_writers.h
```diff
@@ -4478,8 +4481,9 @@ public static class %
 
     bool write_abi_interface(writer& w, TypeDef const& type)
     {
-        XLANG_ASSERT(get_category(type) == category::interface_type);
+         XLANG_ASSERT(get_category(type) == category::interface_type);
         auto type_name = write_type_name_temp(w, type, "%", true);
+        //std::cout << type_name << std::endl << std::endl << std::endl;
```
Scottj1s, line 4469: cruft

jkoritzinsky:
We should make sure to ship it in a separate package then. Otherwise people might get TypeLoadExceptions if they make a library targeting .NET Standard 2.0 and an application targeting .NET 5.0-Windows.

Scottj1s:
> #include <iostream>

cruft? (std::cout)

---
Refers to: cswinrt/code_writers.h:6 in 471979e. [](commit_id = 471979e418209e825d8298354cadf4b916dc007f, deletion_comment = False)

UnitTest/TestComponentCSharp_Tests.cs
```diff
@@ -525,15 +525,17 @@ public void TestComposedNonBlittableStruct()
             Assert.True(val.bools.x);
             Assert.False(val.bools.y);
             Assert.True(val.bools.z);
-        }
-
+        }
+
+#if NETCOREAPP2_0
```
Scottj1s, line 530: is the failing test?  if so can you add a comment/TODO?

UnitTest/TestComponentCSharp_Tests.cs
```diff
@@ -775,7 +777,7 @@ public void TestClassGeneric()
             {
                 var obj = objs[i];
                 Assert.Same(obj, TestObject);
-                Assert.Equal(TestObject.ThisPtr, objs[i].ThisPtr);
+                Assert.Equal(TestObject, objs[i]);
```
Scottj1s, line 780: how come the ThisPtr check was removed?

Scottj1s:
> using System;

I don't see any IDynamicInterfaceCastable tests - can we add a few?

---
Refers to: UnitTest/TestComponentCSharp_Tests.cs:1 in 471979e. [](commit_id = 471979e418209e825d8298354cadf4b916dc007f, deletion_comment = False)

WinRT.Runtime/IInspectable.cs
```diff
@@ -16,7 +16,7 @@ public enum TrustLevel
     // IInspectable
     [ObjectReferenceWrapper(nameof(_obj))]
     [Guid("AF86E2E0-B12D-4c6a-9C5A-D7AA65101E90")]
-    public class IInspectable
+    public partial class IInspectable
```
Scottj1s, line 19: partial only to combine with IInspectable.net5.cs?  #if NET5_0 might be simpler

WinRT.Runtime/Interop/IAgileReference.netstandard2.0.cs
```diff
@@ -1,213 +1,213 @@
...
+        public struct Vftbl
+        {
+            public global::WinRT.Interop.IUnknownVftbl IUnknownVftbl;
+            private void* _Resolve;
+            public delegate* stdcall<IntPtr, ref Guid, out IntPtr, int> Resolve { get => (delegate* stdcall<IntPtr, ref Guid, out IntPtr, int>)_Resolve; set => _Resolve = value; }
+
+            public static readonly Vftbl AbiToProjectionVftable;
+            public static readonly IntPtr AbiToProjectionVftablePtr;
+
+#if NETSTANDARD2_0
```
Scottj1s, line 49: cruft: everywhere we've forked sources, we can get rid of conditional compilation

Scottj1s:
many of the *.net5.cs/*.netstandard2.0.cs splits seem unnecessary (the sources are identical).  are we expecting them to diverge?  if not, can we remove the duplication?

UnitTest/TestComponentCSharp_Tests.cs
ujjwalchadha: This is not, but this test uses the ABI type directly which is not possible with IDIC anymore so I removed it from .net5

Scottj1s:
> We should make sure to ship it in a separate package then. Otherwise people might get TypeLoadExceptions if they make a library targeting .NET Standard 2.0 and an application targeting .NET 5.0-Windows.

I'm fine with that for RTM - we have a few validation scenarios that are targeting netstandard2.0 that I want to keep enabled in the meantime.

Scottj1s (approved):
:shipit:

WinRT.Runtime/Interop/IAgileReference.net5.cs
```diff
@@ -0,0 +1,194 @@
...
+        public struct Vftbl
+        {
+            public global::WinRT.Interop.IUnknownVftbl IUnknownVftbl;
+            private void* _Resolve;
+            public delegate* stdcall<IntPtr, ref Guid, out IntPtr, int> Resolve { get => (delegate* stdcall<IntPtr, ref Guid, out IntPtr, int>)_Resolve; set => _Resolve = value; }
+
+            public static readonly Vftbl AbiToProjectionVftable;
+            public static readonly IntPtr AbiToProjectionVftablePtr;
+
+#if NETSTANDARD2_0
```
manodasanW, line 51: Do we need netstandard compat in a net5 file?

manodasanW: How much diff is there to merge both files?

WinRT.Runtime/Projections/Bindable.net5.cs
```diff
@@ -0,0 +1,1395 @@
...
+
+                };
+                var nativeVftbl = (IntPtr*)ComWrappersSupport.AllocateVtableMemory(typeof(Vftbl), Marshal.SizeOf<global::WinRT.IInspectable.Vftbl>() + sizeof(IntPtr) * 3);
+                Marshal.StructureToPtr(AbiToProjectionVftable, (IntPtr)nativeVftbl, false);
+                AbiToProjectionVftablePtr = (IntPtr)nativeVftbl;
+            }
+
+
+            [UnmanagedCallersOnly]
+
```
manodasanW, line 103: nit: blank line?

WinRT.Runtime/Interop/IAgileReference.net5.cs
ujjwalchadha: I think I fixed it in a newer commit.

WinRT.Runtime/Projections/Bindable.net5.cs
ujjwalchadha: Should I remove these? These appear as these files had conditional statements on NET5 and NETSTANDARD2.0 and I splitted the files through a script

WinRT.Runtime/Projections/IEnumerable.net5.cs
```diff
@@ -0,0 +1,638 @@
...
+                        m_firstItem = false;
+                        _MoveNext();
+                    }
+
+                    return m_hasCurrent;
+                }
+            }
+
+            public bool _MoveNext()
+            {
```
manodasanW, line 368: Is it possible for someone to call movenext first without calling current or hascurrent or is it just us calling it?  If it is possible, would that make m_firstItem wrong?

WinRT.Runtime/Projections/IPropertyValue.net5.cs
```diff
@@ -0,0 +1,1913 @@
...
+            T[] coercedArray = new T[dataArray.Length];
+            for (int i = 0; i < dataArray.Length; ++i)
+            {
+                try
+                {
+                    coercedArray[i] = UnboxValue<T>(dataArray.GetValue(i));
+                }
+                catch (InvalidCastException elementCastException)
+                {
+                    //global::System.Exception e = new InvalidCastException(string.Format(SR.InvalidCast_WinRTIPropertyValueArrayCoersion, value.GetType(), typeof(T[]).Name, i, elementCastException.Message), elementCastException.HResult);
```
manodasanW, line 236: is commented code intentional?

WinRT.Runtime/Projections/Bindable.net5.cs
manodasanW: there is quite a bit of those, fine to leave as is.

manodasanW (approved):

UnitTest/TestComponentCSharp_Tests.cs
jkoritzinsky: There's no longer a `ThisPtr` property exposed because it makes the public surface area not match the WinMD contract and it isn't entirely accurate for composed types. There's no good reason to use the `ThisPtr` property, so I removed it early in the design.

WinRT.Runtime/IInspectable.cs
jkoritzinsky: Given that IInspectable implements an extra interface in .NET 5.0, we'd have to add quite a few `#if NET5_0` lines throughout the file. It's easier IMO to have two separate files.

PR #374, opened 2020-08-25
Initial WinRT.Host.dll implementation 

Scottj1s:
Includes WinRT.Host.dll (native) and WinRT.Host.Shim.dll (managed) components of the hosting support for C# component authoring, as well as host tests.  Project integration will come along later, when authoring is in.

Note: host tests are currently failing on azure devops.  Will take a bit of debt to get this in, and then address the test failures.


jkoritzinsky:
I think we should be able to remove the WinRT.Host.Shim assembly completely.

The code-generation for the hosting side should provide the native `GetActivationFactory` entrypoint with the ABI types and marked as `[UnmanagedCallersOnly]` in a well-known type name (which looks to already be the case for a managed signature) and the WinRT.Host.dll native dll should call `load_assembly_and_get_function_pointer` directly to that method with a delegate type of `UNMANAGEDCALLERSONLY_METHOD`. This will completely remove the assembly loading ALC issues in `WinRT.Host.Shim` and result in one less file on disk.

Additionally, the generated `GetActivationFactory` implementations should be hard-coded and shouldn't use any reflection to discover type names. I noticed that the test cases use `Type.GetType` to resolve the type names. Since we know which types are provided (and even in those cases directly do an equality comparison for a particular name), we should just use `typeof(TypeName)`, or better yet, make the implementation of `ActivationFactory` generic with a `where T : new()` constraint to completely remove reflection.

Scottj1s:
> I think we should be able to remove the WinRT.Host.Shim assembly completely.
> 
> The code-generation for the hosting side should provide the native `GetActivationFactory` entrypoint with the ABI types and marked as `[UnmanagedCallersOnly]` in a well-known type name (which looks to already be the case for a managed signature) and the WinRT.Host.dll native dll should call `load_assembly_and_get_function_pointer` directly to that method with a delegate type of `UNMANAGEDCALLERSONLY_METHOD`. This will completely remove the assembly loading ALC issues in `WinRT.Host.Shim` and result in one less file on disk.
> 
> Additionally, the generated `GetActivationFactory` implementations should be hard-coded and shouldn't use any reflection to discover type names. I noticed that the test cases use `Type.GetType` to resolve the type names. Since we know which types are provided (and even in those cases directly do an equality comparison for a particular name), we should just use `typeof(TypeName)`, or better yet, make the implementation of `ActivationFactory` generic with a `where T : new()` constraint to completely remove reflection.

Thanks!  The primary reason for the shim is to support scenarios where the target assembly has been renamed.  In that case, we have no way of knowing the assembly's internal display name, and cannot correctly formulate the assembly-qualified typename to load (second param of load_assembly_and_get_function_pointer).  If we think that's an edge case not worth supporting, I'll remove the shim.  Thoughts?

jkoritzinsky:
I don’t believe we support that edge case in the comhost. I’d recommend that if that scenario needs to be supported, that we add that information to the data file that defines the activatable classes (since that will already require updates to the deps.json).

Scottj1s:
> I don’t believe we support that edge case in the comhost. I’d recommend that if that scenario needs to be supported, that we add that information to the data file that defines the activatable classes (since that will already require updates to the deps.json).

Works for me - thx!

Scottj1s:
@jkoritzinsky - a few additional issues with removing the shim:

1. load_assembly_and_get_function_pointer is case sensitive, so even a target assembly rename that changes case will fail to load.  I had forgot that I encountered that early on.  That's a bit more brittle than a user changing the file name to enable reg-free lookup (which should arguably be done by setting the project's AssemblyName to ensure consistent display/file naming)

2. Each target is loaded into a unique ALC.  This causes a fault on subsequent calls (e.g., a later unit test case) to ComWrappers.RegisterForTrackerSupport (_comWrappers being null).  We could catch and ignore that exception, but I wonder if we should be isolating in the first place.

jkoritzinsky:
> @jkoritzinsky Jeremy Koritzinsky FTE - a few additional issues with removing the shim:
> 
> 1. load_assembly_and_get_function_pointer is case sensitive, so even a target assembly rename that changes case will fail to load.  I had forgot that I encountered that early on.  That's a bit more brittle than a user changing the file name to enable reg-free lookup (which should arguably be done by setting the project's AssemblyName to ensure consistent display/file naming)

This could be (and should be) also tracked in the data file as the COM host does.

> 2. Each target is loaded into a unique ALC.  This causes a fault on subsequent calls (e.g., a later unit test case) to ComWrappers.RegisterForTrackerSupport (_comWrappers being null).  We could catch and ignore that exception, but I wonder if we should be isolating in the first place.

I didn't realize that `WinRT.Runtime` and `Microsoft.Windows.SDK.NET` are copied app-local with the SDK work. That does make this more annoying. There's good arguments both for and against isolating, but the global registration kinda pushes the argument towards not isolating WinRT.Runtime since that already forces a break in the isolation.

The easiest way to share it would be to keep the shim and instead of calling `Assembly.LoadFrom()`, create a new ALC that uses the `AssemblyDependencyResolver` to resolve all dependencies except `WinRT.Runtime`. The `WinRT.Runtime` assembly should be resolved in the default ALC, likely with an event handler registered on the default ALC to resolve to it. This does end up with the problem that the version of WinRT.Runtime that ships with the first loaded WinRT component into the process becomes the only loaded version in the process, but that's already the case (the first instance wins) for the global ComWrappers instance registered for tracker support. Although you're still likely to get type conflicts if a particular C#/WinRT projection is loaded in multiple times until we have the IDynamicInterfaceCastable work in.

If we want to avoid most of these type identity issues, we can instead harden WinRT.Runtime for the scenario where there is already a global tracker support instance registered. That would avoid the type identity issues unless the runtime ends up creating wrappers via the global instance (in which case, the first WinRT.Runtime assembly wins)

cc: @AaronRobinsonMSFT for any ideas around solving this.

HostTest/test.cpp
```diff
@@ -0,0 +1,142 @@
+// This unit test exercises WinRT.Host.dll, which provides hosting for
+// runtime components written in C#.  For details on the behavior of 
+// the host, see the spec.
+// 
+// Each test case specifies:
+// 1. A runtime class to be activated: 
+//		ProbyByClass, ProbeByHost for expected successes
```
manodasanW, line 7: typo: ProbeByClass.

WinRT.Host/NetHost.targets
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup Label="Globals">
+    <ResolveNuGetPackages>false</ResolveNuGetPackages>
+    <BuildPassReferences>true</BuildPassReferences>
+    <NetHostUseAsStatic>true</NetHostUseAsStatic>
```
manodasanW, line 6: Who is meant to set this?  Is this intended for us or for someone to override?

WinRT.Host/WinRT.Host.cpp
```diff
@@ -0,0 +1,396 @@
...
+    }
+
+    hostfxr_initialize_for_runtime_config = (hostfxr_initialize_for_runtime_config_fn)::GetProcAddress(lib, "hostfxr_initialize_for_runtime_config");
+    hostfxr_get_runtime_delegate = (hostfxr_get_runtime_delegate_fn)::GetProcAddress(lib, "hostfxr_get_runtime_delegate");
+    hostfxr_close = (hostfxr_close_fn)::GetProcAddress(lib, "hostfxr_close");
+    hostfxr_set_error_writer = (hostfxr_set_error_writer_fn)::GetProcAddress(lib, "hostfxr_set_error_writer");
+
+    if (!is_hostfxr_loaded())
+    {
+        winrt::throw_last_error();
```
manodasanW, line 137: Do we expect throw last error to be overridden if any one of those failed then the other after it succeeded.  Or will it be all fail or all pass.

WinRT.Host/NetHost.targets
Scottj1s: this is an internal setting for winrt.host.dll, to eliminate yet another dll distribution (nethost.dll)

WinRT.Host/WinRT.Host.cpp
Scottj1s: ah, good catch

PR #378, opened 2020-08-26
track source metadata for authoring projections

### LINKED ISSUE #371: Project WinMD assembly information in consuming projection
In order to be able to author a component in C# and then produce a WinMD from it, you need to be able to resolve all the types from a projection referenced by a C# component back to its WinMD assembly.  To do that, this information first needs to be projected.

Scottj1s:
fixes #371 

Jeremy - any concern with piggybacking the source metadata (sans path, extension) onto WindowsRuntimeType for authoring support?  trying to economize on attributes

### DIFF
--- a/Strings.props
+++ b/Strings.props
@@ -4,6 +4,9 @@
     <PrepareForBuildDependsOn>
       UpdateStringSources;UpdateStringsHeader;$(PrepareForBuildDependsOn)
     </PrepareForBuildDependsOn>
+    <ComputeCompileInputsTargets>
+      AddStringSources;$(ComputeCompileInputsTargets)
+    </ComputeCompileInputsTargets>
   </PropertyGroup>
 
   <UsingTask TaskName="WriteStringSources"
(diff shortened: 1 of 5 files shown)


Strings.props
```diff
@@ -4,6 +4,9 @@
     <PrepareForBuildDependsOn>
       UpdateStringSources;UpdateStringsHeader;$(PrepareForBuildDependsOn)
     </PrepareForBuildDependsOn>
+    <ComputeCompileInputsTargets>
```
Scottj1s, line 7: this is just to hide the glut of generated string sources in the cswinrt project node (the previous technique stopped working with a VS update)

WinRT.Runtime/Projections/Type.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace ABI.System
 {
-    [WindowsRuntimeType]
+    [WindowsRuntimeType("Windows.UI.Xaml")]
```
manodasanW, line 8: Is this accurate?

Scottj1s: oops, meant to fix that - should be Windows.UI.Xaml.Interop

manodasanW: This should be the assembly right not the namespace?

Scottj1s: oh, had it right to begin with - yes, the assembly.  e.g., C:\Windows\System32\WinMetadata\Windows.UI.Xaml.winmd

Issue #309, opened 2020-06-11
Enable production (authoring) support of WinRT types


Scottj1s:
This would replace the existing winmdexp tooling experience, to enable C# developers to define WinRT types (all kinds).  These types could be used internally by the app generating them, or externally from a packaged runtime component.  A source generator might be useful for reflecting on WinRT types as they're defined, and creating backing metadata and implementation.

BenJKuhn:
Tagging for RC1, but work will start sooner, some pieces may deliver earlier.

stevenbrix:
@AdamBraden, what is the MSBuild property that you need to set in the project for this?

angelazhangmsft:
Authoring support is enabled in preview currently

PR #383, opened 2020-08-28
Adding WinMD generator

manodasanW:
Adding a source generator which parses a C# project and generates a WinMD with the public types from it.  The WinMD generator isn't complete and there is some missing scenarios in it currently, but it generates a WinMD for most of the basic scenarios.  Portions of the WinMD generation code can be simplified to use the functions which generate WinMD types from the symbol  throughout but that would be addressed in a future PR.

Also adding a sample to test the source generator.

Contributes to #309


Authoring/AuthoringSample/AuthoringSample.csproj
```diff
@@ -0,0 +1,18 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net5.0</TargetFramework>
+    <LangVersion>preview</LangVersion>
+    <AssemblyVersion>1.0.0.0</AssemblyVersion>
+    <EnableCsWinRTLogging>true</EnableCsWinRTLogging>
```
Scottj1s, line 7: do you expect to make this permanent?  Can we rename to CsWinRTEnableLogging?

Authoring/AuthoringSample/AuthoringSample.csproj
```diff
@@ -0,0 +1,18 @@
...
+
+  <PropertyGroup>
+    <TargetFramework>net5.0</TargetFramework>
+    <LangVersion>preview</LangVersion>
+    <AssemblyVersion>1.0.0.0</AssemblyVersion>
+    <EnableCsWinRTLogging>true</EnableCsWinRTLogging>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <CompilerVisibleProperty Include="AssemblyName" />
```
Scottj1s, line 11: I assume AssemblyName has a reasonable default value?

Authoring/AuthoringSample/AuthoringSample.csproj
manodasanW: Yea I think we are going to need some form of logging mode.  I can rename it.

Authoring/WinRT.SourceGenerator/Generator.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        private string GetAssemblyVersion(SourceGeneratorContext context)
+        {
+            context.AnalyzerConfigOptions.GlobalOptions.TryGetValue("build_property.AssemblyVersion", out var assemblyVersion);
+            return assemblyVersion;
+        }
+
+        private string GetWinmdOutputFile(SourceGeneratorContext context)
+        {
+            string winmdDir = Path.Combine(Directory.GetCurrentDirectory(), "Generated Files");
```
Scottj1s, line 27: GeneratedFiles can be defaulted like this, but should also support a project definition (as cppwinrt does)

Scottj1s: One reason for that is cswinrt can generate different code based on the -tpm switch, in which case we'd want to generated sources under $IntDir

Authoring/WinRT.SourceGenerator/Generator.cs
```diff
@@ -0,0 +1,90 @@
...
+            Directory.CreateDirectory(winmdDir);
+            return Path.Combine(winmdDir, GetAssemblyName(context) + ".winmd");
+        }
+
+        private void GenerateWinMD(MetadataBuilder metadataBuilder, string outputFile)
+        {
+            Logger.Log("Writing " + outputFile);
+            var managedPeBuilder = new ManagedPEBuilder(
+                new PEHeaderBuilder(imageCharacteristics: Characteristics.ExecutableImage),
```
Scottj1s, line 36: you don't have to OR in Characteristics.Dll ?

Authoring/WinRT.SourceGenerator/Logger.cs
```diff
@@ -0,0 +1,33 @@
...
+namespace Generator
+{
+    class Logger
+    {
+        public static void Initialize(SourceGeneratorContext context)
+        {
+            context.AnalyzerConfigOptions.GlobalOptions.TryGetValue("build_property.EnableCsWinRTLogging", out var enableLogging);
+            if(enableLogging != null && bool.Parse(enableLogging))
+            {
+                string generatedDir = Path.Combine(Directory.GetCurrentDirectory(), "Generated Files");
```
Scottj1s, line 13: this should be factored into a shared function, GetGeneratedFiles

Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
```diff
@@ -0,0 +1,1791 @@
...
+        }
+    }
+
+    class TypeDeclaration
+    {
+        public readonly ISymbol Node;
+        public TypeDefinitionHandle Handle;
+        public string DefaultInterface;
+
+        public Dictionary<ISymbol, List<MethodDefinitionHandle>> MethodDefinitions;
```
Scottj1s, line 73: can inline assign these

Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
```diff
@@ -0,0 +1,1791 @@
...
+            {
+                MethodReferences[node] = new List<EntityHandle>();
+            }
+
+            MethodReferences[node].Add(handle);
+        }
+
+        public List<MethodDefinitionHandle> GetMethodDefinitions()
+        {
+            return MethodDefinitions.Values.SelectMany(list => list).ToList();
```
Scottj1s, line 118: can't just use MethodDefinitions.Values.ToList() ?

Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
```diff
@@ -0,0 +1,1791 @@
...
+        }
+
+        public List<EntityHandle> GetMethodReferences()
+        {
+            return MethodReferences.Values.SelectMany(list => list).ToList();
```
Scottj1s, line 123: same as above

Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
manodasanW: No, that makes it a list of lists.  I am trying to merge all the method definitions created for each symbol (i.e. property, method, event, etc.) into 1 list.

Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
```diff
@@ -0,0 +1,1791 @@
...
+                isInterfaceParent,
+                method.MethodKind == MethodKind.Constructor,
+                true,
+                true);
+            currentTypeDeclaration.AddMethod(method, methodDefinitionHandle);
+        }
+
+        void AddFactoryMethod(INamedTypeSymbol classSymbol, IMethodSymbol method)
+        {
+            Logger.Log("ading factory method: " + method.Name);
```
Scottj1s, line 1305: typo

Authoring/AuthoringSample/AuthoringSample.csproj
manodasanW: Yes in this sample the default seems to be AuthoringSample.

Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
```diff
@@ -0,0 +1,1791 @@
...
+            AddSynthesizedInterface(
+                classDeclaration,
+                SynthesizedInterfaceType.Factory,
+                classMembersFromInterfaces);
+
+            AddSynthesizedInterface(
+                classDeclaration,
+                SynthesizedInterfaceType.Default,
+                classMembersFromInterfaces);
+        }
```
Scottj1s, line 1554: overridable and composable?

manodasanW: on the todo, will be added once I got the cswinrt side working for basic components.  Will add a todo comment.

Authoring/WinRT.SourceGenerator/Generator.cs
manodasanW: Agree on this, going to defer implementing this until we figure out the right place to write to.

Authoring/WinRT.SourceGenerator/Generator.cs
manodasanW: Updated and added a few more flags.

Issue #380, opened 2020-08-27
Make WinRT.ActivationFactory public


AdamBraden:
Currently this is internal, which prevents the ability to convert this code

eg:

```
    [System.Runtime.InteropServices.Guid("39E050C3-4E74-441A-8DC0-B81104DF949C")]
    [System.Runtime.InteropServices.InterfaceType(System.Runtime.InteropServices.ComInterfaceType.InterfaceIsIInspectable)]
    public interface IUserConsentVerifierInterop
    {
        IAsyncOperation<UserConsentVerificationResult> RequestVerificationForWindowAsync(IntPtr appWindow, [MarshalAs(UnmanagedType.HString)] string Message, [In] ref Guid riid);
    }

    //Helper to initialize UserConsentVerifier
    public static class UserConsentVerifierInterop
    {
        public static IAsyncOperation<UserConsentVerificationResult> RequestVerificationForWindowAsync(IntPtr hWnd, string Message)
        {
            IUserConsentVerifierInterop userConsentVerifierInterop = (IUserConsentVerifierInterop)WindowsRuntimeMarshal.GetActivationFactory(typeof(UserConsentVerifier));
            Guid guid = typeof(IAsyncOperation<UserConsentVerificationResult>).GUID;

            return userConsentVerifierInterop.RequestVerificationForWindowAsync(hWnd, Message, ref guid);

        }
    }
```
Need to be able to do something like this:
```
IUserConsentVerifierInterop userConsentVerifierInterop = new WinRT.ActivationFactory<UserConsentVerifier>().As<Windows.Foundation.IGetActivationFactory>().As<IUserConsentVerifierInterop>();

Scottj1s:
if we do #379, we shouldn't need to make ActivationFactory public.  Are there any other scenarios where we'd need to cast to an interop from a factory/static interface?  If so, we can add an As<> for those too.

PR #385, opened 2020-08-28
Project static As<> casting methods for access to factories for interops, etc

### LINKED ISSUE #379: Expose static As<>() method on static class projections to support interop
motivating scenario is Windows.Security.Credentials.UI.UserConsentVerifier, which is a static class, which nests an internal activation factory and casts that to IUserConsentVerifierStatics.  Users may sometimes require access to this class for interop (or other) usage.

Scottj1s:
fixes #379, #380


UnitTest/TestComponentCSharp_Tests.cs
```diff
@@ -535,6 +535,50 @@ public void TestGenericCast()
...
+            void GetIids(out int iidCount, out IntPtr iids);
+            void GetRuntimeClassName(out IntPtr className);
+            void GetTrustLevel(out TrustLevel trustLevel);
+
+            void ToString(out IntPtr hstr);
+        }
+
+        [ComImport]
+        [Guid("39E050C3-4E74-441A-8DC0-B81104DF949C")]
+        [InterfaceType(ComInterfaceType.InterfaceIsIInspectable)]
```
manodasanW, line 553: I see above comment says ComInterfaceType.InterfaceIsIInspectable is not supported, but is in use here.  Am I misunderstanding something?

Scottj1s: I'll clarify the comment

PR #388, opened 2020-08-31
Update to .NET 5 Preview 8 and update compiler toolset package

jkoritzinsky:
Update to .NET 5 Preview 8 and make changes needed to support the dogfood scenario (TPM of Windows but we don't want the prebuilt SDK projection).

Update function pointer syntax to the new syntax.

Fix bugs in usage of UnmanagedCallersOnly that were discovered via the new compiler errors from the updated toolset package.


jkoritzinsky:
Sounds good!

Also, heads up I forgot to put into the PR description: Due to bugs in the Appx targets, I had to disable the additional build tooling for the WinUIDesktopSample, so it doesn't have compiled XAML or anything else added by additional targets. This was the workaround suggested by @stevenbrix for the "target platform" issue.

PR #402, opened 2020-09-03
[WIP] Add put_ methods for property set accessors, for Powershell binary compat

Scottj1s:
validating with partners before committing


Scottj1s:
closing PR as this doesn't appear to be a simply binary no-build drop in

PR #403, opened 2020-09-04
Adding source generator to nuget package.

manodasanW:
Adding source generator to nuget package.  Setting CsWinRTComponent to true enables it.

### DIFF
--- a/Authoring/AuthoringSample/AuthoringSample.csproj
+++ b/Authoring/AuthoringSample/AuthoringSample.csproj
@@ -5,12 +5,14 @@
     <LangVersion>preview</LangVersion>
     <AssemblyVersion>1.0.0.0</AssemblyVersion>
     <CsWinRTEnableLogging>true</CsWinRTEnableLogging>
+    <CsWinRTComponent>true</CsWinRTComponent>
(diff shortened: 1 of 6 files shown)


Authoring/AuthoringSample/AuthoringSample.csproj
```diff
@@ -5,12 +5,14 @@
     <LangVersion>preview</LangVersion>
     <AssemblyVersion>1.0.0.0</AssemblyVersion>
     <CsWinRTEnableLogging>true</CsWinRTEnableLogging>
+    <CsWinRTComponent>true</CsWinRTComponent>
```
Scottj1s, line 8: is this to make enabling source generation explicit (not invoked from simply referencing the cswinrt nuget)?

Scottj1s (approved):

Authoring/AuthoringSample/AuthoringSample.csproj
manodasanW: Right, this is to make not anyone with a reference to the cswinrt nuget generate an component projection for it as someone may reference it to create a consumption projection.  Specifying this property signifies you are a cswinrt authored component and triggers the source generator.

PR #416, opened 2020-09-18
Authoring part 2

manodasanW:
- Added call from source generator to CsWinRT to generate sources and passing those sources back to the source generator
- Updated CsWinRT to generate sources to project the authored component into WinRT
- Changed CsWinRT to distinguish between the authored / projected type and the CCW type which is used to project an authored component into WinRT and made use of it to generate CCW sources for authored types
- Currently supports most of the common different types in a WinMD as per sample, but there are still some types that need to be added (generics, attributes, etc)
- Added a test to use C++/WinRT to consume the authored component to demonstrate the working functionality
- Fixed several bugs in WinMD generator
- Added logic to wrap an authored type with the CCW type when it is detected as an type that we project

Contributes to #309


Scottj1s (approved):
coming along!

Issue #338, opened 2020-06-29
Apply Strong-Name Signing to winrt.runtime.dll 


Scottj1s:
Based on .Net Core best practice, winrt.runtime.dll should be [strong-name signed](https://github.com/dotnet/runtime/blob/master/docs/project/strong-name-signing.md).

Modern frameworks can automatically apply binding redirects, so the burden of managing strict bindings is no longer an issue.  And since strong-name signing is viral, winrt.runtime.dll must be strong-name signed to permit referencing from another strong-name signed assembly.

Points for consideration:  

- Does it matter that there are two versions of winrt.runtime.dll (netstandard2 and net5)?  Each would have a distinct hash.  It's possible both could be loaded in a single process, in which case unique strong-name signatures would be required.
- Should winrt.runtime.dll be [publicly signed](https://github.com/dotnet/runtime/blob/master/docs/project/public-signing.md), as .Net Core assemblies are?  Or does true strong-name signing offer greater benefits, with only the requirement of distributing the private key?

joperezr:
I maintain the https://github.com/dotnet/iot repo and currently trying to fix https://github.com/dotnet/iot/issues/1091 which involves updating the way we call WinRT code to use CsWinRT instead. I'm currently blocked in doing so because of this issue as I am getting:
```
CSC : error CS8002: Referenced assembly 'Microsoft.Windows.SDK.NET, Version=10.0.18362.3, Culture=neutral, PublicKeyToken=null' does not have a strong name.
```

Without this repo producing strong-name signed assemblies, we are currently blocked with onboarding this. cc: @jkoritzinsky

Scottj1s:
PR #420

PR #420, opened 2020-09-21
Strong-Name Signing to winrt.runtime.dll

karenbtlai:
- Add binary key and strong-name signing property to `WinRT.Runtime.csproj`
-  Resulting .dll when investigated with ILSpy

winrtRuntimeDll


Scottj1s:
fixes #338

Scottj1s (approved):
thank you!

jkoritzinsky (approved):

PR #427, opened 2020-09-22
Update the Microsoft.NET.Compilers.Toolset package to have Roslyn respect UnmanagedCallersOnly

jkoritzinsky:
Update Microsoft.NET.Compilers.Toolset to a version that has a compiler that enforces the calling convention match rules for UnmanagedCallersOnly.

Includes a few fixes for INPC/INCC that were caught by this validation.


WinRT.Runtime/Projections/INotifyCollectionChanged.cs
```diff
@@ -52,14 +52,29 @@ static unsafe Vftbl()
...
-                token = default;
+                fixed (global::WinRT.EventRegistrationToken* ptoken = &token)
+                {
+                    return Do_Abi_add_CollectionChanged_0(thisPtr, handler, ptoken);
+                }
+            }
+#endif
+
+#if !NETSTANDARD2_0
+            [UnmanagedCallersOnly]
```
Scottj1s, line 68: our code-gen now puts this out:
	            [UnmanagedCallersOnly(CallConvs = new [] { typeof(CallConvStdcall) })]

any concern with the custom mappings not doing likewise?

jkoritzinsky: Nope. The manual type mappings have either been updated to handle this correctly, use the Unmanaged calling convention on .NET 5 (which maps to stdcall on Windows), or are generic and as a result still use delegates.

Scottj1s (approved):
thanks!

PR #432, opened 2020-09-23
Upgrade Authoring source generator to use new Roslyn API.

jkoritzinsky:
In #427 we updated the Roslyn toolset package for the repo.

Between the version we were using and the new version, the source generator API was updated. This PR updates our source generator to build against the new API.


manodasanW (approved):

manodasanW:
Thanks, just ran into this when I synced master.

PR #433, opened 2020-09-24
IXamlServiceProvider projection

### LINKED ISSUE #358: Project WinRT IXamlServiceProvider interface to System.IServiceProvider
See the following new Xaml API, which is designed to match System.IServiceProvider: https://github.com/microsoft/microsoft-ui-xaml-specs/pull/84

In a cs/winrt projection, this should show up as System.IServiceProvider, analogous to the special projection of INotifyPropertyChanged from Microsoft.UI.Xaml.Data to System.ComponentModel.

Scottj1s:
fixes #358


TestComponentCSharp/Class.cpp
```diff
@@ -1225,6 +1226,26 @@ namespace winrt::TestComponentCSharp::implementation
         return winrt::make<native_properties1>();
     }
 
+    // TODO: when the public WinUI nuget supports IXamlServiceProvider, just use the projection
+    struct __declspec(uuid("68B3A2DF-8173-539F-B524-C8A2348F5AFB")) IServiceProviderInterop : ::IInspectable
```
manodasanW, line 1230: Does ::IInspectable not work throughout.  Do we need WF:: in the other places?

manodasanW (approved):

TestComponentCSharp/Class.cpp
Scottj1s: I had to #include <inspectable.h>, as a temporary measure to mock winrt::implements<IXamlServiceProvider>, until the public WinUI winmds have that type.  That in turn brought an ambiguous definition of IInspectable into the global namespace, colliding with the C++/WinrT version, so I had to qualify the latter.

PR #439, opened 2020-09-24
transform use of await to use Tasks and waits

j0shuams:
This fixes an issue the test runner had with running async test methods


Scottj1s (approved):
you could keep the awaits by just making each [Fact] an invoker around an async test body (like the existing async tests)

Scottj1s (approved):

PR #441, opened 2020-09-28
always create global.json and prerelease targets files on every build

Scottj1s:
this makes it easier to update .NET SDK, or to recreate older builds


build.cmd
```diff
@@ -20,16 +20,13 @@ powershell -NoProfile -ExecutionPolicy unrestricted -Command ^
 -AzureFeed 'https://dotnetcli.blob.core.windows.net/dotnet' "
 
 :globaljson
-rem User expected to provide global.json with allowPrerelease=true
-if not exist %~dp0global.json (
```
j0shuams, line 24: Why was there a condition originally?

Scottj1s: to give devs the flexibility of moving up to a newer .NET SDK before the build did - but it's proved more trouble, as it basically locks in the current version when the build itself moves up

j0shuams (approved):

PR #444, opened 2020-09-28
Fixed broken tests, bad merge, regression in RC guid calculation (ProjectedRuntimeClass attrib), and made vector view GRCN more reliable

Scottj1s:


UnitTest/TestComponentCSharp_Tests.cs
```diff
@@ -60,7 +60,7 @@ async Task InvokeWriteBufferAsync()
             await stream.WriteAsync(buffer);
         }
 
-        [Fact (Skip = "flakey")]
+        [Fact]
```
Scottj1s, line 63: @j0shuams is confirming whether this test needed to be disabled (shouldn't have - has been passing for a while)

j0shuams: Per builds, I think this was unnecessarily disabled. This change is good with me.

UnitTest/TestComponentCSharp_Tests.net5.cs
```diff
@@ -18,7 +18,7 @@ public TestCSharpNet5()
             TestObject = new Class();
         }
 
-        [Fact (Skip = "flakey")]
+        [Fact(Skip = "flakey")]
```
j0shuams, line 21: @Scottj1s did I misunderstand that these tests were passing now?

cswinrt/code_writers.h
```diff
@@ -2313,7 +2313,20 @@ event % %;)",
                 set_simple_marshaler_type(m, type);
                 break;
             case category::interface_type:
-                m.marshaler_type = "MarshalInterface<" + m.param_type + ">";
+                if (settings.netstandard_compat) {
+                    if (get_mapped_type(type.TypeNamespace(), type.TypeName()) &&
+                        type.TypeNamespace() == "Windows.Foundation.Collections")
```
BenJKuhn, line 2318: why does this not need to set m.marshaler_type, like the rest of the cases?

cswinrt/code_writers.h
```diff
@@ -3517,7 +3530,21 @@ remove => %.Unsubscribe(value);
                     }
                     break;
                 case category::interface_type:
```
BenJKuhn, line 3533: indenting issue? Inconsistent with previous case statement.

j0shuams (approved):

BenJKuhn (approved):

Scottj1s:
thanks!!!

PR #446, opened 2020-09-29
Implement GCPressure

Scottj1s:
for some reason, rebasing caused the [old PR](https://github.com/microsoft/CsWinRT/pull/288) to get confused.  recreating


Scottj1s:
fixes #282

Scottj1s:
can no longer repro original test failures

cswinrt/code_writers.h
```diff
@@ -4714,6 +4714,15 @@ wrapped_type_name);
         auto base_semantics = get_type_semantics(type.Extends());
         auto derived_new = std::holds_alternative<object_type>(base_semantics) ? "" : "new ";
 
+        auto gc_pressure_amount = 0;
```
manodasanW, line 4717: do we want same logic in write_class?  note this is the netstandard variant of it.

Scottj1s: ah, rebase surprise, thx

Scottj1s: and now it's crashing test cases again - at least things are consistent

Scottj1s:
fixes #584

Scottj1s:
@manodasanW please advise on how to deal with API changes.

Scottj1s:
After discussing with @AaronRobinsonMSFT, could implement a new ComWrappers.HandsOffWrapper API in .NET 6, but that's a year from now, and handling disposed objects on the fly by resurrecting them enables better perf by reusing wrappers.

UnitTest/TestComponent_Tests.cs
```diff
@@ -736,15 +736,7 @@ public void Collections_ReadOnly_List_Call()
...
-        // Nota Bene: this test case must always remain the final one
-        [Fact]
-        public void Z_Check_Coverage()
-        {
-            Tests.Simple();
-            //Assert.Equal((double)Tests.Percentage, (double)100);
-        }
-
-        [Fact]
+        [Fact(Skip = "System.Runtime.InteropServices.COMException : Class not registered")]
```
Scottj1s, line 739: disabling for now to unblock

ujjwalchadha (approved):

WinRT.Runtime/ApiCompatBaseline.net5.0.txt
```diff
@@ -1 +1,3 @@
-Total Issues: 0
```
manodasanW, line 1: I was initially a bit confused why this was considered a breaking API compat change as the change is adding a default interface method which I would assume should be fine from a compat perspective.  From looking into it, the rules for default interface methods in the api compat tooling are not enabled by default due to it seems like it might not be fully supported by VB and might affect consumers with LangVersion before 8.

I don't think either of those should affect us and due to that, I think we can enable the optional rules for default interface members by adding the below to our winrt.runtime.csproj and then this shouldn't get flagged anymore.

```xml
    <ApiCompatArgs>$(ApiCompatArgs) --allow-default-interface-methods</ApiCompatArgs>
    <MatchingRefApiCompatArgs>$(MatchingRefApiCompatArgs) --allow-default-interface-methods</MatchingRefApiCompatArgs>
```

WinRT.Runtime/MatchingRefApiCompatBaseline.net5.0.txt
```diff
@@ -1,3 +1,6 @@
 Compat issues with assembly WinRT.Runtime:
 TypesMustExist : Type 'System.Numerics.VectorExtensions' does not exist in the reference but it does exist in the implementation.
-Total Issues: 1
+MembersMustExist : Member 'public System.Boolean WinRT.IObjectReference.Resurrect()' does not exist in the reference but it does exist in the implementation.
```
manodasanW, line 3: These will require an assembly version bump, but I wonder if we can avoid it by making the APIs internal?  Any reason they should be public?

UnitTest/TestComponent_Tests.cs
Scottj1s: re-enabled with rebase

WinRT.Runtime/ObjectReference.cs
```diff
@@ -136,6 +136,18 @@ protected virtual void Dispose(bool disposing)
             disposed = true;
         }
 
+        public bool Resurrect()
```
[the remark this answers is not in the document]
Scottj1s: good call - will add locking

WinRT.Runtime/ApiCompatBaseline.net5.0.txt
Scottj1s: that works - thx

WinRT.Runtime/MatchingRefApiCompatBaseline.net5.0.txt
Scottj1s: good call - done

manodasanW (approved):

PR #450, opened 2020-10-01
User/jlarkin/fixasynctests

j0shuams:
The Readtoend async unit test has been causing problems for our build pipeline, and I think it is because of it reaches into the file system. This test is really just covering that AsBuffer() works now, and there is [another test explicitly covering AsBuffer](https://github.com/microsoft/CsWinRT/blob/79631971fe907543a4b89f2022c8e81cae1635c8/UnitTest/TestComponentCSharp_Tests.cs#L59). 

I propose we cut the ReadToEnd test. I ran the pipeline twice with this change and got green builds.


Scottj1s:
I wanted to wait until we got a bit more clarity on xunit parallel execution issues - these tests fall into a new test class that could be racing against TestWinrt, TestCSharp.  But looking at the pipeline logs, I still don't see any proof that this is a parallel race, so go ahead and remove the redundant test.

Scottj1s (approved):

PR #455, opened 2020-10-05
Angzhang/breakingchange

angelazhangmsft:
list of breaking changes for RC2, linked from the readme Developer Guidance section


README.md
```diff
@@ -52,12 +52,16 @@ After a successful command-line build, the cswinrt.sln can be launched from the
 
 **Note:**  By default, projection projects only generate source files for Release configurations, where cswinrt.exe can execute in seconds.  To generate projection sources for Debug configurations, set the project property GenerateTestProjection to 'true'.  In either case, existing projection sources under the "Generated Files" folder will still be compiled into the projection assembly.  This configuration permits a faster inner loop in Visual Studio.
 
-**Note:** When building the first time, the build clones the testwinrt project. The build depends on this project, but msbuild doesn't consistently pick up the change when the repo is first cloned repo. You may need to build a second time if this is your first build to address the timing issue.
+**Note:** When building the first time, the build clones the testwinrt project. The build depends on this project, but msbuild doesn't consistently pick up the change when the repo is first cloned. You may need to build a second time if this is your first build to address the timing issue.
```
Scottj1s, line 55: "msbuild doesn't consistently ..." 
I haven't seen anything like this.  As long as build.cmd is used, it shouldn't be a problem - was that the issue?

docs/breakingchanges-rc2.md
```diff
@@ -0,0 +1,15 @@
+# Breaking Changes for C#/WinRT in RC2
+
+1.  With IDynamicInterfaceCastable support, all projected interfaces can now use a simple C#-style cast.  Unprojected interfaces must use the `.As<T>()` extension method.
+
+2.  Calling any of the [RuntimeReflectionExtension](https://docs.microsoft.com/dotnet/api/system.reflection.runtimereflectionextensions?view=net-5.0) methods is likely to cause issues. The runtime type of a projected class now carries around some state information and helper methods, and should not be dynamically reflected on.
+
+3. Some WinRT types may have differences in default values when projected. This may raise issues if relying on default values.
```
Scottj1s, line 7: should we be specific here (e.g., DataTime)

docs/breakingchanges-rc2.md
```diff
@@ -0,0 +1,15 @@
...
+
+4. The latest Roslyn compiler is required for C# projects, which come with the following NuGet package:
+
+    ```xml
+    <PackageReference Include="Microsoft.Net.Compilers.Toolset" Version="3.8.0-4.20472.6"
+    ```
+
+5. Because of strong-name signing, you cannot combine pre-RC2 and post-RC2 projection assemblies. For example, with WinUI and the Windows SDK, both would need to move up to latest (RC2) version together.
```
Scottj1s, line 15: "to the latest"

docs/breakingchanges-rc2.md
```diff
@@ -0,0 +1,15 @@
+# Breaking Changes for C#/WinRT in RC2
```
Scottj1s, line 1: Did you want to mention the 'add' event accessor changing return type as well?  (from EventRegistrationToken to void)

Scottj1s:
@stevenbrix Anything to add, since you've been deep in the weeds on this stuff - main concern is message out the function pointer syntax change coordination (roslyn compiler etc - do we have that right?)

stevenbrix:
Overall it looks good. I'm not quite sure who the audience is, but would it be worth mentioning that the latest cswinrt+.net isn't compatible with WinUI Preview 2?

README.md
Scottj1s: ok, let's remove it

docs/breakingchanges-rc2.md
```diff
@@ -0,0 +1,19 @@
...
+
+3. Some WinRT types may have differences in default values when projected. This may raise issues if relying on default values. For example, `Windows.Foundation.DateTime` is projected to `System.DateTimeOffset` in C# and these types have different default values.
+
+4. The latest Roslyn compiler is required for C# projects, which come with the following NuGet package:
+
+    ```xml
+    <PackageReference Include="Microsoft.Net.Compilers.Toolset" Version="3.8.0-4.20472.6"
+    ```
+
+5. Because of strong-name signing, you cannot combine pre-RC2 and post-RC2 projection assemblies. For example, with WinUI and the Windows SDK, both would need to move up to the latest (RC2) version together.
```
Scottj1s, line 15: The latest version of C#/WinRT, supporting .NET 5 RC2, has added strong-name signing to the runtime assembly, winrt.runtime.dll.  Assemblies built with the previous version of winrt.runtime.dll are not compatible with this change. 
 This includes Microsoft.Windows.SDK.Net.dll (the Windows SDK), Microsoft.WinUI.dll (WinUI), and any other C#/WinRT projection assemblies.  Builds that combine pre-RC2 and post-RC2 projection assemblies may fail with a CS0012, and mention of a missing reference to an assembly with PublicIKeyToken=null.  This indicates an older assembly without strong name signing.  The solution is to update all dependencies to their .NET 5 RC2 versions.

Scottj1s (approved):

PR #464, opened 2020-10-07
propagate assignment of _inner to base classes

### LINKED ISSUE #460: Derived types not resolving ThisPtr polymorphically
Discovered when porting XAML Islands on WPF tests to .NET 5

Repro:
```
void Test(UIElement e)
{
  // if e points to a derived type, like Microsoft.UI.Xaml.Controls.Grid, 
  // then the operator== invoked on UIElement consults the _inner.ThisPtr,
  // and when _inner is null, causes a NullReferenceException
  if(e != null)
  {
    // ...
  }
}
```

Solution:
The ThisPtr should be polymorphic, resolving to the concrete type (and it's _inner)

This also points out the need for test coverage of derived types

Scottj1s:
fixes #460


cswinrt/code_writers.h
```diff
@@ -4925,6 +4936,25 @@ IObjectReference IWinRTObject.NativeObject => _inner;)");
                 w.write(R"(
 global::System.Collections.Concurrent.ConcurrentDictionary<global::System.RuntimeTypeHandle, IObjectReference> IWinRTObject.QueryInterfaceCache { get; } = new();)");
                 }
+
```
manodasanW, line 4939: Couple lines above this, I find myself wondering whether _defaultLazy = new Lazy<%>(() => (%)new IInspectable(((IWinRTObject)this).NativeObject)); should be using _inner or not, what do you think?

Scottj1s: all references to _inner should be safe, with SetInner initializing them from the most derived ctor, so should be fine

manodasanW: I agree with that, I was referring to the use of NativeObject in the default lazy initialization as it comes from IWinRTObject and since all classes implement the interface, I think the one from the child will be used.  The cast after probably makes it the right default interface type, but was wondering if we should just give it _inner if it is now correct.

Scottj1s: oh, was that another workaround? I missed it. Yes will change that to _inner

manodasanW: i am not sure if it was a workaround, it was part of the IDIC pr itself.

Scottj1s: also the one on line 4621

Issue #437, opened 2020-09-24
Performance of TryUnwrapObject can be improved by caching attributes.


bartekk8:
When looking into XAML's Dependency Property performance, I noticed that there's a significant overhead coming from using reflection to get custom attributes in WinRT.ComWrappersSupport.TryUnwrapObject.  If these were cached, it could potentially reduce time spent in this method.  Stack info below:

![image](https://user-images.githubusercontent.com/59668723/94179562-89725900-fe51-11ea-9ffa-b33ab756bab8.png)

jkoritzinsky:
We may be able to improve this scenario for .NET 5 by not even having to look at custom attributes if we enhance the `IWinRTObject` interface that is being added in #369.

AdamBraden:
Leaving for .NET5 GA to consider Jeremy's suggestion.

PR #470, opened 2020-10-09
Fixing scenarios where default interface isn't passed for custom projections and perf improvements

### LINKED ISSUE #360: MarshalInterface.FromAbi doesn't handle case where ptr isn't the default interface
MarshalInterface.FromAbi calls MarshalInspectable.FromAbi which checks the runtimeclass and calls FromAbi on it which assumes the ptr is for the default interface.  But it can be for another interface such as IStringable.  This then ends up calling the wrong function in some cases as vtable is wrong.

Array_Stringable_Call test will repro the issue if you undo the temporary fix put in for that instance https://github.com/microsoft/CsWinRT/blob/631bb364e79139c33296e47a8622052ab0b66259/WinRT.Runtime/Projections/Uri.cs#L134

ObjectReference<IUnknownVftbl>.FromAbi(ptr).As<ABI.Windows.Foundation.IUriRuntimeClassVftbl>(); -> ObjectReference<ABI.Windows.Foundation.IUriRuntimeClassVftbl>.FromAbi(ptr);

Ensure that fixes for these don't regress:
- [ ] #410 
- [ ] #362

manodasanW:
- If the default interface wasn't passed for custom projections, like IStringable for URI, you can end up in crashes or unexpected results.  This PR addresses it by making MarshalInspectable generic and consulting the type information statically known from the caller to make a decision on whether to use the runtimeclass name reported by the object or the statically known type.
- If the statically known type is implemented by the runtimeclass, the runtimeclass is still used to allow conversion to class types which would otherwise be not possible.  But if it isn't or the runtimeclass name is not a valid type, the static type is used instead.
- As part of making MarshalInspectable generic and passing the type information to CreateObject, we are also able to reduce QIs for scenarios where the runtimeclass name is misreported like DataPackage / DataPackageView.
- Added a cache for the custom attributes retrieved when unwrapping objects as getting the custom attribute was found to be expensive and the result was going to be the same each for time for a specific type.  This can probably be further improved for .NET 5 as mentioned in the issue, but it got tricky with composable types, so putting this fix out for now.

Resolves #360 
Contributes to #437

### DIFF
--- a/UnitTest/TestComponentCSharp_Tests.cs
+++ b/UnitTest/TestComponentCSharp_Tests.cs
@@ -22,6 +22,8 @@ using System.Collections;
 using WinRT.Interop;
 using System.Runtime.InteropServices;
 using System.Runtime.InteropServices.WindowsRuntime;
+using Windows.Security.Cryptography;
+using Windows.Security.Cryptography.Core;
 
 namespace UnitTest
 {
@@ -1780,6 +1782,16 @@ namespace UnitTest
             TestObject.CopyProperties(nativeProperties);
 
             Assert.Equal(TestObject.ReadWriteProperty, nativeProperties.ReadWriteProperty);
+        }
+
+        [Fact]
+        public void TestNonProjectedRuntimeClass()
+        {
+            string key = ".....";
+            IBuffer keyMaterial = CryptographicBuffer.ConvertStringToBinary(key, BinaryStringEncoding.Utf8);
+            MacAlgorithmProvider mac = MacAlgorithmProvider.OpenAlgorithm(MacAlgorithmNames.HmacSha1);
+            CryptographicKey cryptoKey = mac.CreateKey(keyMaterial);
+            Assert.NotNull(cryptoKey);
         }
     }
 }
--- a/UnitTest/TestComponentCSharp_Tests.cs
+++ b/UnitTest/TestComponentCSharp_Tests.cs
@@ -87,9 +87,13 @@ namespace UnitTest
                 (object sender, Uri value) => Assert.Equal(managedUri, value);
             TestObject.RaiseUriChanged();
 
-            var uri2 = MarshalInspectable.FromAbi(ABI.System.Uri.FromManaged(managedUri));
+            var uri2 = MarshalInspectable<System.Uri>.FromAbi(ABI.System.Uri.FromManaged(managedUri));
             var str2 = uri2.ToString();
             Assert.Equal(full_uri, str2);
+
+            var uri3 = MarshalInspectable<object>.FromAbi(ABI.System.Uri.FromManaged(managedUri));
+            var str3 = uri3.ToString();
+            Assert.Equal(full_uri, str3);
         }
 
         [Fact]
@@ -1615,9 +1619,13 @@ namespace UnitTest
         [Fact]
         public void TestUnwrapInspectable()
         {
-            using var objRef = MarshalInspectable.CreateMarshaler(TestObject);
+            using var objRef = MarshalInspectable<object>.CreateMarshaler(TestObject);
             var inspectable = IInspectable.FromAbi(objRef.ThisPtr);
             Assert.True(ComWrappersSupport.TryUnwrapObject(inspectable, out _));
+
+            using var objRef2 = MarshalInspectable<Class>.CreateMarshaler(TestObject);
+            var inspectable2 = IInspectable.FromAbi(objRef2.ThisPtr);
+            Assert.True(ComWrappersSupport.TryUnwrapObject(inspectable2, out _));
         }
 
         [Fact]
--- a/UnitTest/TestComponentCSharp_Tests.cs
+++ b/UnitTest/TestComponentCSharp_Tests.cs
@@ -1784,6 +1784,7 @@ namespace UnitTest
             Assert.Equal(TestObject.ReadWriteProperty, nativeProperties.ReadWriteProperty);
         }
 
+        // Test scenario where type reported by runtimeclass name is not a valid type (i.e. internal type).
         [Fact]
         public void TestNonProjectedRuntimeClass()
         {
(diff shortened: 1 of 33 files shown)


UnitTest/TestComponentCSharp_Tests.cs
```diff
@@ -1773,5 +1783,15 @@ public void TestInterfaceObjectMarshalling()
 
             Assert.Equal(TestObject.ReadWriteProperty, nativeProperties.ReadWriteProperty);
         }
+
+        [Fact]
```
Scottj1s, line 1787: what is this testing, related to generic MarshalInspectable?  can you add a comment?

WinRT.Runtime/ComWrappersSupport.cs
```diff
@@ -328,6 +342,44 @@ private static bool IsIReferenceArray(Type implementationType)
             return CreateFactoryForImplementationType(runtimeClassName, implementationType);
         }
 
+        internal static string GetRuntimeClassForTypeCreation(IInspectable inspectable, Type staticallyDeterminedType)
+        {
+            string runtimeClassName = inspectable.GetRuntimeClassName(noThrow: true);
+            if (staticallyDeterminedType != null && staticallyDeterminedType != typeof(object))
+            {
+                // We have a static type which we can use to construct the object.  But, we can't just it for all scenarios
```
Scottj1s, line 350: just use it

WinRT.Runtime/ComWrappersSupport.cs
```diff
@@ -28,6 +28,7 @@ public static partial class ComWrappersSupport
     {
         private readonly static ConcurrentDictionary<string, Func<IInspectable, object>> TypedObjectFactoryCache = new ConcurrentDictionary<string, Func<IInspectable, object>>();
         private readonly static ConditionalWeakTable<object, object> CCWTable = new ConditionalWeakTable<object, object>();
+        private readonly static ConcurrentDictionary<Type, MemberInfo> TypeObjectRefFieldCache = new ConcurrentDictionary<Type, MemberInfo>();
```
Scottj1s, line 31: is this Bartek's suggestion?

WinRT.Runtime/ComWrappersSupport.net5.cs
```diff
@@ -50,26 +52,32 @@ internal static unsafe InspectableInfo GetInspectableInfo(IntPtr pThis)
...
-        public static object CreateRcwForComObject(IntPtr ptr)
+        public static T CreateRcwForComObject<T>(IntPtr ptr)
         {
             if (ptr == IntPtr.Zero)
             {
-                return null;
+                return default;
             }
 
+            if (typeof(T) != typeof(object))
```
Scottj1s, line 62: why the if check?

WinRT.Runtime/Projections/Uri.cs
```diff
@@ -112,7 +112,7 @@ public static IObjectReference CreateMarshaler(global::System.Uri value)
                 return null;
             }
 
-            using var uri = ObjectReference<IUnknownVftbl>.FromAbi(ptr).As<ABI.Windows.Foundation.IUriRuntimeClassVftbl>();
+            using var uri = ObjectReference<ABI.Windows.Foundation.IUriRuntimeClassVftbl>.FromAbi(ptr);
```
Scottj1s, line 115: workaround removed!

Scottj1s (approved):
looks good - just a few minor suggestions

WinRT.Runtime/ComWrappersSupport.cs
manodasanW: This is indeed an area where bartek saw high costs (retrieving custom attributes).  And one of the approaches we discussed was doing caching here to see if we can improve performance.

jkoritzinsky: I still wonder if we can enhance `IWinRTObject` to avoid all of the reflection usage altogether (at least for the .NET 5.0 projection). That way we don't need to add another cache.

manodasanW: I agree I think we should still look at if we can enhance IWinRTObject and eliminate it for .NET 5 which is why I am not closing the issue.  I got this change out for now as I wanted to get this PR out with some improvements as it was getting large, then do that as an additive thing when I got it working.

My initial attempt was trying to just use NativeObject from IWinRTObject as the result when a bool was set which is set in the classes which have objreferencewrapper but it broke in composable scenarios where a managed type implementing a projected type also had a NativeObject set but that wouldn't have been previously used.  I need to think about how to handle that scenario.  Was there any other enhancements you were thinking we need to make to IWinRTObject to achieve it and that may help with that issue?

WinRT.Runtime/ComWrappersSupport.cs
```diff
@@ -328,6 +342,44 @@ private static bool IsIReferenceArray(Type implementationType)
             return CreateFactoryForImplementationType(runtimeClassName, implementationType);
         }
 
+        internal static string GetRuntimeClassForTypeCreation(IInspectable inspectable, Type staticallyDeterminedType)
+        {
+            string runtimeClassName = inspectable.GetRuntimeClassName(noThrow: true);
```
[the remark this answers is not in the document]
manodasanW: If it throws, we return null and caller handles it by just using the iinspecatable rather than casting it to a type.  This is what we did for the .NET 5 projection before.  Looks like .NET standard 2 projection might have thrown before in this case, but I don't think that was intentional as they are also have logic to return iinspectable but runtimeclass name is null.

WinRT.Runtime/ComWrappersSupport.net5.cs
manodasanW: Will remove as it is now redundant.  It also now check for it inside of GetRuntimeClassForTypeCreation.

WinRT.Runtime/ComWrappersSupport.cs
jkoritzinsky: One option would be to add a new `bool` property in `IWinRTObject` named `IsComposed`. By default, it would return `false`. We'd update the codegen for composable runtime classes so for a given class `Composable`, it is implemented as follows:

```csharp
bool IWinRTObject.IsComposed => this.GetType() != typeof(Composable);
```

Since each composable class would override this, calls to it would always used the most-derived implementation. Therefore, we'd correctly be able to detect if an object is composed (and if we can unwrap it).

manodasanW: That is actually a clever idea and I think will work and looks like what we do in the constructor already for another thing.  Will give it a try.

PR #476, opened 2020-10-09
net5-windows* tfm support and a few minor fixes

### LINKED ISSUE #472: Windows-specific TFMs are not supported
cswinrt.exe does an explicit check for 'net5.0' so doesn't support net5.0-Windows, net5.0-windows10.0.19041.0, etc.  also need to fix props/targets with similar checks

Scottj1s:
fixes #472


Projections/TestHost.ProbeByClass/TestHost.ProbeByClass.csproj
```diff
@@ -1,9 +1,29 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <AssemblyName>TestHost.ProbeByClass</AssemblyName>
+    <TargetFrameworks>netstandard2.0;;net5.0</TargetFrameworks>
+    <Platforms>x64;x86</Platforms>
   </PropertyGroup>
 
-  <Import Project="../Test/Test.csproj"/>
+  <ItemGroup>
```
manodasanW, line 8: Any reason why we decided to rebuild the test projection here rather than consume it like before?

Scottj1s: it was creating an msbuild warning

manodasanW (approved):

PR #478, opened 2020-10-10
ref const projection support

### LINKED ISSUE #456: `ref const`-Parameter not handled correctly
Given my previous issues with winrt related APIs, I discovered, that `ref const` parameters are _not_ handled correctly.
For following IDL
```
struct Argument
{
	Int32 Parameter;
};

interface IHandler
{
	void Handle(ref const Argument arg);
};
```
I expect the CsWinRT projection to generate an interface that respects this statement:
> A const reference parameter is used for efficient input parameter passing of large value types (but not simple types such as integral or floating-point numbers, nor enums). Methods receive such parameters as a reference to the actual value

This states that `ref const`-parameters are used for _input_ values passed _by reference_. Following interface is generated by CsWinRT `0.1.0-prerelease.200928.7`:
```
public interface IHandler
{
    void Handle(Argument arg);
}
```
The `Argument`-type is translated correctly:
```
public struct Argument: IEquatable<Argument>
{
    public int Parameter;

    public Argument(int _Parameter)
    {
        Parameter = _Parameter; 
    }

    public static bool operator ==(Argument x, Argument y) => x.Parameter == y.Parameter;
    public static bool operator !=(Argument x, Argument y) => !(x == y);
    public bool Equals(Argument other) => this == other;
    public override bool Equals(object obj) => obj is Argument that && this == that;
    public override int GetHashCode() => Parameter.GetHashCode();
}
```

The parameter should either be projected as `ref` (as .NET Core 3.1 does it) or `in` (as I expect an _input parameter_ to be handled).

Previous Issues (for sake of linking)
- [winrt-related/#162](https://github.com/MicrosoftDocs/winrt-related/issues/162)
- [winrt-related/#149](https://github.com/MicrosoftDocs/winrt-related/issues/149)

Scottj1s:
fixes #456


cswinrt/code_writers.h
```diff
@@ -3410,6 +3448,17 @@ remove => %.Unsubscribe(value);
...
                 }
+                else if (category == param_category::ref)
+                {
+                    if (!use_pointers)
+                    {
+                        format_string = "in __%";
+                    }
+                    else
+                    {
+                        format_string = use_pointers ? "*%" : "in %";
```
manodasanW, line 3459: use_pointers will never be false due to this is in the else of if !use_pointers.  And interestingly we set a different format_string between both?

Scottj1s: ah, thanks - missed that refactoring.  yes, with pointers, we can just deref and pass the param directly.  but C# doesn't permit in/ref/out in lambdas, so we have to capture (and unfortunately, lose the perf gain of pass by ref for netstandard)

manodasanW (approved):

PR #479, opened 2020-10-11
Improving perf of unwrapping object

### LINKED ISSUE #437: Performance of TryUnwrapObject can be improved by caching attributes.

manodasanW:
Improving perf of unwrapping objects by removing the use of reflection to get custom attributes to check for an ObjectReferenceWrapper and rather using IWinRTObject in NET 5 to get the same object.

- Everywhere we had an ObjectReferenceWrapper attribute and had implemented IWinRTObject, we now have a bool which indicates it has a native object to unwrap.
- The only exception to this is composable types which check whether the type has been composed / extended by a custom C# type and if not, it returns true and otherwise false which matches the previous behavior.
- For places which didn't have an ObjectReferenceWrapper attribute,  but implemented IWinRTObject,  it returns false.
- There are couple of custom projected ABI types which have the ObjectReferenceWrapper attribute, but don't implement IWinRTObject.  From inspection of them, they seem to project to C# system class types which means we are not probably unwrapping them back to the native object.  The ones which projected to C# system interface types were all dynamicinterfacecastable implementations other than for one which I converted.
- NativeDelegateWrapper is the other one I noticed which has an ObjectReferenceWrapper, but doesn't implement IWinRTObject.  I tried to implement a couple of tests which tried to cause them to unwrap to see whether I should implement  IWinRTObject on them (i.e. returning a delegate from a C++/WinRT component and then passing it back as a delegate and as an object).  In the first case, delegate marshaling doesn't seem to unwrap the native object at all and just creates a CCW on our managed delegate wrapper.  Is that intended or a bug?  And in the second case where I tried to pass the delegate as an object, it did unwrap before, but it then crashes as C++/WinRT delegates doesn't implement IInspectable apparently.  With my changes as present, it no longer unwraps the delegate, but creates a CCW on the wrapper which happens to succeed on the IInspectable QI and due to that doesn't crash,  This is unintentionally parity with delegate marshaling, but if  we expect it should be unwrapped, we should make NativeDelegateWrapper implement IWinRTObject and update delegate marshaling to also do unwrapping if any.  Thoughts?

Fixes #437


WinRT.Runtime/ComWrappersSupport.net5.cs
```diff
@@ -80,8 +80,29 @@ public static T CreateRcwForComObject<T>(IntPtr ptr)
                 ABI.System.Nullable<Type> nt => (T)(object) nt.Value,
                 _ => (T) rcw
             };
-        }
-
+        }
+
+        public static bool TryUnwrapObject(object o, out IObjectReference objRef)
+        {
+            // The unwrapping here needs to be in exact type match in case the user
```
Scottj1s, line 87: "an exact"

WinRT.Runtime/ComWrappersSupport.netstandard2.0.cs
```diff
@@ -75,8 +76,53 @@ public static T CreateRcwForComObject<T>(IntPtr ptr)
...
+                if (projectedClass is object && projectedClass.DefaultInterfaceProperty != null)
+                {
+                    return type.GetProperty(projectedClass.DefaultInterfaceProperty, BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.DeclaredOnly);
+                }
+
+                return null;
+            });
+
+
+            if (objRefField is FieldInfo field)
```
Scottj1s, line 112: you could avoid the type checks by keeping a cache of System.Funcs (factories) that take in o and return objRef

Scottj1s (changes requested):
change to TypeObjectRefAccesorCache = new ConcurrentDictionary<Type, Func<object, IObjectReference>>

Scottj1s:
> * should we make NativeDelegateWrapper implement IWinRTObject and update delegate marshaling to also do unwrapping

my naive answer is "yes" - any downsides?

WinRT.Runtime/ComWrappersSupport.netstandard2.0.cs
Scottj1s: I guess it was this way to start with - but would still help perf a bit

manodasanW: Yea I already have the changes to address this feedback.  Working on fixing delegate marshaling.

Scottj1s (approved):

PR #484, opened 2020-10-12
Fix system.type's incorrect data when projecting to itself

ujjwalchadha:


TestComponentCSharp/Class.cpp
```diff
@@ -157,6 +157,40 @@ namespace winrt::TestComponentCSharp::implementation
...
+    {
+        return _typeProperty;
+    }
+
+    winrt::hstring Class::GetTypePropertyAbiName()
+    {
+        return _typeProperty.Name;
+    }
+
+    winrt::hstring Class::GetTypePropertyKind()
```
Scottj1s, line 176: note: can just return wchar_t* literals

Scottj1s (approved):

PR #487, opened 2020-10-13
Remove WinRT.WeakReference for .NET 5

### LINKED ISSUE #377: remove WinRT.WeakReference
Now that the .NET runtime supports `IWeakReference`, the `WinRT.WeakReference` type should be removed and we should ensure any docs remove references of the necessary workaround.

manodasanW:
- For .NET 5, we no longer need our own weak reference and can use system.weakreference
- For .NET standard 2.0, I kept it as the test which tried to rehydrate from the native weak reference rather than managed was failing (WeakReferenceOfNativeObjectRehydratedAfterWrapperIsCollected).  It was passing with .NET 5.  I am assuming this is new functionality handled by the .NET 5 version.  @jkoritzinsky? 
- Xaml still has a dependency on it so to keep our repo building I put a copy there until we get a version without the dependency.

Fixes #377


UnitTest/TestComponentCSharp_Tests.cs
```diff
@@ -1594,27 +1600,27 @@ public void CCWOfListOfManagedType()
         public void WeakReferenceOfManagedObject()
         {
             var properties = new ManagedProperties(42);
-            WinRT.WeakReference<IProperties1> weakReference = new WinRT.WeakReference<IProperties1>(properties);
-            weakReference.TryGetTarget(out var propertiesStrong);
+            WeakRefNS::WeakReference<IProperties1> weakReference = new WeakRefNS::WeakReference<IProperties1>(properties);
```
jkoritzinsky, line 1603: Are you sure this is supposed to be a double-colon instead of a period separator here?

jkoritzinsky (commented):
Yes, the rehydrating of a weak reference for C#/WinRT was added in .NET 5 along with ComWrappers. We'll need to keep the WinRT.WeakReference type for .NET Standard 2.0 since the runtime doesn't know about the .NET Standard 2.0 wrappers.

One question, other than that looks good.

UnitTest/TestComponentCSharp_Tests.cs
manodasanW: Both do seem to build, but I flipped it as . seems to be more common.

jkoritzinsky (approved):

Scottj1s (approved):

PR #492, opened 2020-10-14
Don't generate a Vftbl nested type for non-generic interfaces (.NET 5 only)

jkoritzinsky:
To help reduce metadata size, this PR changes non-generic interfaces to not emit a Vftbl nested type and instead cast and dereference from `ThisPtr` directly and directly set entries on the `AbiToProjectionVtablePtr` vtable.

This reduces the Windows SDK projection from 28168 kb to 24340 kb (13% reduction) and reduces the WinUI projection from 12624 kb to 10713 kb (15% reduction).

In addition, since this PR adds first class support for types without Vftbl fields, it makes WinRT.Runtime a bit more resilient for linking.

It also reduces a bit of reflection in IWinRTObject and SingleInterfaceOptimizedObject.

Since this PR doesn't mandate when a Vftbl type is available, it is backward compatible with projections that generate Vftbl types on non-generic interfaces (and intermixing between ones that do and ones that don't).


Scottj1s:
This is great, Jeremy - thx!  We'll want to build private cswinrt, Windows & WinUI projections to ensure no regressions before including in RTM.

jkoritzinsky:
Awesome! Can you build the privates and share them out to the right people?

jkoritzinsky:
cc: @AaronRobinsonMSFT

Scottj1s:
> Awesome! Can you build the privates and share them out to the right people?

Yes, we're independently validating with Xaml Control Gallery, on private builds of cswinrt, Windows projection, WinUI projection

Scottj1s (approved):

Scottj1s:
validation looks good!

Issue #334, opened 2020-06-27
WindowsRuntimeBufferExtensions.ToArray incorrectly checks Capacity


AaronRobinsonMSFT:
**Issue moved from dotnet/runtime - see https://github.com/dotnet/runtime/issues/23383**

The following code will throw an ArgumentException:

```c#
var buffer = new Windows.Storage.Streams.Buffer(0);
var array = buffer.ToArray();
```

WindowsRuntimeBufferExtensions.ToArray checks the buffer capacity as a precondition, which isn't used in the method.  Obviously the expected behavior would be to return an array of length zero.

I also can't find a unit test for this case.

weltkante:
For what its worth this edge case is also present in other methods of `WindowsRuntimeBufferExtensions`. (Count is allowed to be zero but then the index-check against the capacity is off by one to actually let the code through. This allows a count of zero on every position except the last, which is odd.)

j0shuams:
[This ](https://github.com/dotnet/runtime/issues/23383#issuecomment-446370673) comment suggests making the condition in ToArray a strict greater-than.

j0shuams:
The other methods this check happens in is CopyTo. I tried the following code, with no change to the exception condition:
```
var buffer = new Windows.Storage.Streams.Buffer(0);
byte[] bytes = { };
buffer.CopyTo(bytes);
```

And there was no error, so I think changing the condition there is unnecessary.

weltkante:
The fix was incomplete, there are various extension methods in WindowsRuntimeBufferExtensions which still throw incorrectly.

j0shuams:
@weltkante Can you be more descriptive and/or give examples? Of the remaining public methods: AsBuffer, CopyTo and GetByte, I do not see a situation like you describe of count being zero causing an index check against capacity to incorrectly throw.

weltkante:
The cause of the false checks are incorrect `<=` checks - in particular of the style `if (source.Length <= sourceIndex) throw new ArgumentException(...);`

Source index can be at the end of the buffer when count is zero, see my [comment above](https://github.com/microsoft/CsWinRT/issues/334#issuecomment-659266068), so many of the extension methods incorrectly throw in this case even though the do not throw if length is zero at any place except the end of buffer.

j0shuams:
@weltkante thanks for your reply, do you think changing to `<` is enough or should the `<=` be anded (`&&`) with `count != 0` ?

weltkante:
[...]

Issue #497, opened 2020-10-14
WindowsRuntimeBufferExtensions.CopyTo throwing InvalidCast 


j0shuams:
The following code causes an Invalid Cast Exception: invalid cast from 'WinRT.IInspectable' to 'Windows.Storage.Streams.IBufferByteAccess'

```
var buffer = new Windows.Storage.Streams.Buffer(2); 
var array  = buffer.ToArray(1,1);
```

Call stack:
    IObjectReference.AsInterface\[TInterface\]() line 80
    CastExtensions.As[TInterface](Object value) line 45
    WindowsRuntimeBufferExtensions.GetPointerAtOffset(IBuffer buffer, UInt32 offset) line 8136
    WindowsRuntimeBufferExtensions.CopyTo(IBuffer source, UInt32 sourceIndex, Byte[] destination, Int32 destinationIndex, Int32 count) line 7834
    WindowsRuntimeBufferExtensions.ToArray(IBuffer source, UInt32 sourceIndex, Int32 count) line 7796

j0shuams:
PR #504  introduces tests that are blocked by this issue. I've attributed "Skip" to these tests, and the PR that closes this issue should include removing the "Skip" attribute.

PR #502, opened 2020-10-15
Add Buffer Length Check in ToArray and CopyTo

j0shuams:
In WindowsRuntimeBufferExtension, the existing implementation of ToArray does not allow for creating an empty array, i.e. one of length 0. Yet this behavior is not improper and in certain cases is desirable. We are updating the method to handle this special case. Similarly, the CopyTo method of the same class should support copying to an empty array, from an empty buffer -- for certain special cases. 

This PR...
* updates the ToArray and CopyTo methods to explicitly check if the capacity of the source Buffer is 0
* includes unit tests that check for these edge cases
* moves unit tests out of the net5 specific test file, into the base test file by using the NET5_0 directive
* closes https://github.com/microsoft/CsWinRT/issues/334


cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBufferExtensions.cs
```diff
@@ -110,6 +110,7 @@ public static void CopyTo(this byte[] source, int sourceIndex, IBuffer destinati
         public static byte[] ToArray(this IBuffer source)
         {
             if (source == null) throw new ArgumentNullException(nameof(source));
+            if (source.Capacity == 0) return Array.Empty<byte>();
```
manodasanW, line 113: capacity or length?

j0shuams: Buffers can only be constructed with a capacity, so I think the idea is that empty buffers should become empty arrays...

manodasanW: I would assume a buffer with a capacity of 2 would result in a ToArray of 0 length.

[This](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.windowsruntime.windowsruntimebufferextensions.toarray?view=dotnet-plat-ext-3.1#System_Runtime_InteropServices_WindowsRuntime_WindowsRuntimeBufferExtensions_ToArray_Windows_Storage_Streams_IBuffer_) was the previous documentation for this API which also indicates similarly.

cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBufferExtensions.cs
```diff
@@ -139,13 +141,15 @@ public static void CopyTo(this IBuffer source, byte[] destination)
...
             if (destination == null) throw new ArgumentNullException(nameof(destination));
+            if (source.Length == 0 && destination.Length == 0) return;
 
             CopyTo(source, 0, destination, 0, checked((int)source.Length));
         }
 
 
         public static void CopyTo(this IBuffer source, uint sourceIndex, byte[] destination, int destinationIndex, int count)
         {
+            if (source.Length == 0 && destination.Length == 0) return;
```
manodasanW, line 152: You should move this after the null checks, otherwise we can get null pointer here.

cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBufferExtensions.cs
```diff
@@ -118,6 +119,7 @@ public static byte[] ToArray(this IBuffer source)
         public static byte[] ToArray(this IBuffer source, uint sourceIndex, int count)
         {
             if (source == null) throw new ArgumentNullException(nameof(source));
+            if (source.Length == 0) return Array.Empty<byte>();
```
manodasanW, line 122: I feel like this scenario should technically error.  We are told to convert to an array with the array starting with data from source index to a length of count.   In the length == 0 case and count != 0, we return a empty array which might unexpected to the caller and maybe even against what the API was documented as.

We should probably do source.length == 0 and count == 0, return empty.  otherwise throw exception like before.

cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBufferExtensions.cs
```diff
@@ -139,13 +141,15 @@ public static void CopyTo(this IBuffer source, byte[] destination)
         {
             if (source == null) throw new ArgumentNullException(nameof(source));
             if (destination == null) throw new ArgumentNullException(nameof(destination));
+            if (source.Length == 0 && destination.Length == 0) return;
```
manodasanW, line 144: I think we probably only need to check source.length==0 here.

cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBufferExtensions.cs
manodasanW: We also probably should do source.length == 0 && count == 0 rather than destination.length

Scottj1s (approved):

Scottj1s:
while you're in this code, maybe also take care of #497

cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBufferExtensions.cs
```diff
@@ -140,6 +141,9 @@ public static void CopyTo(this IBuffer source, byte[] destination)
             if (source == null) throw new ArgumentNullException(nameof(source));
             if (destination == null) throw new ArgumentNullException(nameof(destination));
 
+            // If buffer and array are empty, nothing to copy
```
manodasanW, line 144: buffer is empty

manodasanW (approved):

PR #504, opened 2020-10-15
Fix capacity checks in BufferExtensions

### LINKED ISSUE #334: WindowsRuntimeBufferExtensions.ToArray incorrectly checks Capacity

j0shuams:
This PR adds support for 0 length slices of buffers and arrays for buffer extension methods. There was another issue found that showed an overflow vulnerability that has been fixed by adding a check. Unit tests have been added that verify taking 0 length slices of buffers and arrays works the same when done from the middle, beginning and end. 

I also did these same tests with length 1 slices to ensure that errors would be thrown properly in the case that the selection would be out of bounds. 

Right now most of these unit tests are have to be skipped until the fix for #497 gets merged.
I grabbed the change from that work (a new version of `IBufferByteAccess.cs`), and all tests with `Skip` attribute pass when the new version is used. I left a note on the 497 page to remove the attribute when before merging. 

Closes #334


cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBufferExtensions.cs
```diff
@@ -154,8 +150,7 @@ public static void CopyTo(this IBuffer source, uint sourceIndex, byte[] destinat
             if (destination == null) throw new ArgumentNullException(nameof(destination));
             if (count < 0) throw new ArgumentOutOfRangeException(nameof(count));
             if (destinationIndex < 0) throw new ArgumentOutOfRangeException(nameof(destinationIndex));
-            if (source.Length == 0 && count == 0) return;
-            if (source.Capacity <= sourceIndex) throw new ArgumentException(global::Windows.Storage.Streams.SR.Argument_BufferIndexExceedsCapacity);
+            if (source.Capacity < sourceIndex) throw new ArgumentException(global::Windows.Storage.Streams.SR.Argument_BufferIndexExceedsCapacity);
             if (source.Capacity - sourceIndex < count) throw new ArgumentException(global::Windows.Storage.Streams.SR.Argument_InsufficientSpaceInSourceBuffer);
             if (destination.Length <= destinationIndex) throw new ArgumentException(global::Windows.Storage.Streams.SR.Argument_IndexOutOfArrayBounds);
```
weltkante, line 155: `<=` operator is suspicious for failures when copying a zero length span to the end of the destination buffer

weltkante (commented):
besides the one I commented `GetWindowsRuntimeBuffer(this MemoryStream underlyingStream, int positionInStream, int length)` also has a `<=` operator which is supicious for throwing when retrieving a zero sized buffer at end of memory stream capacity

cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBufferExtensions.cs
```diff
@@ -85,7 +85,7 @@ public static void CopyTo(this byte[] source, int sourceIndex, IBuffer destinati
             if (destination == null) throw new ArgumentNullException(nameof(destination));
             if (count < 0) throw new ArgumentOutOfRangeException(nameof(count));
             if (sourceIndex < 0) throw new ArgumentOutOfRangeException(nameof(sourceIndex));
-            if (source.Length <= sourceIndex) throw new ArgumentException(global::Windows.Storage.Streams.SR.Argument_IndexOutOfArrayBounds, nameof(sourceIndex));
+            if (source.Length < sourceIndex) throw new ArgumentException(global::Windows.Storage.Streams.SR.Argument_IndexOutOfArrayBounds, nameof(sourceIndex));
```
j0shuams, line 88: I actually think this should be reverted -- we don't want to try and index into the array at it's length value, since arrays are 0-based.

j0shuams:
@weltkante 

After some thought, I don't think we should support indices that are equal to the length of the buffer or array, as this is technically out of memory bounds. You mentioned the invariant was `offest <= buffer.Length` but I believe the invariant is actually `offset < buffer.Length` since buffers are 0-based. The special case for the indices being equal to length are when length is 0 and count is 0, which the original PR supports. 

Please provide an example scenario / code that you believe should work that doesn't

weltkante:
> After some thought, I don't think we should support indices that are equal to the length of the buffer or array, as this is technically out of memory bounds.

You are mistaken and mixing up the concepts of element access and span access. A span of length zero is valid at the end of a buffer, for example `Insert` methods will happily allow an `index == buffer.Length` because `Insert` methods are not about element access, they take a span of (implied) length zero as argument, which is the position after the last element.

> You mentioned the invariant was offest <= buffer.Length but I believe the invariant is actually offset < buffer.Length since buffers are 0-based.

This is incorrect, `offset < buffer.Length` is the invariant for slot access, which implies `count = 1` or at least `count > 0`

The full mathematical invariant is `0 <= offset && offset <= ending && ending <= buffer.Length` which needs `<=` in all positions, I hope you can agree with that.

From this invariant the individual checks are derived, needing four checks in total instead of three since you have `offset + count` instead of `ending` and computers do overflow arithmetic. The four checks usually are
* `offset >= 0` (throw if `offset < 0`)
* `count >= 0` (throw if `count < 0`)
* `offset + count <= buffer.Length` can't easily be checked in one comparison due to overflow being possible, so its split, which also conveniently lets you assign blame to single variables instead of the sum
  * `offset <= buffer.Length` (throw if `offset > buffer.Length`)
  * `count <= buffer.Length - offset` (throw if `count > buffer.Length - offset`)
the split can be done in other ways, with `offset` and `count` swapped, or other variations

The invariant for _indexed slot access_ you have in mind is `0 <= index && index < buffer.Length` but actually is a special-case of a span access check with count one (since you are accessing a single slot). It is equivalent to `0 <= index && index + 1 <= buffer.Length` when you substitute `count = 1` in the more general span invariant. Of course nobody uses the `+ 1` version of the invariant, everyone writes `<` instead of `<=`. The inverse throwing condition becomes `index >= buffer.Length`, but it is for indexed access.

> The special case for the indices being equal to length are when length is 0 and count is 0, which the original PR supports.

You can of course solve the currently present bugs by special casing zero, but its a bit inelegant. Its not my codebase so I'm not arguing over it, if you find it easier to reason about special casing zero instead of reasoning about the invariants (which I agree can be hard if you aren't used to it) then feel free to do so. Maintainability of the codebase is more important than elegance, and the more people can easily understand it, the better.

Note though, that in the previous PR the early-exit on zero count did appear in a spot where some argument checks were skipped, i.e. when special casing zero without reordering the other checks you may accidently change the semantics and no longer throw on some invalid arguments. Not critical to me personally, since the arguments are invalid anyways, just mentioning it because its a semantic change to how it behaved previously and may prevent detecting bugs in callers.

> Please provide an example scenario / code that you believe should work that doesn't

I'll make the list whith scenarios that don't work, will take some time and should be ready later today.

j0shuams:
Thank you for your reply!

> You are mistaken and mixing up the concepts of element access and span access. A span of length zero is valid at the end of a buffer, for example Insert methods will happily allow an index == buffer.Length because Insert methods are not about element access, they take a span of (implied) length zero as argument, which is the position after the last element.

> If you do not like the Insert example, consider a general list.Replace(index, count, items) method (which replaces the given span by the given items, possibly a different amount). This API would be taking a span as input and a zero-size span at the end is perfectly valid, and in fact common, as it represents the "append" operation. If the replace operation is providing undo-capability it will want to take a snapshot of the to-be-replaced span. I hope this can serve as a reasonable example where "read zero sized span at end of buffer" happens.

This example doesn't quite make sense to me because you are using List as the example data-structure, which can grow to arbitrary size; arrays and buffers have fixed lengths. I am concerned that if we change the `<=` to `<` we now allow the possibility for someone to pass in indices beyond the the bounds when they also pass a non-zero count. Which is leading me to think we should special case for `count == 0` instead. 

> The full mathematical invariant for spans is 0 <= offset && offset <= ending && ending <= buffer.Length which needs <= in all positions, I hope you can agree with that. (Note that offset is inclusive while ending is exclusive when specifying it this way.)

If ending is exclusive shouldn't it be `ending < buffer.Length` ? 

I am writing some test cases now to exemplify my concerns.

UnitTest/TestComponentCSharp_Tests.cs
```diff
@@ -53,6 +53,150 @@ public void TestEmptyBufferToArray()
...
+            // Do something with buffer and arr
+            // CopyTo(this byte[] source, int sourceIndex, IBuffer destination, uint destinationIndex, int count)
+            arr.CopyTo(3, buf, 3, 0);
+        }
+        [Fact(Skip="errors properly")]
+        public void TestBufferCheck11()
+        {
+            IBuffer buf = new Windows.Storage.Streams.Buffer(3);
+            byte[] arr = new byte[] { 0x01, 0x02, 0x03 };
+            arr.CopyTo(3, buf, 3, 1); // errors
```
j0shuams, line 69: This case is interesting in comparison to the above, as I anticipated it to go through but an exception did get thrown.

weltkante:
> This example doesn't quite make sense to me because you are using List as the example data-structure, which can grow to arbitrary size

I think you misunderstood the example, the buffers would be an implementation detail of implementing such a list. Immutable lists and strings do not grow at all, yet they have such an API. It is common to implement file editors by having a tree of buffers, and span based APIs are the fundamental building block there. Having to code around the defect of the framework not allowing zero length spans at end of a buffer is just annoying. Its possible to easily work around, I have done so in the past, but you have to litter the callers with checks that wouldn't be necessary otherwise.

Or did you ever see anyone doing length checks when calling `string.substring` or `string.replace` ? These "just work" because zero length spans are correctly implemented. In fact there are many .NET APIs which correctly implement zero ranges at the end of their collection, e.g. spans, arrays, strings, lists, etc. all have correct range checks (as far as I'm aware) for APIs which allow to take ranges.

> If ending is exclusive shouldn't it be ending < buffer.Length ?

No, that would be if `ending` where _inclusive_ .

If you need to make yourself familiar with range indexing first, take the string API as an example, `string.substring` and `string.replace` both operate on range parameters and should have proper validation. Strings are immutable and its perfectly reasonable for someone to implement something similar on top of arrays or buffers, wanting to use these framework APIs, yet they are broken and you cannot use them without adding additional checks on every single call site.

I've created a bunch of test cases, not covering the previous PR, but instead what is still looking wrong in the current source. I've also included examples for other APIs which have correct range checking. Its really just these extensions that are broken because their range checks are wrong. Everyone else seems to have gotten it right.

<details><summary>test cases failing</summary>

```csharp
using System;
using System.IO;
using System.Linq;
using System.Runtime.InteropServices.WindowsRuntime;
using Microsoft.VisualStudio.TestTools.UnitTesting;

// add 'System.Memory' nuget package for span support

namespace UnitTestProject1
{
    [TestClass]
    public class UnitTest1
    {
        [TestMethod]
        public void CopyArrayToBufferWithZeroCountAtEnd()
        {
            byte[] array = { 0xA1, 0xA2, 0xA3 };
            var buffer = WindowsRuntimeBuffer.Create(10);
            array.CopyTo(3, buffer, 10, 0);
        }

        [TestMethod]
        public void CopyArrayToBufferWithZeroCountAtEnd_WorksWithArrays()
        {
            byte[] array = { 0xA1, 0xA2, 0xA3 };
            byte[] buffer = new byte[10];
            Array.Copy(array, 3, buffer, 10, 0);
        }

        [TestMethod]
        public void CopyBufferToArrayWithZeroCountAtEnd()
        {
            byte[] array = { 0xA1, 0xA2, 0xA3 };
            var buffer = array.AsBuffer();
            var target = new byte[10];
            buffer.CopyTo(3, target, 10, 0);
        }

        [TestMethod]
        public void CopyBufferToArrayWithZeroCountAtEnd_WorksWithSpans()
        {
            byte[] array = { 0xA1, 0xA2, 0xA3 };
            byte[] buffer = new byte[10];
            array.AsSpan(3).CopyTo(buffer.AsSpan(10, 0));
        }

        [TestMethod]
        public void CopyBufferToBufferWithZeroCountAtEnd()
        {
            byte[] array = { 0xA1, 0xA2, 0xA3 };
            var sourceBuffer = array.AsBuffer();
            var targetBuffer = WindowsRuntimeBuffer.Create(10);
            sourceBuffer.CopyTo(3, targetBuffer, 10, 0);
        }

        [TestMethod]
        public void CopyBufferToBufferWithZeroCountAtEnd_WorksWithSpans()
        {
            byte[] array = { 0xA1, 0xA2, 0xA3 };
            var sourceBuffer = array.AsSpan();
            var targetBuffer = new byte[10].AsSpan();
            sourceBuffer.Slice(3).CopyTo(targetBuffer.Slice(10, 0));
        }

        [TestMethod]
        public void BufferToArrayWithZeroCountAtEnd()
        {
            byte[] array = { 0xA1, 0xA2, 0xA3 };
            var result = array.AsBuffer().ToArray(3, 0);
            Assert.IsNotNull(result);
            Assert.AreEqual(0, result.Length);
        }

        [TestMethod]
        public void BufferToArrayWithZeroCountAtEnd_WorksWithSpans()
        {
            byte[] array = { 0xA1, 0xA2, 0xA3 };
            var result = array.AsSpan().Slice(3, 0).ToArray();
            Assert.IsNotNull(result);
            Assert.AreEqual(0, result.Length);
        }

        [TestMethod]
        public void BufferToArrayWithZeroCountAtEnd_WorksInLinq()
        {
            byte[] array = { 0xA1, 0xA2, 0xA3 };
            var result = array.ToList().Skip(3).Take(0).ToArray();
            Assert.IsNotNull(result);
            Assert.AreEqual(0, result.Length);
        }

        [TestMethod]
        public void GetWindowsRuntimeBufferWithZeroCountAtEnd()
        {
            byte[] array = { 0xA1, 0xA2, 0xA3 };
            var stream = new MemoryStream(array);
            var result = stream.GetWindowsRuntimeBuffer(3, 0);
            Assert.IsNotNull(result);
            Assert.AreEqual(0, result.Length);
        }

        [TestMethod]
        public void GetWindowsRuntimeBufferWithZeroCountAtEnd_WorksWithSpan()
        {
            byte[] array = { 0xA1, 0xA2, 0xA3 };
            var stream = new MemoryStream(array);
            var result = stream.ToArray().AsSpan(3, 0).ToArray();
            Assert.IsNotNull(result);
            Assert.AreEqual(0, result.Length);
        }
    }
}
```

</details>

UnitTest/TestComponentCSharp_Tests.cs
weltkante: why do you expect this to go through? you are attempting to copy into range `[3;4)` using `[offset;ending)` notation, and 4 is clearly beyond array length 3.

cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBufferExtensions.cs
```diff
@@ -426,7 +420,7 @@ public static Stream AsStream(this IBuffer source)
         public static byte GetByte(this IBuffer source, uint byteOffset)
         {
             if (source == null) throw new ArgumentNullException(nameof(source));
-            if (source.Capacity <= byteOffset) throw new ArgumentException(global::Windows.Storage.Streams.SR.Argument_BufferIndexExceedsCapacity, nameof(byteOffset));
+            if (source.Capacity < byteOffset) throw new ArgumentException(global::Windows.Storage.Streams.SR.Argument_BufferIndexExceedsCapacity, nameof(byteOffset));
```
weltkante, line 423: This check was actually correct, you are reading a byte, that is count=1, so from the discussion means `<=` is the correct check (alternatively `source.Capacity < byteOffset + 1`)

weltkante (changes requested):

UnitTest/TestComponentCSharp_Tests.cs
j0shuams: I don't expect it to go through, I wanted to make sure the change in condition checks didn't allow this to go through

cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBufferExtensions.cs
j0shuams: Yes I agree

cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBufferExtensions.cs
weltkante: No the check is correctly modified, compare with my examples in the PR comment, for example `Array.Copy` allows this to go through. The next two following lines check the upper bound again and prevent overflows, what you check here is that `source.Length - sourceIndex` doesn't overflow, otherwise this check wouldn't be needed at all.

weltkante: Yup, you can get past the safety checks.

```csharp
var array = new byte[3];
var buffer = WindowsRuntimeBuffer.Create(1);
array.CopyTo(0, buffer, 0xFFFFFFFE, 2);
```

This fails in `System.Buffer.BlockCopy` because it casts to `int`, but the `else` branch uses `Marshal.Copy` and will corrupt memory I believe.

weltkante:
There is an [unrelated bug](https://github.com/microsoft/CsWinRT/pull/504#discussion_r506627120) I discovered while reviewing, which is actually security relevant because it can corrupt memory, should this get a separate issue? Its not covered by #334 (which wants to reduce checks, not add new checks). Obviously it can just be fixed in this PR.

j0shuams:
> There is an [unrelated bug](https://github.com/microsoft/CsWinRT/pull/504#discussion_r506627120) I discovered while reviewing, which is actually security relevant because it can corrupt memory, should this get a separate issue? Its not covered by #334 (which wants to reduce checks, not add new checks). Obviously it can just be fixed in this PR.

I made the change, the check was already present in the CopyTo for IBuffer to IBuffer, strangely enough. Nice catch

weltkante:
@j0shuams did you forget to push? I'm not seeing the changes.

j0shuams:
@weltkante I am about to, I was comparing the unit tests I have been developing with the ones you provided to make sure I was handling proper cases.

cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBufferExtensions.cs
```diff
@@ -381,7 +376,7 @@ public static IBuffer GetWindowsRuntimeBuffer(this MemoryStream underlyingStream
             if (length < 0)
                 throw new ArgumentOutOfRangeException(nameof(length));
 
-            if (underlyingStream.Capacity <= positionInStream)
+            if (underlyingStream.Capacity < positionInStream)
```
j0shuams, line 379: Not sure this is a needed change.

weltkante: I have that one in my test list though (`GetWindowsRuntimeBufferWithZeroCountAtEnd`) and wouldn't expect it to throw, if you first convert via ToArray then it doesn't throw (but makes a copy of the buffer). If the conversion is implicit in the method call it shouldn't throw either.

j0shuams: Interesting, I have a test just like that one and it fails but not on this exception, but on the UnauthorizedAccessException check after

j0shuams: and yes if I undo this change, that test does fail.

Scottj1s:
@weltkante thank you for your attention on this PR - much appreciated

Scottj1s:
@j0shuams is this PR ready for review?

cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBufferExtensions.cs
```diff
@@ -118,9 +119,8 @@ public static byte[] ToArray(this IBuffer source)
         public static byte[] ToArray(this IBuffer source, uint sourceIndex, int count)
         {
             if (source == null) throw new ArgumentNullException(nameof(source));
-            if (source.Length == 0 && count == 0) return Array.Empty<byte>();
             if (count < 0) throw new ArgumentOutOfRangeException(nameof(count));
-            if (source.Capacity <= sourceIndex) throw new ArgumentException(global::Windows.Storage.Streams.SR.Argument_BufferIndexExceedsCapacity);
+            if (source.Capacity < sourceIndex) throw new ArgumentException(global::Windows.Storage.Streams.SR.Argument_BufferIndexExceedsCapacity);
```
Scottj1s, line 123: shouldn't these checks be against source.Length, and not source.Capacity?

cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBufferExtensions.cs
```diff
@@ -154,10 +151,9 @@ public static void CopyTo(this IBuffer source, uint sourceIndex, byte[] destinat
             if (destination == null) throw new ArgumentNullException(nameof(destination));
             if (count < 0) throw new ArgumentOutOfRangeException(nameof(count));
             if (destinationIndex < 0) throw new ArgumentOutOfRangeException(nameof(destinationIndex));
-            if (source.Length == 0 && count == 0) return;
-            if (source.Capacity <= sourceIndex) throw new ArgumentException(global::Windows.Storage.Streams.SR.Argument_BufferIndexExceedsCapacity);
+            if (source.Capacity < sourceIndex) throw new ArgumentException(global::Windows.Storage.Streams.SR.Argument_BufferIndexExceedsCapacity);
```
Scottj1s, line 154: these should be source.Length as well

cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBufferExtensions.cs
```diff
@@ -154,10 +151,9 @@ public static void CopyTo(this IBuffer source, uint sourceIndex, byte[] destinat
...
             if (source.Capacity - sourceIndex < count) throw new ArgumentException(global::Windows.Storage.Streams.SR.Argument_InsufficientSpaceInSourceBuffer);
-            if (destination.Length <= destinationIndex) throw new ArgumentException(global::Windows.Storage.Streams.SR.Argument_IndexOutOfArrayBounds);
+            if (destination.Length < destinationIndex) throw new ArgumentException(global::Windows.Storage.Streams.SR.Argument_IndexOutOfArrayBounds);
```
Scottj1s, line 156: and shouldn't these be destination.Capacity, and not Length

cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBufferExtensions.cs
```diff
@@ -191,10 +187,9 @@ public static void CopyTo(this IBuffer source, uint sourceIndex, IBuffer destina
         {
             if (source == null) throw new ArgumentNullException(nameof(source));
             if (destination == null) throw new ArgumentNullException(nameof(destination));
-            if (count == 0) return;
-            if (source.Capacity <= sourceIndex) throw new ArgumentException(global::Windows.Storage.Streams.SR.Argument_BufferIndexExceedsCapacity);
+            if (source.Capacity < sourceIndex) throw new ArgumentException(global::Windows.Storage.Streams.SR.Argument_BufferIndexExceedsCapacity);
```
Scottj1s, line 190: source.Length

Scottj1s:
>             if (source.Capacity <= byteOffset) throw new ArgumentException(global::Windows.Storage.Streams.SR.Argument_BufferIndexExceedsCapacity, nameof(byteOffset));

source.Length here too

---
Refers to: cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBufferExtensions.cs:424 in d721226. [](commit_id = d7212263d8c2159c401e85135e9b624891c02878, deletion_comment = False)

cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBufferExtensions.cs
j0shuams: The API documentation suggests Capacity, https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.windowsruntime.windowsruntimebufferextensions.toarray?view=dotnet-plat-ext-3.1 

I realize it also suggests `<=` instead of `<` so I think I need to see about getting the docs updated there.

Scottj1s (changes requested):
generally, shouldn't all bounds checks look at:

- Length for source, to ensure valid bytes are read
- Capacity for destination, to ensure sufficient space

cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBufferExtensions.cs
j0shuams: Here is documentation for this function, I think an array only has a Length property not Capacity.

https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.windowsruntime.windowsruntimebufferextensions.copyto?view=dotnet-plat-ext-3.1#System_Runtime_InteropServices_WindowsRuntime_WindowsRuntimeBufferExtensions_CopyTo_Windows_Storage_Streams_IBuffer_System_UInt32_System_Byte___System_Int32_System_Int32_

cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBufferExtensions.cs
j0shuams: Okay I was basing off of documentation/pre-existing, but the only argument I see for capacity is if you know the buffer is partially full, you could make a blank array from the unused part (a span starting at an index greater than length but less than capacity).

weltkante:
I'm not sure you can make a breaking change from `Capacity` to `Length` as far as `IBuffer` is concerned. When writing algorithms the `Length` property may not be up to date in intermediate steps when you take copies/snapshots, so since this API always allowed access to the full buffer it should continue to do so for compatibility reasons.

cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBufferExtensions.cs
weltkante: Buffers which require setting the "length of valid data" explicitly (e.g. native `IBuffer` implementations) may not have their `Length` up to date during all times when this API is called. Changing it to be more restrictive and not allow access to the full buffer would be a breaking change I don't think you should take.

Scottj1s:
> @Scottj1s I'm not sure you can make a breaking change from `Capacity` to `Length` as far as `IBuffer` is concerned. When implementing data transformations using `IBuffer` the `Length` property may not always be up to date in intermediate steps when you take copies/snapshots, so since this API always allowed access to the full buffer it should continue to do so for compatibility reasons.

We're already committed to breaking changes here, with the new conditions on params, beyond the the Length check.  This code (and any client code) has to be able to rely on IBuffer.Length being stable.

Scottj1s:
@weltkante thanks again for the invaluable feedback.  we're going to run with this and will be on the lookout for any regressions.

Scottj1s (approved):

weltkante:
Making previously throwing cases non-throwing and handled correctly is IMHO less breaking than starting to throw exceptions in previously working cases, but its your call.

PR #505, opened 2020-10-16
Fix our auto generation target to not cause race with sourcelink and add sourcelink for winrt.runtime

manodasanW:
- Added sourcelink for winrt.runtime
- In our auto generation of the projection, we were adding the sources before CoreCompile.  But sourcelink indicates any sources it includes should come before BeforeCompile as they also have a BeforeTargets of CoreCompile.  In my sample runs even though both of us are before CoreCompile, our targets still ran before sourcelink and was fine.  But I feel like this order is not defined and can be not deterministic, so this changes moves our source generation to be before BeforeCompile.  This change helps projection generators who may use sourcelink.
- Also making our source generation be dependent on CsWinRTRemoveWinMDReferences as it consumes the winmds it gathers
- ResolveAssemblyReferencesDependsOn seems like had a typo which probably dropped all the other depends, so fixing that.


Scottj1s (approved):
thanks - feel free to close PR 330 as well

PR #512, opened 2020-10-17
Update additional projections to use IDIC

ujjwalchadha:


UnitTest/TestComponentCSharp_Tests.cs
```diff
@@ -122,6 +122,30 @@ public void TestDynamicInterfaceCastingOnInvalidInterface()
...
+        [Fact]
+        public void TestBuffer()
+        {
+            var buffer = new Windows.Storage.Streams.Buffer(2);
+            var array = buffer.ToArray(0, 2);
+            Assert.Equal(2, array.Length);
+        }
+
+        [Fact]
+        public async void TestStorageFile()
```
j0shuams, line 135: I think our pipeline is going to throw a fit on these tests. From what I've read, xunit and `async void` test methods are not friends. I have some example tests that show how to pull out the async part into an async method, then the [Fact] attributed test method can just invoke it. 

Take a look at this async method and the test that follows it: https://github.com/microsoft/CsWinRT/blob/f91e9624d200d9b863757a8bcffa338f2fc8203c/UnitTest/TestComponentCSharp_Tests.cs#L92

cswinrt/strings/additions/Windows.Storage.Streams/IBufferByteAccess.cs
```diff
@@ -13,8 +13,10 @@ internal interface IBufferByteAccess
 namespace ABI.Windows.Storage.Streams
 {
     using global::System;
-    using global::System.Runtime.InteropServices;
-
+    using global::System.Runtime.InteropServices;
+//TODO: Update netstandard version by adding objectreferenceattribute
```
j0shuams, line 17: is this a todo for this PR or in the future? if future, can we open an issue for it?

UnitTest/TestComponentCSharp_Tests.cs
Scottj1s: you cannot make a test case async - look at the other pattern we use for async tests, where the [Fact]-attributed test case invokes an async DoIt function that awaits, synchronously Wait()-ing the result

cswinrt/strings/additions/Windows.Storage/IStorageFolderHandleAccess.cs
```diff
@@ -24,6 +24,8 @@ namespace ABI.Windows.Storage
 {
     using global::Microsoft.Win32.SafeHandles;
     using global::System;
+//TODO: Update netstandard version by adding objectreferenceattribute 
+#if NETSTANDARD2_0
```
Scottj1s, line 28: can we scope the #if blocks down to just the differences and not the entire class?  will make future maintenance easier

cswinrt/strings/additions/Windows.Storage.Streams/IBufferByteAccess.cs
Scottj1s: if possible, let's address it in this PR while you have context, etc

---
In reply to: [506975159](https://github.com/microsoft/CsWinRT/pull/512#discussion_r506975159) [](ancestors = 506975159)

Scottj1s (approved):

PR #513, opened 2020-10-17
Add an entry for MUX.ICustomProperty

### LINKED ISSUE #509: Microsoft.UI.Xaml.Data.ICustomProperty not being excluded from projection
so it's showing up in winrt.runtime.dll and microsoft.winui.dll.  an entry should be made to cswinrt\helpers.h, and breaking change relnote posted

j0shuams:
Not so sure I did this right...

Closes #509

### DIFF
--- a/WinRT.Runtime/Projections.cs
+++ b/WinRT.Runtime/Projections.cs
@@ -1,4 +1,5 @@
-﻿using System;
+﻿using ABI.Microsoft.UI.Xaml.Data;
+using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Collections.Specialized;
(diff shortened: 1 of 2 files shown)


manodasanW:
If we are doing a custom projection for it, should ICustomProperty ABI class implement from ICustomProperty?  As it doesn't seem [to](https://github.com/microsoft/CsWinRT/blob/5bd189172ec14f462d05edd3bc0c07a9ee1b9aa1/WinRT.Runtime/Projections/ICustomPropertyProvider.net5.cs#L29)  And we don't handle the non Do_ABI part.  @Scottj1s any idea?

Scottj1s:
There's more to this work than just the entries.  It looks like the manual projection is missing several things, like the ABI inheritance Mano mentions, DynamicInterfaceCastableImplementation, etc.  We should also consider compat issues of removing the type from WinUI now.  May not be an issue, but I want to validate.  I'm not sure why we even needed the manual projection to begin with.  This may not meet the bar for GA as it's just a duplicate type.

PR #517, opened 2020-10-18
Benkuhn/api tests

BenJKuhn:
Initial set of "real world" API unit tests. It's not a lot of code, but covers a wide range of concepts. async, enumeration, properties. buffers, objects, collections, interfaces. 

It's a little hard to design automation-friendly tests that use real APIs, since many APIs want real functionality (note the storage device test is disabled by default, because it only passes if you have a USB thumb drive plugged in). 

But at least we can always count on Mark Zbikowski ("MZ" prefix in PE format). ;)

### DIFF
--- a/UnitTest/ApiCompatTests.cs
+++ b/UnitTest/ApiCompatTests.cs
@@ -1,3 +1,8 @@
+
+// Workstation tests require local resources or configuration that may not be available in automation.
+// These can be enabled for additional local testing.
+// #define ENABLE_WORKSTATION_TESTS
+
 using System;
 using System.Collections;
 using System.Collections.Generic;
@@ -40,6 +45,10 @@ namespace UnitTest
             Assert.NotEqual("", cal.DayOfWeekAsString());
         }
 
+#if ENABLE_WORKSTATION_TESTS 
+        // Even with exception handling fall-backs, this test fails on x64 builds in automation 
+        // runs. Leaving disabled for now in automation, root cause unknown.
+
         // This test fails if geolocation is disabled, as is often the case in automation servers.
         async Task InvokeGeolocation()
         {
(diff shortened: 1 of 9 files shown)


UnitTest/ApiCompatTests.cs
```diff
@@ -0,0 +1,161 @@
...
+            Assert.NotEqual("", cal.DayOfWeekAsString());
+        }
+
+        async Task InvokeGeolocation()
+        {
+            // Actual position doesn't matter (unless you're on a weather bouy in 
+            // the Gulf of Guinea. Then this might give a false failure). 
+            // This test is just validating the results of calling a simply async OS API.
+            Geolocator locator = new Geolocator();
+            Geoposition pos = await locator.GetGeopositionAsync();
```
manodasanW, line 49: Hopefully this doesn't throw in the lab due to location being turned off or something, but a run will tell us whether it will or not.

manodasanW: Did a run and it seems it does fail.

 Failed UnitTest.TestAPIs.TestGeolocation [22 ms]
  Error Message:
   System.AggregateException : One or more errors occurred. (The service cannot be started, either because it is disabled or because it has no enabled devices associated with it. (0x80070422))
---- System.Runtime.InteropServices.COMException : The service cannot be started, either because it is disabled or because it has no enabled devices associated with it. (0x80070422)
  Stack Trace:
     at System.Threading.Tasks.Task.Wait(Int32 millisecondsTimeout, CancellationToken cancellationToken)
   at System.Threading.Tasks.Task.Wait(Int32 millisecondsTimeout)
   at UnitTest.TestAPIs.TestGeolocation() in D:\a\1\s\UnitTest\ApiCompatTests.cs:line 57

BenJKuhn: Thanks for checking. I wondered about that. Updating to give a :free pass" on this error.

manodasanW (approved):

PR #522, opened 2020-10-19
Build breaks

Scottj1s:
fixed authoring break from vtbl optimization, temporarily disabled authoring tests and api tests failing on build agent


j0shuams (approved):

Scottj1s:
unfortunately, we took the vtbl optimization work based on a private build, just before authoring tests were enabled - temporarily disabling to unblock the build

cswinrt/code_writers.h
```diff
@@ -5080,7 +5080,8 @@ public static unsafe void DisposeAbiArray(object box) => MarshalInspectable<obje
                             {
                                 w.write(
                                     ".As<IUnknownVftbl>(GuidGenerator.GetIID(typeof(%).GetHelperType()))",
-                                    bind<write_type_name>(get_type_semantics(get_default_interface(type)), typedef_name_type::Projected, false));
+                                    bind<write_type_name>(get_type_semantics(get_default_interface(type)), 
+                                        settings.component ? typedef_name_type::ABI : typedef_name_type::Projected, false));
```
manodasanW, line 5084: I feel like you want typedef_name_type::ccw here which should handle component and non component.

UnitTest/ApiCompatTests.cs
```diff
@@ -79,9 +79,7 @@ public void TestGeolocation()
         {
             Assert.True(InvokeGeolocation().Wait(1000));
         }
-#endif
 
-#if ENABLE_WORKSTATION_TESTS
```
BenJKuhn, line 84: The first #if is unmatched in this PR. Intentional?

Scottj1s: the #ifs are matched - the PR doesn't show it well

PR #523, opened 2020-10-19
better authoring fix

Scottj1s:


UnitTest/TestComponentCSharp_Tests.cs
```diff
@@ -1572,6 +1572,14 @@ public void TestVector4TypeMapping()
             Assert.True(TestObject.Vector4Property == vector4);
         }
 
+        [Fact]
+        public void BufferToArrayWithZeroCountAtEnd_WorksWithSpans()
+        {
+            byte[] array = { 0xA1, 0xA2, 0xA3 };
+            var result = array.AsSpan().Slice(3, 0);
+            //Assert.IsNotNull(result);
```
manodasanW, line 1580: intentional?

manodasanW (approved):

UnitTest/TestComponentCSharp_Tests.cs
Scottj1s: nope (my you're fast!)

PR #526, opened 2020-10-19
Added short circuit for zero count

Scottj1s:
after parameter validation, but before getting pointer offset, to avoid assert


comp.rsp
```diff
@@ -0,0 +1 @@
+ -component -input "D:\git\cswinrt3\Authoring\AuthoringSample\Generated Files\net5.0\AuthoringSample.winmd" -input 10.0.18362.0 -include AuthoringSample -output "C:\Users\sjones.REDMOND\AppData\Local\Temp\CsWinRT\a4nwanat.ym0" -verbose
```
manodasanW, line 1: intentional?

Scottj1s: sigh, no (thx again!)

PR #533, opened 2020-10-21
Fixed removal of dll references that report IsWinmdFile == true

Scottj1s:
Some references required by csc are inadvertently removed because they record true for being WinMD files, but actually have a DLL extension.

### DIFF
--- a/nuget/Microsoft.Windows.CsWinRT.targets
+++ b/nuget/Microsoft.Windows.CsWinRT.targets
@@ -27,11 +27,11 @@ Copyright (C) Microsoft Corporation. All rights reserved.
   <Target Name="CsWinRTRemoveWinMDReferences" BeforeTargets="ResolveReferences;BeforeCompile" AfterTargets="AfterResolveReferences">
     <ItemGroup>
       <!--Move winmd references into private item group to prevent subsequent winmd reference errors-->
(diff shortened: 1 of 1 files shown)


manodasanW (approved):

nuget/Microsoft.Windows.CsWinRT.targets
```diff
@@ -27,11 +27,11 @@ Copyright (C) Microsoft Corporation. All rights reserved.
   <Target Name="CsWinRTRemoveWinMDReferences" BeforeTargets="ResolveReferences;BeforeCompile" AfterTargets="AfterResolveReferences">
     <ItemGroup>
       <!--Move winmd references into private item group to prevent subsequent winmd reference errors-->
-      <CsWinRTRemovedReferences Include="@(ReferencePath)" Condition="'%(ReferencePath.WinMDFile)' == 'true'" />
+      <CsWinRTRemovedReferences Include="@(ReferencePath)" Condition="'%(ReferencePath.Extension)' == '.winmd'" />
```
j0shuams, line 30: Is ReferencePath an object with fields `WinMDFile` and `Extension`? Or did you change the `WinMDFile` field to `Extension`?

AdamBraden:
Not sure I understand the root issue - dlls were getting marked IsWinMDFile=true. Who is doing this, are they winmd files with the wrong extension? If they are not winmd files, doesn't that point to a VS or Net5 problem?

PR #536, opened 2020-10-23
Fixing read only dictionary lookup.

### LINKED ISSUE #535: Accessing Properties dictionary of DeviceInformation causes InvalidCastException
The following code results in an InvalidCastException on the properties read. 

```
var ports = await DeviceInformation.FindAllAsync(SerialDevice.GetDeviceSelector(), new string[] { "System.DeviceInterface.Serial.PortName" });
foreach (var port in ports)
{
    object o = port.Properties["System.DeviceInterface.Serial.PortName"];
}

'Invalid cast from 'WinRT.IInspectable' to 'ABI.System.Collections.Generic.IEnumerable`1[System.Collections.Generic.KeyValuePair`2[System.String,System.Object]]'.'
```

An exception should not be thrown here, especially since this is very simple code.

System is running 5.0.100-rc.2.20479.15, and I'm using TFM net5.0-windows10.0.17763.0

manodasanW:
Fixes #535

### DIFF
--- a/UnitTest/TestComponentCSharp_Tests.cs
+++ b/UnitTest/TestComponentCSharp_Tests.cs
@@ -194,6 +194,24 @@ namespace UnitTest
             Assert.Equal(vector, uriVector);
         }
 
+        async Task LookupPorts()
+        {
+            var ports = await Windows.Devices.Enumeration.DeviceInformation.FindAllAsync(
+                Windows.Devices.SerialCommunication.SerialDevice.GetDeviceSelector(),
+                new string[] { "System.ItemNameDisplay" });
+            foreach (var port in ports)
+            {
+                object o = port.Properties["System.ItemNameDisplay"];
+                Assert.NotNull(o);
+            }
+        }
+
+        [Fact]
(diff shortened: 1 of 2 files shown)


UnitTest/TestComponentCSharp_Tests.cs
```diff
@@ -297,8 +297,26 @@ public void TestVectorCastConversion()
...
+            {
+                object o = port.Properties["System.ItemNameDisplay"];
+                Assert.NotNull(o);
+            }
+        }
+
+        [Fact]
+        public void TestReadOnlyDictionaryLookup()
+        {
+            Assert.True(LookupPorts().Wait(10000));
```
Scottj1s, line 317: any reason the timeout is so long?

manodasanW: Reduced it, I think I previously needed a higher timeout as I had a slightly different logic for lookup before and had breakpoints set.  But 1000ms seems to work fine.

Scottj1s (approved):

PR #551, opened 2020-10-29
Type.FromManaged returning an HSTRING reference, rather than allocating an HSTRING, causing crashes

### LINKED ISSUE #550: ICustomPropertyProvider::get_Type returns an object with HSTRING lifetime issues
reported internally by DCPP team.

Scottj1s:
fixes #550

### DIFF
--- a/WinRT.Runtime/Projections/Type.cs
+++ b/WinRT.Runtime/Projections/Type.cs
@@ -29,7 +29,7 @@ namespace ABI.System
             }
         }
 
-        public static Marshaler CreateMarshaler(global::System.Type value)
+        private static (String Name, TypeKind Kind) ToAbi(global::System.Type value)
         {
             TypeKind kind = TypeKind.Custom;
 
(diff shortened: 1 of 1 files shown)


WinRT.Runtime/Projections/Type.cs
```diff
@@ -85,8 +91,13 @@ public static Type GetAbi(Marshaler m)
             *(Type*)dest.ToPointer() = GetAbi(arg);
 
         public static Type FromManaged(global::System.Type value)
-        {
-            return GetAbi(CreateMarshaler(value));
+        {
+            var abi = ToAbi(value);
+            return new Type
+            {
+                Name = MarshalString.FromManaged(abi.Name),
```
manodasanW, line 98: The effect of this change I see is that we are now doing WindowsCreateString rather than WindowsCreateStringReference in this scenario.  Do we know why WindowsCreateStringReference doesn't work here, but works above?  And shouldn't MarshalString.GetAbi work for both scenarios assuming it isn't WindowsCreateStringReference just failing?

Scottj1s: WindowsCreateStringReference is wrong and was causing crashes, because there's nothing to hold the marshaler for the lifetime of the returned Type object.  In other words, the entire pattern is wrong.  We were using the 'in' pattern of creating a marshaler and then passing its value across the ABI.  This is an 'out' pattern, where we must allocate the entire object for the ABI.

ujjwalchadha (approved):

BenJKuhn (approved):

WinRT.Runtime/Projections/Type.cs
manodasanW: thanks, that makes sense.

PR #552, opened 2020-10-29
Ensure that xamlprecompile also adds generated sources

### LINKED ISSUE #530: CsWinRTIncludeProjection should be added to CoreCompileDependsOn
Currently, the target `CsWinRTIncludeProjection` is listed as `BeforeTargets="BeforeCompile"`, but this doesn't cause it to occur before every instance of the `CSC` task - e.g., there's also the target `XamlPreCompile`, which it isn't guaranteed to come before.  Instead of the BeforeTargets, we should add `CsWinRTIncludeProjection` to the property `CoreCompileDependsOn`, which will cause it to be called prior to any compilation.

Scottj1s:
fixes #530

### DIFF
--- a/nuget/Microsoft.Windows.CsWinRT.targets
+++ b/nuget/Microsoft.Windows.CsWinRT.targets
@@ -15,6 +15,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <CsWinRTGenerateProjection Condition="'$(CsWinRTGenerateProjection)' == ''">true</CsWinRTGenerateProjection>
     <AllowUnsafeBlocks Condition="'$(CsWinRTComponent)' == 'true'">true</AllowUnsafeBlocks>
     <AllowUnsafeBlocks Condition="$(CsWinRTEnabled)">true</AllowUnsafeBlocks>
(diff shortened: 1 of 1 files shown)


nuget/Microsoft.Windows.CsWinRT.targets
```diff
@@ -15,6 +15,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <CsWinRTGenerateProjection Condition="'$(CsWinRTGenerateProjection)' == ''">true</CsWinRTGenerateProjection>
     <AllowUnsafeBlocks Condition="'$(CsWinRTComponent)' == 'true'">true</AllowUnsafeBlocks>
     <AllowUnsafeBlocks Condition="$(CsWinRTEnabled)">true</AllowUnsafeBlocks>
+    <CoreCompileDependsOn>CsWinRTIncludeProjection;CsWinRTRemoveWinMDReferences;$(CoreCompileDependsOn)</CoreCompileDependsOn>
```
manodasanW, line 18: I assume CsWinRTRemoveWinMDReferences is not needed as CsWinRTIncludeProjection already depends on it?  But including it doesn't change behavior either, so probably fine.

manodasanW (approved):

PR #559, opened 2020-11-02
Fixed IDictionary.Clear EntryPointNotFoundException

ujjwalchadha:


WinRT.Runtime/Projections/IDictionary.net5.cs
```diff
@@ -791,7 +791,17 @@ public static ObjectReference<Vftbl> ObjRefFromAbi(IntPtr thisPtr)
             }
         }
 
-        unsafe void global::System.Collections.Generic.ICollection<global::System.Collections.Generic.KeyValuePair<K, V>>.Clear()
+        void global::Windows.Foundation.Collections.IMap<K, V>.Clear()
```
Scottj1s, line 794: why didn't this trigger a compile error (not implementing the interface)?

ujjwalchadha: This should have been a compilation error indeed but seems to be a bug. The reason it did not complain was that Clear method was defined and present but as `System.Collections.Generic.ICollection<global::System.Collections.Generic.KeyValuePair<K, V>>.Clear`. This means the ICollection's Clear method was overridden, but IMap's Clear method was not.

But now when I think of it again, I see that we have 2 definitions of the Clear method so I'm not sure if this is the right way to do it because if we call the Clear method on class's default interface and not from IMap or ICollection, this might be an ambiguous call as to which Clear method to call.

Scottj1s: let's try to understand root cause a bit better - why we didn't get a compiler error with only one explicit interface implementation

jkoritzinsky: There’s currently no compile errors for not implementing all the interface members in a default interface used as an IDynamicInterfaceCastable implementation. I opened an issue a bit ago to make an analyzer for that.

Scottj1s (approved):

WinRT.Runtime/Projections/IDictionary.net5.cs
jkoritzinsky: I think we need both of these methods to ensure that there's implementations for all of the methods.

Issue #514, opened 2020-10-17
NullReferenceExceptions - audit code for potential null derefs


Scottj1s:
With some recent NREs on string compares, a comprehensive check for nulls in the code would be beneficial.

Some techniques to automate, vs a manual scan of the code:

Ensure that fxcop is flagging public strings via CA1062:
https://docs.microsoft.com/en-ca/dotnet/fundamentals/code-analysis/quality-rules/ca1062?view=vs-2019

Run a few Roslyn analyzers:
http://pihrt.net/Roslynator/Analyzers

Enable (really, disable) nullable references:
https://devblogs.microsoft.com/dotnet/try-out-nullable-reference-types/

The last option is the most comprehensive and would detect any unguarded nullables (not just strings), which may induce a bit of churn, but would probably be worthwhile.

AdamBraden:
Create a branch with this enabled and check in

j0shuams:
TL;DR: Cost outweighs benefit. About 1k warnings to handle in WinRT.Runtime, and more to be discovered once the first wave of annotations are added (e.g. newly annotated functions are used in the return statement of other functions, which would also now need annotations). More changes would be needed as well, and there is not much to gain from enabling null references. 

Did due diligence and played around with nullable references enabled. Caused about 100k warnings to be generated, with about 1% of those living in WinRT.Runtime. Having gone in and added `?` to return types of ABI functions like `MarshalInspectable.FromAbi`, the functions that use the ABI functions will in-turn need `?`. So there will be more work than just the 1%, once changes start to be made. And this is just for the manually projected code, to get proper changes in for generated code we would need to update `code_writers.h` to maintain state/context about what type is being written so it would know whether to add a `?` to the end of a type name. So, there's a lot of work and not a lot of benefit.

PR #560, opened 2020-11-03
Enable nullable reference checks in all but test and sample projects

j0shuams:
Follow-up from discussion we had on #514

### DIFF
--- a/Authoring/WinRT.SourceGenerator/WinRT.SourceGenerator.csproj
+++ b/Authoring/WinRT.SourceGenerator/WinRT.SourceGenerator.csproj
@@ -15,7 +15,6 @@
     <Description>C#/WinRT Authoring Source Generator Preview $(VersionString)</Description>
     <AssemblyTitle>C#/WinRT Authoring Source Generator Preview v$(VersionString)</AssemblyTitle>
     <Copyright>Copyright (c) Microsoft Corporation. All rights reserved.</Copyright>
-    <Nullable>enable</Nullable>
   </PropertyGroup>
 
   <ItemGroup>
--- a/Authoring/WinRT.SourceGenerator/WinRT.SourceGenerator.csproj
+++ b/Authoring/WinRT.SourceGenerator/WinRT.SourceGenerator.csproj
@@ -15,6 +15,7 @@
     <Description>C#/WinRT Authoring Source Generator Preview $(VersionString)</Description>
     <AssemblyTitle>C#/WinRT Authoring Source Generator Preview v$(VersionString)</AssemblyTitle>
     <Copyright>Copyright (c) Microsoft Corporation. All rights reserved.</Copyright>
+    <Nullable>enable</Nullable>
   </PropertyGroup>
 
   <ItemGroup>
--- a/Projections/Test/Test.csproj
+++ b/Projections/Test/Test.csproj
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0;net5.0</TargetFrameworks>
     <Platforms>x64;x86</Platforms>
-  </PropertyGroup>
+   </PropertyGroup>
   
   <ItemGroup>
     <ProjectReference Include="..\..\TestComponentCSharp\TestComponentCSharp.vcxproj" />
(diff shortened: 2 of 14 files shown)


cswinrt/cswinrt.vcxproj
```diff
@@ -7,6 +7,7 @@
     <Keyword>Win32Proj</Keyword>
     <ProjectName>cswinrt</ProjectName>
     <RootNamespace>cswinrt</RootNamespace>
+    <Nullable>enable</Nullable>
```
manodasanW, line 10: This is a C++ project, I assume nullable doesn't have any effect as it is C#?

Authoring/WinRT.SourceGenerator/WinRT.SourceGenerator.csproj
```diff
@@ -15,6 +15,7 @@
     <Description>C#/WinRT Authoring Source Generator Preview $(VersionString)</Description>
     <AssemblyTitle>C#/WinRT Authoring Source Generator Preview v$(VersionString)</AssemblyTitle>
     <Copyright>Copyright (c) Microsoft Corporation. All rights reserved.</Copyright>
+    <Nullable>enable</Nullable>
```
manodasanW, line 18: Should we set this in our common props for all csproj similar to what we do for some other properties?

Projections/Test/Test.csproj
```diff
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0;net5.0</TargetFrameworks>
     <Platforms>x64;x86</Platforms>
-  </PropertyGroup>
+   </PropertyGroup>
```
manodasanW, line 6: Did we not set it here?

WinRT.Host/WinRT.Host.vcxproj
```diff
@@ -25,6 +25,7 @@
     <ProjectGuid>{7e33bcb7-19c5-4061-981d-ba695322708a}</ProjectGuid>
     <RootNamespace>WinRTHost</RootNamespace>
     <WindowsTargetPlatformVersion>10.0</WindowsTargetPlatformVersion>
+    <Nullable>enable</Nullable>
```
manodasanW, line 28: This is also a C++ project.

manodasanW:
So did this all build fine with nullable enabled?

Authoring/WinRT.SourceGenerator/WinRT.SourceGenerator.csproj
Scottj1s: right - the article suggests:
"Note that you can also apply the Nullable property to a Directory.build.props file if that fits your workflow better."

Projections/Test/Test.csproj
j0shuams: I didn't originally because I didn't think it would make much sense to enable it in test projects. But this morning I had a new thought that it wouldn't be bad to add it in these places. Not sure how this whitespace change got committed, if that's what you're asking.

cswinrt/cswinrt.vcxproj
j0shuams: The idea was to not miss strings directory (which is C#)

Scottj1s:
I don't think it's enough just to enable nullables.  I would expect some changes in Marshalers.cs as well.  E.g., MarshalString.FromAbi will never return a null string, so doesn't require any annotation changes (e.g., string?).  But MarshalInterface<T>.FromAbi can return null if the ABI pointer is null, so should probably be returning a T?, and null-checks enforced downstream.  Same goes for MarshalInspectable<T>.FromAbi, which returns 'default' for T (which is null for a reference type).  I would think some of this would be caught by the compiler with nullables enabled, but that may not apply evenly (e.g., to generics).  Some experimenting to ensure known scenarios will generate warnings would be good.

j0shuams:
@manodasanW 
Yes the changes do not cause any errors or failures, they just write a lot (> 100k) of warnings. I'm not so sure about checking-in with nullable enabled as it about doubles the build time 

@Scottj1s Yes that is more of what I was thinking would be needed. I will look into that now.

j0shuams:
Did work to see what changes would be needed and decided there's more work to be done than benefit to be gained. Issue #514 closed with comment

PR #563, opened 2020-11-04
COM Interop guide

Scottj1s:
Lots of material in [System.Runtime.InteropServices](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices?view=netcore-3.1), so I expect this guide will grow.

### DIFF
--- a/docs/interop.md
+++ b/docs/interop.md
@@ -78,11 +78,11 @@ public void SetString(IntPtr hstr);
 // public void GetString([MarshalAs(UnmanagedType.HString)] out String s);
 public void GetString(IntPtr out hstr);
 // ...
-// Marshal System.String as fast-pass HSTRING reference
+// Marshal HSTRING to System.String
 IntPtr hstr;
 GetString(out hstr);
 var str = MarshalString.FromAbi(hstr);
-// Marshal HSTRING to System.String
+// Marshal System.String as fast-pass HSTRING reference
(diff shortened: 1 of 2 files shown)


docs/interop.md
```diff
@@ -0,0 +1,151 @@
...
+## Unsupported
+The following Marshal CCW/RCW-related functions are not supported in C#/WinRT:
+```csharp
+AreComObjectsAvailableForCleanup,
+ChangeWrapperHandleStrength
+CleanupUnusedObjectsInCurrentContext
+CreateAggregatedObject
+CreateWrapperOfType
+GetComObjectData
+SetComObjectData
```
jkoritzinsky, line 145: The IWinRTObject.AdditionalTypeData dictionary could be considered to replace these two members.

docs/interop.md
```diff
@@ -0,0 +1,151 @@
...
+GetEndComSlot
+GetIDispatchForObject
+GetStartComSlot
+GetUniqueObjectForIUnknown
+IsComObject
```
jkoritzinsky, line 150: An `is IWinRTObject` expression is the equivalent of `IsComObject`.

docs/interop.md
Scottj1s: oh sure - good suggestion

AdamBraden (approved):

PR #564, opened 2020-11-05
Fixed #553 and further issues in native to managed calls

### LINKED ISSUE #553: IMap.GetView() & IVector.GetView() call from native side on managed IDictionary & IList implementations respectively result in InvalidCastException
Repro steps: 
1. Disable TestWinRT clone from git in build.cmd
2. In TestWinRT, in TestRunner.cpp, in function Collection3Call, add the following line to consume the managed IDictionary object's Do_Abi_GetView_ method:
```
TEST_REQUIRE_N(L"Collection", 3, c.GetView().Size() == 3);
```
3. Run build.cmd and check result for test: `UnitTest.TestWinRT.Collections_Dictionary_Call`

It fails with the following exception:
```
[xUnit.net 00:00:03.19]       System.InvalidCastException : Specified cast is not valid.
[xUnit.net 00:00:03.19]       Stack Trace:
[xUnit.net 00:00:03.20]            at System.Runtime.InteropServices.Marshal.ThrowExceptionForHR(Int32 errorCode)
[xUnit.net 00:00:03.20]         C:\Users\ujchadha\source\repos\CsWinRT\WinRT.Runtime\ObjectReference.cs(55,0): at WinRT.IObjectReference.As[T](Guid iid)
[xUnit.net 00:00:03.20]         C:\Users\ujchadha\source\repos\CsWinRT\WinRT.Runtime\ObjectReference.cs(51,0): at WinRT.IObjectReference.As[T]()
[xUnit.net 00:00:03.20]         C:\Users\ujchadha\source\repos\CsWinRT\WinRT.Runtime\Marshalers.cs(845,0): at WinRT.MarshalInterface`1.CreateMarshaler(T value)
[xUnit.net 00:00:03.20]         C:\Users\ujchadha\source\repos\CsWinRT\WinRT.Runtime\Marshalers.cs(861,0): at WinRT.MarshalInterface`1.FromManaged(T value)
[xUnit.net 00:00:03.20]         C:\Users\ujchadha\source\repos\CsWinRT\WinRT.Runtime\Projections\IDictionary.net5.cs(610,0): at ABI.System.Collections.Generic.IDictionary`2.Vftbl.Do_Abi_GetView_3(IntPtr thisPtr, IntPtr& __return_value__)
[xUnit.net 00:00:03.20]         --- End of stack trace from previous location ---
[xUnit.net 00:00:03.20]         C:\Users\ujchadha\source\repos\CsWinRT\WinRT.Runtime\ExceptionHelpers.cs(65,0): at WinRT.ExceptionHelpers.ThrowExceptionForHR(Int32 hr)
[xUnit.net 00:00:03.20]         C:\Users\ujchadha\source\repos\CsWinRT\Projections\Test\Generated Files\net5.0\TestComponent.cs(8838,0): at ABI.TestComponent.ITests.global::TestComponent.ITests.Collection3Call(Collection3Handler handler)
[xUnit.net 00:00:03.20]         C:\Users\ujchadha\source\repos\CsWinRT\UnitTest\TestComponent_Tests.cs(677,0): at UnitTest.TestWinRT.Collections_Dictionary_Call()
```

ujjwalchadha:
Fixes #553 
The fix for that is simply adding `[global::WinRT.WindowsRuntimeType]` attribute to the interfaces

But there's another issue that comes up once `GetView()` is able to return the underlying `ToAbiHelper` object: When the native side tries to call methods on the underlying object (which is `ToAbiHelper` instance and not the actual instance of the object on C# side), the Do_Abi calls assume the underlying object also implements the C# equivalent of the interface and calls those methods (In this case the Do_Abi method would assume ToAbiHelper also implements IReadOnlyDictionary and thus would call the IReadOnlyDictionary methods instead of IMapView methods. But since ToAbiHelper does not implement IReadOnlyDictionary, the call fails).

To fix it, we need to either return the underlying object to the native side or make `ToAbiHelper` implement both `IReadOnlyDictionary` and `IMapView` (i.e. both managed and native side equivalent of the interface). I chose the first approach as it seemed a more logical thing to do: the native side should have reference to the managed side object and not some helper object. Also, there's no good motivation to add duplicate logic in ToAbiHelper. But to make this fix, I had to change the interface definitions for the projected types so as to have a proper return type which the managed code can create (because the managed side object to be returned would implement the managed side interface- IReadOnlyDictionary, not IMapView)

### DIFF
--- a/TestComponentCSharp/ManualProjectionTestClasses.cpp
+++ b/TestComponentCSharp/ManualProjectionTestClasses.cpp
@@ -1,8 +1,7 @@
 #include "pch.h"
 #include "ManualProjectionTestClasses.h"
-#include "IBindableIteratorTest.g.cpp"
-#include "IDictionaryTest.g.cpp"
-#include "IDisposableTest.g.cpp"
+#include "CustomBindableIteratorTest.g.cpp"
+#include "CustomDisposableTest.g.cpp"
 
 namespace winrt
 {
@@ -21,84 +20,31 @@ namespace winrt
 
 namespace winrt::TestComponentCSharp::implementation
 {
-	IBindableIteratorTest::IBindableIteratorTest()
+	CustomBindableIteratorTest::CustomBindableIteratorTest()
 	{
 		
 	}
 
-	bool IBindableIteratorTest::MoveNext()
+	bool CustomBindableIteratorTest::MoveNext()
 	{
 		return true;
 	}
 
-	Windows::Foundation::IInspectable IBindableIteratorTest::Current()
+	Windows::Foundation::IInspectable CustomBindableIteratorTest::Current()
 	{
 		return Windows::Foundation::PropertyValue::CreateInt32(27861);
 	}
 
-	bool IBindableIteratorTest::HasCurrent()
+	bool CustomBindableIteratorTest::HasCurrent()
 	{
 		return true;
 	}
-	IDictionaryTest::IDictionaryTest()
+
+	CustomDisposableTest::CustomDisposableTest()
 	{
-	}
-	hstring IDictionaryTest::Lookup(hstring key)
-	{
-		return _map.Lookup(key);
-	}
-	bool IDictionaryTest::HasKey(hstring key)
-	{
-		return _map.HasKey(key);
-	}
-	Windows::Foundation::Collections::IMapView<hstring, hstring> IDictionaryTest::GetView()
-	{
-		return _map.GetView();
-	}
-	bool IDictionaryTest::Insert(hstring key, hstring value)
-	{
-		return _map.Insert(key, value);
-	}
-	void IDictionaryTest::Remove(hstring key)
-	{
-		_map.Remove(key);
-	}
-	void IDictionaryTest::Clear()
-	{
-		_map.Clear();
-	}
-	int IDictionaryTest::Size()
-	{
-		return _map.Size();
 	}
 
-	Windows::Foundation::Collections::IIterator<Windows::Foundation::Collections::IKeyValuePair<hstring, hstring>> IDictionaryTest::First()
-	{
-		return _map.First();
-	}
-
-	bool IDictionaryTest::Consume(Windows::Foundation::Collections::IMap<hstring, hstring> map)
-	{
-		map.Clear();
-
-		map.Insert(L"key.cpp", L"value.cpp");
-		if (map.Size() != 1)
-		{
-			return false;
-		}
-		if (!map.HasKey(L"key.cpp"))
-		{
-			return false;
-		}
-		
-		return true;
-		
-	}
-	
-	IDisposableTest::IDisposableTest()
-	{
-	}
-	void IDisposableTest::Close()
+	void CustomDisposableTest::Close()
 	{
 	}
 }
--- a/TestComponentCSharp/ManualProjectionTestClasses.cpp
+++ b/TestComponentCSharp/ManualProjectionTestClasses.cpp
@@ -3,21 +3,6 @@
 #include "CustomBindableIteratorTest.g.cpp"
 #include "CustomDisposableTest.g.cpp"
 
-namespace winrt
-{
-	template <typename T, typename Allocator = std::allocator<T>>
-	Windows::Foundation::Collections::IVectorView<T> single_threaded_vector_view(std::vector<T, Allocator>&& values = {})
-	{
-		return make<impl::input_vector_view<T, std::vector<T, Allocator>>>(std::move(values));
-	}
-
-	template <typename K, typename V, typename Allocator = std::allocator<K>>
-	Windows::Foundation::Collections::IMapView<K, V> single_threaded_map_view(std::map<K, V, Allocator>&& values = {})
-	{
-		return make<impl::input_map_view<K, V, std::map<K, V, Allocator>>>(std::move(values));
-	}
-}
-
 namespace winrt::TestComponentCSharp::implementation
 {
 	CustomBindableIteratorTest::CustomBindableIteratorTest()
--- a/TestComponentCSharp/ManualProjectionTestClasses.h
+++ b/TestComponentCSharp/ManualProjectionTestClasses.h
@@ -1,68 +1,33 @@
 #pragma once
-#include "IBindableIteratorTest.g.h"
-#include "IDictionaryTest.g.h"
-#include "IDisposableTest.g.h"
+#include "CustomBindableIteratorTest.g.h"
+#include "CustomDisposableTest.g.h"
 
 namespace winrt::TestComponentCSharp::implementation
 {
-	struct IBindableIteratorTest : IBindableIteratorTestT<IBindableIteratorTest>
+	struct CustomBindableIteratorTest : CustomBindableIteratorTestT<CustomBindableIteratorTest>
 	{
-		IBindableIteratorTest();
+		CustomBindableIteratorTest();
 		bool MoveNext();
 		Windows::Foundation::IInspectable Current();
 		bool HasCurrent();
 	};
 
-	struct IDictionaryTest : IDictionaryTestT<IDictionaryTest>
+	struct CustomDisposableTest : CustomDisposableTestT<CustomDisposableTest>
 	{
-	private:
-		Windows::Foundation::Collections::IMap<hstring, hstring> _map{
-			winrt::single_threaded_map<hstring, hstring>()
-		};
-	public:
-		IDictionaryTest();
-		hstring Lookup(hstring key);
-		bool HasKey(hstring key);
-		Windows::Foundation::Collections::IMapView<hstring, hstring> GetView();
-		bool Insert(hstring key, hstring value);
-		void Remove(hstring key);
-		void Clear();
-		int Size();
-		Windows::Foundation::Collections::IIterator<Windows::Foundation::Collections::IKeyValuePair<hstring, hstring>> First();
-
-		static bool Consume(Windows::Foundation::Collections::IMap<hstring, hstring> map);
-	};
-
-	struct IDisposableTest : IDisposableTestT<IDisposableTest>
-	{
-		IDisposableTest();
+		CustomDisposableTest();
 		void Close();
 	};
 }
 
 namespace winrt::TestComponentCSharp::factory_implementation
 {
-	struct IBindableIteratorTest : IBindableIteratorTestT<IBindableIteratorTest, implementation::IBindableIteratorTest, Windows::Foundation::IStringable>
+	struct CustomBindableIteratorTest : CustomBindableIteratorTestT<CustomBindableIteratorTest, implementation::CustomBindableIteratorTest>
 	{
-		hstring ToString()
-		{
-			return L"IBindableIteratorTest";
-		}
+
 	};
 
-	struct IDictionaryTest : IDictionaryTestT<IDictionaryTest, implementation::IDictionaryTest, Windows::Foundation::IStringable>
+	struct CustomDisposableTest : CustomDisposableTestT<CustomDisposableTest, implementation::CustomDisposableTest>
 	{
-		hstring ToString()
-		{
-			return L"IDictionaryTest";
-		}
-	};
 
-	struct IDisposableTest : IDisposableTestT<IDisposableTest, implementation::IDisposableTest, Windows::Foundation::IStringable>
-	{
-		hstring ToString()
-		{
-			return L"IDisposableTest";
-		}
 	};
 }
--- a/TestComponentCSharp/TestComponentCSharp.idl
+++ b/TestComponentCSharp/TestComponentCSharp.idl
@@ -341,29 +341,16 @@ namespace TestComponentCSharp
         void Observe(Microsoft.UI.Xaml.Interop.IBindableObservableVector vector);
     }
 
-    //[default_interface]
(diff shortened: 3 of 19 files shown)


WinRT.Runtime/Projections.cs
```diff
@@ -70,6 +70,8 @@ static Projections()
             CustomTypeToHelperTypeMappings.Add(typeof(IMapView<,>), typeof(ABI.System.Collections.Generic.IReadOnlyDictionary<,>));
             CustomTypeToHelperTypeMappings.Add(typeof(IVectorView<>), typeof(ABI.System.Collections.Generic.IReadOnlyList<>));
             CustomTypeToHelperTypeMappings.Add(typeof(global::Microsoft.UI.Xaml.Interop.IBindableVector), typeof(ABI.System.Collections.IList));
+            CustomTypeToHelperTypeMappings.Add(typeof(IIterable<>), typeof(global::ABI.System.Collections.Generic.IEnumerable<>));
```
Scottj1s, line 73: what bug was this fixing?  should we now remove the above registrations?
            RegisterCustomAbiTypeMappingNoLock(typeof(IEnumerable<>), typeof(ABI.System.Collections.Generic.IEnumerable<>), "Windows.Foundation.Collections.IIterable`1");
            RegisterCustomAbiTypeMappingNoLock(typeof(IEnumerator<>), typeof(ABI.System.Collections.Generic.IEnumerator<>), "Windows.Foundation.Collections.IIterator`1");

WinRT.Runtime/Projections/IDictionary.net5.cs
```diff
@@ -14,12 +14,13 @@
...
 namespace Windows.Foundation.Collections
 {
+    [global::WinRT.WindowsRuntimeType]
     [Guid("3C2925FE-8519-45C1-AA79-197B6718C1C1")]
     interface IMap<K, V> : IIterable<IKeyValuePair<K, V>>
     {
         V Lookup(K key);
         bool HasKey(K key);
-        IMapView<K, V> GetView();
+        IReadOnlyDictionary<K, V> GetView();
```
Scottj1s, line 23: the pattern in the PR seems to be to return nested projected types from WinRT interfaces:
IIterable->IEnumerator
IMap->IReadOnlyDictionary

what's the reason for that?

Scottj1s:
you mention tests now passing -
can those be included in the PR?
should we add IVector.GetView, IMap.GetView test cases to TestComponent, assert that they're equivalent to the source collections?

WinRT.Runtime/Projections.cs
ujjwalchadha: This fixes a NullReferenceException that occurs while registering the helper type of IIterable in the FromManaged call when the native code asks for it

WinRT.Runtime/Projections/IDictionary.net5.cs
ujjwalchadha: As mentioned in the description, this is required to enable the managed code to return a reference of the managed object to the native side instead of an instance of `ToAbiHelper` which indirectly points to the managed object.

TestComponentCSharp/TestComponentCSharp.idl
```diff
@@ -341,6 +341,31 @@ namespace TestComponentCSharp
         void Observe(Microsoft.UI.Xaml.Interop.IBindableObservableVector vector);
     }
 
+    //[default_interface]
```
Scottj1s, line 344: cruft?

TestComponentCSharp/TestComponentCSharp.idl
```diff
@@ -341,6 +341,31 @@ namespace TestComponentCSharp
...
+    //runtimeclass IBindableVectorTest : Microsoft.UI.Xaml.Interop.IBindableVector
+    //{
+    //    //IBindableVectorTest();
+    //}
+
+    [default_interface]
```
Scottj1s, line 350: these runtime classes should not start with 'I' (implies interface)

TestComponentCSharp/ManualProjectionTestClasses.h
```diff
@@ -0,0 +1,68 @@
...
+		IDisposableTest();
+		void Close();
+	};
+}
+
+namespace winrt::TestComponentCSharp::factory_implementation
+{
+	struct IBindableIteratorTest : IBindableIteratorTestT<IBindableIteratorTest, implementation::IBindableIteratorTest, Windows::Foundation::IStringable>
+	{
+		hstring ToString()
```
Scottj1s, line 47: looks like all these ToString overrides were copied from ComImports (same for NonAgileClass)?  They should be removed, since we're not testing them.

TestComponentCSharp/TestComponentCSharp.idl
```diff
@@ -341,6 +341,31 @@ namespace TestComponentCSharp
...
+    runtimeclass IBindableIteratorTest : Microsoft.UI.Xaml.Interop.IBindableIterator
+    {
+        IBindableIteratorTest();
+    }
+
+    [default_interface]
```
Scottj1s, line 356: shouldn't this be in TestWinRT.TestComponent?

UnitTest/TestComponentCSharp_Tests.cs
```diff
@@ -2173,5 +2173,53 @@ public void TestNonProjectedRuntimeClass()
...
+        public void TestIBindableIterator()
+        {
+            IBindableIteratorTest bindableIterator = new IBindableIteratorTest();
+            Assert.True(bindableIterator.MoveNext());
+            Assert.True(bindableIterator.HasCurrent);
+            Assert.Equal(27861, bindableIterator.Current);
+        }
+
+        [Fact]
+        public void TestIDictionary()
```
Scottj1s, line 2187: move to TestComponent?

TestComponentCSharp/ManualProjectionTestClasses.cpp
```diff
@@ -0,0 +1,104 @@
+#include "pch.h"
+#include "ManualProjectionTestClasses.h"
+#include "IBindableIteratorTest.g.cpp"
+#include "IDictionaryTest.g.cpp"
+#include "IDisposableTest.g.cpp"
+
+namespace winrt
```
Scottj1s, line 7: consider moving from here and class.cpp into pch.h, with a TODO to replace with latest cppwinrt

TestComponentCSharp/TestComponentCSharp.idl
ujjwalchadha: Yup. Actually, I think we can get rid of it now because I think I added the tests for IDictionary APIs in the existing Dictionary Test in TestWinRT

WinRT.Runtime/Projections/IDictionary.net5.cs
```diff
@@ -14,12 +14,13 @@
...
 namespace Windows.Foundation.Collections
 {
+    //Need to rethink how to name/define this interface
     [Guid("3C2925FE-8519-45C1-AA79-197B6718C1C1")]
     interface IMap<K, V> : IIterable<IKeyValuePair<K, V>>
     {
         V Lookup(K key);
         bool HasKey(K key);
-        IMapView<K, V> GetView();
+        IReadOnlyDictionary<K, V> GetView();
```
Scottj1s, line 23: comment on this line specifically - that combining IMap & IReadOnlyDictionary needs redesign

WinRT.Runtime/Projections/IEnumerable.net5.cs
```diff
@@ -15,11 +15,14 @@
 
 namespace Windows.Foundation.Collections
 {
+
     [Guid("FAA585EA-6214-4217-AFDA-7F46DE5869B3")]
     internal interface IIterable<T>
     {
-        IIterator<T> First();
+        IEnumerator<T> First();
```
Scottj1s, line 22: comment that combining IIterable & IEnumerator needs redesign

WinRT.Runtime/Projections/IEnumerable.netstandard2.0.cs
```diff
@@ -15,11 +15,14 @@
 
 namespace Windows.Foundation.Collections
 {
+
     [Guid("FAA585EA-6214-4217-AFDA-7F46DE5869B3")]
     internal interface IIterable<T>
     {
-        IIterator<T> First();
+        IEnumerator<T> First();
```
Scottj1s, line 22: comment that combining IIterable & IEnumerator needs redesign

WinRT.Runtime/Projections/IList.net5.cs
```diff
@@ -14,11 +14,12 @@
 
 namespace Windows.Foundation.Collections
 {
+
     [Guid("913337E9-11A1-4345-A3A2-4E7F956E222D")]
     interface IVector<T> : IIterable<T>
     {
         T GetAt(uint index);
-        IVectorView<T> GetView();
+        IReadOnlyList<T> GetView();
```
Scottj1s, line 22: comment that combining IVector & IReadOnlyList needs redesign

WinRT.Runtime/Projections/IList.netstandard2.0.cs
```diff
@@ -14,11 +14,12 @@
 
 namespace Windows.Foundation.Collections
 {
+
     [Guid("913337E9-11A1-4345-A3A2-4E7F956E222D")]
     interface IVector<T> : IIterable<T>
     {
         T GetAt(uint index);
-        IVectorView<T> GetView();
+        IReadOnlyList<T> GetView();
```
Scottj1s, line 22: comment needs redesign

TestComponentCSharp/ManualProjectionTestClasses.cpp
ujjwalchadha: TestWinRT already has a test for IDictionary but it did not test the APIs (Didn't call all the methods on it). I made this component initially but later added all the  API tests in the existing test in TestComponent.cs

Scottj1s (changes requested):

WinRT.Runtime/Projections/IReadOnlyDictionary.net5.cs
```diff
@@ -349,9 +349,15 @@ private ConstantSplittableMap(global::System.Collections.Generic.KeyValuePair<K,
                     return kvArray;
                 }
 
+                public uint Size => (uint)(lastItemIndex - firstItemIndex + 1);
+
+                public global::System.Collections.Generic.IEnumerable<K> Keys => throw new NotImplementedException();
```
manodasanW, line 354: Can this split map be passed to C# for a Dictionary parameter and be expected to respond to the Keys and Values property?

WinRT.Runtime/Projections.cs
Scottj1s: They should be left in place and the non-public registrations on CustomTypeToHelperTypeMappings removed - we can revisit later.

TestComponentCSharp/ManualProjectionTestClasses.cpp
Scottj1s: your response seems to be unrelated to my question, which is about single_threaded_vector_view, single_threaded_map_view

Scottj1s (approved):

WinRT.Runtime/Projections/IReadOnlyDictionary.net5.cs
```diff
@@ -351,9 +351,31 @@ private ConstantSplittableMap(global::System.Collections.Generic.KeyValuePair<K,
...
 
-                public global::System.Collections.Generic.IEnumerable<K> Keys => throw new NotImplementedException();
+                public global::System.Collections.Generic.IEnumerable<K> Keys  
+                {
+                    get
+                    {
+                        K[] keys = new K[items.Length];
+                        for (var i = 0; i < items.Length; i++)
+                        {
+                            keys[i] = items[i].Key;
```
manodasanW, line 361: Should you take into account firstItemIndex and lastItemIndex?

WinRT.Runtime/Projections/IReadOnlyDictionary.net5.cs
```diff
@@ -351,9 +351,31 @@ private ConstantSplittableMap(global::System.Collections.Generic.KeyValuePair<K,
...
+                    }
+                }
 
-                public global::System.Collections.Generic.IEnumerable<V> Values => throw new NotImplementedException();
+                public global::System.Collections.Generic.IEnumerable<V> Values
+                {
+                    get
+                    {
+                        V[] values = new V[items.Length];
+                        for (var i = 0; i < items.Length; i++)
```
manodasanW, line 372: Also here.

WinRT.Runtime/Projections/IReadOnlyDictionary.net5.cs
```diff
@@ -351,9 +351,31 @@ private ConstantSplittableMap(global::System.Collections.Generic.KeyValuePair<K,
 
                 public uint Size => (uint)(lastItemIndex - firstItemIndex + 1);
 
-                public global::System.Collections.Generic.IEnumerable<K> Keys => throw new NotImplementedException();
+                public global::System.Collections.Generic.IEnumerable<K> Keys
```
manodasanW, line 354: I wonder if we should modify DictionaryValueCollection to work also for this scenario and make use of it here similar to in IDictionary.  A couple reasons why I am thinking that is technically Keys and Values collection apparently cannot be modified which DictionaryValueCollection handles, but I think below we will allow that.  And if someone does modify it now, I wonder if we can get into bad states.

PR #568, opened 2020-11-08
Use custom marshalers when type is erased with MarshalInspectable<object>

### LINKED ISSUE #567: Custom marshalers are not being used when marshaling as object
When a public type with a custom marshaler is marshaled as an object, it's not being correctly marshaled.  E.g., passing a System.Uri to a property of type Object causes a generic CCW to be created around the Uri, versus marshaling it to Windows::Foundation::Uri, causing cast exception in native code.

Scottj1s:
fixes #567

### DIFF
--- a/TestComponentCSharp/Class.cpp
+++ b/TestComponentCSharp/Class.cpp
@@ -442,6 +442,10 @@ namespace winrt::TestComponentCSharp::implementation
     }
     void Class::ObjectProperty(WF::IInspectable const& value)
     {
+        if (auto uri = value.try_as<Windows::Foundation::Uri>())
+        {
+            _uri = uri;
+        }
         _object = value;
     }
     void Class::RaiseObjectChanged()
--- a/WinRT.Runtime/Marshalers.cs
+++ b/WinRT.Runtime/Marshalers.cs
@@ -932,13 +932,19 @@ namespace WinRT
                 return null;
             }
 
-            if (o is Uri uri)
-            {
-                return ABI.System.Uri.CreateMarshaler(uri);
-            }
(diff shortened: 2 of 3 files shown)


WinRT.Runtime/Marshalers.cs
```diff
@@ -936,6 +936,16 @@ public static IObjectReference CreateMarshaler(T o, bool unwrapObject = true)
             {
                 return objRef.As<IInspectable.Vftbl>();
             }
+            var publicType = o.GetType();
+            Type helperType = Projections.FindCustomHelperTypeMapping(publicType, true);
+            if(helperType != null)
+            {
+                var parms = new[] { Expression.Parameter(typeof(object), "arg") };
+                var createMarshaler = Expression.Lambda<Func<object, IObjectReference>>(
```
manodasanW, line 944: We may want to open an issue to look at how expensive this is both here and below in the FromAbi case and whether we should consider caching it to improve performance.

manodasanW (approved):

TestComponentCSharp/Class.cpp
```diff
@@ -432,6 +432,10 @@ namespace winrt::TestComponentCSharp::implementation
     }
     void Class::ObjectProperty(WF::IInspectable const& value)
     {
+        if (auto uri = value.try_as<Windows::Foundation::Uri>())
```
stevenbrix, line 435: Is this code necessary for the Uri to be marshalled properly, or is this just a test?

Scottj1s: this is a test to ensure that a Uri passed as Object is still properly marshalled, such that we can cast it to IUriRuntimeClass.  (that was the original bug)

PR #573, opened 2020-11-10
Fixed ConstantSplittableMap's Keys and Values to return read only col…

ujjwalchadha:
…lections


WinRT.Runtime/Projections/IReadOnlyDictionary.netstandard2.0.cs
```diff
@@ -359,28 +361,12 @@ private ConstantSplittableMap(global::System.Collections.Generic.KeyValuePair<K,
...
-                    get
-                    {
-                        K[] keys = new K[items.Length];
-                        for (var i = 0; i < items.Length; i++)
-                        {
-                            keys[i] = items[i].Key;
-                        }
-                        return keys;
-                    }
+                    get => new ReadOnlyDictionaryKeyCollection(this);
```
manodasanW, line 364: I assume you want to do same for net5?

manodasanW (approved):

PR #577, opened 2020-11-11
Update sample for net5 GA

angelazhangmsft:
- remove compilers.toolset package
- update readme requirements
- add windowsmetadata property
- remove ".sln" workaround (bug fixed in VS 16.8)

### DIFF
--- a/Samples/Net5ProjectionSample/Directory.Build.props
+++ b/Samples/Net5ProjectionSample/Directory.Build.props
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup>
-    <BuildOutDir>$([MSBuild]::NormalizeDirectory('$(SolutionDir.Replace("$(SolutionFileName)\", ""))', '_build', '$(Platform)', '$(Configuration)'))</BuildOutDir>
+    <BuildOutDir>$([MSBuild]::NormalizeDirectory('$(SolutionDir)', '_build', '$(Platform)', '$(Configuration)'))</BuildOutDir>
     <OutDir>$([MSBuild]::NormalizeDirectory('$(BuildOutDir)', '$(MSBuildProjectName)', 'bin'))</OutDir>
     <IntDir>$([MSBuild]::NormalizeDirectory('$(BuildOutDir)', '$(MSBuildProjectName)', 'obj'))</IntDir>
   </PropertyGroup>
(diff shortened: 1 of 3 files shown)


Samples/Net5ProjectionSample/SimpleMathProjection/SimpleMathProjection.csproj
```diff
@@ -24,17 +25,4 @@
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
   </PropertyGroup>
 
-  <!--The C# compiler package below is not needed with Visual Studio 16.8 Preview 4-->
-  <ItemGroup>
-    <PackageReference Include="Microsoft.Net.Compilers.Toolset" Version="3.8.0-4.20472.6" />
-    <PackageReference Include="Microsoft.Windows.SDK.Contracts" Version="10.0.19041.1" />
```
manodasanW, line 30: Do you want to remove the SDK contracts package?

angelazhangmsft: Yes - I think since I added <CsWinRTWindowsMetadata>10.0.19041.0<CsWinRTWindowsMetadata> to the project file, we don't need SDK contracts package. Not sure what's the preferred way to specify windows metadata but specifying the CsWinRT specific property seems to make more sense?

manodasanW: Well we support multiple ways to specify the windows SDK metadata.
- Via the installed Windows SDK which is what setting CsWinRTWindowsMetadata currently does
- Via the SDK contracts package which we were doing before
- And hopefully soon via the TFM which probably becomes the default way if a Windows SDK TFM is used.

I am not sure which version of the Windows SDK is installed on the build machines, so we may want to consider keeping the contracts package.  We can also try doing a build for this branch and seeing what happens.

manodasanW (approved):

Samples/Net5ProjectionSample/Directory.Build.props
```diff
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup>
-    <BuildOutDir>$([MSBuild]::NormalizeDirectory('$(SolutionDir.Replace("$(SolutionFileName)\", ""))', '_build', '$(Platform)', '$(Configuration)'))</BuildOutDir>
+    <BuildOutDir>$([MSBuild]::NormalizeDirectory('$(SolutionDir)', '_build', '$(Platform)', '$(Configuration)'))</BuildOutDir>
```
Scottj1s, line 4: ok, but generally, we don't remove workarounds (especially if they carry no risk) - users could still try to open the sample in the version of VS with the break.

Scottj1s (approved):

PR #586, opened 2020-11-17
Update interop.md

Scottj1s:
elaborated com import requirements

### DIFF
--- a/docs/interop.md
+++ b/docs/interop.md
@@ -1,7 +1,7 @@
 # COM Interop Guide
 
 ## Overview
-With .NET 5, direct support of WinRT has been removed from the C# language and runtime, including most of the functionality previously provided in [System.Runtime.InteropServices](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices?view=netcore-3.1).  In some cases, C#/WinRT provides equivalent functionality. In other cases, functionality may no longer be supported. This article provides a migration guide for interop scenarios in C#/WinRT.  
+With .NET 5, direct support of WinRT has been removed from the C# language and CLR, including most of the functionality previously provided in [System.Runtime.InteropServices](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices?view=netcore-3.1).  In some cases, C#/WinRT provides equivalent functionality. In other cases, functionality may no longer be supported. This article provides a migration guide for interop scenarios in C#/WinRT.  
 
 ## Summary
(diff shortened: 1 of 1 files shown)


docs/interop.md
```diff
@@ -71,35 +71,39 @@ TInterop interop = objRef.AsInterface<TInterop>();
 ```
 
 ### Custom Marshaling
-When defining a [ComImport](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.comimportattribute?view=netcore-3.1) interop interface, marshaling of WinRT types must be manual (not using the runtime).  
+**Note:** When defining a [ComImport](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.comimportattribute?view=netcore-3.1) interop interface, WinRT parameters and return values must be passed by their ABI types, and marshaling must be done manually (not using the CLR).  For example, reference types like strings and interfaces must be passed as IntPtr.  Blittable value types can be passed directly. Non-blittable value types must have separate ABI and projected definitions, and marshaling between these values must be done manually.
```
manodasanW, line 74: Do we need our own documentation (list out) on what we consider a Blittable type or does .NET documentation cover it?

docs/interop.md
```diff
@@ -71,35 +71,39 @@ TInterop interop = objRef.AsInterface<TInterop>();
...
 ```
 
 #### IInspectables
-The runtime no longer supports marshaling IInspectable-based ComImport interfaces. But these can be approximated with IUnknown:
+**Note:** The CLR no longer supports marshaling IInspectable-based ComImport interfaces. Casting to an IInspectable interface will succeed, but method calls will crash in the CLR. Instead, IInspectable interfaces can be approximated with IUnknown, by explicitly defining the GetIids, GetRuntimeClassName, and GetTrustLevel methods:
 ```csharp
 [ComImport]
 [InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
 [Guid("68B3A2DF-8173-539F-B524-C8A2348F5AFB")]
 internal unsafe interface IServiceProviderInterop
```
manodasanW, line 103: Should we have a before and after?

docs/interop.md
Scottj1s: added link to blittable .NET docs

docs/interop.md
Scottj1s: added a bit of before/after code & docs

docs/interop.md
```diff
@@ -108,8 +114,21 @@ internal unsafe interface IServiceProviderInterop
...
+
+// Cast a class or object to an interop interface using the .As<>() method described above
+var serviceProvider = someObject.As<IServiceProviderInterop>();
+IntPtr service;
+serviceProvider.GetService(IntPtr.Zero, out service);
+// See Create RCW section below for marshaling an object pointer from the ABI 
+
 ```
 
 ## Create RCW
```
manodasanW, line 134: Do we want to add details on when the respective FromAbi function should be used (i.e. ABI.Namespace.T.FromABI vs MarshalGeneric<T>.FromABI).   The other 2 are pretty obvious.

manodasanW (approved):

PR #588, opened 2020-11-18
Update StreamOperationsImplementation

### LINKED ISSUE #580: NetFxToWinRtStreamAdapter.WriteAsync fails with ArgumentOutOfRangeException when buffer is empty
NOTE: This bug was originally filed [here](https://github.com/dotnet/runtime/issues/31564), but is was never fixed, and I was surprised that exactly the same code existed in this new repo. The bug is reproduceable on just released .Net 5. So I'm duplicating that issue here with hope it will be fixed eventually.

This bug exists since .Net Framework implementation. Basically, when a WinRT stream is an adapter around a .Net `Stream` and you attempt to write an empty **native** `IBuffer` to that stream, you get an `ArgumentOutOfRangeException`. Here is an example (.Net Core 3.0 console app):

``` csharp
static async Task Main()
{
    using var fileStream = File.OpenWrite("TestFile.txt");
    using var winRtStream = fileStream.AsOutputStream();

    // This works OK.
    await fileStream.WriteAsync(ReadOnlyMemory<byte>.Empty);
    Console.WriteLine("fileStream.WriteAsync completed OK");

    var winRtbuffer = new Windows.Storage.Streams.Buffer(capacity: 0);

    // This fails!
    await winRtStream.WriteAsync(winRtbuffer);
    Console.WriteLine("winRtStream.WriteAsync completed OK");
}
```
This gives the following when running:

```
fileStream.WriteAsync completed OK
Unhandled exception. System.ArgumentOutOfRangeException: Positive number required. (Parameter 'bufferSize')
Actual value was 0.
   at System.IO.StreamHelpers.ValidateCopyToArgs(Stream source, Stream destination, Int32 bufferSize)
   at System.IO.Stream.CopyToAsync(Stream destination, Int32 bufferSize, CancellationToken cancellationToken)
   at Windows.Storage.Streams.StreamOperationsImplementation.<>c__DisplayClass2_0.<<WriteAsync_AbstractStream>b__1>d.MoveNext()
```
Unless I'm missing something, there is no point in not supporting empty buffer writes. Actually, if the `IBuffer` is a wrapper for managed memory, all works fine (because internally the write is done differently in this case). 

It's easy to find the root of the problem. [Here](https://github.com/microsoft/CsWinRT/blob/66da8e1d1eb84cb962a1d0fbcea688a1e2838eff/cswinrt/strings/additions/Windows.Storage.Streams/StreamOperationsImplementation.cs#L205)  `Stream.CopyAsync` method is used, which doesn't support zero buffer length and crashes. The code above that line doesn't check if `bytesToWrite` is zero and uses it as the buffer size:
```csharp
 int buffSize = 0x4000;
 if (bytesToWrite < buffSize)
      buffSize = (int)bytesToWrite;
```

While it may seem silly to try to write an empty buffer to a stream, Writer code may calculate the amount to write dynamically and not check for zero length. Actually, I have a real world scenario in which this bug affects me: I'm using UWP's `Windows.Media.Transcoding.MediaTranscoder` API to transcode audio files, passing it .Net file streams. `Transcoder` often tries to write an empty buffer, so the code almost always crashes.

j0shuams:
Adds a condition when writing to a WinRT buffer to skip copying when 0 bytes to copy. 
Also adds a unit test for this case that was failing previously. 

Fixes #580

### DIFF
--- a/cswinrt/strings/additions/Windows.Storage.Streams/StreamOperationsImplementation.cs
+++ b/cswinrt/strings/additions/Windows.Storage.Streams/StreamOperationsImplementation.cs
@@ -201,8 +201,9 @@ namespace Windows.Storage.Streams
                     int buffSize = 0x4000;
                     if (bytesToWrite < buffSize)
                         buffSize = (int)bytesToWrite;
-
-                    await dataStream.CopyToAsync(stream, buffSize, cancelToken).ConfigureAwait(continueOnCapturedContext: false);
+                    
+                    if (buffSize > 0)
+                        await dataStream.CopyToAsync(stream, buffSize, cancelToken).ConfigureAwait(continueOnCapturedContext: false);
 
                     if (progressListener != null)
                         progressListener.Report((uint)bytesToWrite);
--- a/UnitTest/TestComponentCSharp_Tests.cs
+++ b/UnitTest/TestComponentCSharp_Tests.cs
@@ -318,6 +318,22 @@ namespace UnitTest
         }
 
 #if NET5_0
+        async Task InvokeStreamWriteZeroBytes()
+        {
+            var random = new Random(42);
+            byte[] data = new byte[256];
+            random.NextBytes(data);
+
+            using var stream = new InMemoryRandomAccessStream().AsStream();
+            await stream.WriteAsync(data, 0, 0);
+            await stream.WriteAsync(data, data.Length, 0);
+        }
+
+        [Fact]
+        public void TestStreamWriteZeroByte()
+        {
+            Assert.True(InvokeStreamWriteZeroBytes().Wait(1000));
+        }
 
         async Task InvokeStreamWriteAsync()
         {
--- a/UnitTest/TestComponentCSharp_Tests.cs
+++ b/UnitTest/TestComponentCSharp_Tests.cs
@@ -318,6 +318,24 @@ namespace UnitTest
         }
 
 #if NET5_0
+
+        async Task InvokeStreamWriteAsync()
+        {
+            using var fileStream = File.OpenWrite("TestFile.txt");
+            using var winRTStream = fileStream.AsOutputStream();
+
+            var winRTBuffer = new Windows.Storage.Streams.Buffer(capacity: 0);
+
+            await winRTStream.WriteAsync(winRTBuffer);
+            Assert.True(true);
+        }
+
+        [Fact]
+        public void TestStreamWriteAsync()
+        {
+            Assert.True(InvokeStreamWriteAsync().Wait(1000));
+        }
+
         [Fact]
         public void TestAsStream()
         {


cswinrt/strings/additions/Windows.Storage.Streams/StreamOperationsImplementation.cs
```diff
@@ -201,8 +201,9 @@ internal static class StreamOperationsImplementation
                     int buffSize = 0x4000;
```
manodasanW, line 201: To confirm, WriteAsync in the if part doesn't have issues with 0 bytesToWrite right?

j0shuams: Do you mean CopyToAsync? It does, it is not designed to support 0 bytes

manodasanW: I was referring to a couple lines above:                     await stream.WriteAsync(data, offset, bytesToWrite, cancelToken).ConfigureAwait(continueOnCapturedContext: false);

j0shuams: Yes, I believe this method is testing that WriteAsync: https://github.com/microsoft/CsWinRT/blob/3a28fb4fcc240bae8f89145fb028e2b63226d596/UnitTest/TestComponentCSharp_Tests.cs#L337

I added the same line but specified writing 0 bytes to front and end of the array, and no errors were thrown. 

I'm adding a unit test for that as well

manodasanW (approved):

cswinrt/strings/additions/Windows.Storage.Streams/StreamOperationsImplementation.cs
```diff
@@ -201,8 +201,9 @@ internal static class StreamOperationsImplementation
...
                     if (bytesToWrite < buffSize)
                         buffSize = (int)bytesToWrite;
-
-                    await dataStream.CopyToAsync(stream, buffSize, cancelToken).ConfigureAwait(continueOnCapturedContext: false);
+                    
+                    if (buffSize > 0)
```
Scottj1s, line 205: is the buffSize check needed?  what happens without it?

manodasanW: Without it, the crash in the bug this PR is fixing will hit  I believe [this](https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.copytoasync?view=net-5.0#System_IO_Stream_CopyToAsync_System_IO_Stream_System_Int32_System_Threading_CancellationToken_) is datastream.CopyToAsync which indicates it throws when zero.

Scottj1s (approved):

PR #595, opened 2020-11-19
Add docs on authoring components

j0shuams:
When working on an issue related to authoring components, I took notes on things I was doing to get it working. This doc is an effort to share these and help any customers get started quickly.


docs/authoring.md
```diff
@@ -0,0 +1,108 @@
...
+    
+    <!-- the manifest -->
+    <Manifest Include="PosnConsole.exe.manifest">
+      <DeploymentContent>true</DeploymentContent>
+    </Manifest>
+</ItemGroup> 
+```
+
+### Manifest and RuntimeConfig
+You'll need to author some files to assist the hosting process by the native app: `YourNativeApp.exe.manifest` and `WinRT.Host.runtimeconfig.json`.
```
AdamBraden, line 65: The manifest is only required for unpackaged apps, correct? If so, start with the WinRT.Host.runtimeconfig.json. Then add a comment like: "if your app is not packaged with MSIX then be sure to include your activatable class registrations in a fusion manifest file 'YourNativeApp.exe.manifest'".

docs/authoring.md
```diff
@@ -0,0 +1,108 @@
...
+  <PropertyGroup>
+    <!-- The Windows SDK metadata will need to update as new versions release -->
+    <CsWinRTWindowsMetadata>10.0.19041.0</CsWinRTWindowsMetadata>
+    <CsWinRTComponent>true</CsWinRTComponent>
+    <CsWinRTEnableLogging>true</CsWinRTEnableLogging>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <GeneratedFilesDir Condition="'$(GeneratedFilesDir)'==''">$([MSBuild]::NormalizeDirectory('$(MSBuildProjectDirectory)', '$(IntermediateOutputPath)', 'Generated Files'))</GeneratedFilesDir>
+  </PropertyGroup>
+```
+And don't forget to include a `PackageReference` to `Microsoft.Windows.CsWinRT`. You can add the nuget feed found in the C#/WinRT README as a nuget source, and install the package in Visual Studio via the `Manage NuGet Packages for Solution` option.
```
manodasanW, line 21: I don't think you need the nuget feed part as our packages are on nuget.org.

docs/authoring.md
```diff
@@ -0,0 +1,108 @@
...
+  </Target>  
+  
+</Project>
+
+```
+
+## References
+Here are some resources that demonstrate authoring C#/WinRT components and the changes discussed above.
+1. https://github.com/AdamBraden/MyRandom
+2. https://github.com/microsoft/CsWinRT/tree/master/Samples/Net5ProjectionSample
```
angelazhangmsft, line 107: This sample shows generating the projection assembly for a C++/WinRT component, don't think it has to do with authoring?

docs/authoring.md
```diff
@@ -0,0 +1,108 @@
...
+```
+And don't forget to include a `PackageReference` to `Microsoft.Windows.CsWinRT`. You can add the nuget feed found in the C#/WinRT README as a nuget source, and install the package in Visual Studio via the `Manage NuGet Packages for Solution` option. 
+
+
+## Known Authoring Issues
+1. There are some programs you could write as/in your component that aren't available in the Windows Runtime. 
+We are working on implementing diagnostics in our tool that will catch these errors before a winmd is generated for your component.
+
+2. Not all C# types have been mapped to runtime types, but we are working on completing this coverage. 
+
```
manodasanW, line 29: 3. Composable support

docs/authoring.md
```diff
@@ -0,0 +1,108 @@
...
+  1. update the consuming app's project file (e.g. `YourApp.vcxproj`)
+  2. add a manifest file and runtimeconfig file 
+  3. add a target for copying the required DLLs.
+
+
+### Project File
+The following item group demonstrates the needed modifications to the project file (1. above):
+
+```
+<ItemGroup>
```
manodasanW, line 44: Might want to convert these to UI steps as it isn't a direct copy paste for folks to follow.

docs/authoring.md
```diff
@@ -0,0 +1,108 @@
...
+In your C++ app, add a `Directory.Build.targets` file that copies over the necessary DLLs: 
+```
+<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <PropertyGroup>
+    <PrepareForRunDependsOn>CopyHostAssets;$(PrepareForRunDependsOn)</PrepareForRunDependsOn>
+  </PropertyGroup>
+  
+  <Target Name="CopyHostAssets">
+    <Copy SourceFiles="Path\To\Nuget\Packages\microsoft.windows.cswinrt\DownloadedVersion\native\$(Platform)\WinRT.Host.dll"
```
manodasanW, line 79: May want to make DownloadedVersion a variable.

docs/authoring.md
manodasanW: Well even for packaged scenarios, I think we may still need to do tooling work to make it do the right thing like specify our host.

docs/authoring.md
j0shuams: I added "composable type support is still in progress" -- do you have something different you want to say here?

docs/authoring.md
j0shuams: addressed this, but might not be exactly what you had in mind

docs/authoring.md
j0shuams: Don't quite understand what you mean by "UI steps", but I got rid of the item group and gave a more general description. Let me know what you think of it.

docs/authoring.md
```diff
@@ -0,0 +1,93 @@
+# Authoring Components
+
+## Overview
+C#/WinRT provides support for authoring Windows Runtime components. You can write a library in C#, and use C#/WinRT's source generator to get a winmd that any WinRT compatible language can use. For example, a library written in C# can be used by a C++ program, via C#/WinRT and C++/WinRT, with just a few tweaks to the C++ project.
+
+
+## Authoring the C# Component
+To create a library, select the Class Library (.NET Core) template in Visual Studio. 
+Make sure the `<TargetFramework>` for the project is `net5.0-windows10.0.19041.0`.
```
angelazhangmsft, line 9: For the TFM, it could also target net5.0-windows10.0.18362.0 and net5.0-windows10.0.17763.0 correct?

j0shuams: Yeah that's true. How do we describe the need for specifying `net5.0-windows10.0.xxxxx.0` instead of just `net5.0` ?

j0shuams:
It seems if users don't specify `AssemblyVersion` in their authored C# component, the CsWinRT SourceGenerator will throw an exception and the compilation fails. Should we have some kind of target/property that sets this property to `1.0.0.0` if it is empty?

docs/authoring.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Known Authoring Issues
+1. There are some programs you could write as/in your component that aren't available in the Windows Runtime. 
+We are working on implementing diagnostics in our tool that will catch these errors before a winmd is generated for your component.
+
+2. Not all C# types have been mapped to runtime types, but we are working on completing this coverage. 
+
+3. Composable type support is still in progress
+
+## Using the authored component in C++
```
angelazhangmsft, line 32: Is using the authored component in a C# app also possible or is that not relevant? and would there be different modifications for that scenario or just adding a reference to the winmd?

docs/authoring.md
angelazhangmsft: For UI steps:

Manifest and Runtime Config - Right click on the project node under Solution Explorer, click Add -> New Item. Search for the "Text File" template and name your file "YourNativeApp.exe.manifest" (and repeat the same for WinRT.Host.runtimeconfig.json).

Project File - Right click on "References" under the project node, click Add Reference and Browse for the winmd file generated by the authoring project.

also maybe rename Project File to "Add reference to component" or something similar?

docs/authoring.md
j0shuams: I think it is possible; I believe that is what @AdamBraden is doing [here](https://github.com/AdamBraden/MyRandom/blob/64676c3b9f2a16fde36c158904e343626814e284/CsTestApp/CsTestApp.csproj#L9)

Which makes me think the consuming C# app just needs a project reference to the authored component? Or perhaps a package reference if they aren't to live in the same directory/tree

docs/authoring.md
```diff
@@ -0,0 +1,94 @@
...
```
angelazhangmsft, line 32: add #include statement in pch.h file

docs/authoring.md
```diff
@@ -0,0 +1,146 @@
...
+You'll need to author some files to assist the hosting process by the native app: `YourNativeApp.exe.manifest` and `WinRT.Host.runtimeconfig.json`. 
+
+To do this, **in Visual Studio**, right click on the project node on the "Solution Explorer" window, click "Add", then "New Item". Search for the "Text File" template and name your file "YourNativeApp.exe.manifest".
+Repeat this for the "WinRT.Host.runtimeconfig.json" file. 
+
+This process adds the nodes `<Manifest Include=... >` and `<None Include=... >` to your native app's project file -- **you need to update these to have `<DeploymentContent>true</DeploymentContent>` for them to be placed in the output directory with your executable**.  
+
+You should read the [hosting docs](https://github.com/microsoft/CsWinRT/blob/master/docs/hosting.md) as well, for more information on these files.
+
+You'll need to use [C++/WinRT](https://docs.microsoft.com/en-us/windows/uwp/cpp-and-winrt-apis/intro-to-using-cpp-with-winrt) to consume the `winmd` too, so make sure you have it installed, and have added `#include <winrt/YourAuthoredLibrary.h>` to the file `pch.h` of the native app.
```
j0shuams, line 70: @angelazhangmsft is this good?

docs/authoring.md
manodasanW: C# apps can consume it and it is just a regular project/package reference like they would for any other .NET library.

docs/authoring.md
```diff
@@ -0,0 +1,147 @@
...
+
+
+## References
+Here are some resources that demonstrate authoring C#/WinRT components and the details discussed in this document.
+1. https://github.com/microsoft/CsWinRT/tree/master/Authoring/AuthoringConsumptionTest
+2. https://github.com/AdamBraden/MyRandom
+
+
+## Authoring the C# Component
+To create a library, select the Class Library (.NET Core) template in Visual Studio. C#/WinRT projects require SDK specific .NET frameworks.
```
manodasanW, line 14: Consider SDK specific -> Windows API version specific

docs/authoring.md
```diff
@@ -0,0 +1,147 @@
...
+  </PropertyGroup>
+```
+And don't forget to include a `PackageReference` to `Microsoft.Windows.CsWinRT`!
+
+
+## Known Authoring Issues
+1. There are some programs you could write as/in your component that aren't available in the Windows Runtime. 
+We are working on implementing diagnostics in our tool that will catch these errors before a winmd is generated for your component.
+
+2. Not all C# types have been mapped to runtime types, but we are working on completing this coverage.
```
manodasanW, line 40: runtime -> WinRT.  Not all mappable native C# types haves have been mapped to WinRT types,...

docs/authoring.md
manodasanW: That is good.

docs/authoring.md
```diff
@@ -0,0 +1,147 @@
...
+Modifications needed: 
+  1. Reference to your authored component in the consuming app
+  2. Add a manifest file and runtimeconfig file [for native consumption only] 
+  3. Use a MSBuild targets for copying necessary DLLs
+
+
+### Add the reference to your authored component
+**In Visual Studio** Under the Project node, right click on "References", click "Add Reference", then "Browse" and add the `.winmd` file generated in your authored component. 
+
+For consumption by native components, you'll also need to add `Include` statements for your hosting `runtimeconfig.json` file and the consuming app's `exe.manifest` file. More information on these files can be found in the hosting docs.
```
manodasanW, line 55: Isn't this what the next section also talks about?

docs/authoring.md
```diff
@@ -0,0 +1,147 @@
...
+  <PropertyGroup>
+    <PrepareForRunDependsOn>CopyHostAssets;$(PrepareForRunDependsOn)</PrepareForRunDependsOn>
+  </PropertyGroup>
+  
+  <PropertyGroup>
+    <CsWinRTVersion>1.1.0</CsWinRTVersion>
+  </PropertyGroup>
+  
+  <Target Name="CopyHostAssets">
+    <Copy SourceFiles="Path\To\Nuget\Packages\microsoft.windows.cswinrt\$(CsWinRTVersion)\native\$(Platform)\WinRT.Host.dll"
```
manodasanW, line 122: $(NuGetPackageRoot) might also work.

manodasanW:
> It seems if users don't specify `AssemblyVersion` in their authored C# component, the CsWinRT SourceGenerator will throw an exception and the compilation fails. Should we have some kind of target/property that sets this property to `1.0.0.0` if it is empty? Or add a note in these docs to add said property to their project?

For now they need to specify it, but there is a default we should detect somehow or set ourselves.

docs/authoring.md
```diff
@@ -0,0 +1,147 @@
...
+    <!-- the manifest -->
+    <Manifest Include="PosnConsole.exe.manifest">
+      <DeploymentContent>true</DeploymentContent>
+    </Manifest>
+  </ItemGroup> 
+```
+
+### Copying DLLs Target
+
+1. In the project folder for the component you are authoring, you will need to update an ItemGroup so MSBuild places the proper WinRT Dlls in the output directory. This is needed by the targets file we need to add to C++ apps to copy the WinRT Dlls to the native app's output directory.
```
manodasanW, line 95: I believe you meant property rather than ItemGroup?

docs/authoring.md
j0shuams: @adambraden [how is this](https://github.com/microsoft/CsWinRT/commit/4c9c94b3c1432ad694fcf296403a5459f1e05b67)

docs/authoring.md
```diff
@@ -0,0 +1,151 @@
...
+
+Accepted `<TargetFramework>` properties |
+--- |
+`net5.0-windows10.0.19041.0` |
+`net5.0-windows10.0.18362.0` |
+`net5.0-windows10.0.17763.0` |
+
+The library you are authoring should specify the following properties in its project file: 
+```
+  <PropertyGroup>
```
angelazhangmsft, line 24: I found I also needed to add the following line to my PropertyGroup, otherwise my hosting cpp app can't find WinRT.Runtime.dll:
<CopyLocalLockFileAssemblies Condition="$(CsWinRTComponent)==true">true</CopyLocalLockFileAssemblies>

angelazhangmsft: nvm - didn't see it was under the section Copying Dlls target. But could also move it to the project file?

docs/authoring.md
```diff
@@ -0,0 +1,151 @@
...
+### Add the reference to your authored component
+**In Visual Studio** Under the Project node, right click on "References", click "Add Reference", then "Browse" and add the `.winmd` file generated in your authored component. 
+
+### For native app (C++) consumption
+As part of the native (C++) support, the WinRT Hosting dlls (WinRT.Host and WinRT.Host.Shim) need to be in the same folder as the native executable. 
+For now, users need a special target of their own so MSBuild can place the hosting dlls in the correct place. But soon we will implement this so that a package reference to C#/WinRT in the authored component is all that is needed.
+
+You'll need to author some files to assist the hosting process by the native app: `YourNativeApp.exe.manifest` and `WinRT.Host.runtimeconfig.json`. 
+
+If your app is packaged with MSIX, then you don't need to the manifest file, otherwise you need to include your activatable class registrations in the manifest file.
```
manodasanW, line 65: don't need to **include** the?

manodasanW (approved):

manodasanW:
You should mention that it is still in preview somewhere.

PR #600, opened 2020-11-20
Consumed inheritance test projects from TestWinRT and added basic test cases

ujjwalchadha:


UnitTest/TestComponent_Tests.cs
```diff
@@ -741,5 +743,22 @@ public void Z_Check_Coverage()
...
+
+        [Fact]
+        public void TestComposable()
+        {
+            HierarchyA hierarchyA = new HierarchyD();
+            hierarchyA.HierarchyA_Method();
+            
+            HierarchyD hierarchyD = new HierarchyD();
+            hierarchyD.HierarchyD_Method();
+
```
manodasanW, line 755: Any tests for overrides and verifying right function got called?

ujjwalchadha: I believe that if there is no error, it has to be the correct method because each class defines just one method at the moment. For example, `HierarchyC` extends from `HierarchyB` but it defines only `HierarchyC_Method`. So if I call `HierarchyB_Method` on a `HierarchyC` object and it succeeds, it has to be the correct method being called because there is just one.

We can however change that and override the methods in TestWinRT and see if the correct methods are being called still. But since this PR was just a porting exercise, this test for now just checks the basic functionality of the ported components. I can add those tests in a future PR after we update TestWinRT.

Scottj1s: I think the emphasis on the C++/WinRT hierarchy tests was to support Fast ABI, and not necessarily to assert that overridden functions are called.  Agree with @manodasanW - let's include assertions of overridden calls in this PR - then we don't have to revisit.

Scottj1s (approved):

Issue #554, opened 2020-10-30
VTable memory should be handled by Type not Object


AaronRobinsonMSFT:
The following table:

https://github.com/microsoft/CsWinRT/blob/fab6067c83e9063a8c0eda387b4c250ca5a4da62/WinRT.Runtime/ComWrappersSupport.net5.cs#L161

stores memory for vtables on the `object` level instead of the `Type` level. This is creating an enormous amount of unnecessary memory pressure on the application. Converting this to be on the `Type` rather than `object` would make CCW creation faster and reduce memory usage.

/cc @jkoritzinsky

AaronRobinsonMSFT:
I was just reminded we created a new API in .NET 5 for this very purpose - [`RuntimeHelpers.AllocateTypeAssociatedMemory()`](https://docs.microsoft.com/dotnet/api/system.runtime.compilerservices.runtimehelpers.allocatetypeassociatedmemory).

PR #601, opened 2020-11-20
Add vtable entry for covariant IEnumerable<object> for WinRT types too.

### LINKED ISSUE #592: System.InvalidCastException on CollectionView enumeration
**Describe the bug**

I create CollectionViewSource and enumerate items in View.

```
CollectionViewSource source = new CollectionViewSource() { Source = new string[] { "str1", "str2", "str3" } };
            foreach(var item in source.View) {
            }
```
Code failed with "System.InvalidCastException: 'Specified cast is not valid.'" exception

image

Exception details:

```
System.InvalidCastException
  HResult=0x80004002
  Message=Specified cast is not valid.
  Source=WinRT.Runtime
  StackTrace:
   at WinRT.ExceptionHelpers.ThrowExceptionForHR(Int32 hr)
```

Call stack:

```
 	winrt.runtime.dll!WinRT.ExceptionHelpers.ThrowExceptionForHR(int hr)
winrt.runtime.dll!ABI.System.Collections.Generic.IEnumerable<object>.Windows.Foundation.Collections.IIterable<object>.First()
 	winrt.runtime.dll!ABI.System.Collections.Generic.IEnumerable<object>.FromAbiHelper.GetEnumerator()
 	winrt.runtime.dll!ABI.System.Collections.Generic.IEnumerable<System.__Canon>.GetEnumerator()
 	winrt.runtime.dll!ABI.System.Collections.Generic.IList<System.__Canon>.FromAbiHelper.GetEnumerator()
 	Microsoft.WinUI.dll!ABI.Microsoft.UI.Xaml.Data.ICollectionView.GetEnumerator()
	AttachedCollections.dll!AttachedCollections.MainWindow.MainWindow() Line 19	C#
```

**Version Info**

Windows 10 1903 (OS Build 18362.1082)

**NuGet package version:**

WinUI 3.0.0-preview2.200713.0

**Windows app type:**

| UWP              | Win32            |
| :--------------- | :--------------- |
|| Yes|

| Device form factor | Saw the problem? |
| :----------------- | :--------------- |
| Desktop            | Yes |

### LINKED ISSUE #598: passing array of derived type to IIterable parameter results in System.InvalidCastException
I tried to call a winrt method that has a parameter of type Windows.Foundation.Collections.IIterable<BaseClass>, where BaseClass is a runtime class.

From C#, calling this method with an array of type BaseClass works as expected. If instead the method is called with an array of type SubClass, where SubClass is a runtime class inheriting from BaseClass, this results in an InvalidCastException.

The same code works as expected when using .NET framework.

manodasanW:
Adding a more complete covariant support.  For covariant interfaces, we get all the possible types that can be passed for the generic and create a vtable entry for them.  Along with that, if the generic is itself a covariant interface, we do the same for it.

For now, I haven't removed the previous version of this function as it was a public function.

Fixes #592 
Fixes #598

### DIFF
--- a/TestComponentCSharp/Class.cpp
+++ b/TestComponentCSharp/Class.cpp
@@ -481,6 +481,34 @@ namespace winrt::TestComponentCSharp::implementation
     {
         _objectIterableChanged.remove(token);
     }
+    IIterable<IIterable<WF::Point>> Class::IterableOfPointIterablesProperty()
+    {
+        return _pointIterableIterable;
+    }
+    void Class::IterableOfPointIterablesProperty(IIterable<IIterable<WF::Point>> const& value)
+    {
+        for (auto pointList : value)
+        {
+            for (auto point : pointList)
+            {
+            }
+        }
+        _pointIterableIterable = value;
+    }
+    IIterable<IIterable<WF::IInspectable>> Class::IterableOfObjectIterablesProperty()
+    {
+        return _objectIterableIterable;
+    }
+    void Class::IterableOfObjectIterablesProperty(IIterable<IIterable<WF::IInspectable>> const& value)
+    {
+        for (auto objectList : value)
+        {
+            for (auto object : objectList)
+            {
+            }
+        }
+        _objectIterableIterable = value;
+    }
     Uri Class::UriProperty()
     {
         return _uri;
--- a/TestComponentCSharp/Class.cpp
+++ b/TestComponentCSharp/Class.cpp
@@ -487,9 +487,9 @@ namespace winrt::TestComponentCSharp::implementation
     }
     void Class::IterableOfPointIterablesProperty(IIterable<IIterable<WF::Point>> const& value)
     {
-        for (auto pointList : value)
+        for (auto points : value)
         {
-            for (auto point : pointList)
+            for (auto point : points)
             {
             }
         }
@@ -501,9 +501,9 @@ namespace winrt::TestComponentCSharp::implementation
     }
     void Class::IterableOfObjectIterablesProperty(IIterable<IIterable<WF::IInspectable>> const& value)
     {
-        for (auto objectList : value)
+        for (auto objects : value)
         {
-            for (auto object : objectList)
+            for (auto object : objects)
             {
             }
         }
--- a/UnitTest/TestComponentCSharp_Tests.cs
+++ b/UnitTest/TestComponentCSharp_Tests.cs
@@ -707,14 +707,6 @@ namespace UnitTest
             var objectUriArray = new object[] { new Uri("http://github.com") };
             TestObject.ObjectIterableProperty = objectUriArray;
             Assert.True(TestObject.ObjectIterableProperty.SequenceEqual(objectUriArray));
-
-            var listOfListOfUris = new List<List<Uri>>() {
-                new List<Uri>{ new Uri("http://aka.ms/cswinrt"), new Uri("http://github.com") },
-                new List<Uri>{ new Uri("http://aka.ms/cswinrt") },
-                new List<Uri>{ new Uri("http://aka.ms/cswinrt"), new Uri("http://microsoft.com") }
-            };
-            TestObject.IterableOfObjectIterablesProperty = listOfListOfUris;
-            Assert.True(TestObject.IterableOfObjectIterablesProperty.SequenceEqual(listOfListOfUris));
         }
 
         [Fact]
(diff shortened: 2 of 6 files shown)


manodasanW:
I realized there were other covariance scenarios still broken, made an attempt at solving them.

TestComponentCSharp/Class.cpp
```diff
@@ -481,6 +481,34 @@ namespace winrt::TestComponentCSharp::implementation
     {
         _objectIterableChanged.remove(token);
     }
+    IIterable<IIterable<WF::Point>> Class::IterableOfPointIterablesProperty()
+    {
+        return _pointIterableIterable;
+    }
+    void Class::IterableOfPointIterablesProperty(IIterable<IIterable<WF::Point>> const& value)
+    {
+        for (auto pointList : value)
```
Scottj1s, line 490: nit: not a list (rename to points, objects)

WinRT.Runtime/Projections.cs
```diff
@@ -247,6 +249,92 @@ public static bool TryGetCompatibleWindowsRuntimeTypeForVariantType(Type type, o
...
+
+            // https://docs.microsoft.com/en-us/archive/blogs/ericlippert/computing-a-cartesian-product-with-linq
+            IEnumerable<IEnumerable<Type>> newArgumentsList = new[] { Enumerable.Empty<Type>() };
+            foreach (var compatibleTypesForGeneric in compatibleTypesPerGeneric)
+            {
+                newArgumentsList =
+                  from seq in newArgumentsList
+                  from item in compatibleTypesForGeneric
+                  select seq.Concat(new[] { item });
+            }
```
jkoritzinsky, line 332: LINQ usage like this can cause heavy allocations and lead to a performance hit. I recommend to either not use LINQ or to also add caching of the interface list at the type level (see #554)

UnitTest/TestComponentCSharp_Tests.cs
```diff
@@ -687,7 +687,34 @@ public void TestObjectCasting()
...
+
+            var uriArray = new Uri[] { new Uri("http://aka.ms/cswinrt"), new Uri("http://github.com") };
+            TestObject.ObjectIterableProperty = uriArray;
+            Assert.True(TestObject.ObjectIterableProperty.SequenceEqual(uriArray));
+
+            var objectUriArray = new object[] { new Uri("http://github.com") };
+            TestObject.ObjectIterableProperty = objectUriArray;
+            Assert.True(TestObject.ObjectIterableProperty.SequenceEqual(objectUriArray));
+
+            var listOfListOfUris = new List<List<Uri>>() {
```
Scottj1s, line 711: shouldn't this go in TestCovariance?

WinRT.Runtime/Projections.cs
```diff
@@ -206,6 +207,7 @@ private static bool IsTypeWindowsRuntimeTypeNoArray(Type type)
                 || type.GetCustomAttribute<WindowsRuntimeTypeAttribute>() is object;
         }
 
+        // Use TryGetCompatibleWindowsRuntimeTypesForVariantType instead.
```
Scottj1s, line 210: is the comment a TODO?

WinRT.Runtime/Projections.cs
```diff
@@ -247,6 +249,92 @@ public static bool TryGetCompatibleWindowsRuntimeTypeForVariantType(Type type, o
...
+
+                if (argumentCovariantObject)
+                {
+                    compatibleTypesForGeneric.AddRange(GetCompatibleTypes(genericArguments[i]));
+                }
+
+                compatibleTypesPerGeneric.Add(compatibleTypesForGeneric);
+            }
+
+            // https://docs.microsoft.com/en-us/archive/blogs/ericlippert/computing-a-cartesian-product-with-linq
```
Scottj1s, line 324: hmm, this looks strangely familiar ;-)
https://github.com/microsoft/CsWinRT/compare/jkoritzinsky/multiple-PIIDs

WinRT.Runtime/Projections.cs
manodasanW: No, the issue is this API was marked public even though we used it is only internally.  Since we don't anyone else to use it or in cross module scenarios, we can accept the breaking change and remove the old version of the API.  But before I did that, I wanted to make sure there were concerns with that, so I instead put a comment.

UnitTest/TestComponentCSharp_Tests.cs
manodasanW: Moved.

WinRT.Runtime/Projections.cs
Scottj1s: should we fold the caching suggestion in with #554 work then? (to unblock adding this fix for 1.1)

Scottj1s (approved):

WinRT.Runtime/Projections.cs
manodasanW: Yea it is basically the same, it is the non-linq version.  Moved to that based on the other comment about linq can cause high memory usage.

WinRT.Runtime/Projections.cs
manodasanW: We definitely should consider caching this, but for now I moved it to the non-linq version.  In reality, for the types in our scenario which have covariant generics, there shouldn't too many generics for that type.

PR #608, opened 2020-11-29
Handle spaces in paths for nuspec and build

### LINKED ISSUE #607: Space In User Name Causes an Error When Building Projects That Use the CsWinRT Nuget Package
More generally this will occur if the Nuget package path happens to have any space.

For example, when building the XamlControlsGallery.Desktop solution I get:
`The command "C:\Users\Jamie Last Name\.nuget\packages\microsoft.windows.cswinrt\1.0.1\cswinrt.exe @"obj\generatedcswinrt.rsp"" exited with code 9009.`

Detailed logs finally gave it away:
```
1>  Task "Exec"
1>    C:\Users\Jamie Last Name\.nuget\packages\microsoft.windows.cswinrt\1.0.1\cswinrt.exe @"obj\generatedcswinrt.rsp"
1>    'C:\Users\Jamie' is not recognized as an internal or external command,
1>    operable program or batch file.
```

I'll post a PR with the fix shortly.

jdasilva:
I verified this by running a build and using the contents of the target file to replace those in the nuget cache folder and successfully built the [XamlControlsGallery.Desktop solution](https://github.com/microsoft/Xaml-Controls-Gallery/tree/winui3preview/XamlControlsGallery).

I also found and fixed some issues from spaces in my user name in the build script.

Fixes #607

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -11,13 +11,13 @@ set path=%DOTNET_ROOT%;%path%
 powershell -NoProfile -ExecutionPolicy unrestricted -Command ^
 [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12; ^
 &([scriptblock]::Create((Invoke-WebRequest -UseBasicParsing 'https://dot.net/v1/dotnet-install.ps1'))) ^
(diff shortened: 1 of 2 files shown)


Scottj1s (approved):
thanks!

PR #615, opened 2020-12-01
Authoring Support: Add WinRT Diagnostics

j0shuams:
#### Background
Previously (before C#/WinRT) `Winmdexp`  tool was used to generate a winmd file from C# code, for a WRC.  See [this link](https://docs.microsoft.com/en-us/previous-versions/hh977010(v=vs.110)) for a list of the scenarios `Winmdexp` looks for.

I am working on adding diagnostic checks to the C#/WinRT source generator to catch these same cases that `Winmdexp` did.  As I have been working, my additions are growing in size. So, I thought it would be good to integrate them into our primary branch in pieces. For one thing, if there's any suggestions that affect the design/code in a large way, it's better to catch that now than down the road when even more code has been written. 

#### The Work 

This PR includes the work to add diagnostics for users authoring C#/WinRT components. Essentially, users could write some C# code that is not valid for a Windows Runtime Component (WRC). 

The code I am introducing in this PR is 
- `DiagnosticUtils` and `DiagnosticHelpers` , new files in the WinRT.SourceGenerator project that handles all of the diagnostic checks
- `DiagnosticRules`, another new file that holds all the `DiagnosticDescriptor` used for the rules (diagnostic messages)
- a unit test app called `DiagnosticTests` , built with positive and negative data driven tests
- updates to the existing `Generator` file in the SourceGenerator project, with calls to my functions 

Contributes to #309


src/Authoring/TestDiagnostics/ArrayParameterTests.cs
```diff
@@ -0,0 +1,132 @@
+﻿using System.Runtime.InteropServices;
+
+namespace TestDiagnostics
+{
+    /* TODO: 
+     * what happens if you put another random attribute on an array param? 
+     * check in WRC3 project ...
+    */
+    public class ReadOnlyArray : System.Attribute
```
manodasanW, line 9: FYI, You can see the attribute I have defined in my PR.

src/Authoring/TestDiagnostics/ArrayParameterTests.cs
```diff
@@ -0,0 +1,132 @@
...
+        public ReadOnlyArray() { }
+    }
+
+    public class WriteOnlyArray : System.Attribute
+    {
+        public WriteOnlyArray() { }
+    }
+
+    /*
+    public interface IHaveAMethodNamedArray_Valid
```
manodasanW, line 20: Is all of this commented out due to the attribute or some other reason?

src/Authoring/TestDiagnostics/Scenarios.cs
```diff
@@ -0,0 +1,60 @@
+﻿using ABI.System.Numerics;
```
manodasanW, line 1: I wouldn't expect you to be using ABI namespaces, you probably want the projected ones in the non ABI namespace?

src/Authoring/WinRT.SourceGenerator/DiagnosticRules.cs
```diff
@@ -0,0 +1,172 @@
...
+            return new DiagnosticDescriptor(
+                id: id,
+                title: title,
+                messageFormat: messageFormat,
+                category: "Usage",
+                /* Warnings dont fail command line build; winmd generation is prevented regardless of severity.
+                * Make this error when making final touches on this deliverable. */
+                defaultSeverity: DiagnosticSeverity.Warning,
+                isEnabledByDefault: true,
+                helpLinkUri: "https://docs.microsoft.com/en-us/previous-versions/hh977010(v=vs.110)");
```
manodasanW, line 32: This URL is fine for now, but we will eventually probably need our own version.

src/Authoring/WinRT.SourceGenerator/Generator.cs
```diff
@@ -1,186 +1,263 @@
...
-            Logger.Log(cswinrtProcess.StandardError.ReadToEnd());
-            cswinrtProcess.WaitForExit();
-
-            foreach(var file in Directory.GetFiles(outputDir, "*.cs", SearchOption.TopDirectoryOnly))
-            {
-                Logger.Log("Adding " + file);
-                context.AddSource(Path.GetFileNameWithoutExtension(file), SourceText.From(File.ReadAllText(file), Encoding.UTF8));
-            }
-
-            Directory.Delete(outputDir, true);
```
manodasanW, line 110: Probably don't want to comment this out - I moved it behind a flag in my PR.

src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
```diff
@@ -0,0 +1,619 @@
...
+{
+    public class WinRTRules
+    {
+
+        #region ModifierHelpers
+        private bool SyntaxTokenIs(SyntaxToken stx, string str) { return stx.Value.Equals(str); }
+        
+        private bool ModifiersContains(SyntaxTokenList modifiers, string str)
+        {
+            foreach (var modifier in modifiers)
```
manodasanW, line 17: Technically this is similar to modifiers.Any(modifier => modifier.ValueText == str);

src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
```diff
@@ -0,0 +1,619 @@
...
+            {
+                if (SyntaxTokenIs(modifier, str)) { return true; }
+            }
+            return false;
+        }
+
+        private bool PropertyIsPublic(PropertyDeclarationSyntax p) { return ModifiersContains(p.Modifiers, "public"); }
```
manodasanW, line 24: I think all these have the same base class of MemberDeclarationSyntax which you might be able to use to simplify this to 1 function.

src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
```diff
@@ -0,0 +1,619 @@
...
+        /// Keeps track of the method (via name + arity) and whether it was declared with the DefaultOverload attribute</param>
+        /// <param name="overloadsWithoutAttributeMap">
+        ///     Keeps track of the methods that are overloads but don't have the DefaultOverload attribute (yet)
+        ///     Used after this function executes, hence the reference parameter 
+        ///     </param>
+        /// <param name="classId">The class the method lives in -- used for creating the diagnostic</param>
+        /// <param name="context">The SourceGenerator context the code lives in</param>
+        /// <returns>True iff multiple overloads of a method are found, where more than one has been designated as the default overload</returns>
+        private bool CheckOverloadAttributes(MethodDeclarationSyntax method,
+            ref Dictionary<string, bool> methodHasAttributeMap,
```
manodasanW, line 89: Any reason to pass these by ref?  https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/ref#passing-an-argument-by-reference-an-example

In C#, objects are by default passed by reference. I believe ref here is only helpful if you plan on changing the object value itself here.

src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
```diff
@@ -0,0 +1,619 @@
...
+        /// </summary>
+        /// <param name="context">Object to raise diagnostic on</param>
+        /// <param name="classDeclaration">look for constructors of this class</param>
+        /// <returns>True if multiple constructors of the same arity exist for the given class</returns>
+        public bool HasMultipleConstructorsOfSameArity(ref GeneratorExecutionContext context, ClassDeclarationSyntax classDeclaration)
+        {
+            IEnumerable<ConstructorDeclarationSyntax> constructors = classDeclaration.ChildNodes().OfType<ConstructorDeclarationSyntax>();
+
+            // A true entry means a constructor of that arity has been seen
+            Dictionary<int, bool> aritiesSeenSoFar = new Dictionary<int, bool>();
```
manodasanW, line 197: Is the check being done here basically not the equivalent of a list.contains or set.contains?

src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
```diff
@@ -0,0 +1,619 @@
...
+        /// Class to check for operator declarations 
+        /// operator declarations are just like method declarations except they use the `operator` keyword 
+        /// </param>
+        /// <returns>True iff an operator is overloaded by the given class</returns>
+        public bool OverloadsOperator(ref GeneratorExecutionContext context, ClassDeclarationSyntax classDeclaration)
+        {
+            var operatorDeclarations = classDeclaration.DescendantNodes().OfType<OperatorDeclarationSyntax>();
+            foreach (var op in operatorDeclarations)
+            { 
+                context.ReportDiagnostic(Diagnostic.Create(DiagnosticRules.OperatorOverloadedRule, op.GetLocation(), op.OperatorToken));
```
manodasanW, line 260: We should probably remove the return true and then change return false to return found or return operatorDeclarations.Count != 0

src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
```diff
@@ -0,0 +1,619 @@
...
+        }
+
+        #endregion
+
+        #region Parameters
+
+        private bool ParamHasInOrOutAttribute(ParameterSyntax param)
+        {
+            string baseString = ""; //"System.Runtime.InteropServices.";
+            return ParamHasAttribute(baseString + "In", param) || ParamHasAttribute(baseString + "Out", param);
```
manodasanW, line 331: Do we need to handle case where it has namespace and where it doesn't?  maybe we want to do a symbol check rather than name check?

src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
```diff
@@ -0,0 +1,619 @@
...
+            
+            /* we can't throw the diagnostic as soon as we see a second overload without an attribute, 
+             *   as there could be a third overload with the default attribute
+             * we use this map to keep track of these cases, and after we have checked all methods, 
+             *   we check the elements of the map to raise diagnostics for those that didn't get attributed */
+            Dictionary<string, Diagnostic> overloadsWithoutAttributeMap = new Dictionary<string, Diagnostic>();
+
+            foreach (MethodDeclarationSyntax method in methods.Where(MethodIsPublic))
+            {
+                found |= CheckParamsForArrayAttributes(method, ref context);
```
manodasanW, line 467: I wonder if once one of these evaluates to true, whether the other functions get called or not?

src/Authoring/TestDiagnostics/TestDiagnostics.csproj
```diff
@@ -0,0 +1,28 @@
...
+    <LangVersion>preview</LangVersion>
+    <AssemblyVersion>1.0.0.0</AssemblyVersion>
+    <CsWinRTEnableLogging>true</CsWinRTEnableLogging>
+    <CsWinRTComponent>true</CsWinRTComponent>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <GeneratedFilesDir Condition="'$(GeneratedFilesDir)'==''">$([MSBuild]::NormalizeDirectory('$(MSBuildProjectDirectory)', '$(IntermediateOutputPath)', 'Generated Files'))</GeneratedFilesDir>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <CompilerVisibleProperty Include="AssemblyName" />
```
manodasanW, line 15: These are also set by our cswinrt targets when CsWinRTComponent = true so shouldn't be needed anymore.

src/Authoring/TestDiagnostics/Scenarios.cs
```diff
@@ -0,0 +1,60 @@
...
+using System;
+using System.Runtime.ExceptionServices;
+using Windows.Foundation;
+using Windows.Foundation.Numerics;
+using Windows.Security.Cryptography.Core;
+using Windows.Web.Syndication;
+
+namespace TestDiagnostics
+{
+    /*
```
manodasanW, line 12: I think I see why you have all the invalid scenarios commented out (so that they don't trigger and break build).  But I think we still want to be able to run tests to validate none of the checks regress so we should consider creating a compilation context in a text for a certain source text and looking for diagnostics.

src/Authoring/WinRT.SourceGenerator/Generator.cs
```diff
@@ -1,186 +1,263 @@
...
+                var interfaces = nodes.OfType<InterfaceDeclarationSyntax>().Where(winrtRules.InterfaceIsPublic);
+                var structs = nodes.OfType<StructDeclarationSyntax>();
+
+                // Used in the checking of structure fields 
+                List<string> classNames = new List<string>();
+
+                /* Check all classes */
+                foreach (ClassDeclarationSyntax classDeclaration in classes)
+                {
+                    classNames.Add(classDeclaration.Identifier.ToString());
```
manodasanW, line 160: Does identifier include namespace (I believe not)?  You might want the qualified name with namespace as it might cause issues when doing comparisons later depending on whether they expect namespace or not.

src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
```diff
@@ -0,0 +1,619 @@
...
+
+        /// <summary>
+        /// keeps track of the arity of all constructors, and reports the diagnostic (and exits) as soon as a two constructors of the same arity are seen.
+        /// </summary>
+        /// <param name="context">Object to raise diagnostic on</param>
+        /// <param name="classDeclaration">look for constructors of this class</param>
+        /// <returns>True if multiple constructors of the same arity exist for the given class</returns>
+        public bool HasMultipleConstructorsOfSameArity(ref GeneratorExecutionContext context, ClassDeclarationSyntax classDeclaration)
+        {
+            IEnumerable<ConstructorDeclarationSyntax> constructors = classDeclaration.ChildNodes().OfType<ConstructorDeclarationSyntax>();
```
manodasanW, line 194: Do we need a public check as I would assume if they were private, it might be fine.

src/Authoring/WinRT.SourceGenerator/Generator.cs
```diff
@@ -1,186 +1,263 @@
...
+            {
+                var model = context.Compilation.GetSemanticModel(tree); 
+                var nodes = tree.GetRoot().DescendantNodes();
+                
+                var classes = nodes.OfType<ClassDeclarationSyntax>().Where(winrtRules.ClassIsPublic); 
+                var interfaces = nodes.OfType<InterfaceDeclarationSyntax>().Where(winrtRules.InterfaceIsPublic);
+                var structs = nodes.OfType<StructDeclarationSyntax>();
+
+                // Used in the checking of structure fields 
+                List<string> classNames = new List<string>();
```
manodasanW, line 155: Should this class name list be built earlier?  Can class names from different syntax trees be used in other syntax trees?

src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
```diff
@@ -0,0 +1,619 @@
...
+
+            if (ModifiersContains(field.Modifiers, "const"))
+            {
+                context.ReportDiagnostic(Diagnostic.Create(DiagnosticRules.StructHasConstFieldRule, field.GetLocation(), structDeclaration.Identifier));
+                found |= true;
+            }
+
+            foreach (var variable in field.DescendantNodes().OfType<VariableDeclarationSyntax>())
+            {
+                var typeStr = variable.Type.ToString();
```
manodasanW, line 599: Can this be a TypeKind check whether we check like variable.Type.TypeKind == class or something.  This is assuming no classes are allowed.

src/Authoring/TestDiagnostics/ArrayParameterTests.cs
j0shuams: Yeah I saw that yesterday 👍

src/Authoring/TestDiagnostics/ArrayParameterTests.cs
j0shuams: Yeah my unit testing story is a little funky -- but your comment about making contexts is spot-on, I think that will improve it a lot.

src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
j0shuams: I'll try this out, thanks!

src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
j0shuams: I actually was using a list and doing list.contains previously, and moved to a dictionary because i thought it would be faster haha.

src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
j0shuams: I'll try this out

src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
manodasanW: You can use HashSet which should be pretty similar in performance to dictionary for what you are doing.

src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
j0shuams: Good idea!

src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
j0shuams: When I started out, I wasn't passing by ref and diagnostics weren't sticking to the context, so I wanted to change the context value to have the diagnostic I reported. I'll see what happens if I take `ref` out again

src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
j0shuams: I changed it but didn't change the check style, let me know what you think

src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
j0shuams: What do you mean?

src/Authoring/TestDiagnostics/TestDiagnostics.csproj
j0shuams: this project got deprecated in latest changes

src/Authoring/TestDiagnostics/Scenarios.cs
j0shuams: yep coming now

src/Authoring/WinRT.SourceGenerator/Generator.cs
j0shuams: QualifiedName isn't a property we can get, but I think I can look at the syntax tree again and grab the namespace and add that to the string -- sound good?

src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
j0shuams: Unfortunately, at this point in the analysis the type is "IdentifierSyntax", because it is not a known type it is considered an identifier for a variable.

src/Authoring/WinRT.SourceGenerator/Generator.cs
j0shuams: I'm not sure how to make a test with multiple syntax trees, yet. Are you thinking we do one pass to gather all (qualified) class names (and interface?) and then use that here, instead of building it for each syntax tree...

src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
j0shuams: Taking them all out only causes 42 (out of 257) tests to fail, I'll investigate and see if we can get by without `ref`

src/Authoring/WinRT.SourceGenerator/Generator.cs
j0shuams: I think I found how to get the symbol with the namespace on it, using the INamedTypeSymbol returned by `model.GetDeclaredSymbol(classDeclaration)`

src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
j0shuams: Related to the `INamedTypeSymbol` comment, we might be able to do something similar here

src/Authoring/WinRT.SourceGenerator/Generator.cs
j0shuams: I did the idea i proposed

src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
j0shuams: Discussed offline -- we are catching qualified and unqualified here now

src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
j0shuams: I have a second PR for after this one and I have it working with no explicit `ref` params

src/Authoring/DiagnosticTests/DiagnosticTests.csproj
```diff
@@ -0,0 +1,20 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
```
manodasanW, line 4: net5-windows10.0...?

j0shuams: It works with both, fwiw; I've updated it

src/Authoring/DiagnosticTests/NegativeData.cs
```diff
@@ -0,0 +1,2142 @@
...
+    public interface GenIface<T> 
+    { 
+        int Foo(T input); 
+    }
+}";
+
+        private const string InterfaceInheritsException = @"
+namespace Test 
+{ 
+    public interface IfaceWithExceptions : System.Exception
```
manodasanW, line 58: Technically this isn't valid C# code.

j0shuams: Yeah I thought I had removed this, thanks for catching it

src/Authoring/WinRT.SourceGenerator/DiagnosticRules.cs
```diff
@@ -45,16 +41,15 @@ private static DiagnosticDescriptor MakeRule(string id, string title, string mes
         public static DiagnosticDescriptor ParameterNamedValueRule = MakeRule(
             "WME1092",
             "Parameter Named Value Rule",
-            ("The method {0} has a parameter named {1} which is the same as the default return value name. "
```
manodasanW: Open question on if we support the attribute, but we can keep this diagnostic for now.

src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
```diff
@@ -0,0 +1,494 @@
...
+                    Report(ref context, DiagnosticRules.NonArrayMarked, method.GetLocation(), method.Identifier, param.Identifier);
+                    found |= true;
+                }
+            }
+
+            return found;
+        }
+
+        public bool HasInvalidMethods<T>(ref GeneratorExecutionContext context, IEnumerable<MethodDeclarationSyntax> methodDeclarations, SyntaxToken typeId)
+            where T : TypeDeclarationSyntax
```
manodasanW, line 395: Are you using this generic somewhere?

manodasanW (approved):

src/Authoring/WinRT.SourceGenerator/DiagnosticRules.cs
j0shuams: I think you meant this for DefaultOverload ? I noticed you have multiple overloads of same arity but no attribute, added the attribute since we are blocking those scenarios. 

The diagnostic you commented on is for the parameter name conflict (__retval)

PR #630, opened 2020-12-09
Adding missing Marshaler array functions.

### LINKED ISSUE #629: IVectorView GetMany results in error for classes
GetMany function results in error for class vectors, interface vectors, and object vectors.  Structs seem to be fine.

manodasanW:
GetMany uses Marshaler<T> to get various marshal array functions, but they were only set for structs.  Setting it for the other types.

Fixes #629


src/Tests/TestComponentCSharp/TestComponentCSharp.idl
```diff
@@ -241,6 +241,14 @@ namespace TestComponentCSharp
         Windows.Foundation.Collections.IVectorView<IProperties1> GetInterfaceVector();
         [noexcept] Windows.Foundation.Collections.IVectorView<Class> GetClassVector();
 
+        Windows.Foundation.Collections.IVectorView<Class> GetClassVectorSubset(Windows.Foundation.Collections.IVectorView<Class> classVector, Int32 startIndex);
```
Scottj1s, line 244: shouldn't this all go into TestComponent?

Scottj1s:
>         internal static readonly Func<T[], object> CreateMarshalerArray;

thanks for catching all these - this stuff is a pain

---
Refers to: src/WinRT.Runtime/Marshalers.cs:438 in bcbe8d2. [](commit_id = bcbe8d2ad42bd0d1ca510fa6f88af85f0bdd242f, deletion_comment = False)

src/Tests/TestComponentCSharp/TestComponentCSharp.idl
manodasanW: Some of this can go in TestComponent, but not all such as the Blittable ones which are only defined here.  I will look at moving the others.

manodasanW: Actually realized testwinrt has them too, so will look at moving them all.

Scottj1s (approved):
thanks!

PR #631, opened 2020-12-09
Authoring support for C# built-in types to WinRT types

manodasanW:
- Authoring support for mapping C# built-in types to WinRT types (both as part of type reference in param / return type, but also for implementing them)
- Added support for generics
- Added support for arrays
- Various bug fixes
- Addded an authoring impl class for authored structs / enums which don't need a wrapper to hold onto attribute information which is used as part of the WinRT type check
- Added support for out parameters
- Added tests to exercise these scenarios

Contributes to #309

### DIFF
--- a/src/Authoring/AuthoringConsumptionTest/AuthoringConsumptionTest.exe.manifest
+++ b/src/Authoring/AuthoringConsumptionTest/AuthoringConsumptionTest.exe.manifest
@@ -11,9 +11,25 @@
         threadingModel="both"
         xmlns="urn:schemas-microsoft-com:winrt.v1" />
     <activatableClass
+        name="AuthoringSample.CustomReadOnlyDictionary"
+        threadingModel="both"
+        xmlns="urn:schemas-microsoft-com:winrt.v1" />
+    <activatableClass
+        name="AuthoringSample.CustomVector"
+        threadingModel="both"
+        xmlns="urn:schemas-microsoft-com:winrt.v1" />
+    <activatableClass
+        name="AuthoringSample.CustomVectorView"
+        threadingModel="both"
+        xmlns="urn:schemas-microsoft-com:winrt.v1" />        
+    <activatableClass
         name="AuthoringSample.CustomWWW"
         threadingModel="both"
         xmlns="urn:schemas-microsoft-com:winrt.v1" />
+    <activatableClass
+        name="AuthoringSample.DisposableClass"
+        threadingModel="both"
+        xmlns="urn:schemas-microsoft-com:winrt.v1" />    
     <activatableClass
         name="AuthoringSample.TestClass"
         threadingModel="both"
--- a/src/Authoring/AuthoringConsumptionTest/AuthoringConsumptionTest.exe.manifest
+++ b/src/Authoring/AuthoringConsumptionTest/AuthoringConsumptionTest.exe.manifest
@@ -7,6 +7,10 @@
         threadingModel="both"
         xmlns="urn:schemas-microsoft-com:winrt.v1" />
     <activatableClass
+        name="AuthoringSample.CustomDictionary"
+        threadingModel="both"
+        xmlns="urn:schemas-microsoft-com:winrt.v1" />
+    <activatableClass
         name="AuthoringSample.CustomWWW"
         threadingModel="both"
         xmlns="urn:schemas-microsoft-com:winrt.v1" />
--- a/src/Authoring/AuthoringSample/AuthoringSample.csproj
+++ b/src/Authoring/AuthoringSample/AuthoringSample.csproj
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFramework>net5.0</TargetFramework>
@@ -9,16 +9,10 @@
     <CsWinRTComponent>true</CsWinRTComponent>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <GeneratedFilesDir Condition="'$(GeneratedFilesDir)'==''">$([MSBuild]::NormalizeDirectory('$(MSBuildProjectDirectory)', '$(IntermediateOutputPath)', 'Generated Files'))</GeneratedFilesDir>
+    <CsWinRTKeepGeneratedSources>true</CsWinRTKeepGeneratedSources>
   </PropertyGroup>
 
-  <ItemGroup>
-    <CompilerVisibleProperty Include="AssemblyName" />
-    <CompilerVisibleProperty Include="AssemblyVersion" />
-    <CompilerVisibleProperty Include="CsWinRTComponent" />
-    <CompilerVisibleProperty Include="CsWinRTEnableLogging" />
-    <CompilerVisibleProperty Include="GeneratedFilesDir" />
-    <CompilerVisibleProperty Include="CsWinRTExe" />
-    
+  <ItemGroup>    
     <ProjectReference Include="..\..\Projections\Windows\Windows.csproj" />
     <ProjectReference Include="..\WinRT.SourceGenerator\WinRT.SourceGenerator.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" />
     <ProjectReference Include="..\..\cswinrt\cswinrt.vcxproj" />
--- a/src/Authoring/AuthoringSample/AuthoringSample.csproj
+++ b/src/Authoring/AuthoringSample/AuthoringSample.csproj
@@ -9,7 +9,8 @@
     <CsWinRTComponent>true</CsWinRTComponent>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <GeneratedFilesDir Condition="'$(GeneratedFilesDir)'==''">$([MSBuild]::NormalizeDirectory('$(MSBuildProjectDirectory)', '$(IntermediateOutputPath)', 'Generated Files'))</GeneratedFilesDir>
-    <CsWinRTKeepGeneratedSources>true</CsWinRTKeepGeneratedSources>
+    <!-- Enable to diagnose generation issues -->
+    <!-- <CsWinRTKeepGeneratedSources>true</CsWinRTKeepGeneratedSources> -->
   </PropertyGroup>
 
   <ItemGroup>    
--- a/src/Authoring/WinRT.SourceGenerator/Generator.cs
+++ b/src/Authoring/WinRT.SourceGenerator/Generator.cs
@@ -14,6 +14,20 @@ namespace Generator
     [Generator]
     public class SourceGenerator : ISourceGenerator
     {
+        private static readonly string ArrayAttributes = @"
+namespace System.Runtime.InteropServices.WindowsRuntime
+{
+    [global::System.AttributeUsage(System.AttributeTargets.Parameter, AllowMultiple = false, Inherited = false)]
+    internal sealed class ReadOnlyArrayAttribute : global::System.Attribute
+    {
+    }
+
+    [global::System.AttributeUsage(System.AttributeTargets.Parameter, AllowMultiple = false, Inherited = false)]
+    internal sealed class WriteOnlyArrayAttribute : global::System.Attribute
+    {
+    }
+}";
+
         private string _tempFolder;
 
         private static string GetAssemblyName(GeneratorExecutionContext context)
@@ -170,19 +184,7 @@ namespace Generator
 
             try
             {
-                context.AddSource("System.Runtime.InteropServices.WindowsRuntime", SourceText.From(@"
-namespace System.Runtime.InteropServices.WindowsRuntime
-{
-    [global::System.AttributeUsage(System.AttributeTargets.Parameter, AllowMultiple = false, Inherited = false)]
-    internal sealed class ReadOnlyArrayAttribute : global::System.Attribute
-    {
-    }
-
-    [global::System.AttributeUsage(System.AttributeTargets.Parameter, AllowMultiple = false, Inherited = false)]
-    internal sealed class WriteOnlyArrayAttribute : global::System.Attribute
-    {
-    }
-}", Encoding.UTF8));
+                context.AddSource("System.Runtime.InteropServices.WindowsRuntime", SourceText.From(ArrayAttributes, Encoding.UTF8));
 
                 string assembly = GetAssemblyName(context);
                 string version = GetAssemblyVersion(context);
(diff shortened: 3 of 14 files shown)


manodasanW:
You probably should disable whitespace diff when reviewing this PR.

src/Authoring/AuthoringConsumptionTest/AuthoringConsumptionTest.exe.manifest
```diff
@@ -1,18 +1,38 @@
...
+<?xml version="1.0" encoding="utf-8"?>
+<assembly manifestVersion="1.0" xmlns="urn:schemas-microsoft-com:asm.v1">
+  <assemblyIdentity version="1.0.0.0" name="CppTestApp"/>
+  <file name="WinRT.Host.dll">
+    <activatableClass
+        name="AuthoringSample.BasicClass"
+        threadingModel="both"
+        xmlns="urn:schemas-microsoft-com:winrt.v1" />
+    <activatableClass
+        name="AuthoringSample.CustomDictionary"
```
j0shuams, line 10: We have to do this for every public class/type they author right? Not for this PR, but maybe in the future we can add something that writes this manifest file based on the authored class

src/Authoring/WinRT.SourceGenerator/Generator.cs
```diff
@@ -1,186 +1,226 @@
...
+
+        private static bool GetKeepGeneratedSources(GeneratorExecutionContext context)
+        {
+            context.AnalyzerConfigOptions.GlobalOptions.TryGetValue("build_property.CsWinRTKeepGeneratedSources", out var keepGeneratedSourcesStr);
+            return keepGeneratedSourcesStr != null && bool.TryParse(keepGeneratedSourcesStr, out var keepGeneratedSources) && keepGeneratedSources;
+        }
+
+        private static string GetCsWinRTWindowsMetadata(GeneratorExecutionContext context)
+        {
+            context.AnalyzerConfigOptions.GlobalOptions.TryGetValue("build_property.CsWinRTWindowsMetadata", out var cswinrtWindowsMetadata);
```
j0shuams, line 63: Do we want a function on contexts and strings that returns the `out var` ? we do this same thing here and in a couple other places (iscswinrtcomponent, getgeneratedfilesdir, ...)

src/Authoring/WinRT.SourceGenerator/Generator.cs
```diff
@@ -1,186 +1,226 @@
...
+            };
+
+            try
+            {
+                using var cswinrtProcess = Process.Start(processInfo);
+                Logger.Log(cswinrtProcess.StandardOutput.ReadToEnd());
+                Logger.Log(cswinrtProcess.StandardError.ReadToEnd());
+                cswinrtProcess.WaitForExit();
+
+                if (cswinrtProcess.ExitCode != 0)
```
j0shuams, line 119: love it!

src/Authoring/WinRT.SourceGenerator/Generator.cs
```diff
@@ -1,186 +1,226 @@
...
+            if (!IsCsWinRTComponent(context))
+            {
+                return;
+            }
+
+            Logger.Initialize(context);
+
+            try
+            {
+                context.AddSource("System.Runtime.InteropServices.WindowsRuntime", SourceText.From(@"
```
j0shuams, line 173: Have you tried this out in a test app? It has no problem "resolving" the attribute? Just curious

src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
```diff
@@ -1,2180 +1,2682 @@
...
+{
+    class Parameter
+    {
+        public Symbol Type;
+        public string Name;
+        public ParameterAttributes Attributes;
+        public bool ByRef;
+
+        public Parameter(Symbol type, string name, ParameterAttributes attributes)
+            :this(type, name, attributes, attributes == ParameterAttributes.Out)
```
j0shuams, line 23: what is this? a constraint on the parameters?

src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
```diff
@@ -1,2180 +1,2682 @@
...
+            : this(new Symbol(type), name, attributes)
+        {
+        }
+
+        public Parameter(EntityHandle type, string name, ParameterAttributes attributes)
+            : this(new Symbol(type), name, attributes)
+        {
+        }
+
+        public Parameter(IParameterSymbol parameterSymbol)
```
j0shuams, line 45: could you add some comments? I don't quite see what this function is doing. I think it is doing the check on valid read/write only array attribute usage?

src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
```diff
@@ -1,2180 +1,2682 @@
...
+            return methodDefinitionHandle;
+        }
+
+        public override void VisitMethodDeclaration(MethodDeclarationSyntax node)
+        {
+            Logger.Log("method: " + node.Identifier.ValueText);
+            var methodSymbol = Model.GetDeclaredSymbol(node);
+            if (!IsPublicNode(node) || currentTypeDeclaration.CustomMappedSymbols.Contains(methodSymbol))
+            {
+                Logger.Log("method skipped");
```
j0shuams, line 700: do we want to say the name? might help debugging

src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
```diff
@@ -1,2180 +1,2682 @@
...
+                metadataBuilder.GetOrAddString(propertyName),
+                metadataBuilder.GetOrAddBlob(propertySignature));
+            currentTypeDeclaration.AddProperty(symbol, propertyDefinitonHandle);
+
+            if (hasSetMethod)
+            {
+                string setMethodName = "put_" + propertyName;
+                var setMethod = AddMethodDefinition(
+                    setMethodName,
+                    new Parameter[] { new Parameter(type, "value", ParameterAttributes.In) },
```
j0shuams, line 784: Do we need to consider naming conflicts with user params?

src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
```diff
@@ -1,2180 +1,2682 @@
...
+                else
+                {
+                    var memberReferenceHandle = AddMethodReference(name, parameters, returnType, symbol, false);
+                    currentTypeDeclaration.AddMethodReference(symbol, memberReferenceHandle);
+                }
+            }
+
+            void AddProperty(string name, Symbol type, bool setProperty)
+            {
+                if(isDefinition)
```
j0shuams, line 1002: space between `if` and `(`

src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
```diff
@@ -1,2180 +1,2682 @@
...
+                    false,
+                    true,
+                    false,
+                    true
+                )
+            );
+
+            int numParameters = node.ParameterList.Parameters.Count;
+            Parameter[] parameters = new Parameter[numParameters];
+            for (int idx = 0; idx < numParameters; idx++)
```
j0shuams, line 1858: want to pull this loop out into a function that we can call? seems like we do it a lot. just curious

src/Authoring/AuthoringConsumptionTest/AuthoringConsumptionTest.exe.manifest
manodasanW: I agree that we should look at automating this later on when we add more project support.

src/Authoring/WinRT.SourceGenerator/Generator.cs
manodasanW: I was thinking about that before, but I hadn't as these were just 2 line functions that doing this change will only reduce it by a line and I couldn't use it across the different return types consistently.  If I add more such string reutrn functions, it might be worth considering to do to avoid repeating the TryGetValue call in code.

src/Authoring/WinRT.SourceGenerator/Generator.cs
manodasanW: The test app builds fine, but intellisense does complain about it.  I might end up removing it from here and moving it to winrt.runtime later, but I wanted to avoid adding another DLL that we need to worry about.  I am hoping that the intellisense issue goes away with VS / source generator improvements.

src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
manodasanW: This is just calling the other constructor (public Parameter(Symbol type, string name, ParameterAttributes attributes, bool byRef))

src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
manodasanW: This is just a constructor for the scenarios where we have IParameterSymbol as what we use to construct this object.  The attribute check is there to make sure we consult it to determine whether to make the param out or in.  But I will add a comment.

src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
manodasanW: That is logged 3 lines ahead (Logger.Log("method: " + node.Identifier.ValueText);)

src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
manodasanW: I am hoping naming conflicts will be caught by diagnostics before we even get here.  We can in the future maybe support overriding the name if allowed by WinRT as an enhancement.

src/Authoring/WinRT.SourceGenerator/Generator.cs
```diff
@@ -1,186 +1,226 @@
...
+namespace System.Runtime.InteropServices.WindowsRuntime
```
j0shuams, line 174: Do we want this as a private static string and just refer to it? it would be hard to adjust this in the future since it is kind of adhoc here. Also, just curious, did you try injecting this code in the Initialize function (instead of Execute) ?

manodasanW: The latter is not supported today, but there is a proposed issue that I am tracking in the rosyln repo for this.  Ideally if that feature is added, I would move it to initialize as it makes the most sense there.

src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
manodasanW: Good idea, addressed.

src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
manodasanW: reformatted entire document.

src/Authoring/AuthoringSample/AuthoringSample.csproj
```diff
@@ -1,28 +1,22 @@
...
+  <PropertyGroup>
+    <TargetFramework>net5.0</TargetFramework>
+    <Platforms>x64;x86</Platforms>
+    <LangVersion>preview</LangVersion>
+    <AssemblyVersion>1.0.0.0</AssemblyVersion>
+    <CsWinRTEnableLogging>true</CsWinRTEnableLogging>
+    <CsWinRTComponent>true</CsWinRTComponent>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <GeneratedFilesDir Condition="'$(GeneratedFilesDir)'==''">$([MSBuild]::NormalizeDirectory('$(MSBuildProjectDirectory)', '$(IntermediateOutputPath)', 'Generated Files'))</GeneratedFilesDir>
+    <CsWinRTKeepGeneratedSources>true</CsWinRTKeepGeneratedSources>
```
Scottj1s, line 12: >CsWinRTKeepGeneratedSources [](start = 5, length = 27)

is this primarily a debugging tool? allowing you to hack the sources?

src/Authoring/AuthoringSample/AuthoringSample.csproj
```diff
@@ -1,28 +1,22 @@
...
-    <CompilerVisibleProperty Include="CsWinRTExe" />
-    
-    <ProjectReference Include="..\..\Projections\Windows\Windows.csproj" />
-    <ProjectReference Include="..\WinRT.SourceGenerator\WinRT.SourceGenerator.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" />
-    <ProjectReference Include="..\..\cswinrt\cswinrt.vcxproj" />
-    <ProjectReference Include="..\..\WinRT.Runtime\WinRT.Runtime.csproj" />
-  </ItemGroup>
-
-</Project>
+<Project Sdk="Microsoft.NET.Sdk">
```
Scottj1s, line 1: should this be moved to the Samples folder?

src/Authoring/AuthoringSample/AuthoringSample.csproj
manodasanW: Its more like a diagnostics thing where if for some reason the generated sources fails to compile.  I probably should disable it in the checked in version to avoid temp sources kept.

src/Authoring/AuthoringSample/AuthoringSample.csproj
manodasanW: I probably want to rename this to be a test, this component is more for test purposes than to serve as a sample.  I will address that in a future PR.

Scottj1s:
>         private static readonly Dictionary<string, MappedType> MappedCSharpTypes = new Dictionary<string, MappedType>()

Ideally, the 3 sources of mapping data would derived from a single source.  Meantime, can we add comments to all 3 to indicate the other two need to be kept in sync?
[https://github.com/microsoft/CsWinRT/blob/1b4ff32b5770a7901432e562065e75edcb844aaa/src/cswinrt/helpers.h#L450](https://github.com/microsoft/CsWinRT/blob/1b4ff32b5770a7901432e562065e75edcb844aaa/src/cswinrt/helpers.h#L450)
[https://github.com/microsoft/CsWinRT/blob/1b4ff32b5770a7901432e562065e75edcb844aaa/src/WinRT.Runtime/Projections.cs#L25](https://github.com/microsoft/CsWinRT/blob/1b4ff32b5770a7901432e562065e75edcb844aaa/src/WinRT.Runtime/Projections.cs#L25)

---
Refers to: src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs:271 in e3f6fd1. [](commit_id = e3f6fd1f2b643616b2f525c8664eb5610a07c9b7, deletion_comment = False)

Scottj1s (approved):
exciting!

PR #637, opened 2020-12-11
Disable VS tracking

### LINKED ISSUE #636: C#/WinRT doesn't properly incremental build
If I change an .idl file in my C++ project, when I do an incremental build, the C#/WinRT projection isn't updated. This is the opposite problem of #519.

Repro project: [CS-Desktop2.zip](https://github.com/microsoft/CsWinRT/files/5680966/CS-Desktop2.zip)

Repro steps:

1. Do a full build
2. In the `RuntimeComponent1` project, change `Class.idl` to take a constructor which accepts an `Int32`:
   `Class(Int32 prop);`
3. Do an incremental build

**Expected results**
Build fails, since call sites in C# code is using the default constructor, which no longer exists

**Actual results**
Build succeeds

Scottj1s:
until/unless cswinrt invocations can produce tlog entries to support VS incremental builds

fixes #636

### DIFF
--- a/nuget/Microsoft.Windows.CsWinRT.targets
+++ b/nuget/Microsoft.Windows.CsWinRT.targets
@@ -16,6 +16,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <AllowUnsafeBlocks Condition="'$(CsWinRTComponent)' == 'true'">true</AllowUnsafeBlocks>
     <AllowUnsafeBlocks Condition="$(CsWinRTEnabled)">true</AllowUnsafeBlocks>
     <CoreCompileDependsOn>CsWinRTIncludeProjection;CsWinRTRemoveWinMDReferences;$(CoreCompileDependsOn)</CoreCompileDependsOn>
+    <TrackFileAccess>false</TrackFileAccess> 
   </PropertyGroup>
   
   <!-- Remove WinRT.Host.Shim.dll references -->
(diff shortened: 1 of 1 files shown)


nuget/Microsoft.Windows.CsWinRT.targets
```diff
@@ -16,6 +16,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <AllowUnsafeBlocks Condition="'$(CsWinRTComponent)' == 'true'">true</AllowUnsafeBlocks>
     <AllowUnsafeBlocks Condition="$(CsWinRTEnabled)">true</AllowUnsafeBlocks>
     <CoreCompileDependsOn>CsWinRTIncludeProjection;CsWinRTRemoveWinMDReferences;$(CoreCompileDependsOn)</CoreCompileDependsOn>
+    <TrackFileAccess>false</TrackFileAccess>
```
manodasanW, line 19: I wonder if we should do a condition on if not CsWinRTComponent as authoring may and probably not be broken for incremental builds.

Scottj1s: I can do that, but why do you think authoring would not be affected?  It has to run cswinrt.exe as well, potentially with inputs from project references.

manodasanW: I was thinking authoring might not be affected because of 2 reasons.
- If the winmd of the component changed, then the generated C# files generated by the CsWinRT tool that were passed as generated files to the source generator would have been seen as changed as the source generator is what writes the winmd too.
- If a dependent winmd from a project reference changed, it would mean the C#/WinRT projection of that winmd referenced by the authored component would also have been updated and it would be seen as a change.

Scottj1s: OK, added the condition - we can always tighten later if need by

manodasanW (approved):

PR #639, opened 2020-12-14
Update versioning scenarios error messages

angelazhangmsft:
Clarify errors that differ based on which assemblyversion is incremented


Scottj1s (approved):
this is great - thanks for the thorough coverage

PR #643, opened 2020-12-20
Add final diagnostics, refactoring

j0shuams:
* add namespace diagnostics
* move all diagnostic code out of generator.cs
* use helper class for collecting all authored types
* add invalid type diagnostics
* check type symbol kinds instead of strings where possible
* split diagnostic code into two files for readability
* comment fixups
* no more pass by ref, no more returning bool to `found` var
* less generic functions
* more tests
* turn some one-off functions into local functions
* better signature checking algorithm 

* there is still one test (and a variation of it) failing still, but this might not be an issue
* authoring sample built once fine then complained about ReadOnlyArray / WriteOnlyArray attribute not being found, will try again 

I have a couple issues on the roslyn-sdk repo for things that seem like they should work
https://github.com/dotnet/roslyn-sdk/issues/678
https://github.com/dotnet/roslyn-sdk/issues/677


src/Authoring/DiagnosticTests/NegativeData.cs
```diff
@@ -1,75 +1,109 @@
 namespace DiagnosticTests
 {
-    public partial class TestDiagnostics
+    public partial class UnitTesting
     {
 
-        // Cases to add -- WIP
-        private const string StructWithInterfaceField = @"
-namespace Test 
+        // namespace tests -- WIP
```
j0shuams, line 6: update comment

src/Authoring/DiagnosticTests/PositiveData.cs
```diff
@@ -1,10 +1,68 @@
-namespace DiagnosticTests
-{
-    public partial class TestDiagnostics
+namespace DiagnosticTests
+{
+    public partial class UnitTesting
+    {
+        // oops
```
j0shuams, line 5: fixed

src/Authoring/WinRT.SourceGenerator/DiagnosticHelpers.cs
```diff
@@ -0,0 +1,367 @@
...
+                
+                // Nothing can be marked `ref`
+                if (HasModifier(param, SyntaxKind.RefKeyword))
+                { 
+                    Report(WinRTRules.RefParameterFound, method.GetLocation(), param.Identifier); 
+                }
+                
+                if (ParamHasInOrOutAttribute(param))
+                {
+                    // recommend using ReadOnlyArray or WriteOnlyArray
```
j0shuams, line 145: "instead of in/out"

j0shuams, line 145: "instead of in/out"

src/Authoring/AuthoringSample/Program.cs
```diff
@@ -462,7 +460,7 @@ public bool ContainsKey(string key)
             return _dictionary.ContainsKey(key);
         }
 
-        public void CopyTo(KeyValuePair<string, BasicStruct>[] array, int arrayIndex)
+        public void CopyTo([System.Runtime.InteropServices.WindowsRuntime.ReadOnlyArray] KeyValuePair<string, BasicStruct>[] array, int arrayIndex)
```
manodasanW, line 463: We might need to improve our diagnostics to detect whether the function belongs to a custom type mapping.  In this case, this function is part of a custom mapping interface which means this function as is won't be projected.  But it would be removed and a mapping of it would be projected.

src/Authoring/AuthoringSample/Program.cs
```diff
@@ -471,6 +469,7 @@ public void CopyTo(KeyValuePair<string, BasicStruct>[] array, int arrayIndex)
             return _dictionary.GetEnumerator();
         }
 
+        [DefaultOverload]
         public bool Remove(string key)
```
manodasanW, line 473: Same here, this function is part of a custom mapped interface.

src/Authoring/AuthoringSample/Program.cs
```diff
@@ -629,7 +628,7 @@ public bool Contains(DisposableClass item)
             return _list.Contains(item);
         }
 
-        public void CopyTo(DisposableClass[] array, int arrayIndex)
+        public void CopyTo([System.Runtime.InteropServices.WindowsRuntime.ReadOnlyArray] DisposableClass[] array, int arrayIndex)
```
manodasanW, line 631: see other comment.

src/Authoring/DiagnosticTests/PositiveData.cs
```diff
@@ -1,10 +1,67 @@
...
+                new Uri(""http://microsoft.com"")
+                };
+        }
+    }
+}";
+
+
+        // Namespaces
+        private const string Valid_NestedNamespace = @"
+// Assuming the winmd is RuntimeComponent, the error gets thrown since namespace `A` has no common prefix with RuntimeComponent
```
manodasanW, line 39: Isn't this a positive test case file?

j0shuams: Yes this is a valid test, the comment is a copy-paste error, fixed in last commit

src/Authoring/WinRT.SourceGenerator/DiagnosticHelpers.cs
```diff
@@ -0,0 +1,365 @@
...
+        /// <param name="typeSymbol">type that might inherit</param>
+        /// <param name="typeToCheck">Inherited interface or class</param>
+        private bool ImplementsInterface(INamedTypeSymbol typeSymbol, string typeToCheck)
+        {
+            if (typeSymbol == null)
+            {
+                return false;
+            }
+
+            if (typeSymbol.BaseType != null)
```
manodasanW, line 107: Technically the base type can have another base type.  Does that matter in the scenarios this is used in?

src/Authoring/WinRT.SourceGenerator/DiagnosticHelpers.cs
```diff
@@ -0,0 +1,365 @@
...
+            {
+                return typeSymbol.BaseType.MetadataName == typeToCheck || typeSymbol.BaseType.ToString() == typeToCheck;
```
manodasanW, line 109: A type can have a base type and also implement other interfaces.  Should we check interfaces if this is false?

src/Authoring/WinRT.SourceGenerator/DiagnosticHelpers.cs
```diff
@@ -0,0 +1,365 @@
...
+        /// <returns>True iff namespace is disjoint from the assembly name</returns>
+        private bool IsInvalidNamespace(INamespaceSymbol @namespace, string assemblyName)
+        {
+            var contain = @namespace;
+            while (!contain.ContainingNamespace.IsGlobalNamespace)
+            {
+                contain = contain.ContainingNamespace;
+            }
+
+            return !contain.Name.Equals(assemblyName) && !@namespace.Name.Equals(assemblyName);
```
manodasanW, line 199: What if it is A.B.C.type and A.B.WinMD or something like that.

src/Authoring/WinRT.SourceGenerator/DiagnosticHelpers.cs
```diff
@@ -0,0 +1,365 @@
...
+        { 
+            "System.Array",
+            "System.Collections.Generic.Dictionary`2",
+            "System.Collections.Generic.List`1", 
+            "System.Collections.Generic.KeyValuePair"
+        };
+
+        private readonly static string[] WIPInvalidTypes =
+        {
+            "System.Linq.Enumerable`1",
```
manodasanW, line 352: Should we normalize the types as technically they can already include System and then do Linq...

src/Authoring/WinRT.SourceGenerator/DiagnosticHelpers.cs
```diff
@@ -0,0 +1,365 @@
...
+        {
+            "System.Exception",
+            "IAsyncAction",
+            "IAsyncActionWithProgress`1",
+            "IAsyncOperation`1",
+            "IAsyncOperationWithProgress`2",
+        };
+
+        private readonly static string[] InvalidTypes = 
+        {
```
manodasanW, line 343: Don't need to address right now, but we need to do a not valid type check rather than an invalid type check.  But can be addressed later on scenarios that we identify if not already handled.

src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
```diff
@@ -1,494 +1,380 @@
...
-                bool hasWriteOnlyArray = ParamHasAttribute("System.Runtime.InteropServices.WindowsRuntime.WriteOnlyArray", param);
-                bool isOutputParam = ParamMarkedOutput(param);
+            HashSet<string> simplifiedNames = new HashSet<string>();
 
-                // Nothing can be marked `ref`
-                if (ParamMarkedRef(param))
+            foreach (var namespaceSymbol in definedNamespaces)
+            {
+                string upperNamed = namespaceSymbol.ToString().ToUpper();
+                if (simplifiedNames.Contains(upperNamed))
```
manodasanW, line 188: What about 2 namespaces defined multiple times with the same case?  Is that a scenario to consider here?

src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
```diff
@@ -1,494 +1,380 @@
...
-            if (ModifiersContains(field.Modifiers, "const"))
+            if (field.Modifiers.Where(m => m.IsKind(SyntaxKind.ConstKeyword)).Any())
             {
-                Report(ref context, DiagnosticRules.StructHasConstFieldRule, field.GetLocation(), structId);
-                found |= true;
+                Report(WinRTRules.StructHasConstFieldRule, field.GetLocation(), structId);
             }
 
+            // helper function to see if a string is in the list of typenames
+            bool SymbolSetHasString(string typeStr) { return typeNames.Where(sym => sym.ToString().Contains(typeStr)).Any(); }
```
manodasanW, line 309: To confirm, is this looking for a partial match due to contains?

src/Authoring/WinRT.SourceGenerator/Generator.cs
```diff
@@ -178,103 +173,30 @@ private void GenerateWinMD(MetadataBuilder metadataBuilder, string outputFile)
...
-                    }
-                    if (!fields.Any())
-                    {
-                        context.ReportDiagnostic(Diagnostic.Create(DiagnosticRules.StructWithNoFieldsRule, structDeclaration.GetLocation(), model.GetDeclaredSymbol(structDeclaration)));
-                        found |= true;
-                    }
-                }
-            }
-            return found;
+            WinRTScanner winrtScanner = new WinRTScanner(context, assemblyName);
```
manodasanW, line 182: Consider WinRTComponentScanner or WinRTComponentDiagnostics?

j0shuams: I like it! Thanks

src/Authoring/WinRT.SourceGenerator/Generator.cs
```diff
@@ -178,103 +173,30 @@ private void GenerateWinMD(MetadataBuilder metadataBuilder, string outputFile)
...
+            winrtScanner.FindDiagnostics();
+            return winrtScanner.Found();
         }
 
         public void Execute(GeneratorExecutionContext context)
         {
+            /*
```
j0shuams, line 189: take this out before final merge, can we use a directive here like `#if NET5_0`

src/Authoring/WinRT.SourceGenerator/DiagnosticHelpers.cs
j0shuams: It seems like there is some ambiguity about the `BaseType` field of type symbols. For type symbols of classes, any interface implemented is actually a basetype (e.g. IAsyncAction is not in the symbol's `AllInterfaces` field, but in its `BaseType` field).

For an interface type symbol, this is not true, and the `IAsyncAction` symbol **is** in the `AllInterfaces` field. So the loop is for interfaces, when we want to check what a class implements, we need it's `BaseType`, but an interface has a null `BaseType`. 

And class type symbols need a `BaseType` check, to find `System.Exception` for example. 

It is also curious that calling `GetTypeByMetadataName("Windows.Foundation.IAsyncAction")` returns `null` instead of the type symbol we might expect. This is why we have to do the `==` check.

src/Authoring/WinRT.SourceGenerator/DiagnosticHelpers.cs
j0shuams: I had moved the interface loop up to before we check the base type. Not sure what you mean by check interfaces if this is false

src/Authoring/WinRT.SourceGenerator/DiagnosticHelpers.cs
j0shuams: I am, I build up a string of the containing namespace(s) for the type before comparing it to these strings. I don't have a test for if a user defined a class named Enumerable and uses it, so I will add that

src/Authoring/WinRT.SourceGenerator/DiagnosticHelpers.cs
j0shuams: Went ahead and changed the name, but not sure I understand the difference in not valid and invalid

src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
j0shuams: If the namespace name is valid (e.g. is the assembly name or a namespace prefix with the assembly name) then that's fine. Otherwise it will get the "disjoint namespace" diagnostic, meaning the types won't be accessible. I'll add a test for this thanks

src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
j0shuams: I found an alternative approach that doesn't use this kind of check

src/Authoring/WinRT.SourceGenerator/DiagnosticHelpers.cs
j0shuams: Yes I have tests for these and have added one that more obviously tests what you are saying -- they are blocked by issue https://github.com/dotnet/roslyn/issues/50112 

As the namespace C will have "{global namespace}" as its ContainingNamespace, instead of "A.B"

src/Authoring/WinRT.SourceGenerator/Generator.cs
j0shuams: tried this and setting it in the diagnostictest project properties didn't seem to apply, so it didn't work...

src/Authoring/AuthoringSample/Program.cs
j0shuams: I've updated it to now look for IList, IDictionary, and IReadOnlyDictionary, and filter out these interfaces' methods and properties from those that get scanned

src/Authoring/WinRT.SourceGenerator/DiagnosticHelpers.cs
```diff
@@ -330,6 +353,111 @@ private string SimplifySyntaxTypeString(string syntaxType)
...
+            SpecialType.System_UInt64,
+            SpecialType.System_Int16,
+            SpecialType.System_Int32,
+            SpecialType.System_Int64,
+            SpecialType.System_Enum,
+        };
+
+        private static readonly Dictionary<string, List<string>> MappedCSharpTypeMethods = new Dictionary<string, List<string>>()
+        {
+            { "System.DateTimeOffset", new List<string> { "Add", "AddDays", "AddHours", "AddMilliseconds", "AddMinutes", "AddMonths", "AddSeconds", "AddTicks", "AddYears", "Compare", "CompareTo",
```
manodasanW, line 373: You only need the interfaces as those are the ones that can be implemented by component classes.  The custom mapped classes I believe are all sealed so they are not used for extending.

src/Authoring/WinRT.SourceGenerator/DiagnosticHelpers.cs
```diff
@@ -330,6 +353,111 @@ private string SimplifySyntaxTypeString(string syntaxType)
...
+                "Equals", "EqualsExact", "FromFileTime", "FromUnixTimeMilliseconds", "FromUnixTimeSeconds", "GetHashCode", "Parse", "ParseExact", "Subtract", "ToFileTime", "ToLocalTime",
+                "ToOffset", "ToString", "ToUniversalTime", "ToUnixTimeMilliseconds", "ToUnixTimeSeconds", "TryFormat", "TryParse", "TryParseExact" }  },
+            { "System.Exception", new List<string> { "GetBaseException", "GetObjectData", "GetType", "ToString" } },
+            { "System.IDisposable", new List<string> { "Dispose" } },
```
manodasanW, line 377: Any way to make the method lookup generic.  I am concerned about overloaded methods with different # of parameters that are from the component.

src/Authoring/WinRT.SourceGenerator/DiagnosticHelpers.cs
```diff
@@ -330,6 +353,111 @@ private string SimplifySyntaxTypeString(string syntaxType)
...
+            { "System.Collections.Generic.IEnumerable`1", new List<string> { "GetEnumerator" } },
+            { "System.Collections.Generic.KeyValuePair`2", new List<string> { "Deconstruct", "ToString" } },
+            { "System.Collections.Generic.IReadOnlyDictionary`2", new List<string> { "ContainsKey", "TryGetValue" } },
+            { "System.Collections.Generic.IDictionary`2", new List<string> { "Add", "Clear", "Contains", "CopyTo", "ContainsKey", "Remove", "TryGetValue" } },
+            { "System.Collections.Generic.IList`1", new List<string> {"Add", "Clear", "Contains", "IndexOf", "Insert", "Remove", "RemoveAt", "CopyTo" } },
+            { "System.Collections.IList`1", new List<string> {"Add", "Clear", "Contains", "IndexOf", "Insert", "Remove", "RemoveAt", "CopyTo" } }
+        };
+
+        private static readonly Dictionary<string, List<string>> MappedCSharpTypeProperties = new Dictionary<string, List<string>>()
+        {
```
manodasanW, line 422: Same here.

src/Authoring/WinRT.SourceGenerator/DiagnosticRules.cs
```diff
@@ -0,0 +1,150 @@
...
+        /// <param name="messageFormat">string, describes the diagnostic -- formatted with {0}, ... -- 
+        /// such that data can be passed in for the code the diagnostic is reported for</param>
+        private static DiagnosticDescriptor MakeRule(string id, string title, string messageFormat)
+        {
+            return new DiagnosticDescriptor(
+                id: id,
+                title: title,
+                messageFormat: messageFormat,
+                category: "Usage",
+                /* Warnings dont fail command line build; winmd generation is prevented regardless of severity.
```
manodasanW: Did you plan on addressing this or future PR?

j0shuams: Yes we want to do it now, thanks

src/Authoring/WinRT.SourceGenerator/DiagnosticHelpers.cs
j0shuams: Addressed

src/Authoring/WinRT.SourceGenerator/DiagnosticRules.cs
j0shuams: addressed

src/Authoring/WinRT.SourceGenerator/DiagnosticHelpers.cs
j0shuams: addressed

src/Authoring/WinRT.SourceGenerator/Generator.cs
```diff
@@ -1,327 +1,247 @@
...
+            // shouldn't need workaround once we can pass AnalyzerConfigOptionsProvider in DiagnosticTests.Helpers.cs
+            string assemblyName = GetAssemblyName(context) ?? "DiagnosticTests";
+            WinRTComponentScanner winrtScanner = new WinRTComponentScanner(context, assemblyName);
+            winrtScanner.FindDiagnostics();
+            return winrtScanner.Found();
+        }
+
+        public void Execute(GeneratorExecutionContext context)
+        {
+            /*
```
j0shuams, line 188: when we merge this pr, i will take these out before merging the main topic pr

src/Authoring/WinRT.SourceGenerator/DiagnosticHelpers.cs
```diff
@@ -47,8 +47,7 @@ private void ReportDiagnostic(Diagnostic d)
             Flag();
             _context.ReportDiagnostic(d);
         }
-
-        private bool SymEq(ISymbol sym1, ISymbol sym2) { return SymbolEqualityComparer.Default.Equals(sym1, sym2); }
+        private bool SymEq(ISymbol sym1, ISymbol sym2) { return sym1.Equals(sym2); /*SymbolEqualityComparer.Default.Equals(sym1, sym2);*/ }
```
manodasanW, line 50: Is this intentional?  Do you plan on going back to Equals as is?

src/Authoring/WinRT.SourceGenerator/DiagnosticHelpers.cs
```diff
@@ -368,97 +367,58 @@ private string SimplifySyntaxTypeString(string syntaxType)
             SpecialType.System_Enum,
         };
 
-        private static readonly Dictionary<string, List<string>> MappedCSharpTypeMethods = new Dictionary<string, List<string>>()
+        private static readonly HashSet<string> MappedCSharpTypes = new HashSet<string>()
```
manodasanW, line 370: Any chance for both this list and the other, we can use the one already in WinRTTypeWriter.cs.  We can move it to a helper file if needed.  Is there any difference between both lists other than this only has the keys?

src/Authoring/WinRT.SourceGenerator/DiagnosticHelpers.cs
```diff
@@ -368,97 +367,58 @@ private string SimplifySyntaxTypeString(string syntaxType)
...
+            "System.Collections.Generic.ICollection`1",
+            "System.Collections.Generic.IReadOnlyCollection`1",
+            "System.Collections.IEnumerable",
+            "System.Collections.IEnumerator",
+            "System.Collections.IList"
         };
 
-        private static readonly string[] nonWindowsRuntimeInterfaces = 
+        private static readonly HashSet<string> nonWindowsRuntimeInterfaces = new HashSet<string>()
         {
```
manodasanW, line 422: Should the other 4 types here be qualified names similar to the first one?

src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
```diff
@@ -121,6 +100,34 @@ public void FindDiagnostics()
...
+                }
+                return false;
+            }
+
+            foreach (var @interface in typeSymbol.AllInterfaces.
+                        Where(symbol => MappedCSharpTypes.Contains(QualifiedName(symbol)) ||
+                                        ImplementedInterfacesWithoutMapping.Contains(QualifiedName(symbol))))
+            {
+                var mems = @interface.GetMembers();
+                methods = methods.Where(m => !mems.Where(mem => mem.Name == m.Identifier.Text).Any());
```
manodasanW, line 126: I feel like you would want a symbol comparison here as what happens if 2 functions have same name, but have a diff # of parameters and one is from a custom type mapping while the other isn't.

src/Authoring/WinRT.SourceGenerator/DiagnosticHelpers.cs
j0shuams: I went back to the original, thanks for catching this.

src/Authoring/WinRT.SourceGenerator/DiagnosticHelpers.cs
j0shuams: yes we are sharing now

src/Authoring/WinRT.SourceGenerator/DiagnosticHelpers.cs
j0shuams: Well gettypebymetadataname fails for the qualified ones, and I checked one of our tests for these interfaces, and when the class implements it, the interface's containing namespace is global (unless qualified when giving the class the base type), which means qualified names won't help, just metadata name.

src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
j0shuams: addressed

src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
```diff
@@ -109,22 +110,20 @@ string QualifiedName(INamedTypeSymbol sym)
...
-                                        ImplementedInterfacesWithoutMapping.Contains(QualifiedName(symbol))))
+                        Where(symbol => WinRTTypeWriter.MappedCSharpTypes.ContainsKey(QualifiedName(symbol)) ||
+                                        WinRTTypeWriter.ImplementedInterfacesWithoutMapping.Contains(QualifiedName(symbol))))
             {
-                var mems = @interface.GetMembers();
-                methods = methods.Where(m => !mems.Where(mem => mem.Name == m.Identifier.Text).Any());
-                properties = properties.Where(p => !mems.Where(mem => SameProp(mem.Name, p.Identifier.Text)).Any()); 
+                var InterfaceMembers = @interface.GetMembers();
+
+                bool ValidMember(MemberDeclarationSyntax member)
```
manodasanW, line 119: might be performant to do a loop over interface members and build a list of class members using FindImplementationForInterfaceMember and then do a remove all or where not contains.

src/Authoring/WinRT.SourceGenerator/DiagnosticHelpers.cs
manodasanW: My only concern was if someone used the full typename or somewhere whether we will miss it.  But if we are comparing metadata name, it probably is fine. Just seemed inconsistent with System.Exception.  And what happens if someone had their own version of IAsyncAction in their own namespace?

manodasanW (approved):

src/Authoring/WinRT.SourceGenerator/DiagnosticHelpers.cs
j0shuams: Yeah we do catch full name, but I bet we won't let it happen if they write IAsyncAction, but should they be allowed to write IAsyncAction? they are are encouraged (in windmdexp) to use IAsyncInfo class for exception code

manodasanW: Well I meant IAsyncAction from some other namespace (different type).

j0shuams: I meant is Exception a valid winrt type? implementing IAsyncInfo would mean having a property that is an Exception.. Maybe there's an edge case there with them also writing their own Exception class?

manodasanW: I think it is valid, but probably want to confirm.  You probably also want to check if inheriting is allowed, but I think not.

PR #651, opened 2021-01-05
Authoring: Diagnostic for private getter

j0shuams:
There was a diagnostic missed for private getters, I am adding a positive and negative test case for private setter and private getter on properties (respectively). My change is to look at the method accessibility on all property symbols, via the `CheckProperties` function -- which is the new name of `CheckPropertySignature`. 

The function `CheckPropertySignature` used in checking classe **and** interfaces, and this check really only needs to happen for classes. Originally, I thought to take the loop out of the helper function and leave it in the class loop of `FindDiagnostics`. But felt like looping over properties twice wasn't essential. Open to feedback on whether to do this check for interfaces or not.

### DIFF
--- a/src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
+++ b/src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
@@ -64,9 +64,10 @@ namespace Generator
 
                     // check for things in nonWindowsRuntimeInterfaces
                     ImplementsInvalidInterface(classSymbol, @class);
-                    
-                    CheckPropertySignature(props, classId);
-                    
+                   
+                    CheckProperties(props, classId);
+                   
+                    // check types -- todo: check for !valid types
                     CheckMethods(publicMethods, classId);
                 }
 
@@ -87,7 +88,7 @@ namespace Generator
 
                     ImplementsInvalidInterface(interfaceSym, @interface);
                     
-                    CheckPropertySignature(props, @interface.Identifier);
+                    CheckProperties(props, @interface.Identifier);
                     
                     CheckMethods(methods, @interface.Identifier);
                 }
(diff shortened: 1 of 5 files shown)


src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
```diff
@@ -222,13 +223,18 @@ private void CheckMethods(IEnumerable<MethodDeclarationSyntax> methodDeclaration
...
         /// <returns>True iff any of the invalid array types are used in any of the propertyy signatures in the given class</returns>
-        private void CheckPropertySignature(IEnumerable<PropertyDeclarationSyntax> props, SyntaxToken typeId)
+        private void CheckProperties(IEnumerable<PropertyDeclarationSyntax> props, SyntaxToken typeId)
         {
             foreach (var prop in props)
             {
                 var propSym = GetModel(prop.SyntaxTree).GetDeclaredSymbol(prop);
                 var loc = prop.GetLocation();
 
+                if (propSym.GetMethod.DeclaredAccessibility.Equals(Accessibility.Private))
```
manodasanW, line 233: You probably want to check != public as internal and so on probably falls in the same case as private.

j0shuams: addressed, thanks!

src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
```diff
@@ -222,13 +223,18 @@ private void CheckMethods(IEnumerable<MethodDeclarationSyntax> methodDeclaration
...
         /// <returns>True iff any of the invalid array types are used in any of the propertyy signatures in the given class</returns>
-        private void CheckPropertySignature(IEnumerable<PropertyDeclarationSyntax> props, SyntaxToken typeId)
+        private void CheckProperties(IEnumerable<PropertyDeclarationSyntax> props, SyntaxToken typeId)
         {
             foreach (var prop in props)
             {
                 var propSym = GetModel(prop.SyntaxTree).GetDeclaredSymbol(prop);
                 var loc = prop.GetLocation();
 
+                if (!propSym.GetMethod.DeclaredAccessibility.Equals(Accessibility.Public))
```
manodasanW, line 233: Does this error if they didn't have a get method at all?  We probably want to catch that case too.

Scottj1s (approved):

src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
j0shuams: No it doesn't. I added a unittest for the scenario you are describing and it passes with no other changes. I think this is because all methods are considered public unless otherwise declared

manodasanW (approved):

PR #652, opened 2021-01-05
Composable type support and bug fixes

manodasanW:
- Added support for authoring types which extend composable types
- As part of adding support for it, made a design change where we no longer instantiate an instance of the wrapper class, but rather use it only for generating the v-tables.  As an effect of this change, the class is used rather than the default interface for non-interface members in the FindObject calls in the ABI interface.  This default interface was the main reason why the wrapper class was needed, but using the class in equivalent of it would achieve the same given both scenarios don't support proxies for the interface.  In addition, this probably has the effect of improving performance given we no longer need to instantiate a wrapper type.
- Added support for passing additional WinMDs of the dependent libraries to the CsWinRT tool.  Currently this is passed manually, but hope is to automate it later if possible.
- Added WinUI test app which consumed some authored types which extended the composable types.
- Added support for static classes and classes without default constructor
- Added support for private set
- Added IManagedActivationFactory for calling static constructor which is implemented by the generated activation factory and called in .XAML scenarios
- Fixed issue where WindowsRuntimeType attribute didn't have source metadata passed for scenarios where it is used in classes

Contributes to #309


j0shuams (approved):

src/Authoring/AuthoringConsumptionTest/packages.config
```diff
@@ -1,5 +1,7 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="Microsoft.googletest.v140.windesktop.msvcstl.static.rt-dyn" version="1.8.1.3" targetFramework="native" />
+  <package id="Microsoft.VCRTForwarders.140" version="1.0.7" targetFramework="native" />
   <package id="Microsoft.Windows.CppWinRT" version="2.0.201113.7" targetFramework="native" />
+  <package id="Microsoft.WinUI" version="3.0.0-preview2.200713.0" targetFramework="native" />
```
Scottj1s, line 6: should keep in sync with MicrosoftWinUIVersion (3.0.0-preview3.201113.0)
could automate by generating packages.config, or leave comments in both places

Scottj1s:
> <Project Sdk="Microsoft.NET.Sdk">

still think this should either move under Samples, or be changed to a unit test

---
Refers to: src/Authoring/AuthoringSample/AuthoringSample.csproj:1 in a96c9fe. [](commit_id = a96c9fe6c04b95cd771cad542fde12717ce6b0de, deletion_comment = False)

src/Authoring/AuthoringWinUITest/AuthoringWinUITest/PropertySheet.props
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
```
Scottj1s, line 1: can delete this

src/Authoring/AuthoringWinUITest/AuthoringWinUITest/Directory.Build.targets
```diff
@@ -0,0 +1,15 @@
+<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <PropertyGroup>
+    <PrepareForRunDependsOn>CopyTestAssets;$(PrepareForRunDependsOn)</PrepareForRunDependsOn>
+  </PropertyGroup>
+
+  <Target Name="CopyTestAssets">
```
Scottj1s, line 7: comment temporary until nuget supports (and tests like this should pull in that support as if the nuget was referenced)

Scottj1s:
> <?xml version="1.0" encoding="utf-8"?>

we should try to keep to the new repo structure - Samples, Tests, etc

---
Refers to: src/Authoring/AuthoringWinUITest/AuthoringWinUITest/AuthoringWinUITest.vcxproj:1 in a96c9fe. [](commit_id = a96c9fe6c04b95cd771cad542fde12717ce6b0de, deletion_comment = False)

src/Authoring/AuthoringWinUITest/AuthoringWinUITest/App.idl
```diff
@@ -0,0 +1,3 @@
+namespace AuthoringWinUITest
```
Scottj1s, line 1: can delete

src/cswinrt/code_writers.h
```diff
@@ -5905,15 +5961,21 @@ IntPtr instance = _factory.ActivateInstance();
 return ObjectReference<IInspectable.Vftbl>.Attach(ref instance).As<I>();
 }
 
+public void RunClassConstructor()
```
Scottj1s, line 5964: I don't see any tests/samples that exercise this.  But could we do away with IManagedActivationFactory (and the breaking change) by just calling RunClassConstructor from a static constructor of the activation factory?

manodasanW: The WinUI test triggers it.  In the below code, when you set custom properties in a XAML file, XAML QIs for IManagedActivationFactory and does a call to RunClassConstructor.

```       <custom:CustomButton x:Name="myButton" Click="myButton_Click" OverrideEntered="False">Custom</custom:CustomButton>```

I will try removing the interface and calling it as part of ActivationFactory and see what happens, but WinUI may break I think.

Scottj1s: Looks like WinUI calls it [here](https://dev.azure.com/microsoft/WinUI/_git/microsoft-ui-xaml-lift?path=%2Fdxaml%2Fxcp%2Fcore%2Fmetadata%2FReflectionAPI.cpp&version=GBmaster&line=467&lineEnd=468&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents).  If the activation factory type is accessed (causing the static ctor to execute), and the cast to IManagedActivationFactory then fails, that might still be a reliable mechanism.

src/Authoring/AuthoringConsumptionTest/packages.config
manodasanW: Updated version.  Turns out there is a couple other places using preview 2 still which I will look at updating all in another PR.  Might be a bit harder to automate given the vcxproj need to be updated, but will take a look.

src/cswinrt/code_writers.h
manodasanW: Went ahead and removed it as we don't see a disadvantage or scenario that might be impacted by doing that and allows us to not increment assembly version.

manodasanW:
> > 
> 
> still think this should either move under Samples, or be changed to a unit test
> 
> Refers to: src/Authoring/AuthoringSample/AuthoringSample.csproj:1 in [a96c9fe](https://github.com/microsoft/CsWinRT/commit/a96c9fe6c04b95cd771cad542fde12717ce6b0de). [](commit_id = [a96c9fe](https://github.com/microsoft/CsWinRT/commit/a96c9fe6c04b95cd771cad542fde12717ce6b0de), deletion_comment = False)

I agree, I got a separate branch with the rename and move of all the authoring tests to the test folder.  Will create a PR for it once this PR is merged.

src/Authoring/AuthoringConsumptionTest/packages.config
Scottj1s: another technique would be to assert the WinUI version in the root directory.build.props/targets  - doesn't auto-config everything, but achieves the same end by breaking the build

Scottj1s (approved):
looks good - won't hold you hostage any longer - we can move sample/test projects later

PR #666, opened 2021-01-14
Authoring bug fixes and xaml type mapping

### LINKED ISSUE #662: Authoring with user control - diagnostics catching Xaml generated file
**Describe the bug**
Authoring diagnostics catches that one of Xaml's generated files doesn't have the default overload attribute on their classes.  It used to be that as long as the attribute was set on the interface (which it is), it was implied the same attribute on class.

**To Reproduce**

1. Add empty User Control WinUI template to authoring component project
2. Build results in error: Severity    Code    Description    Project    File    Line    Source    Suppression State    Tool
Error    WME1085    In class XamlMetaDataProvider: The 1-parameter overloads of GetXamlType must have exactly one method specified as the default overload by decorating it with Windows.Foundation.Metadata.DefaultOverloadAttribute.    UserControlLibrary    C:\Users\angzhang\source\repos\WinUII3Desktopnternal\UserControlLibrary\obj\Debug\net5.0-windows10.0.18362.0\XamlTypeInfo.g.cs    46    Build        Csc

**Version Info**

1.1.1-prerelease.210111.5

manodasanW:
- Finished up incomplete XAML type mappings
- Added support for propagating class member attributes to the respective static / default interface member
- Added support for propagating interface member attribute to the class member
- Fixed bugs with ordering of interface impls

Fixes #662 and contributes to #309

### DIFF
--- a/src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
+++ b/src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
@@ -140,7 +140,8 @@ namespace Generator
         public Dictionary<string, List<ISymbol>> MethodsByName = new Dictionary<string, List<ISymbol>>();
         public Dictionary<ISymbol, string> OverloadedMethods = new Dictionary<ISymbol, string>();
         public List<ISymbol> CustomMappedSymbols = new List<ISymbol>();
-        public List<ISymbol> SymbolsWithAttributes = new List<ISymbol>();
+        public HashSet<ISymbol> SymbolsWithAttributes = new HashSet<ISymbol>();
+        public Dictionary<ISymbol, ISymbol> ClassInterfaceMemberMapping = new Dictionary<ISymbol, ISymbol>();
 
         public TypeDeclaration()
             :this(null)
@@ -1851,7 +1852,7 @@ namespace Generator
                 metadataBuilder.GetOrAddBlob(attributeSignature));
         }
 
-        private void AddCustomAttributes(IList<AttributeData> attributes, EntityHandle parentHandle)
+        private void AddCustomAttributes(IEnumerable<AttributeData> attributes, EntityHandle parentHandle)
         {
             foreach (var attribute in attributes)
             {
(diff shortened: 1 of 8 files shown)


src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
```diff
@@ -2425,6 +2580,54 @@ void AddType(INamedTypeSymbol type)
             }
         }
 
+        void AddCustomAttributes(TypeDeclaration typeDeclaration, string interfaceName = null)
```
Scottj1s, line 2583: given any thought to reversing SupportedOSPlatform?  It could always map to UniversalApiContract of the appropriate nearest version (for which the OS version is less than or equal to the one supplied with SupportedOSPlatform).  Then, other language projections could based their own downlevel version checks on the Contract (a C#/WinRT does).

src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
```diff
@@ -274,24 +284,25 @@ public MappedType(Func<ISymbol, (string, string, string, bool, bool)> multipleMa
         {
```
Scottj1s, line 284: where is Windows.Foundation.Metadata.GuidAttribute <--> System.Runtime.InteropServices.Guid handled?  i.e., how does an authored interface specify an explicit guid?

manodasanW: We don't currently have support for explicitly specifying a GUID, will open a issue to track that.

Scottj1s: are there any other custom attribute mappings we need to capture in an issue?

src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
manodasanW: I thought the contract attributes were only meant for Windows APIs as per [this](https://docs.microsoft.com/en-us/uwp/midl-3/predefined-attributes#the-contract-attribute).

Scottj1s: I suppose so - doesn't sound like there's any way then to enable a component author to assert an OS requirement

src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
manodasanW: Based on looking [here](https://github.com/microsoft/CsWinRT/blob/86895f639886517484ecc8db06ed036d3460dc96/src/cswinrt/code_writers.h#L1291) and the list of attributes, the only other special one is GCPressure which I don't believe applies here.  There is also AllowMultiple, but that is part of authoring attributes support.

Scottj1s: sounds good

Scottj1s (approved):

src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
manodasanW: I wonder if we can detect the target framework used and any SupportedOSPlatform and then when we generate our targets file which they include, we add something that checks if the targetplatformversion in the consumer is compatible with that.  I will open a issue to track investigating that.

PR #669, opened 2021-01-15
Fix issue where the same component can be loaded in multiple ALC and cause cast issues 

### LINKED ISSUE #633: Hosting support loading same binary in multiple assembly loader context causing type cast issues
'[A]ABI.Impl.AuthoringSample.DisposableClass cannot be cast to [B]ABI.Impl.AuthoringSample.DisposableClass.   where A and B are both same binary, but are in different assembly loader context.

manodasanW:
Previously if you activate multiple types from a component, you can run into cast issues when you pass things between them.  This was because each type activation from a component was going into its own ALC.  Now we have a singleton pattern where for a given component binary, it gets loaded in the same ALC preventing the cast issues.  There is still a possibility if multiple components are in use and they pass dependent authored types between each other, we can possibly still run into similar issues which I am not addressing as part of this PR.

Also changing bin directory of wapproj projects to _build folder.

Fixes #633

### DIFF
--- a/src/Authoring/WinRT.Host.Shim/Module.cs
+++ b/src/Authoring/WinRT.Host.Shim/Module.cs
@@ -2,14 +2,8 @@
 // to simplify deployment
 
 using System;
-using System.Collections.Generic;
-using System.IO;
+using System.Collections.Concurrent;
 using System.Reflection;
-using System.Runtime.CompilerServices;
-using System.Runtime.InteropServices;
-using System.Text;
-using Windows.Foundation;
-using WinRT;
 
 #if !NETSTANDARD2_0
 using System.Runtime.Loader;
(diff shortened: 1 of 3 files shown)


src/Authoring/WinRT.Host.Shim/Module.cs
```diff
@@ -69,13 +63,14 @@ private static class ActivationLoader
...
-        {
+        {
+            private static readonly ConcurrentDictionary<string, ActivationLoader> ALCMapping = new ConcurrentDictionary<string, ActivationLoader>();
             private AssemblyDependencyResolver _resolver;
 
             public static Assembly LoadAssembly(string targetAssembly)
             {
-                var loader = new ActivationLoader(targetAssembly);
-                return loader.LoadFromAssemblyPath(targetAssembly);
+                return ALCMapping.GetOrAdd(targetAssembly, (_) => new ActivationLoader(targetAssembly))
```
j0shuams, line 72: Slick!

j0shuams (approved):

Scottj1s (approved):

PR #671, opened 2021-01-15
Fix path to generated winmd

j0shuams:
My PR to update nuget targets had a bug when trying to copy the generated winmd to the output (bin) directory. This should fix it (tested locally)

### DIFF
--- a/nuget/Microsoft.Windows.CsWinRT.targets
+++ b/nuget/Microsoft.Windows.CsWinRT.targets
@@ -143,14 +143,14 @@ $(CsWinRTFilters)
   <Target Name="CsWinRTPlaceWinMDInOutputFolder"
           Condition="'$(CsWinRTComponent)' == 'true'" 
           BeforeTargets="AfterBuild">
-      <ItemGroup>
-        <CsWinRTGeneratedWinMD Include="$(GeneratedFilesDir)\net5**\$(AssemblyName).winmd"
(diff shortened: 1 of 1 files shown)


Scottj1s (approved):

nuget/Microsoft.Windows.CsWinRT.targets
```diff
@@ -143,10 +143,14 @@ $(CsWinRTFilters)
   <Target Name="CsWinRTPlaceWinMDInOutputFolder"
           Condition="'$(CsWinRTComponent)' == 'true'" 
           BeforeTargets="AfterBuild">
-    <Copy SourceFiles="Generated Files\$(AssemblyName).winmd"
-          DestinationFolder="$(TargetDir)"
-          UseHardlinksIfPossible="false"
-          SkipUnchangedFiles="true" />
+    <ItemGroup>
+      <CsWinRTGeneratedWinMD Include="$(MSBuildProjectDirectory)\Generated Files\net5**\$(AssemblyName).winmd"
```
manodasanW, line 147: Consider CsWinRTGeneratedFilesDir

manodasanW: Actually for now you might want GeneratedFilesDir as this is what source generator uses.

manodasanW (approved):

PR #673, opened 2021-01-17
Tactical mem leak fix for aggregation

Scottj1s:
My main concern with this technique is that when _inner is disposed before its owner is finalized, the finalizer has no chance to detach and the refcount is too low by 1.  Since finalizers can run on any thread, in any order, I'm not confident that keeping the inner refcount artificially low is a reliable fix here.  In fact, I think any approach that tries to do so is doomed.  There's just no safe way to defer an AddRef while still holding onto a COM pointer - it's a time bomb.


src/Directory.Build.targets
```diff
@@ -7,6 +7,7 @@
       https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet5/nuget/v3/index.json;
       https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json;
       https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json;
+      $(RestoreSources)
```
AaronRobinsonMSFT, line 10: Is this intentional?

src/WinRT.Runtime/WinRT.Runtime.csproj
```diff
@@ -21,6 +21,7 @@
     <Copyright>Copyright (c) Microsoft Corporation. All rights reserved.</Copyright>
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>key.snk</AssemblyOriginatorKeyFile>
+    <RuntimeIdentifiers>win-x86;win-x64</RuntimeIdentifiers>
```
AaronRobinsonMSFT, line 24: Are there plans for ARM64?

src/WinRT.Runtime/ObjectReference.cs
```diff
@@ -1,305 +1,334 @@
...
+        {
+            get
+            {
+                ThrowIfDisposed();
+                return _thisPtr;
+            }
+        }
+
+#if DEBUG
+        private unsafe uint RefCount
```
AaronRobinsonMSFT, line 32: I don't believe this needs to be marked `unsafe`.

src/WinRT.Runtime/ObjectReference.cs
```diff
@@ -1,305 +1,334 @@
...
+            ComCallData data = default;
+            IntPtr contextCallbackPtr = _contextCallbackPtr;
+
+            var contextCallback = new ABI.WinRT.Interop.IContextCallback(ObjectReference<ABI.WinRT.Interop.IContextCallback.Vftbl>.Attach(ref contextCallbackPtr));
+
+            contextCallback.ContextCallback(_ =>
+            {
+                base.Release();
+                return 0;
+            }, &data, IID_ICallbackWithNoReentrancyToApplicationSTA, 5);
```
AaronRobinsonMSFT, line 317: Perhaps doc what 5 is?

src/WinRT.Runtime/ComWrappersSupport.net5.cs
```diff
@@ -161,7 +166,217 @@ public static void InitializeComWrappers(ComWrappers wrappers = null)
...
+            }
+
+            if (isAggregation)
+            {
+                Marshal.AddRef(newInstance);
+                var outerRC = Marshal.Release(newInstance);
+                if (outerRC != 0)
+                {
+                    // In aggregation scenarios, at this point outer's refcount must be 0
+                    System.Diagnostics.Debugger.Break();
```
AaronRobinsonMSFT, line 359: I thought this will crash a process if a debugger is not attached. Is that desirable? Perhaps check with `Debugger.IsAttached`? Alternatively if a debugger isn't attached throw an exception?

AaronRobinsonMSFT (commented):
Cursory comments - will look in more detail next week.

src/Tests/UnitTest/TestComponentCSharp_Tests.cs
```diff
@@ -2253,96 +2253,145 @@ public void TestNonProjectedRuntimeClass()
...
+                WarningOverridableMethod(); // warning CA1416
+                WarningOverridableProperty = 0; // warning CA1416
+                // see https://github.com/microsoft/cppwinrt/issues/782
+                //WarningOverridableEvent += (object s, Int32 v) => { }; // warning CA1416
+            }
+        }
+
+        // Manual for now - verify that all APIs targeting 19041 generate a warning
+        // Consider invoking roslyn and scraping/analyzing output as an automated test
+        private void Test()
```
manodasanW, line 2364: Not a feedback for this PR, but as an FYI, we can probably automate this test this by creating a C# compilation context and triggering a compilation on it and then checking for diagnostics.  An example of this is the diagnostics test which Joshua wrote which do this.

Scottj1s: Feel free to open an issue with specific suggestion

src/Directory.Build.targets
Scottj1s: yes, opportunistic to allow adding on for other projects (Angela discovered this and reported it as a bug to DevDiv)

src/WinRT.Runtime/WinRT.Runtime.csproj
Scottj1s: This was unintentional

src/WinRT.Runtime/ObjectReference.cs
Scottj1s: needed for ThisPtr: CS0214: Pointers and fixed size buffers may only be used in an unsafe context

src/WinRT.Runtime/ComWrappersSupport.net5.cs
Scottj1s: thanks - will wrap with IsAttached (as well as BreakOnDispose, if I decide to check that it)

src/WinRT.Runtime/ObjectReference.cs
Scottj1s: it's ignored, but will add a comment.  See: https://devblogs.microsoft.com/oldnewthing/20191128-00/?p=103157

PR #674, opened 2021-01-19
Fixed ObjectDisposedException on delegates

### LINKED ISSUE #640: Random exceptions when calling UWP APIs on .NET 5
Original author @a11cf0 . Transferred from https://github.com/dotnet/runtime/issues/46148

### Description
I'm trying to call Windows 10 UWP OCR APIs from a .NET 5 Windows Forms application. The code which i'm using was perfectly working on .NET Core 3.1 but on .NET 5 it's extremely unstable. For some reason the code throws either ObjectDisposedException or sometimes InvalidCastException. this usually happens on the call to GetSoftwareBitmapAsync() but sometimes right on the first await. Also, hen wrapped in a method, it usually works on the first run but on subsequent runs it fails as above. Here's a full code sample tested on both .NET Core 3.1 and .NET 5 with the above results.

```
using System;
using System.Drawing;
using System.Drawing.Imaging;
using System.IO;
using System.Threading.Tasks;

namespace Supergovno
{
    static class Program
    {
        static async Task<int> Main()
        {
            Bitmap Screen = new Bitmap(500, 500);
            Graphics g = Graphics.FromImage(Screen);
            g.CopyFromScreen(1, 1, 00, 0, Screen.Size);
            Bitmap S = new Bitmap(Screen, new Size(500, 500));
            var engine = Windows.Media.Ocr.OcrEngine.TryCreateFromLanguage(new Windows.Globalization.Language("en"));
            using MemoryStream memoryStream = new();
            S.Save(memoryStream, ImageFormat.Bmp);
            var decoder = await Windows.Graphics.Imaging.BitmapDecoder.CreateAsync(memoryStream.AsRandomAccessStream());
            var softwareBitmap = await decoder.GetSoftwareBitmapAsync();
            var ocrResult = await engine.RecognizeAsync(softwareBitmap);
            Console.WriteLine(ocrResult.Text);
            Console.WriteLine("OK!");
            Console.ReadLine();
            return 0;
        }
    }
}
```

### Configuration
I'm using .NET 5.0.101 on Windows 10 X64 10.0.19042.

### Regression?
Yes. Everything is working on .NET Core 3.1.

manodasanW:
A delegate previously associated with a this ptr may be reused by the ComWrappers API even though its native object has already been disposed.  In this case, resurrect them similar to how we did with the non delegate scenario.

Fixes #640

### DIFF
--- a/src/WinRT.Runtime/ComWrappersSupport.net5.cs
+++ b/src/WinRT.Runtime/ComWrappersSupport.net5.cs
@@ -117,7 +117,7 @@ namespace WinRT
             var rcw = ComWrappers.GetOrRegisterObjectForComInstance(thisPtr, CreateObjectFlags.TrackerObject, obj);
 
             // Resurrect IWinRTObject's disposed IObjectReferences, if necessary
-            var target = rcw is System.Delegate del ? del.Target : rcw;
+            var target = rcw is Delegate del ? del.Target : rcw;
             if (target is IWinRTObject winrtObj)
             {
                 winrtObj.Resurrect();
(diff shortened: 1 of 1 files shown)


j0shuams (approved):

Scottj1s (approved):
thanks!

Issue #623, opened 2020-12-08
WinRT bug: System.InvalidOperationException:“Attempt to update previously set global instance.”


taooceros:
### Description

When calling `new PackageManager()`, this error throw. Only the first time calling the constructor throws the error. 

### Configuration
Using .Net 5 with TargetFramework net5.0-windows10.0.19041.0.
The windows OS version is 20H2 x64.

### Regression?

Not sure, but .net core 3.1 with the Windows.SDK.Contract works well.

#### StackTrace

```
   at System.Runtime.InteropServices.ComWrappers.RegisterForTrackerSupport(ComWrappers instance)
   at WinRT.ComWrappersSupport.get_ComWrappers()
   at WinRT.ComWrappersSupport.TryRegisterObjectForInterface(Object obj, IntPtr thisPtr)
   at WinRT.ComWrappersSupport.RegisterObjectForInterface(Object obj, IntPtr thisPtr)
   at Windows.Management.Deployment.PackageManager..ctor()
   at Flow.Launcher.Plugin.Program.Programs.UWP.CurrentUserPackages() in D:\Document\Dotnet\Flow.Launcher\Plugins\Flow.Launcher.Plugin.Program\Programs\UWP.cs:line 200
   at Flow.Launcher.Plugin.Program.Programs.UWP.All() in D:\Document\Dotnet\Flow.Launcher\Plugins\Flow.Launcher.Plugin.Program\Programs\UWP.cs:line 156
```

#### Similar bug
https://github.com/microsoft/CsWinRT/issues/394

angelazhangmsft:
@taooceros What version of .NET5 SDK are you using? This seems to be fixed in the latest (.NET 5.0.100)

taooceros:
> @taooceros What version of .NET5 SDK are you using? This seems to be fixed in the latest (.NET 5.0.100)

Weird, I do use the latest version 5.0.100.

Scottj1s:
@taooceros, this can also occur if there are multiple versions of winrt.runtime.dll loaded in process, each with their own static instance of registration state.  Do you have a repro project?

taooceros:
@Scottj1s check out https://github.com/taooceros/Flow.Launcher
The usage of `new PackageManager` is on here.
https://github.com/taooceros/Flow.Launcher/blob/113def721c6c1de191257113005a4482eb76e8e8/Plugins/Flow.Launcher.Plugin.Program/Programs/UWP.cs#L201-L210
[...]

PR #675, opened 2021-01-19
Implement singleton pattern for ComWrappers

j0shuams:
Make ComWrappers robust/have an idempotent setter via singleton pattern. 

Addresses #623

### DIFF
--- a/src/WinRT.Runtime/ComWrappersSupport.net5.cs
+++ b/src/WinRT.Runtime/ComWrappersSupport.net5.cs
@@ -17,7 +17,7 @@ namespace WinRT
     {
         // Instance field and property for Singleton pattern: ComWrappers `set` method should be idempotent 
         private static DefaultComWrappers _instance;
-        internal static DefaultComWrappers Instance
+        private static DefaultComWrappers DefaultComWrappersInstance
         {
             get 
             {
@@ -44,7 +44,7 @@ namespace WinRT
                     {
                         if (_comWrappers is null)
                         {
-                            _comWrappers = Instance;
+                            _comWrappers = DefaultComWrappersInstance;
                             ComWrappers.RegisterForTrackerSupport(_comWrappers);
                         }
                     }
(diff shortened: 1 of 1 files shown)


Scottj1s (approved):
perfect!

src/WinRT.Runtime/ComWrappersSupport.net5.cs
```diff
@@ -165,6 +169,23 @@ public static void InitializeComWrappers(ComWrappers wrappers = null)
 
     public class DefaultComWrappers : ComWrappers
     {
+        // Instance field and property for Singleton pattern: ComWrappers setter method should be idempotent 
+        private static DefaultComWrappers _instance;
+        public static DefaultComWrappers Instance
```
manodasanW, line 174: Looks like a new public API.  Probably want to make it internal.

src/WinRT.Runtime/ComWrappersSupport.net5.cs
```diff
@@ -15,6 +15,20 @@ namespace WinRT
 {
     public static partial class ComWrappersSupport
     {
+        // Instance field and property for Singleton pattern: ComWrappers `set` method should be idempotent 
+        private static DefaultComWrappers _instance;
+        internal static DefaultComWrappers Instance
```
manodasanW, line 20: for clarity, consider naming it DefaultComWrappersInstance.

manodasanW, line 20: Given we moved it here, probably make it private too.

src/WinRT.Runtime/ComWrappersSupport.net5.cs
```diff
@@ -30,7 +44,7 @@ private static ComWrappers ComWrappers
                     {
                         if (_comWrappers is null)
                         {
-                            _comWrappers = new DefaultComWrappers();
+                            _comWrappers = Instance;
```
manodasanW, line 47: In both cases, I think you want to do the register first and then set the _comWrappers variable.  As I see for a potential race where if the get is called twice in somewhat parallel to each other, we can end up with one getting the comwrappers instance before it has been registered.

manodasanW (approved):

Issue #677, opened 2021-01-20
Write new docs for WRC diagnostics


j0shuams:
Since our diagnostics do what Winmdexp used to do, we should make a new version of this [webpage](https://docs.microsoft.com/en-us/previous-versions/hh977010%28v=vs.110%29)

j0shuams:
There is a link used in WinRTRules.cs that needs to be updated when we get our new webpage

j0shuams:
I'll be updating this [branch](https://github.com/MicrosoftDocs/windows-uwp-pr/tree/jlarkin/update-diagnostic-docs) and will cut a pr into @angelazhangmsft 's branch when I've made all my changes

angelazhangmsft:
Authoring diagnostics docs have been published here - https://docs.microsoft.com/en-us/windows/uwp/csharp-winrt/authoring-diagnostics

PR #685, opened 2021-01-22
Update Diagnostic rules

### LINKED ISSUE #676: Rename "WME" in Diagnostic rules
Our current diagnostic messages all start with "WMExxxx", and changing "WME" to "CsWinRT" would make error messages clearer. The messages we report to users are DiagnosticDescriptors that get attached to Diagnostics and raised by the scanner. In WinRTRules.cs, of SourceGenerator, we need to rename the "WME" prefix of the diagnostic identifier.

j0shuams:
This PR introduces two new files -- AnalzyerReleases.Unshipped.md and AnalyzerReleases.Shipped.md 
These are made from our diagnostic descriptors, and serve as a documentation page for our rules. 
I am also changing the "WME" identifier of the rules to be "CsWinRT", which required a change in our diagnostic unit testing app. 

I have updated our `.gitignore` as well, as there is a new directory that we don't want to track for our winui packaging sample. 

closes #676 , addresses part of #677

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ nuget/Microsoft.Windows.CsWinRT.Prerelease.targets
 *.binlog
 vs_buildtools.exe
 .buildtools
+**(Package)*
--- a/.gitignore
+++ b/.gitignore
@@ -21,4 +21,3 @@ nuget/Microsoft.Windows.CsWinRT.Prerelease.targets
 *.binlog
 vs_buildtools.exe
 .buildtools
-**(Package)*
--- a/src/Authoring/WinRT.SourceGenerator/WinRTRules.cs
+++ b/src/Authoring/WinRT.SourceGenerator/WinRTRules.cs
@@ -18,7 +18,7 @@ namespace WinRT.SourceGenerator
                 category: "Usage",
                 defaultSeverity: DiagnosticSeverity.Error,
                 isEnabledByDefault: true,
-                helpLinkUri: "https://docs.microsoft.com/en-us/previous-versions/hh977010(v=vs.110)");
+                helpLinkUri: "https://github.com/microsoft/CsWinRT/tree/master/src/Authoring/WinRT.SourceGenerator/AnalyzerReleases.Unshipped.md");
         }
 
         public static DiagnosticDescriptor PrivateGetterRule = MakeRule(
--- a/src/Authoring/WinRT.SourceGenerator/WinRTRules.cs
+++ b/src/Authoring/WinRT.SourceGenerator/WinRTRules.cs
@@ -56,9 +56,9 @@ namespace WinRT.SourceGenerator
         public static DiagnosticDescriptor UnsupportedTypeRule = MakeRule(
             "CsWinRT1006",
             "Exposing unsupported type",
-            "The member '{0}' has the type '{1}' in its signature. The type '{1}' is not a valid Windows Runtime type\n"  
-            + "Yet, the type (or its generic parameters) implement interfaces that are valid Windows Runtime types\n" 
-            + "Consider changing the type '{1} in the member signature to one of the following types from System.Collections.Generic:\n{2}");
+            "The member '{0}' has the type '{1}' in its signature. The type '{1}' is not a valid Windows Runtime type. "  
+            + "Yet, the type (or its generic parameters) implement interfaces that are valid Windows Runtime types. " 
+            + "Consider changing the type '{1} in the member signature to one of the following types from System.Collections.Generic: {2}.");
         
         public static DiagnosticDescriptor StructWithNoFieldsRule = MakeRule(
             "CsWinRT1007",
--- a/src/Authoring/WinRT.SourceGenerator/WinRTRules.cs
+++ b/src/Authoring/WinRT.SourceGenerator/WinRTRules.cs
@@ -21,130 +21,122 @@ namespace WinRT.SourceGenerator
                 helpLinkUri: "https://docs.microsoft.com/en-us/previous-versions/hh977010(v=vs.110)");
         }
 
-        public static DiagnosticDescriptor ImplementerUseSameVariableName = MakeRule(
-            "CsWinRT1113", 
-            "Class implementaiton method should use same paramter as interface method",
-            "Class '{0}' implements method '{1}' with parameter '{2}', but '{1}' was declared in interface '{3}' with parameter '{4}'." 
-            + "Parameter names must match exactly in the Windows Runtime.");
-        
         public static DiagnosticDescriptor PrivateGetterRule = MakeRule(
-            "CsWinRT", 
+            "CsWinRT1000", 
             "Property must have public getter",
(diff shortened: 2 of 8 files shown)


src/Authoring/WinRT.SourceGenerator/WinRTRules.cs
```diff
@@ -21,121 +21,130 @@ private static DiagnosticDescriptor MakeRule(string id, string title, string mes
                 helpLinkUri: "https://docs.microsoft.com/en-us/previous-versions/hh977010(v=vs.110)");
         }
 
+        public static DiagnosticDescriptor ImplementerUseSameVariableName = MakeRule(
+            "CsWinRT1113", 
+            "Class implementaiton method should use same paramter as interface method",
```
manodasanW, line 26: typo - implementation.

manodasanW, line 26: parameter name

src/Authoring/WinRT.SourceGenerator/WinRTRules.cs
```diff
@@ -21,121 +21,130 @@ private static DiagnosticDescriptor MakeRule(string id, string title, string mes
...
+            "Class '{0}' implements method '{1}' with parameter '{2}', but '{1}' was declared in interface '{3}' with parameter '{4}'." 
+            + "Parameter names must match exactly in the Windows Runtime.");
+        
         public static DiagnosticDescriptor PrivateGetterRule = MakeRule(
-            "WME", 
+            "CsWinRT",
```
manodasanW, line 31: Is there no # for this.

manodasanW: I see a couple actually without a #.  Do we want them to all have # or no?

src/Authoring/WinRT.SourceGenerator/WinRTRules.cs
```diff
@@ -21,121 +21,130 @@ private static DiagnosticDescriptor MakeRule(string id, string title, string mes
...
             "Class (or interface) is generic",
             "Type {0} is generic. Windows Runtime types cannot be generic.");
 
 
         public static DiagnosticDescriptor UnsealedClassRule = MakeRule(
-            "WME",
+            "CsWinRT",
             "Class is unsealed",
-            "Exporting unsealed types is not supported. Please mark type {0} as sealed.");
+            "Exporting unsealed types is not supported in Windows Runtime. Please mark type {0} as sealed.");
```
manodasanW, line 61: I would remove the in Windows Runtime as technically it is now supported in WinRT, but we still have that limitation.

src/Authoring/WinRT.SourceGenerator/WinRTRules.cs
```diff
@@ -21,121 +21,130 @@ private static DiagnosticDescriptor MakeRule(string id, string title, string mes
...
+            "Classes cannot have multiple constructors of the same arity in the Windows Runtime, class {0} has multiple {1}-arity constructors");
 
         public static DiagnosticDescriptor ParameterNamedValueRule = MakeRule(
-            "WME1092",
+            "CsWinRT1092",
             "Parameter Named Value Rule",
-            "The method {0} has a parameter named {1} which is the same as the default return value name. "
-            + "Consider using another name for the parameter or use the System.Runtime.InteropServices.WindowsRuntime.ReturnValueNameAttribute "
-            + "to explicitly specify the name of the return value.");
+            "The method {0} is used in the Windows Runtime and has a parameter named {1}." +
```
manodasanW, line 88: I am not sure whether in the Windows Runtime makes sense here.

src/Authoring/WinRT.SourceGenerator/WinRTRules.cs
j0shuams: Yeah when we make the docs page I'll have a table of all of them. Was going to assign the number then, can do it now though. I can come up with numbers that aren't the ones we have already -- there's even a couple like 1060(a) that might just be clearer under a new number.

src/Authoring/WinRT.SourceGenerator/WinRTRules.cs
j0shuams: Do we want to remove the diagnostic for this then?

manodasanW: CsWinRT still has the limitation that requires it to be sealed.

j0shuams: Per discussion, will change to CsWinRT.

src/Authoring/WinRT.SourceGenerator/WinRTRules.cs
```diff
@@ -18,124 +18,125 @@ private static DiagnosticDescriptor MakeRule(string id, string title, string mes
                 category: "Usage",
                 defaultSeverity: DiagnosticSeverity.Error,
                 isEnabledByDefault: true,
-                helpLinkUri: "https://docs.microsoft.com/en-us/previous-versions/hh977010(v=vs.110)");
+                helpLinkUri: "https://github.com/microsoft/CsWinRT/tree/master/src/Authoring/WinRT.SourceGenerator/AnalyzerReleases.Unshipped.md");
```
j0shuams, line 21: We may want this to be AnalyzerReleases.Shipped.md, but I am not sure when we consider a version of our SourceGenerator analyzer "released"

src/Authoring/WinRT.SourceGenerator/WinRTRules.cs
```diff
@@ -18,124 +18,125 @@ private static DiagnosticDescriptor MakeRule(string id, string title, string mes
...
+            "Classes cannot have multiple constructors of the same arity in the Windows Runtime, class {0} has multiple {1}-arity constructors");
 
         public static DiagnosticDescriptor ParameterNamedValueRule = MakeRule(
-            "WME1092",
+            "CsWinRT1010",
             "Parameter Named Value Rule",
-            "The method {0} has a parameter named {1} which is the same as the default return value name. "
-            + "Consider using another name for the parameter or use the System.Runtime.InteropServices.WindowsRuntime.ReturnValueNameAttribute "
-            + "to explicitly specify the name of the return value.");
+            "The method {0} is used in the Windows Runtime and has a parameter named {1}." +
```
manodasanW, line 81: Can we just merge both like The parameter name {1} in method {0} is the same as the return value used ...

manodasanW (approved):

.gitignore
```diff
@@ -20,4 +20,5 @@ global.json
 nuget/Microsoft.Windows.CsWinRT.Prerelease.targets
 *.binlog
 vs_buildtools.exe
-.buildtools
\ No newline at end of file
+.buildtools
+**(Package)*
```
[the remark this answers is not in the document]
j0shuams: Hm okay I'll remove this; I put it in because of the pain of restoring many files when you accidentally add it to your commit.

PR #688, opened 2021-01-25
Update nuget targets

j0shuams:
This PR makes two updates to our Authoring/Hosting support targets. 
1) We now automatically make the author's targets file -- if we get complaints about this in the future, then we can make a property that users can set to opt-out of this behavior. 

2) We do not use _PackageFiles anymore and authors do not have to create a targets file nor update their project file to include it. The latter part of this comes from (1), and the former from switching to "Content" type instead of "_PackageFiles". 

I tested this in my app, and it worked fine.

### DIFF
--- a/nuget/Microsoft.Windows.CsWinRT.targets
+++ b/nuget/Microsoft.Windows.CsWinRT.targets
@@ -163,7 +163,7 @@ $(CsWinRTFilters)
       <!-- Custom targets that copy dlls for consumers of the authored component.
            `buildTransitive\` is for PackageReference, `build\` is for packages.config -->
       <_PackageFiles Include="$(CsWinRTPath)buildTransitive\Microsoft.Windows.CsWinRT.Authoring.targets"
-                     PackagePath="buildTransitive\$(AssemblyName).CsWinRT.targets;build\$(AssemblyName).CsWinRT.targets"/>
+                     PackagePath="buildTransitive\$(AssemblyName).targets;build\$(AssemblyName).targets"/>
 
       <!-- WinRT.Host.dll is architecture specific -->
       <!-- x64 --> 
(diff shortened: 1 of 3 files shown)


nuget/Microsoft.Windows.CsWinRT.targets
```diff
@@ -140,50 +140,58 @@ $(CsWinRTFilters)
...
       
       <!-- The authored component gets the WinRT.Runtime and NET.SDK dlls from ReferenceCopyLocalPaths 
-            (because CopyLocalLockFileAssemblies == true); $(TargetPath) has the assembly (e.g. PackageId.dll) -->
-      <_PackageFiles Include="$(CsWinRTPath)lib\net5.0\WinRT.Host.Shim.dll;@(ReferenceCopyLocalPaths);$(TargetPath)"
-                     PackagePath="lib\native"/>
+            (because CopyLocalLockFileAssemblies == true) and $(TargetPath) has the assembly (e.g. PackageId.dll) -->
+      <Content Include="$(CsWinRTPath)lib\net5.0\WinRT.Host.Shim.dll;@(ReferenceCopyLocalPaths);$(TargetPath)"
+               Pack="true"
+               PackageCopyToOutput="true"
+               PackagePath="native" />
```
manodasanW, line 160: Are we intentionally moving these out of the lib folder?

j0shuams: I figured it would be nicer to have a shorter path, and don't think the lib folder is that important for our purposes. Do you?

j0shuams: I restored the lib folder.

manodasanW (approved):

PR #704, opened 2021-01-28
Add support for explicitly implemented interfaces

### LINKED ISSUE #684: Authoring support for explicitly implemented interfaces.

manodasanW:
- Fix issue with implementing events from interfaces
- Add additional logging to assist with diagnosing issues
- Add support for explicitly implemented interfaces
- When a member from an interface is explicitly implemented, it is exposed as a private method in the winmd
- When the CsWinRT projection for it is generated for such members, only the explicit interface implementation is generated and not the public class member which the explicit interface implementation typically points to.

Fixes #684


Scottj1s (approved):
Nice!

PR #706, opened 2021-01-28
Remove unsupported solution configs, make get_testwinrt curdir-independent, assert TestWinRT requirement

### LINKED ISSUE #692: cswinrt.sln doesn't build
Is the SLN file up to date? Looks like there are several problems, e.g. it references a BenchmarkComponent.vcxproj that doesn't exist

The WinUIDesktopSample has build errors too:
```
---------------------------
Microsoft Visual Studio
---------------------------
The debug executable 'C:\Users\asklar\source\repos\cswinrt\src\Samples\WinUIDesktopSample\bin\x86\Debug\net5.0-windows10.0.19041.0\WinUIDesktopSample.exe' specified in the 'WinUIDesktopSample' debug profile does not exist.
---------------------------
OK   
---------------------------
```

Could someone add some details in the CONTRIBUTING.md file as to how one can build this project including tests and samples? Is this supposed to be built for `any cpu` or `x86`, etc.?

Scottj1s:
fixes #692


j0shuams:
I see we are removing Arm and Arm64 -- when I was doing nuget work I know I made sure we included these host dlls. But only if they exist, which made me think eventually we will make these to support arm architectures. I see we are removing those platforms, is the plan to add support for arm{64} in the future or are we sticking with x86/x64?

Scottj1s:
> I see we are removing Arm and Arm64 -- when I was doing nuget work I know I made sure we included these host dlls. But only if they exist, which made me think eventually we will make these to support arm architectures. I see we are removing those platforms, is the plan to add support for arm{64} in the future or are we sticking with x86/x64?

No plans to ever build a cswinrt.exe (or other native components) for ARM32/64

j0shuams:
> > I see we are removing Arm and Arm64 -- when I was doing nuget work I know I made sure we included these host dlls. But only if they exist, which made me think eventually we will make these to support arm architectures. I see we are removing those platforms, is the plan to add support for arm{64} in the future or are we sticking with x86/x64?
> 
> No plans to ever build a cswinrt.exe (or other native components) for ARM32/64. Removing those configs, and Any CPU, prevents confusion for devs when they open the sln

Does this mean we should update our nuget pipeline / build cmd to not consider winrt.host dlls for arm architectures?

j0shuams (approved):

manodasanW:
Opened #708 to track ARM64 support for WinRT.Host.

manodasanW (approved):

PR #709, opened 2021-01-28
Fix namespace diagnostic predicate

### LINKED ISSUE #705: WME1044 for valid assembly name
**Describe the bug**
For assembly names that are specified with a period -- e.g. `MyComponent.Foo` the winmd is named `MyComponent.winmd` instead of `MyComponent.Foo.winmd`. 

**To Reproduce**

1. Change the AuthoringDemo to have assembly and default namespace "AuthoringDemo.Foo"
2. Build
3. inspect winmd filename 

**Expected behavior**

The winmd should be named "AuthoringDemo.Foo.winmd"

**Initial Comments**
This is either a bug in GetAssemblyName in WinRT.SourceGenerator\Generators.cs or a bug in how we verify assembly names with namespaces in WinRT.SourceGenerator\DiagnosticHelpers.cs

j0shuams:
Turns out the `Name` property for a namespace symbol like `A.Foo` returns `Foo` and not `A.Foo` 

This caused a bug where we threw an invalid namespace diagnostic when we shouldn't have. I did have a unit test for this (in theory) but it wasn't failing for some reason. Tested by changing our `AuthoringTest` project to have a namespace/assembly name like `AuthoringTest.Foo`. 

Closes #705


src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
```diff
@@ -339,8 +339,8 @@ private bool IsInvalidNamespace(INamespaceSymbol @namespace, string assemblyName
             {
                 topLevel = topLevel.ContainingNamespace;
```
manodasanW, line 340: Shouldn't we also be checking if AssemblyName != partial top level rather than the most top level only?

manodasanW (approved):

Issue #663, opened 2021-01-11
Authoring/Hosting project tooling support 


angelazhangmsft:
Creating an issue to track the status of end-to-end C# Authoring/Hosting. Here is a list of hacks that we should remove to consider authoring/hosting experience complete:
- [x] Instead of winmd reference from hosting project, allow for project reference to the component - Joshua
- [x] Remove need for `DeploymentContent` set to true in hosting project file (for the manifest and runtimeconfig) - Joshua
- [ ] Automate activatable classes instead of manually creating manifest file - Mano (#789)
- [x] Remove need for creating WinRT.Host.runtimeconfig.json (create a default, and allow override) - Joshua (#735)
- [x] `CsWinRTEnableLogging` in authoring csproj (#687) - Mano
- [x] `CsWinRTWindowsMetadata` (#657) 
- [x] `GeneratedFilesDir` in authoring csproj bug with potential race condition - Mano
- [x] `AssemblyVersion` - can we set a default
- [x] Copy hosting dlls so you don't need Directory.build.targets files (#474)

For packaged apps, there are some additional hacks:
- [x] Manually copying CsWinRT and component assemblies to the Appx folder on deploy (#734)
- [x] Manually editing the appxmanifest (may need VS help), for example:
```xml
<Extensions>
    <Extension Category="windows.activatableClass.inProcessServer">
      <InProcessServer>
        <Path>WinRT.Host.dll</Path>
          <ActivatableClass ActivatableClassId="Net5BgTaskComponent.BgTask" ThreadingModel="both" />
      </InProcessServer>
    </Extension>
  </Extensions>
```

Other existing issues for E2E authoring support include:
- [x] #633 
- [x] #656 
- [ ] #649 
- [x] #662

j0shuams:
Hosting support needs some more work with ProjectReferences, but PackageRef's are handled by my PR #665

j0shuams:
Meeting notes:

Top priority is project reference support, as a few of the later hacks we want to see if they are needed once projref support is in place. 

Second priority: manifest/runtimeconfig

- Mano will work on embedding manifest in authored component, with the hope that consumers wont need to make a manifest file at all.

- Joshua will work on making a generic runtimeconfig, and shipping it with DeploymentContent metadata -- consumers wont need to author one unless they want to make a custom one (by referring to the hosting docs)

angelazhangmsft:
Closing and tracking the remaining items separately

PR #710, opened 2021-01-29
Reduce properties required to be set for authoring

manodasanW:
- Moved to using the CsWinRTGeneratedFilesDir property rather than GeneratedFilesDir
- Ensure GetAssemblyVersion target ran first before source generator to ensure default assembly version was set

Contributes to #663


src/Tests/AuthoringTest/AuthoringTest.csproj
```diff
@@ -3,13 +3,9 @@
   <PropertyGroup>
     <TargetFramework>net5.0</TargetFramework>
     <Platforms>x64;x86</Platforms>
-    <LangVersion>preview</LangVersion>
```
j0shuams, line 6: What was this used for?

j0shuams (approved):

Issue #697, opened 2021-01-27
Payload contains two or more files with the same destination path


asklar:
Related to #693 

- Created a C# NET 5 class library. Reference the WinUI 3 Preview 4 CI 1/26 nuget package.
- Follow the authoring instructions to produce a nuget pkg for the class library
- Created a cppwinrt WinUI 3 app from WinUI 3 Preview 3 VSIX
- update cppwinrt app's WinUI reference to latest internal WinUI nuget preview 4 from 01/26
- Referenced the NET 5 class library nupkg file from the cppwinrt app via a local nuget source
- Build solution

after applying workaround to copy png file, I get:

> Severity	Code	Description	Project	File	Line	Suppression State
Error	APPX1111	Payload contains two or more files with the same destination path 'CoreMessagingXP.dll', but they are different sizes. Source files: 
C:\Users\asklar\source\repos\NET5Lib\packages\Microsoft.WinUI.3.0.0-preview4.210126.6-CI\build\..\runtimes\win10-x64\native\CoreMessagingXP.dll
C:\Users\asklar\source\repos\NET5Lib\packages\NET5Lib.1.0.0\build\..\lib\native\CoreMessagingXP.dll	CppAppDesktop	C:\Users\asklar\source\repos\NET5Lib\packages\Microsoft.WinUI.3.0.0-preview4.210126.6-CI\build\Microsoft.ApplicationModel.Resources.PriGen.targets	2329	

Indeed those files have different sizes; it looks like the C# library is pulling in the ARM binaries from WinUI. Copying the x64 binaries on top of the expanded class library nuget package (i.e. in the apps `packages` folder) allows the exe to build.

Issue #696, opened 2021-01-27
WinMD's implementation should be set as WinRT.Host


manodasanW:
The automatic support should set the implementation as WinRT.Host to allow the correct manifest entries be generated by tooling.

https://github.com/microsoft/CsWinRT/blob/e33691ea5f4bac9baaaaa3be969ebab4772cc90d/nuget/Microsoft.Windows.CsWinRT.Authoring.targets#L22

https://github.com/microsoft/CsWinRT/blob/e33691ea5f4bac9baaaaa3be969ebab4772cc90d/src/Tests/AuthoringWinUITest/AuthoringWinUITest/AuthoringWinUITest.vcxproj#L145

asklar:
confirmed that manually patching the cswinrt target with that property and rebuilding the package makes it possible to load the managed dll and instantiate the managed object

PR #712, opened 2021-01-29
Fix authoring bugs

### LINKED ISSUE #693: Can't find NoiseAsset_256X256_PNG.png
1. Created a C# NET 5 class library. Reference the WinUI 3 Preview 4 CI 1/26 nuget package. 
2. Follow the authoring instructions to produce a nuget pkg for the class library
3. Created a cppwinrt WinUI 3 app from WinUI 3 Preview 3 VSIX 
4. update cppwinrt app's WinUI reference to latest internal WinUI nuget preview 4 from 01/26
5. Referenced the NET 5 class library nupkg file from the cppwinrt app via a local nuget source
6. Build solution

Workaround for now: create an empty png file at that location

j0shuams:
Previously, our targets that automate dll reference updating were too assumptive in what files would be in the authored nupkg. This PR introduces a filter on the dlls we package with an authored winmd/dll, as well as make some structural changes to remove warnings about the nupkg. Some bugs arose because of the assumption that only CsWinRT related dlls would be on our 
reference path when building, when in fact WinUI dlls/assets were being added just by nature of a package reference. 

I also found some things to cleanup in WinRTRules

Closes #693 #696 #697

### DIFF
--- a/src/Authoring/WinRT.SourceGenerator/WinRTRules.cs
+++ b/src/Authoring/WinRT.SourceGenerator/WinRTRules.cs
@@ -36,7 +36,7 @@ namespace WinRT.SourceGenerator
         public static DiagnosticDescriptor NamespacesDifferByCase = MakeRule(
             "CsWinRT1002",
             "Namespace names cannot differ only by case",
-            "Multiple namespaces found with the name '{0}'; namespace names cannot differ only by case in the Windows Runtime.");
+            "Multiple namespaces found with the name '{0}'; namespace names cannot differ only by case in the Windows Runtime");
 
         public static DiagnosticDescriptor NoPublicTypesRule = MakeRule(
             "CsWinRT1003",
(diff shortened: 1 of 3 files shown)


src/Authoring/WinRT.SourceGenerator/WinRTRules.cs
```diff
@@ -124,12 +124,6 @@ private static DiagnosticDescriptor MakeRule(string id, string title, string mes
             "Array signature found with multi-dimensional array, which is not a valid Windows Runtime type",
             "Method '{0}' has a multi-dimensional array of type '{1}' in its signature. Arrays in Windows Runtime method signatures must be one dimensional.");
```
manodasanW, line 126: Is this not needed?

j0shuams: No I used this at one point, but have switched to a more general "not valid type" rule, and this one has been unused for a while.

j0shuams:
Wwhen testing with C# apps via package reference, there are some issues with the target framework of the consumer. Although the changes in the PR are aimed at support for native consumers, I won't push for this to merge until told otherwise and/or the issues with c# consumers are resolved.

manodasanW (approved):

PR #720, opened 2021-02-04
Authoring bug fixes

### LINKED ISSUE #714: Static properties in authored class are not properly generated
Properties are not being written in the activation factory.

manodasanW:
- Fixing issue where properties and events weren't emitted for static classes
- Adding support for explicitly implemented interfaces which are represented as private functions on the class, but accessible through the interface
- Fixing issue with method impls for generic interface members where they could refer to the wrong override if there is multiple classes implementing the interface with different generics due to not keeping track of the generics.
- Fixing issues with interface inheritance

Fixes #714

### DIFF
--- a/src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
+++ b/src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
@@ -935,6 +935,8 @@ namespace Generator
         private void ProcessCustomMappedInterfaces(INamedTypeSymbol classSymbol)
         {
             Logger.Log("writing custom mapped interfaces for " + QualifiedName(classSymbol));
+            Dictionary<INamedTypeSymbol, bool> IsPublicImplementation = new Dictionary<INamedTypeSymbol, bool>();
+
             // Mark custom mapped interface members for removal later.
             // Note we want to also mark members from interfaces without mappings.
             foreach (var implementedInterface in GetInterfaces(classSymbol, true).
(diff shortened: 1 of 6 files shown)


src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
```diff
@@ -935,25 +935,29 @@ internal static (string, string, string, bool, bool) GetSystemTypeCustomMapping(
...
             {
-                string interfaceName = QualifiedName(implementedInterface);
-                Logger.Log("custom mapped interface: " + interfaceName);
+                bool isPubliclyImplemented = false;
+                Logger.Log("custom mapped interface: " + QualifiedName(implementedInterface, true));
                 foreach (var interfaceMember in implementedInterface.GetMembers())
                 {
                     var classMember = classSymbol.FindImplementationForInterfaceMember(interfaceMember);
                     currentTypeDeclaration.CustomMappedSymbols.Add(classMember);
+                    isPubliclyImplemented |= (classMember.DeclaredAccessibility == Accessibility.Public);
```
j0shuams, line 952: if the first member is public, isPubliclyImplemented := true
if the second member is not public, isPubliclyImplemented is still true... Seems like we would want it to be false? so &=?

manodasanW: Thats a good question and I probably should add a comment in the code there.  The issue with custom mapped interfaces is we do not have a 1 to 1 mapping between the members of the mapped from interface and the mapped to interface.  This means we can't say since these 2 members were explicitly implemented while the others weren't, we will find those 2 members in the mapped interface and mark them private.  We need to actually decide whether the entire interface is private (explicitly implemented) or not.  So the decision chosen here based on what I see .NET native doing is if at least one member is not explicitly implemented,  the entire mapped interface is treated as a public (not explicitly implemented).  Only if all the members were explicitly implemented is when we treat the mapped interface as also explicitly implemented.

Scottj1s: could break the loop as soon as isPubliclyImplemented is true

---
In reply to: [571209062](https://github.com/microsoft/CsWinRT/pull/720#discussion_r571209062) [](ancestors = 571209062)

Scottj1s (approved):
:shipit:

src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
manodasanW: I won't be able to break out of the loop as I am using the loop to collect other information too (CustomMappedSymbols).  I can initialize the dictionary entry first outside of the loop to false and then do the |= in the loop and remove isPubliclyImplemented, but its almost equivalent to what is there with additional dictionary lookup.

PR #722, opened 2021-02-05
Update usage and readme

angelazhangmsft:
Making some updates to the readme and usage.md.

- Separate "using" and "building" cswinrt
- Update usage.md with 2 use cases (building projection and component)


docs/usage.md
```diff
@@ -1,12 +1,19 @@
 # Usage
 
-## Component Project
+The [C#/WinRT NuGet package](https://www.nuget.org/packages/Microsoft.Windows.CsWinRT/) provides tooling for the following scenarios:
 
-A component project adds a NuGet reference to C#/WinRT to invoke cswinrt.exe at build time, generate projection sources, and compile these into an interop assembly. For an example of this, see the [Projection Sample](https://github.com/microsoft/CsWinRT/tree/master/src/Samples/Net5ProjectionSample). Command line options can be displayed by running **cswinrt -?**.  The interop assembly is then typically distributed as a NuGet package itself. 
+- [Generate and distribute an interop assembly](#generate-and-distribute-an-interop-assembly)
+- [Author and consume a C#/WinRT component](#author-and-consume-a-cwinrt-component)
```
j0shuams, line 6: Does this link work as expected for you? For me it goes to a page that is failing to render the mark down properly

docs/usage.md
```diff
@@ -1,12 +1,19 @@
...
 
-To invoke cswinrt to generate projection sources for types in the **Contoso** namespace, you need to add the following property to your C# library project file. In this project you would also need to reference the CsWinRT NuGet package and the project-specific `winmd` files you want to project, whether through a NuGet package, project reference or direct reference. By default, the **Windows** and **Microsoft** namespaces are not projected.
+## Generate and distribute an interop assembly
+
+A projection project adds a NuGet reference to C#/WinRT to invoke cswinrt.exe at build time, generate projection sources, and compile these into an interop assembly. For an example of this, see the [Projection Sample](https://github.com/microsoft/CsWinRT/tree/master/src/Samples/Net5ProjectionSample) and associated [walkthrough](https://docs.microsoft.com/windows/uwp/csharp-winrt/net-projection-from-cppwinrt-component). Command line options can be displayed by running **cswinrt -?**.  The interop assembly is then typically distributed as a NuGet package itself for application projects to use.
```
j0shuams, line 12: Nit: add backticks so it gets formatted like `cswinrt.exe`

docs/usage.md
```diff
@@ -1,12 +1,19 @@
...
+
+### Projection Project
```
j0shuams, line 14: -> Consuming Projections?

docs/usage.md
```diff
@@ -1,12 +1,19 @@
...
+## Generate and distribute an interop assembly
+
+A projection project adds a NuGet reference to C#/WinRT to invoke cswinrt.exe at build time, generate projection sources, and compile these into an interop assembly. For an example of this, see the [Projection Sample](https://github.com/microsoft/CsWinRT/tree/master/src/Samples/Net5ProjectionSample) and associated [walkthrough](https://docs.microsoft.com/windows/uwp/csharp-winrt/net-projection-from-cppwinrt-component). Command line options can be displayed by running **cswinrt -?**.  The interop assembly is then typically distributed as a NuGet package itself for application projects to use.
+
+### Projection Project 
+
+To invoke cswinrt to generate projection sources for types in the **Contoso** namespace, you need to add the following property to your C# library project file. In this project you would also need to reference the CsWinRT NuGet package and the project-specific `.winmd` files you want to project, whether through a NuGet package, project reference or direct reference. By default, the **Windows** and **Microsoft** namespaces are not projected.
```
Scottj1s, line 16: we should use C#/WinRT consistently - only CsWinRT if it's actually part of a name (like Microsoft.Windows.CsWinRT.nupkg, cswinrt.exe, etc)

docs/usage.md
Scottj1s: yes, the anchor tag includes "cwinrt" - the "#/" probably needs to be escaped

Scottj1s (changes requested):
Thanks - love this attention to detail

docs/usage.md
angelazhangmsft: Isn't this section about projects that generate a projection?

docs/usage.md
angelazhangmsft: the link works for me

docs/usage.md
j0shuams: Yeah, how about "Generating Projections" ?

docs/usage.md
```diff
@@ -1,21 +1,32 @@
...
   <CsWinRTIncludes>Contoso</CsWinRTIncludes>
 </PropertyGroup>
 ```
 
-To further customize C#/WinRT behavior, refer to the [CsWinRT NuGet documentation](../nuget/readme.md).
+To further customize C#/WinRT behavior, refer to the [C#/WinRT NuGet documentation](../nuget/readme.md).
+
+### Application Project
+
+An interop assembly is typically distributed as a NuGet package for application projects to reference. This package will require a dependency on C#/WinRT to include `WinRT.Runtime.dll` for .NET 5 targets. If the interop assembly is not distributed as a NuGet package, an application project adds references to both the component interop assembly produced above and to C#/WinRT to include the `WinRT.Runtime.dll` assembly. If a third party WinRT component is distributed without an official interop assembly, an application project may add a reference to C#/WinRT to generate its own private component interop assembly.  There are versioning concerns related to this scenario, so the preferred solution is for the third party to publish an interop assembly directly.
```
manodasanW, line 28: consider renaming interop assembly to projection interop assembly.

docs/usage.md
```diff
@@ -1,21 +1,32 @@
...
 
-To invoke cswinrt to generate projection sources for types in the **Contoso** namespace, you need to add the following property to your C# library project file. In this project you would also need to reference the CsWinRT NuGet package and the project-specific `winmd` files you want to project, whether through a NuGet package, project reference or direct reference. By default, the **Windows** and **Microsoft** namespaces are not projected.
+## Generate and distribute an interop assembly
+
+A projection project adds a NuGet reference to C#/WinRT to invoke cswinrt.exe at build time, generate projection sources, and compile these into an interop assembly. For an example of this, see the [Projection Sample](https://github.com/microsoft/CsWinRT/tree/master/src/Samples/Net5ProjectionSample) and associated [walkthrough](https://docs.microsoft.com/windows/uwp/csharp-winrt/net-projection-from-cppwinrt-component). Command line options can be displayed by running `cswinrt -?`.  The interop assembly is then typically distributed as a NuGet package itself for application projects to use.
+
+### Generate Projections
```
manodasanW, line 14: If I recall correctly, CsWinRTIncludes is needed for a projection to be generated right?  Due to that, should we merge both sections and without CsWinRTIncludes, you won't build an interop assembly.

angelazhangmsft: Not sure I understand, should we not show the `CswinRTIncludes` property and just let folks refer to the projection sample for if they want to generate their own projection?

manodasanW: No, we should include it.  I was just saying the flow seemed a bit confusing where we were talking about generating interop assemblies and then generating projections when both are connected to each other.  I feel like both together should be a section where we talk about we generate projection interop assemblies based on the types listed as to be projected in CsWinRTIncludes.

angelazhangmsft: do these changes look ok

manodasanW: Looks good.

manodasanW (approved):

PR #728, opened 2021-02-08
Project reference support (preview)

### LINKED ISSUE #707: ProjectReference Hosting/Authoring Support
**Describe the bug**
Lots of hacks are needed to get authored components working when they are added as a project reference.
Let's update our targets to handle these automatically.

**To Reproduce**

1. Make a C# RTC, make a c++ console app in the same solution
2. add a project reference in the app to the RTC
3. build

**Expected behavior**
The RTC's winmd gets added as a winmd input to cppwinrt, and necessary hosting dll's get put in the outdir.
Developer should only have to build and run their solution, no copying/adding explicit winmd reference.

j0shuams:
This PR moves the existing authoring targets into their own file, and the transitive targets to a new authoring.transitive.targets file, and adds support for project reference.

For the new targets file, we update our nuspec to include the appropriate files. We can undo this if we prefer to just have one "CsWinRT" targets file, given that the new one (Authoring.targets) only has two entries. 

I am also updating the authoring docs to explain what users must do for project reference (add winmd reference, modify vcxproj). As a note about the vcxproj, because dotnet assumes the c++ project to be targeting net framework **if a target framework is not specified**. The only way I could find to work around this was to specify targetframework related properties on the vcxproj. 

Users do still have to add manually add a reference to the winmd, and I am hoping to find a better solution for this by the Release. 

closes #707 

Ideally, by release we have a way to let authored components add winmds to their targetpath. But dotnet 1130 blocks that and thus we require project reference consumers to add a second reference to the winmd generated.

### DIFF
--- a/docs/authoring.md
+++ b/docs/authoring.md
@@ -43,6 +43,7 @@ To make your component available as a NuGet package, it is important to include 
 When you pack your C#/WinRT component the DLLs/WinMD are automatically added to your nupkg, based on a nuspec generated from your project file. 
 
 **If you are going to write your own nuspec**, then you should make sure your nuspec contains the following ```file``` entries for your component ```MyAuthoredComponent``` (note: your TargetFramework may vary). This is so our targets that supply the DLLs for any consumers of your package work.  
+Similarly, any other dependencies, e.g. `Microsoft.WinUI`, will need to be included in your nuspec as well.
 
 ``` nuspec
 <files>
@@ -86,18 +87,15 @@ Examples of this are seen in the code snippet below. This is needed for this pre
 You can then use the Visual Studio UI to add a reference to the C#/WinRT component's `csproj` file, and you also need to add a reference to the WinMD file produced 
 for your component. The WinMD can be found in the output (`bin`) directory and the `Generated Files` directory of the authored component's project.
 
-You'll need to use [C++/WinRT](https://docs.microsoft.com/en-us/windows/uwp/cpp-and-winrt-apis/intro-to-using-cpp-with-winrt) to consume your API. So make sure you have C++/WinRT installed, and have added `#include <winrt/MyAuthoredComponent.h>` to the file `pch.h` of the native app.  
-
 You'll need to author some files to assist the hosting process by the native app: `YourNativeApp.exe.manifest` and `WinRT.Host.runtimeconfig.json`. 
-
 If your app is packaged with MSIX, then you don't need to include the manifest file, otherwise you need to include your activatable class registrations in the manifest file.
 
-To do this, **in Visual Studio**, right click on the project node on the "Solution Explorer" window, click "Add", then "New Item". Search for the "Text File" template and name your file "YourNativeApp.exe.manifest".
-Repeat this for the "WinRT.Host.runtimeconfig.json" file. 
+To do add these files, **in Visual Studio**, right click on the project node on the "Solution Explorer" window, click "Add", then "New Item". 
+Search for the "Text File" template and name your file `YourNativeApp.exe.manifest`.
+Repeat this for the `WinRT.Host.runtimeconfig.json` file. 
+For each item, right-click on it in the "Solution Explorer" window of Visual Studio; then select "Properties" and change the "Content" property to "Yes" using the drop-down arrow on the right -- this ensures it will be added to the output directory of your solution.
 
-This process adds the nodes `<Manifest Include=... >` and `<None Include=... >` to your native app's project file -- **you need to update these to have `<DeploymentContent>true</DeploymentContent>` for them to be placed in the output directory with your executable**.  
-
-You should read the [hosting docs](https://github.com/microsoft/CsWinRT/blob/master/docs/hosting.md) as well, for more information on these files.
+We have some [hosting docs](https://github.com/microsoft/CsWinRT/blob/master/docs/hosting.md) as well, that provide more information on these files.
 
 In summary, here is the fragment of additions made to the native app's project file:
 ``` vcxproj
@@ -108,18 +106,10 @@ In summary, here is the fragment of additions made to the native app's project f
   <TargetFramework>native</TargetFramework>
   <TargetRuntime>Native</TargetRuntime>
 </PropertyGroup>
-
-<ItemGroup>
-    <!-- the runtimeconfig.json -->
-    <None Include="WinRT.Host.runtimeconfig.json">
-      <DeploymentContent>true</DeploymentContent>
-    </None>
-    <!-- the manifest -->
-    <Manifest Include="YourNativeApp.exe.manifest">
-      <DeploymentContent>true</DeploymentContent>
-    </Manifest>
-</ItemGroup> 
 ```
+
+### C++/WinRT
+You'll need to use [C++/WinRT](https://docs.microsoft.com/en-us/windows/uwp/cpp-and-winrt-apis/intro-to-using-cpp-with-winrt) to consume your API. So make sure you have C++/WinRT installed, and have added `#include <winrt/MyAuthoredComponent.h>` to the file `pch.h` of the native app.  
 
 ## Known Authoring Issues
 You can follow along [here](https://github.com/microsoft/CsWinRT/issues/663) as we develop authoring support. 
--- a/docs/authoring.md
+++ b/docs/authoring.md
@@ -85,12 +85,12 @@ Because dotnet will assume a `TargetFramework` for your app that conflicts with 
 Examples of this are seen in the code snippet below. This is needed for this preview version, as we continue working on proper support.
 
 You can then use the Visual Studio UI to add a reference to the C#/WinRT component's `csproj` file, and you also need to add a reference to the WinMD file produced 
-for your component. The WinMD can be found in the output (`bin`) directory and the `Generated Files` directory of the authored component's project.
+for the component. The WinMD can be found in the output (`bin`) directory of the authored component's project.
 
 You'll need to author some files to assist the hosting process by the native app: `YourNativeApp.exe.manifest` and `WinRT.Host.runtimeconfig.json`. 
 If your app is packaged with MSIX, then you don't need to include the manifest file, otherwise you need to include your activatable class registrations in the manifest file.
 
-To do add these files, **in Visual Studio**, right click on the project node on the "Solution Explorer" window, click "Add", then "New Item". 
+To add these files, **in Visual Studio**, right click on the project node on the "Solution Explorer" window, click "Add", then "New Item". 
 Search for the "Text File" template and name your file `YourNativeApp.exe.manifest`.
 Repeat this for the `WinRT.Host.runtimeconfig.json` file. 
 For each item, right-click on it in the "Solution Explorer" window of Visual Studio; then select "Properties" and change the "Content" property to "Yes" using the drop-down arrow on the right -- this ensures it will be added to the output directory of your solution.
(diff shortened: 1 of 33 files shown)


docs/authoring.md
```diff
@@ -101,6 +101,13 @@ You should read the [hosting docs](https://github.com/microsoft/CsWinRT/blob/mas
 
 In summary, here is the fragment of additions made to the native app's project file:
 ``` vcxproj
+<!-- Note: this property group is only required if you are using a project reference -->
```
manodasanW, line 104: Mention this is needed for this preview as we work on proper support.

nuget/Microsoft.Windows.CsWinRT.Authoring.targets
```diff
@@ -1,47 +1,68 @@
...
-  <Target Name="CsWinRTCopyAuthoringDlls" Condition="'$(HostingSupport-IsNative)' == 'true'" Outputs="@(ReferenceCopyLocalPaths)">
-    
-    <ItemGroup Label="Copy Dlls needed for hosting/using authored components">
-      <ReferenceCopyLocalPaths Include="$(HostingSupport-NativeDir)\*.dll" />
+  <Target Name="CsWinRTCopySDKRefDllToOutDir" Condition="'$(CsWinRTComponent)' == 'true'" 
+          AfterTargets="ResolveRuntimePackAssets" BeforeTargets="GetCopyToOutputDirectoryXamlAppDefs">
+    <ItemGroup>
+      <CsWinRTSDKRefDll Include="@(RuntimePackAsset)" Condition="'%(RuntimePackAsset.DestinationSubPath)' == 'Microsoft.Windows.SDK.NET.dll'" />
+      <CsWinRTAuthoringDependencyDlls Include="@(CsWinRTSDKRefDll)" />
+      <_ThisProjectItemstoCopyToOutputDirectory Include="@(CsWinRTSDKRefDll)">
```
manodasanW, line 24: Can we use None include here rather rather something that might be private like _ThisProjectItemstoCopyToOutputDirectory.

docs/authoring.md
```diff
@@ -78,40 +76,40 @@ To generate a NuGet package for the component, you can simply right click on the
...
+If you choose to consume your component through a project reference, then some modifications to the native app's `.vcxproj` file are needed.
+Because dotnet will assume a `TargetFramework` for your app that conflicts with `net5`, we need to specify the `TargetFramwork`, `TargetFrameworkVersion` and `TargetRuntime`. 
+Examples of this are seen in the code snippet below. This is needed for this preview version, as we continue working on proper support.
 
-You'll need to use [C++/WinRT](https://docs.microsoft.com/en-us/windows/uwp/cpp-and-winrt-apis/intro-to-using-cpp-with-winrt) to consume your API. So make sure you have C++/WinRT installed, and have added `#include <winrt/MyAuthoredComponent.h>` to the file `pch.h` of the native app.  
+You can then use the Visual Studio UI to add a reference to the C#/WinRT component's `csproj` file, and you also need to add a reference to the WinMD file produced 
+for your component. The WinMD can be found in the output (`bin`) directory and the `Generated Files` directory of the authored component's project.
```
manodasanW, line 88: _for the component_ to be consistent with the rest of the sentence and because the component might not be the consumers.

manodasanW, line 88: Probably shouldn't mention the generated files one as that is an intermediate output.

docs/authoring.md
```diff
@@ -78,40 +76,40 @@ To generate a NuGet package for the component, you can simply right click on the
...
 
 You'll need to author some files to assist the hosting process by the native app: `YourNativeApp.exe.manifest` and `WinRT.Host.runtimeconfig.json`. 
-
 If your app is packaged with MSIX, then you don't need to include the manifest file, otherwise you need to include your activatable class registrations in the manifest file.
 
-To do this, **in Visual Studio**, right click on the project node on the "Solution Explorer" window, click "Add", then "New Item". Search for the "Text File" template and name your file "YourNativeApp.exe.manifest".
-Repeat this for the "WinRT.Host.runtimeconfig.json" file. 
+To do add these files, **in Visual Studio**, right click on the project node on the "Solution Explorer" window, click "Add", then "New Item".
```
manodasanW, line 93: To add these files (remove do).

docs/authoring.md
```diff
@@ -78,40 +76,40 @@ To generate a NuGet package for the component, you can simply right click on the
...
+<!-- Note: this property group is only required if you are using a project reference, 
+           and is a part of the preview while we work on proper support -->
+<PropertyGroup>
+  <TargetFrameworkVersion>net5.0</TargetFrameworkVersion>
+  <TargetFramework>native</TargetFramework>
+  <TargetRuntime>Native</TargetRuntime>
+</PropertyGroup>
 ```
 
+### C++/WinRT
```
manodasanW, line 111: Maybe make the header C++ consumption.  there is technically other native scenarios that the component can be consumed in.

nuget/Microsoft.Windows.CsWinRT.Authoring.targets
j0shuams: Unfortunately `None` will suffice with MsBuild, but not with a build done in Visual Studio

docs/authoring.md
```diff
@@ -36,23 +36,21 @@ The library you are authoring should specify the following properties in its pro
...
 
-**If you are going to write your own nuspec, i.e. not rely on automatic packaging** then the CsWinRT target that adds the hosting dlls to your package will not run, and you should make sure your nuspec contains the following ```file``` entries for ```MyAuthoredComponent``` (note: your TargetFramework may vary). 
+**If you are going to write your own nuspec**, then you should make sure your nuspec contains the following ```file``` entries for your component ```MyAuthoredComponent``` (note: your TargetFramework may vary). This is so our targets that supply the DLLs for any consumers of your package work.  
+Similarly, any other dependencies, e.g. `Microsoft.WinUI`, will need to be included in your nuspec as well.
 
 ``` nuspec
 <files>
-  <file src="$(TargetDir)MyAuthoredComponent.dll"        target="lib\native\MyAuthoredComponent.dll" />
+  <file src="$(TargetDir)MyAuthoredComponent.dll"        target="lib\$(TargetFramework)\MyAuthoredComponent.dll" />
   <file src="$(TargetDir)MyAuthoredComponent.winmd"      target="winmd\MyAuthoredComponent.winmd" />
```
manodasanW, line 51: Is this target directory still accurate, it looks different from below.

nuget/Microsoft.Windows.CsWinRT.Authoring.targets
```diff
@@ -1,47 +1,68 @@
...
-      <Reference Include="$(HostingSupport-MetadataDir)\*.winmd">
-        <IsWinMDFile>true</IsWinMDFile>
-        <Implementation>WinRT.Host.dll</Implementation>
-      </Reference>
-    </ItemGroup>
+    <None Include="@(CsWinRTAuthoringDependencyDlls)" Link="%(RecursiveDir)%(FileName)%(Extension)" CopyToOutputDirectory="PreserveNewest" />
 
-  </Target>
+    <CsWinRTAuthoringDependencyDlls Condition="Exists('$(CsWinRTPath)runtimes\win-$(Platform)\native\WinRT.Host.dll')" 
+                                    Remove="$(CsWinRTPath)runtimes\win-$(Platform)\native\WinRT.Host.dll" />
```
manodasanW, line 16: Is the remove here from CsWinRTAuthoringDependencyDlls intentional?

docs/authoring.md
j0shuams: Our transitive targets expect to find the dlls in a net5.0* directory of their nupkg

nuget/Microsoft.Windows.CsWinRT.Authoring.targets
```diff
@@ -1,47 +1,68 @@
...
+  </ItemGroup>
 
-  <!-- Happens when resolving references, so the app can be hosted -->
-  <Target Name="CsWinRTCopyAuthoringDlls" Condition="'$(HostingSupport-IsNative)' == 'true'" Outputs="@(ReferenceCopyLocalPaths)">
-    
-    <ItemGroup Label="Copy Dlls needed for hosting/using authored components">
-      <ReferenceCopyLocalPaths Include="$(HostingSupport-NativeDir)\*.dll" />
+  <Target Name="CsWinRTCopySDKRefDllToOutDir" Condition="'$(CsWinRTComponent)' == 'true'" 
+          AfterTargets="ResolveRuntimePackAssets" BeforeTargets="GetCopyToOutputDirectoryXamlAppDefs">
```
manodasanW, line 20: Does GetCopyToOutputDirectoryXamlAppDefs exist in non XAML scenarios?

docs/authoring.md
```diff
@@ -36,23 +37,57 @@ The library you are authoring should specify the following properties in its pro
...
+
+For consuming by "PackageReference", this is all that is required. C++ apps will need to use [C++/WinRT](https://docs.microsoft.com/en-us/windows/uwp/cpp-and-winrt-apis/intro-to-using-cpp-with-winrt) to consume the authored component. So make sure you have C++/WinRT installed, and have added `#include <winrt/MyAuthoredComponent.h>` to the file `pch.h` of the native app.  
+
+## Native Consumption by Project Reference
+
+If you choose to consume your component through a project reference in a native app, then some modifications to the native app's `.vcxproj` file are needed.
+Because dotnet will assume a `TargetFramework` for your app that conflicts with `net5`, we need to specify the `TargetFramwork`, `TargetFrameworkVersion` and `TargetRuntime`. 
+Examples of this are seen in the code snippet below. This is needed for this preview version, as we continue working on proper support.
+
+You will need to add a reference to both the C#/WinRT component's `csproj` file, and the WinMD produced for the component.
```
angelazhangmsft, line 60: does "Add a reference to the C#/WinRT component's csproj file" just mean a project reference?

j0shuams: Yes. Is it confusing by being too explicit?

manodasanW (approved):

nuget/Microsoft.Windows.CsWinRT.nuspec
```diff
@@ -22,9 +22,10 @@
     <file src="$cswinrt_exe$"/>
     <file src="Microsoft.Windows.CsWinRT.props" target="build"/>
     <file src="Microsoft.Windows.CsWinRT.targets" target="build"/>
-    <file src="Microsoft.Windows.CsWinRT.Prerelease*.targets" target="build"/>
     <file src="Microsoft.Windows.CsWinRT.Authoring.targets" target="build"/>
-    <file src="Microsoft.Windows.CsWinRT.Authoring.targets" target="buildTransitive"/>
+    <file src="Microsoft.Windows.CsWinRT.Prerelease*.targets" target="build"/>
+    <file src="Microsoft.Windows.CsWinRT.Authoring.Transitive.targets" target="build"/>
+    <file src="Microsoft.Windows.CsWinRT.Authoring.Transitive.targets" target="buildTransitive"/>
```
jlaanstra, line 28: This is almost certainly not what's intended. The fact that the target doesn't have the package name means this file is gonna be ignored.

j0shuams: @jlaanstra We have targets work in Authoring.targets that renames the file when packaging a custom component, so it doesn't get ignored. 

```
<!-- From the target "CsWinRTIncludeHostDlls" -->
<Content Include="$(CsWinRTPath)buildTransitive\Microsoft.Windows.CsWinRT.Authoring.Transitive.targets" 
               Pack="true" 
               PackagePath="buildTransitive\$(AssemblyName).targets;build\$(AssemblyName).targets" />
```

PR #731, opened 2021-02-08
Fix diagnostic bug for structs

j0shuams:
Before we were checking all structs, but we should only check the public ones. 

A bug came in for the following scenario

```
namespace DiagnosticTests
{
    public sealed class PrivateSetter
    {
        struct ExampleStruct
        {
            int i;
            public void SetValue(int value) { i = value; }
        }
    }
}
```

I tested this scenario in WinMDExp and it is valid. So I added a valid case, and changed the implementation to only raise diagnostics for public structs.

### DIFF
--- a/src/Authoring/WinRT.SourceGenerator/AnalyzerReleases.Unshipped.md
+++ b/src/Authoring/WinRT.SourceGenerator/AnalyzerReleases.Unshipped.md
@@ -23,7 +23,6 @@ CsWinRT1015 | Usage | Error | Do not use `DefaultOverloadAttribute` more than on
 CsWinRT1016 | Usage | Error | Exactly one overload should be marked as DefaultOverload
 CsWinRT1017 | Usage | Error | Array types should be one dimensional, not jagged
 CsWinRT1018 | Usage | Error | Array types should be one dimensional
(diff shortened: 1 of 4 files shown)


manodasanW (approved):

src/Authoring/WinRT.SourceGenerator/AnalyzerReleases.Unshipped.md
```diff
@@ -23,7 +23,6 @@ CsWinRT1015 | Usage | Error | Do not use `DefaultOverloadAttribute` more than on
 CsWinRT1016 | Usage | Error | Exactly one overload should be marked as DefaultOverload
 CsWinRT1017 | Usage | Error | Array types should be one dimensional, not jagged
 CsWinRT1018 | Usage | Error | Array types should be one dimensional
-CsWinRT1019 | Usage | Error | Do not use the `System.Array` type for array parameters or array return values
```
j0shuams, line 26: I took this out in a previous PR and missed this at that time. We don't use this rule anymore, we use InvalidType rule now.

j0shuams:
Closing, this got fixed in #737

PR #736, opened 2021-02-11
Add background task component sample

angelazhangmsft:
- adding authoring sample for background task WinRT component
- described the project modifications in the readme (required for v1.1.2-prerelease.210208.6)
- this sample is not part of cswinrt.sln.


Scottj1s (approved):
awesome!

PR #739, opened 2021-02-16
Perf improvements in CCW creation.

### LINKED ISSUE #438: Improve perf of CreateCCWForObject by caching.
WinRT.ComWrappersSupport.CreateCCWForObject(object) shows up in DP perf investigation traces.  The main cost is associated with calculating VTables (this seems to be partially due to fetching custom attributes).  Ideally (if possible), caching of VTables for type could speed it up.  If that's not possible (due to these changing at runtime), caching interface ptrs would reduce the number of calls to get custom attributes and would still give an improvement.

### LINKED ISSUE #554: VTable memory should be handled by Type not Object

manodasanW:
- Caching the vtable entries per type and as part of that also moving vtable cleanup to per type rather than per object.
- Also moving to store IInspectableInfo per type rather than per object.

Based on a perf and memory analysis, it can be seen the memory consumption for storing the vtable entries is down as expected and in the repro scenario where CCW creation was found to be expensive, it is no longer near the top of the CPU usage and the execution time of the repro seems to have went down by half.

Fixes #438 
Fixes #554

### DIFF
--- a/src/WinRT.Runtime/ComWrappersSupport.net5.cs
+++ b/src/WinRT.Runtime/ComWrappersSupport.net5.cs
@@ -26,7 +26,7 @@ namespace WinRT
             }
         }
 
-        internal static readonly ConditionalWeakTable<object, InspectableInfo> InspectableInfoTable = new ConditionalWeakTable<object, InspectableInfo>();
+        internal static readonly ConditionalWeakTable<Type, InspectableInfo> InspectableInfoTable = new ConditionalWeakTable<Type, InspectableInfo>();
         internal static readonly ThreadLocal<Type> CreateRCWType = new ThreadLocal<Type>();
 
         private static ComWrappers _comWrappers;
@@ -67,7 +67,7 @@ namespace WinRT
         internal static unsafe InspectableInfo GetInspectableInfo(IntPtr pThis)
         {
             var _this = FindObject<object>(pThis);
-            return InspectableInfoTable.GetValue(_this, o => PregenerateNativeTypeInformation(o).inspectableInfo);
+            return InspectableInfoTable.GetValue(_this.GetType(), o => PregenerateNativeTypeInformation(o).inspectableInfo);
         }
 
         public static T CreateRcwForComObject<T>(IntPtr ptr)
@@ -193,7 +193,7 @@ namespace WinRT
 
     public class DefaultComWrappers : ComWrappers
     {
-        private static ConditionalWeakTable<object, VtableEntriesCleanupScout> ComInterfaceEntryCleanupTable = new ConditionalWeakTable<object, VtableEntriesCleanupScout>();
+        private static readonly ConditionalWeakTable<Type, VtableEntries> TypeVtableEntryTable = new ConditionalWeakTable<Type, VtableEntries>();
         public static unsafe IUnknownVftbl IUnknownVftbl => Unsafe.AsRef<IUnknownVftbl>(IUnknownVftblPtr.ToPointer());
 
         internal static IntPtr IUnknownVftblPtr { get; }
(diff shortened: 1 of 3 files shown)


AaronRobinsonMSFT:
FYI @jkotas

Scottj1s (approved):

src/WinRT.Runtime/ComWrappersSupport.net5.cs
```diff
@@ -212,51 +212,57 @@ static unsafe DefaultComWrappers()
...
+                });
+
+                // This should be the last entry as it is included / excluded based on the flags.
+                entries.Add(new ComInterfaceEntry
+                {
+                    IID = typeof(IUnknownVftbl).GUID,
+                    Vtable = IUnknownVftbl.AbiToProjectionVftblPtr
+                });
+
+                ComInterfaceEntry* nativeEntries = (ComInterfaceEntry*)Marshal.AllocCoTaskMem(sizeof(ComInterfaceEntry) * entries.Count);
```
AaronRobinsonMSFT, line 239: I would avoid the manual managed allocation here and use [`RuntimeHelpers.AllocateTypeAssociatedMemory()`](https://docs.microsoft.com/dotnet/api/system.runtime.compilerservices.runtimehelpers.allocatetypeassociatedmemory) instead.

manodasanW: I did look at using it, but I didn't have a static field associated with the type that I could use to store it and I was concerned about storing this address in the ConditionalWeakTable as I wasn't sure whether there is a scenario where the memory allocated from this function might be cleaned up before the entry is removed / invalidated from the weak table.  If you think this latter case won't happen (i.e. I can't retrieve an entry from a ConditionalWeakTable based on type for which the type had already executed FreeMemory for any type associated memory), I can move to using it.

AaronRobinsonMSFT: >i.e. I can't retrieve an entry from a ConditionalWeakTable based on type for which the type had already executed FreeMemory for any type associated memory

Yep. This function was designed for exactly the scenario being performed here. The fix is simple - change the above call and remove the finalizer doing the cleanup in `VtableEntries`.

AaronRobinsonMSFT: This will also reduce pressure on the finalizer queue.

AaronRobinsonMSFT:
>Based on a perf and memory analysis, it can be seen the memory consumption for storing the vtable entries is down as expected and in the repro scenario where CCW creation was found to be expensive, it is no longer near the top of the CPU usage and the execution time of the repro seems to have went down by half.

@manodasanW I appreciate the following request may be annoying, but would it be possible to get a sense of the raw numbers we are looking at here? Like Scenario A vs Scenario B and then % improvement? Memory used and CPU usage are all that is interesting at the moment. It helps the .NET team talk about the investments we make in APIs and scenarios they unlock.

BenJKuhn (approved):

AaronRobinsonMSFT (approved):

manodasanW:
> > Based on a perf and memory analysis, it can be seen the memory consumption for storing the vtable entries is down as expected and in the repro scenario where CCW creation was found to be expensive, it is no longer near the top of the CPU usage and the execution time of the repro seems to have went down by half.
> 
> @manodasanW I appreciate the following request may be annoying, but would it be possible to get a sense of the raw numbers we are looking at here? Like Scenario A vs Scenario B and then % improvement? Memory used and CPU usage are all that is interesting at the moment. It helps the .NET team talk about the investments we make in APIs and scenarios they unlock.

I plan on asking Bartek if he can do an updated run of this scenario which he has been tracking for various CsWinRT versions and the previous built-in version.  I will also look at getting some metrics for memory usage.

AaronRobinsonMSFT:
>I plan on asking Bartek if he can do an updated run of this scenario which he has been tracking for various CsWinRT versions and the previous built-in version. I will also look at getting some metrics for memory usage.

Perfect. That would be most insightful. Thanks!

PR #741, opened 2021-02-17
ensure that value/delegate types are projected before implementing IPropertyValue

### LINKED ISSUE #605: Non projected enums / delegates shouldn't provide IPropertyValue QI
[See](https://github.com/microsoft/microsoft-ui-xaml/issues/3339) for repro

Scottj1s:
fixes #605

### DIFF
--- a/src/WinRT.Runtime/ComWrappersSupport.cs
+++ b/src/WinRT.Runtime/ComWrappersSupport.cs
@@ -124,11 +124,16 @@ namespace WinRT
 
             if (obj is Delegate)
             {
-                entries.Add(new ComInterfaceEntry
+                var delegateType = obj.GetType();
+                var helperType = delegateType.FindHelperType();
+                if (helperType is object)
                 {
-                    IID = GuidGenerator.GetIID(obj.GetType()),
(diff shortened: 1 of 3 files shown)


src/WinRT.Runtime/ComWrappersSupport.cs
```diff
@@ -337,7 +341,24 @@ internal static string GetRuntimeClassForTypeCreation(IInspectable inspectable,
 
         private static bool ShouldProvideIReference(Type type)
         {
-            return type.IsValueType || type == typeof(string) || type == typeof(Type) || type.IsDelegate();
+            static bool IsWindowsRuntimeType(Type type)
+            {
+                // TODO: attribute checks are expensive - cache per-vtable
```
manodasanW, line 346: Is this ToDo still necessary given the higher level caching changes where this function gets called?

manodasanW (approved):

src/WinRT.Runtime/ComWrappersSupport.cs
Scottj1s: no, should be covered by that

Issue #733, opened 2021-02-09
[Diagnostics] Namespace conflict check


j0shuams:
**Describe the bug**
We have some code that checks that namespaces only differ by case, and it seems that is broken.
A bug has come in that two sub-namespaces “AuthoringDemo.Foo.Bar.Dog.Test” and “AuthoringDemo.Foo.Bar.Cat.Test” are causing the "namespaces cannot differ only by case". The issue is it thinks "Test" and "Test" conflict, as it is not checking the parent namespaces -- e.g. "Dog.Test" vs. "Cat.Test".

We should also update the error message as it is confusing to consumers what "namespace differ by case" means

Issue #732, opened 2021-02-09
[Diagnostics] Namespace with no public types causes CsWinRT1001


j0shuams:
**Describe the bug**
Namespaces with no public types get diagnostics raised on them when they shouldn't. We should ignore namespaces that don't have public types.

**To Reproduce**
1. Make a runtime component, say `MyComponent`
2. Have a namespace such as `MyComponent.Foo` and `Utilities`
3. `Utilities` should have no public types, `MyComponent.Foo` should

**Expected behavior**
`Utilities` should be ignored as it will not be used in the Windows Runtime

**Additional context**
When the TypeCollector runs, have it check if a namespace has public types.
Get a repro with two namespaces as described, then make a unit test and update the diagnostics.

j0shuams:
I think #737 fixed this as well

PR #749, opened 2021-02-22
Namespace diagnostic fixes

j0shuams:
This PR addresses two diagnostic issues with namespaces #732 and #733 

I am getting rid of the "TypeCollector" I used to check for the "no public types" diagnostic, and instead checking in the SyntaxReceiver. If a namespace declaration has no public types, we don't add it as a declaration; then on the diagnostic side if we have no declarations, we raise the "no public types" diagnostic. 

I also changed the way we check namespace names for conflicts -- this is because we cannot allow namespaces like "Sample" and "SaMpLe, which only differ by case. We are getting rid of the Contains method and instead looping through the set of namespace names seen so far -- a bit of a pain to not have O(1) method, but I didn't think of another way.

Also making some quality of life/code adjustments like formatting and a readme for adding unit tests.


src/Authoring/WinRT.SourceGenerator/Generator.cs
```diff
@@ -222,9 +224,20 @@ class WinRTSyntaxReciever : ISyntaxReceiver
     {
         public List<MemberDeclarationSyntax> Declarations = new List<MemberDeclarationSyntax>();
 
+        private bool HasSomePublicTypes(SyntaxNode syntaxNode)
+        {
+            bool hasSomePublicTypes = false;
```
manodasanW, line 229: This probably should include enums too and delegates?  Should we just look for any public child node?

src/Authoring/WinRT.SourceGenerator/Generator.cs
```diff
@@ -203,10 +203,12 @@ public class SourceGenerator : ISourceGenerator
     {
         public void Execute(GeneratorExecutionContext context)
         {
+            /*
```
manodasanW, line 206: intentional?

j0shuams: Yes, it's temporary for testing, will be removed in final commit.

src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
```diff
@@ -15,25 +16,31 @@ public WinRTComponentScanner(GeneratorExecutionContext context, string assemblyN
...
-            HasSomePublicTypes();
-
+        {
             WinRTSyntaxReciever syntaxReciever = (WinRTSyntaxReciever)_context.SyntaxReceiver;
+
+            if (!syntaxReciever.Declarations.Any())
+            {
+                Report(WinRTRules.NoPublicTypesRule, null);
+                // exit early?
+                // return;
```
manodasanW, line 38: Anything to do if no types?

src/Authoring/WinRT.SourceGenerator/Generator.cs
j0shuams: Updated, seems to work fine.

src/Authoring/WinRT.SourceGenerator/Generator.cs
```diff
@@ -222,9 +224,16 @@ class WinRTSyntaxReciever : ISyntaxReceiver
...
+        {
+            return syntaxNode.ChildNodes().OfType<MemberDeclarationSyntax>().Any(IsPublic);
+        }
+
         public void OnVisitSyntaxNode(SyntaxNode syntaxNode)
         {
-            if (syntaxNode is not MemberDeclarationSyntax decaralation || !IsPublic(decaralation))
+            // We will lose namespaces in `Declarations` if we don't special case them as exempt from the `IsPublic` check
+            if (syntaxNode is not MemberDeclarationSyntax decaralation || 
+                (syntaxNode is not NamespaceDeclarationSyntax @namespace && !IsPublic(decaralation)))
```
manodasanW, line 236: probably can remove @namespace as we don't use it.

manodasanW (approved):

src/Authoring/WinRT.SourceGenerator/Generator.cs
```diff
@@ -231,9 +232,23 @@ private bool HasSomePublicTypes(SyntaxNode syntaxNode)
...
-            // We will lose namespaces in `Declarations` if we don't special case them as exempt from the `IsPublic` check
-            if (syntaxNode is not MemberDeclarationSyntax decaralation || 
-                (syntaxNode is not NamespaceDeclarationSyntax @namespace && !IsPublic(decaralation)))
+            // Store namespaces separately as we only need to look at them for diagnostics
+            // If we did store them in declarations, we would get duplicate entries in the WinMD,
+            // once from the namespace declaration and once from the member's declaration
+            if (syntaxNode is NamespaceDeclarationSyntax @namespace)
+            {
+                if (!HasSomePublicTypes(syntaxNode))
+                {
```
manodasanW, line 241: We might want to just return always if it enters this if namespace as it is not going to meet the checks after.

PR #761, opened 2021-03-07
Use .NET Core 3.1 compatible marshaling for DateTimeOffset.

### LINKED ISSUE #703: DateTimeOffset marshaling doesn't support negative FileTime
**Describe the bug**
https://github.com/microsoft/microsoft-ui-xaml/issues/3889
https://github.com/dotnet/runtime/issues/47523#issuecomment-768808987

Currently `DateTimeOffset` marshaling is using `ToFileTime`, which doesn't support negative value, while WinRT time supports.

huoyaoyuan:
Closes #703

### DIFF
--- a/src/WinRT.Runtime/Projections/SystemTypes.cs
+++ b/src/WinRT.Runtime/Projections/SystemTypes.cs
@@ -67,14 +67,21 @@ namespace ABI.System
 
         public static Marshaler CreateMarshaler(global::System.DateTimeOffset value)
         {
-            return new Marshaler { __abi = new DateTimeOffset { UniversalTime = value.Ticks - ManagedUtcTicksAtNativeZero } };
+            return new Marshaler { __abi = new DateTimeOffset { UniversalTime = value.UtcTicks - ManagedUtcTicksAtNativeZero } };
         }
 
         public static DateTimeOffset GetAbi(Marshaler m) => m.__abi;
 
         public static global::System.DateTimeOffset FromAbi(DateTimeOffset value)
         {
-            return new global::System.DateTimeOffset(value.UniversalTime + ManagedUtcTicksAtNativeZero, global:: System.TimeSpan.Zero);
+            var utcTime = new global::System.DateTimeOffset(value.UniversalTime + ManagedUtcTicksAtNativeZero, global::System.TimeSpan.Zero);
+            var offset = TimeZoneInfo.Local.GetUtcOffset(utcTime);
+            long localTicks = utcTime.Ticks + offset.Ticks;
+            if (localTicks < DateTime.MinValue.Ticks || localTicks > DateTime.MaxValue.Ticks)
+            {
+                throw new ArgumentOutOfRangeException();
+            }
+            return utcTime.ToLocalTime();
         }
 
         public static unsafe void CopyAbi(Marshaler arg, IntPtr dest) =>
--- a/src/WinRT.Runtime/Projections/SystemTypes.cs
+++ b/src/WinRT.Runtime/Projections/SystemTypes.cs
@@ -55,6 +55,11 @@ namespace ABI.System
         // 'System.DateTimeOffset.Ticks' is relative to 01/01/0001
         public long UniversalTime;
 
+        // Numer of ticks counted between 0001-01-01, 00:00:00 and 1601-01-01, 00:00:00.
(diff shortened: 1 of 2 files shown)


src/WinRT.Runtime/Projections/SystemTypes.cs
```diff
@@ -55,21 +55,26 @@ public struct DateTimeOffset
...
-            return new Marshaler { __abi = new DateTimeOffset { UniversalTime = value.ToFileTime() } };
+            return new Marshaler { __abi = new DateTimeOffset { UniversalTime = value.Ticks - ManagedUtcTicksAtNativeZero } };
         }
 
         public static DateTimeOffset GetAbi(Marshaler m) => m.__abi;
 
         public static global::System.DateTimeOffset FromAbi(DateTimeOffset value)
         {
-            return global::System.DateTimeOffset.FromFileTime(value.UniversalTime);
+            return new global::System.DateTimeOffset(value.UniversalTime + ManagedUtcTicksAtNativeZero, global:: System.TimeSpan.Zero);
```
manodasanW, line 77: The time we are getting here is going to be UTC.  It looks like .NET core previously then converted it to local time as seen in the link from the issue and threw an exception for any invalid local times.  We probably should do something similar, but it looks like they used [internal APIs](https://github.com/dotnet/runtime/blob/79ae74f5ca5c8a6fe3a48935e85bd7374959c570/src/libraries/System.Private.CoreLib/src/System/DateTimeOffset.cs#L631) to achieve it.  We can probably achieve something similar by doing something like

```
TimeSpan offset = TimeZoneInfo.Local.GetUtcOffset(utcDateTime);
long localTicks = utcDateTime.Ticks + offset.Ticks;
if (localTicks < DateTime.MinTicks || localTicks > DateTime.MaxTicks)
{
    throw new ArgumentOutOfRangeException();
}

return new global::System.DateTimeOffset(localTicks, offset);
```

src/WinRT.Runtime/Projections/SystemTypes.cs
```diff
@@ -55,21 +55,26 @@ public struct DateTimeOffset
...
         }
 
         public static unsafe void CopyAbi(Marshaler arg, IntPtr dest) =>
```
manodasanW, line 80: Probably want to do something similar below in FromManaged as it uses ToFileTime.

src/WinRT.Runtime/Projections/SystemTypes.cs
```diff
@@ -55,21 +55,26 @@ public struct DateTimeOffset
         // 'System.DateTimeOffset.Ticks' is relative to 01/01/0001
         public long UniversalTime;
 
+        // Numer of ticks counted between 0001-01-01, 00:00:00 and 1601-01-01, 00:00:00.
+        // You can get this through:  (new DateTimeOffset(1601, 1, 1, 0, 0, 1, TimeSpan.Zero)).Ticks;
+        private const long ManagedUtcTicksAtNativeZero = 504911232000000000;
```
manodasanW, line 60: Can you add to this [DateTime test](https://github.com/microsoft/CsWinRT/blob/d361800d2cf6a46d2c7d276126a09bab1d8bc1c4/src/Tests/UnitTest/TestComponentCSharp_Tests.cs#L1912) to exercise a negative value where these changes help.  You can follow the same format where you set the property and then retrieve it and compare.

manodasanW (changes requested):
Thanks @huoyaoyuan for your contribution to making it compatible with .NET core 3.  I left a couple of comments after which I think we should be fully compatible and should be able to merge this.

src/WinRT.Runtime/Projections/SystemTypes.cs
manodasanW: Thanks @huoyaoyuan for addressing the other comments.  I am not sure whether you missed this one.  Right below this function in `FromManaged`, we still use `ToFileTime` as seen below.  That probably should be changed to reflect the same changes you did in `CreateMarshaler`.  Once that is done, I can merge this PR.

```
public static DateTimeOffset FromManaged(global::System.DateTimeOffset value)
{
    return new DateTimeOffset { UniversalTime = value.ToFileTime() };
}
```

manodasanW (approved):

PR #770, opened 2021-03-16
Fixed regression in determining whether to provide IReference implementation, which dropped TimeSpan and a few other value types

Scottj1s:


src/WinRT.Runtime/ComWrappersSupport.cs
```diff
@@ -343,12 +343,13 @@ private static bool ShouldProvideIReference(Type type)
         {
             static bool IsWindowsRuntimeType(Type type)
             {
-                if (type.GetCustomAttribute<WindowsRuntimeTypeAttribute>() is object)
+                if ((type.GetCustomAttribute<WindowsRuntimeTypeAttribute>() is object) ||
+                    WinRT.Projections.IsTypeWindowsRuntimeType(type))
```
j0shuams, line 347: Is it worth it to make a local var for `WinRT.Projections.IsTypeWindowsRuntimeType(type)` since we use it here and at the exit statement (L352)

j0shuams (approved):

src/WinRT.Runtime/ComWrappersSupport.cs
Scottj1s: type gets reassigned between calls

src/WinRT.Runtime/ComWrappersSupport.cs
```diff
@@ -343,12 +343,13 @@ private static bool ShouldProvideIReference(Type type)
         {
             static bool IsWindowsRuntimeType(Type type)
```
manodasanW, line 344: Does WinRT.Projections.IsTypeWindowsRuntimeType do the equivalent of this function?

PR #773, opened 2021-03-18
Fix GC hole in string marshaling

### LINKED ISSUE #768: Access Violation in RoGetActivationFactory
**Describe the bug**

One of our customers has a problem where the application is crashing with Access Violation during initialization of "Windows.UI.Notifications.ToastNotificationManager". It is not immediately obvious whether the problem is in the CsWinRT projection or the native side of the code but it looks more likely to be a platform issue. There seems to be some discrepancy between the strings seen by the managed code and the native code but in both cases they seem to be valid representation in memory.

We'd be happy to share the memory dump privately.

***Stack trace***
```
00000000 00004007 00004007 ====> Exception cxr@0027C9A0
...
07 (Inline) --------     -------- -------- -------- combase!StringCchLengthW+0xb [onecore\external\shared\inc\strsafe.h @ 9098] 
08 (Inline) --------     -------- -------- -------- combase!ApiTelemetryLogger::LogApiCall+0x1f [onecore\com\combase\winrtapilogger\winrtapilogger.cpp @ 130] 
09 0027d204 7508ffda     00000000 00000000 0027d284 combase!_RoGetActivationFactory+0xca5 [onecore\com\combase\winrtbase\winrtbase.cpp @ 1044] 
0a 0027d218 0bc87032     0719f47c 0027d288 0027d284 combase!RoGetActivationFactory+0x1a [onecore\com\combase\winrtbase\winrtbase.cpp @ 1064] 
...
0027D218 0bc87032 (MethodDesc 0db44d50 + 0x4a ILStubClass.IL_STUB_PInvoke(IntPtr, System.Guid ByRef, IntPtr*))
0027D23C 0bc87032 (MethodDesc 0db44d50 + 0x4a ILStubClass.IL_STUB_PInvoke(IntPtr, System.Guid ByRef, IntPtr*))
0027D260 0bc86a5f (MethodDesc 0db435c8 + 0x5f WinRT.WinrtModule.GetActivationFactory(IntPtr)), calling 0bc864ec (stub for WinRT.Platform.RoGetActivationFactory(IntPtr, System.Guid ByRef, IntPtr*))
0027D2A4 0bc861dd (MethodDesc 0db435d8 + 0x9d WinRT.WinrtModule.GetActivationFactory(System.String)), calling 1d7c1cf8
0027D328 0bc86083 (MethodDesc 0db42038 + 0x2b WinRT.BaseActivationFactory..ctor(System.String, System.String)), calling 1d7c1c44
0027D35C 239d26a7 (MethodDesc 239c52a0 + 0x8f Windows.UI.Notifications.ToastNotificationManager+_IToastNotificationManagerStatics..ctor()), calling (MethodDesc 0db42038 + 0 WinRT.BaseActivationFactory..ctor(System.String, System.String))
```

***Variable details***
Part of stack trace dealing with the activation:
```
0027D230 0bc86a5f [InlinedCallFrame: 0027d230] WinRT.Platform.RoGetActivationFactory(IntPtr, System.Guid ByRef, IntPtr*)
0027D26C 0bc86a5f WinRT.WinrtModule.GetActivationFactory(IntPtr)
    PARAMETERS:
        hstrRuntimeClassId (0x0027D29C) = 0x0719f47c

0027D2AC 0bc861dd WinRT.WinrtModule.GetActivationFactory(System.String)
    PARAMETERS:
        runtimeClassId (0x0027D304) = 0x0719f108
```

Dumped variables:
```
 !DumpObj /d 0719f108
Name:        System.String
MethodTable: 067aec98
EEClass:     067ba89c
Size:        112(0x70) bytes
File:        C:\Program Files (x86)\eM Client\System.Private.CoreLib.dll
String:      Windows.UI.Notifications.ToastNotificationManager
Fields:
      MT    Field   Offset                 Type VT     Attr    Value Name
067a73b4  4000212        4         System.Int32  1 instance       49 _stringLength
067a54e0  4000213        8          System.Char  1 instance       57 _firstChar
067aec98  4000211       60        System.String  0   static 06a012b0 Empty

struct HSTRING__ * activatableClassId = 0x0719f47c ""  "(null)"
```

**To Reproduce**

We do not have steps to reproduce. It only happen on customer machine. The application is compiled as `net5.0` with .NET 5.0.104 SDK and calls `Windows.UI.Notifications.ToastNotificationManager.GetTemplateContent(ToastTemplateType.ToastImageAndText01).GetXml()` API.

**Expected behavior**
No crash

**Version Info**
.NET 5.0.4 (SDK 5.0.104)
OS Windows 10 (19041.1.amd64fre.vb_release.191206-1406)

filipnavara:
Fixes #768

Fix GC hole where the whole `MarshalString` object gets moved and makes the `HSTRING` reference invalid
Fix memory leak and possible memory corruption in exception handling in GetActivationFactory

The implementation of `WindowsCreateStringReference` takes the header buffer, fills it and then the returned handle is a pointer to the header buffer. It is thus absolutely essential that the header buffer is not moved while the HSTRING instance is alive. This is explicitly called out in the [documentation](https://docs.microsoft.com/en-us/windows/win32/api/winstring/nf-winstring-windowscreatestringreference):

> The caller must ensure that sourceString and the contents of hstringHeader remain unchanged during the lifetime of the attached HSTRING.

The code was storing the header buffer in unpinned GC memory so the GC was free to move it and it occasionally did so which caused accesses from native code into invalid memory.

Now the solution in the PR is not the most optimal one but it's one that is least disruptive and easy to review. From performance point of view it would make sense to use `ref struct` and rely on the stack being unmovable.

### DIFF
--- a/src/WinRT.Runtime/Marshalers.cs
+++ b/src/WinRT.Runtime/Marshalers.cs
@@ -32,19 +32,15 @@ namespace WinRT
     // as well as passing marshalers to FromAbi by ref so they can be conditionally disposed.
     public class MarshalString
     {
-        public unsafe struct HStringHeader // sizeof(HSTRING_HEADER)
-        {
-            public fixed byte Reserved[24];
-        };
-        public HStringHeader _header;
+        private IntPtr _header;
         public GCHandle _gchandle;
-        private GCHandle _gchandleSelf;
         public IntPtr _handle;
 
         public void Dispose()
         {
             _gchandle.Dispose();
-            _gchandleSelf.Dispose();
+            Marshal.FreeHGlobal(_header);
+            _header = IntPtr.Zero;
         }
 
         public static unsafe MarshalString CreateMarshaler(string value)
@@ -56,11 +52,11 @@ namespace WinRT
             try
             {
                 m._gchandle = GCHandle.Alloc(value, GCHandleType.Pinned);
-                m._gchandleSelf = GCHandle.Alloc(m, GCHandleType.Pinned);
-                fixed (void* chars = value, header = &m._header, handle = &m._handle)
+                m._header = Marshal.AllocHGlobal(24); // sizeof(HSTRING_HEADER)
+                fixed (void* chars = value, handle = &m._handle)
                 {
                     Marshal.ThrowExceptionForHR(Platform.WindowsCreateStringReference(
-                        (char*)chars, value.Length, (IntPtr*)header, (IntPtr*)handle));
+                        (char*)chars, value.Length, (IntPtr*)m._header, (IntPtr*)handle));
                 };
                 return m;
             }
--- a/src/WinRT.Runtime/Marshalers.cs
+++ b/src/WinRT.Runtime/Marshalers.cs
@@ -38,11 +38,13 @@ namespace WinRT
         };
         public HStringHeader _header;
         public GCHandle _gchandle;
+        private GCHandle _gchandleHeader;
         public IntPtr _handle;
 
         public void Dispose()
         {
             _gchandle.Dispose();
+            _gchandleHeader.Dispose();
         }
 
         public static unsafe MarshalString CreateMarshaler(string value)
@@ -54,6 +56,7 @@ namespace WinRT
             try
             {
                 m._gchandle = GCHandle.Alloc(value, GCHandleType.Pinned);
+                m._gchandleHeader = GCHandle.Alloc(m._header, GCHandleType.Pinned);
                 fixed (void* chars = value, header = &m._header, handle = &m._handle)
                 {
                     Marshal.ThrowExceptionForHR(Platform.WindowsCreateStringReference(
--- a/src/WinRT.Runtime/Marshalers.cs
+++ b/src/WinRT.Runtime/Marshalers.cs
@@ -38,13 +38,13 @@ namespace WinRT
         };
         public HStringHeader _header;
         public GCHandle _gchandle;
-        private GCHandle _gchandleHeader;
+        private GCHandle _gchandleSelf;
         public IntPtr _handle;
 
         public void Dispose()
         {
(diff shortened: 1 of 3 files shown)


src/WinRT.Runtime/Marshalers.cs
```diff
@@ -54,6 +56,7 @@ public static unsafe MarshalString CreateMarshaler(string value)
             try
             {
                 m._gchandle = GCHandle.Alloc(value, GCHandleType.Pinned);
+                m._gchandleHeader = GCHandle.Alloc(m._header, GCHandleType.Pinned);
```
jkoritzinsky, line 59: I think this will end up making a boxed copy of the header and pinning that.

filipnavara: D'oh. You could be right. I will double check it and attempt a different fix if that turns out to be the case.

filipnavara: Hopefully fixed by pinning the `MarshalString` object itself. I'm widely uncomfortable with the layout of that class. Just want to have something working with minimal changes before trying to improve on the fix.

jkoritzinsky: Pinning the `MarshalString` object itself should work since it's a class.

filipnavara:
Verified with our customer that the fix helped.

filipnavara:
@angelazhangmsft Please give this issue/PR appropriate attention. It's serious enough that it makes CsWinRT totally unusable for us due to random crashes. I am not pushing for accepting my fix verbatim but it's likely affecting all CsWinRT users to certain extent.

Scottj1s:
@filipnavara - really appreciate this fix.  yes, GC moves had not been considered.

Scottj1s (approved):

manodasanW (approved):

PR #781, opened 2021-03-31
Add tests for memory leak

j0shuams:
This PR consolidates the leak repro samples/tests from @manodasanW 's topic branch into the existing WinUI tests/samples we have. This cuts down on the size of the solution primarily, and also acts as the first step in merging leak fixes.


src/Tests/AuthoringWinUITest/AuthoringWinUITest/AuthoringWinUITest.vcxproj.filters
```diff
@@ -7,7 +7,6 @@
     <Page Include="MainWindow.xaml" />
   </ItemGroup>
   <ItemGroup>
-    <Midl Include="App.idl" />
```
j0shuams, line 10: ?

src/build.cmd
```diff
@@ -124,7 +126,12 @@ call :exec %nuget_dir%\nuget.exe restore %nuget_params% %this_dir%cswinrt.sln
...
-call :exec %msbuild_path%msbuild.exe %cswinrt_build_params% /p:platform=%cswinrt_platform%;configuration=%cswinrt_configuration%;VersionNumber=%cswinrt_version_number%;VersionString=%cswinrt_version_string%;AssemblyVersionNumber=%cswinrt_assembly_version%;GenerateTestProjection=true;BaselineAllAPICompatError=%cswinrt_baseline_breaking_compat_errors%;BaselineAllMatchingRefApiCompatError=%cswinrt_baseline_assembly_version_compat_errors% %this_dir%cswinrt.sln 
+set nosampletarget=""
+if %cswinrt_no_samples%=="true" (
+  :: this is a target in Directory.Build.targets that removes all project files under the Samples directory
+  set nosampletarget="/t:CsWinRTDontBuildSamples"
+)
+call :exec %msbuild_path%msbuild.exe %cswinrt_build_params% /p:platform=%cswinrt_platform%;configuration=%cswinrt_configuration%;VersionNumber=%cswinrt_version_number%;VersionString=%cswinrt_version_string%;AssemblyVersionNumber=%cswinrt_assembly_version%;GenerateTestProjection=true;BaselineAllAPICompatError=%cswinrt_baseline_breaking_compat_errors%;BaselineAllMatchingRefApiCompatError=%cswinrt_baseline_assembly_version_compat_errors% %this_dir%cswinrt.sln %nosampletarget% /bl
```
j0shuams, line 134: We are passing %nosampletarget%, which tries to remove the WinUI samples from the build of the solution for this call, Note we are adding "/bl" here for binlog -- we can customize this to "/bl:CsWinRT.Build.binlog" if we don't want to remove it

Scottj1s:
let's not use the sample for leak tests, as the pipeline won't automate it.  @azchohfi recently implemented mstest support for WinUI.  It should land on nuget.org any day:
https://www.nuget.org/packages/MSTest.TestFramework/

Ideally, we can take the ObjectLifetimeTests, add a few more, and run them as proper unit tests.

manodasanW:
> let's not use the sample for leak tests, as the pipeline won't automate it. @azchohfi recently implemented mstest support for WinUI. It should land on nuget.org any day:
> https://www.nuget.org/packages/MSTest.TestFramework/
> 
> Ideally, we can take the ObjectLifetimeTests, add a few more, and run them as proper unit tests.

I agree with that, but at the same time I think until we have those, we need a set of scenarios that we can quickly run through to demonstrate as we do leak fixes that we haven't regressed anything and I feel like these samples help with that.  What do you think?

Scottj1s:
> > let's not use the sample for leak tests, as the pipeline won't automate it. @azchohfi recently implemented mstest support for WinUI. It should land on nuget.org any day:
> > https://www.nuget.org/packages/MSTest.TestFramework/
> > Ideally, we can take the ObjectLifetimeTests, add a few more, and run them as proper unit tests.
> 
> I agree with that, but at the same time I think until we have those, we need a set of scenarios that we can quickly run through to demonstrate as we do leak fixes that we haven't regressed anything and I feel like these samples help with that. What do you think?

I'm okay with an IOU - let's drop a comment in the samples to that effect

Scottj1s (approved):

j0shuams:
Opened #784 per discussion here

src/Samples/WinUIDesktopSample/XamlPage.xaml
```diff
@@ -0,0 +1,16 @@
...
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:local="using:WinUIDesktopSample"
+    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    mc:Ignorable="d"
+    Background="{ThemeResource ApplicationPageBackgroundThemeBrush}">
+
+    <StackPanel>
+        <Button Content="Navigate to SecondPage" Click="Button_Click" />
+        <TextBlock x:Name="Status">Statu</TextBlock>
```
manodasanW, line 13: This was my typo, should be Status.

src/build.cmd
```diff
@@ -124,7 +124,7 @@ call :exec %nuget_dir%\nuget.exe restore %nuget_params% %this_dir%cswinrt.sln
 
 :build
 echo Building cswinrt for %cswinrt_platform% %cswinrt_configuration%
-call :exec %msbuild_path%msbuild.exe %cswinrt_build_params% /p:platform=%cswinrt_platform%;configuration=%cswinrt_configuration%;VersionNumber=%cswinrt_version_number%;VersionString=%cswinrt_version_string%;AssemblyVersionNumber=%cswinrt_assembly_version%;GenerateTestProjection=true;BaselineAllAPICompatError=%cswinrt_baseline_breaking_compat_errors%;BaselineAllMatchingRefApiCompatError=%cswinrt_baseline_assembly_version_compat_errors% %this_dir%cswinrt.sln 
+call :exec %msbuild_path%msbuild.exe %cswinrt_build_params% /p:platform=%cswinrt_platform%;configuration=%cswinrt_configuration%;VersionNumber=%cswinrt_version_number%;VersionString=%cswinrt_version_string%;AssemblyVersionNumber=%cswinrt_assembly_version%;GenerateTestProjection=true;BaselineAllAPICompatError=%cswinrt_baseline_breaking_compat_errors%;BaselineAllMatchingRefApiCompatError=%cswinrt_baseline_assembly_version_compat_errors% %this_dir%cswinrt.sln /bl
```
manodasanW, line 127: Was /bl intentional?

manodasanW (approved):

src/build.cmd
j0shuams: Yes it is temporary for verifying our new configuration works correctly. Will be removed in final commit

src/Samples/WinUIDesktopSample/XamlPage.xaml
j0shuams: Will get this in final commit too

azchohfi:
https://www.nuget.org/packages/MSTest.TestFramework/2.2.4-preview-20210331-02 is out now.

src/build.cmd
Scottj1s: that's what cswinrt_build_params is for ;-)

Scottj1s (commented):
given Alex's note about winui test support on nuget.org now, can we avoid this churn and just move up directly?

Scottj1s (changes requested):
consider winuitest support in latest MSTest.TestFramework

PR #782, opened 2021-04-01
Address memory leaks

manodasanW:
The changes in this PR addresses the memory leaks observed in the samples from #781.  It is a variant of the changes attempted initially in #673 , but addresses some of the concerns about finalizers on the RCW handling release by moving that to the IObjectReference itself which owns the lifetime.  In addition, it makes sure in aggregated scenarios, any QIs don't increment the COM ref count maintained by the CLR and that for non aggregated scenarios XAML is informed of those references.

From testing, these changes seem to demonstrate an improvement and no leaks are observed in the samples in #781

### DIFF
--- a/src/WinRT.Runtime/ApiCompatBaseline.net5.0.txt
+++ b/src/WinRT.Runtime/ApiCompatBaseline.net5.0.txt
@@ -1,2 +1,4 @@
+Compat issues with assembly WinRT.Runtime:
 MembersMustExist : Member 'public WinRT.MarshalString.HStringHeader WinRT.MarshalString.HStringHeader WinRT.MarshalString._header' does not exist in the implementation but it does exist in the contract.
 TypesMustExist : Type 'WinRT.MarshalString.HStringHeader' does not exist in the implementation but it does exist in the contract.
+Total Issues: 2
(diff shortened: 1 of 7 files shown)


AaronRobinsonMSFT:
First pass looks really good. I will check it out again tomorrow. Two things I would like to see.

1) A test run against some of the WinUI lifetime tests.
2) The performance impact of some of these changes. This is now doing a lot more QI work - which the original .NET Native didn't need to do. We may be able to be clever/smart once we confirm it is correct.

Next step is to get the final .NET fix in as well. We should try and validate that leak scenario still exists. I believe the easiest way is to use the `WeakReference<T>` across the native/managed boundary.

Scottj1s (approved):

BenJKuhn:
(YAY!)

manodasanW:
The 26 WinUI lifetime tests now all pass.

AaronRobinsonMSFT (approved):

AaronRobinsonMSFT:
>I believe the easiest way is to use the WeakReference<T> across the native/managed boundary.

@manodasanW Any luck trying to use that?

BenJKuhn (approved):

manodasanW:
@AaronRobinsonMSFT that is on my list to try out and will get back to you once I do.

PR #809, opened 2021-04-19
Restored AttributeUsage projection

Scottj1s:
Restored the AttributeUsage projection

### DIFF
--- a/src/cswinrt/code_writers.h
+++ b/src/cswinrt/code_writers.h
@@ -1427,7 +1427,9 @@ remove => %.% -= value;
                 }
             }
             // Skip metadata attributes
-            if (attribute_namespace == "Windows.Foundation.Metadata" && attribute_name != "DefaultOverload" && attribute_name != "Overload") continue;
+            if (attribute_namespace == "Windows.Foundation.Metadata" && 
+                attribute_name != "DefaultOverload" && attribute_name != "Overload" && attribute_name != "AttributeUsage") 
+                continue;
             attributes[attribute_full] = std::move(params);
         }
         if (auto&& usage = attributes.find("AttributeUsage"); usage != attributes.end())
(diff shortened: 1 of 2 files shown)


src/cswinrt/code_writers.h
```diff
@@ -1427,7 +1427,9 @@ remove => %.% -= value;
                 }
             }
             // Skip metadata attributes
-            if (attribute_namespace == "Windows.Foundation.Metadata" && attribute_name != "DefaultOverload" && attribute_name != "Overload") continue;
+            if (attribute_namespace == "Windows.Foundation.Metadata" &&
```
j0shuams, line 1430: What do you think of moving the check to a predicate and calling that the function from the `if ()` ?

src/cswinrt/code_writers.h
```diff
@@ -1427,7 +1427,9 @@ remove => %.% -= value;
                 }
             }
             // Skip metadata attributes
-            if (attribute_namespace == "Windows.Foundation.Metadata" && attribute_name != "DefaultOverload" && attribute_name != "Overload") continue;
+            if (attribute_namespace == "Windows.Foundation.Metadata" && 
+                attribute_name != "DefaultOverload" && attribute_name != "Overload" && attribute_name != "AttributeUsage") 
+                continue;
             attributes[attribute_full] = std::move(params);
         }
         if (auto&& usage = attributes.find("AttributeUsage"); usage != attributes.end())
```
manodasanW, line 1435: Any concerns with the below not finding anything given that would probably get skipped:

            bool allow_multiple = attributes.find("Windows.Foundation.Metadata.AllowMultiple") != attributes.end();

Scottj1s: yup - fix coming for that

src/cswinrt/code_writers.h
Scottj1s: yes, that kind of factoring could be helpful if called multiple times.  here, probably just adds unnecessary complexity.

manodasanW (approved):

PR #828, opened 2021-04-28
Add caching for type lookup of runtime class name, to address exception storm perf issue

Scottj1s:
This change improves perf by caching mapping from runtime class to managed type, especially in scenarios where IInspectable::GetRuntimeClassName returns invalid strings like "TrackerCollection<T>".

#fixes 486

### DIFF
--- a/src/WinRT.Runtime/TypeNameSupport.cs
+++ b/src/WinRT.Runtime/TypeNameSupport.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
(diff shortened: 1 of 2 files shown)


Scottj1s:
@azchohfi - FYI, this should help your debug experience

src/WinRT.Runtime/TypeNameSupport.cs
```diff
@@ -25,12 +26,34 @@ enum TypeNameGenerationFlags
     static class TypeNameSupport
     {
         private static List<Assembly> projectionAssemblies = new List<Assembly>();
+        private static ConcurrentDictionary<string, Type> typeNameCache = new ConcurrentDictionary<string, Type>() { ["TrackerCollection<T>"] = null };
```
manodasanW, line 29: Any concern this might keep all encountered types alive and prevent it from being unloaded?

Scottj1s: I thought we'd discuss that last year, and decided that caching types wasn't a concern

manodasanW: Right I think we discussed that when we starting doing something similar with TypeObjectRefFieldCache / TypeObjectRefFuncCache.  We eventually did stop doing that for the net5 projection but that was because we were able to do further optimizations that made it not need a cache and not due to type unloadability concerns.  So this might be fine.  I believe we said then types are rarely unloaded or something.

src/WinRT.Runtime/TypeNameSupport.cs
```diff
@@ -25,12 +26,34 @@ enum TypeNameGenerationFlags
...
         {
             projectionAssemblies.Add(assembly);
         }
 
+        /// <summary>
+        /// Parses and loads the given type name, if not found in the cache.
+        /// </summary>
+        /// <param name="runtimeClassName">The runtime class name to attempt to parse.</param>
+        /// <returns>The type, if found.  Null otherwise</returns>
+        public static Type FindTypeByNameCached(string runtimeClassName)
```
manodasanW, line 41: Should this also be used for https://github.com/microsoft/CsWinRT/blob/ad6adcb997b40cea649cf4156451316e714bc3ea/src/WinRT.Runtime/Projections/Type.cs#L87 or FindTypeByName itself be updated to consult the cache as an internal implementation of the API?

Scottj1s: I skipped that case, as I think we want the exception to propagate back to native code as an HRESULT

manodasanW (approved):

PR #830, opened 2021-04-29
 Project COM interop APIs 

angelazhangmsft:
- Define IInspectable-based interop APIs in IDL and generate projections. 
- Define helper classes implementing the generated interfaces
- Add unit tests for the interop classes

This work will provide definitions for several COM interop interfaces through the Windows SDK projection, including:
```WinRT.Interop.IWindowNative, WinRT.Interop.IInitializeWithWindow, Windows.Storage.Streams.IBufferByteAccess``` and wrappers for all WinRT interop interfaces included in the Universal API Contract (e.g. ```Windows.Security.Credentials.UI.UserConsentVerifierInterop, Windows.UI.ApplicationSettings.AccountsSettingsPaneInterop```)

### CODE REVIEW

File: src/Tests/Interop/Interop.idl [hunk 1]
```diff
@@ -0,0 +1,154 @@
+// Modern IDL 3.0: https://docs.microsoft.com/en-us/uwp/midl-3/intro
+
+namespace Windows.Interop
+{
+    // accountssettingspaneinterop.idl
+   [uuid(D3EE12AD-3865-4362-9746-B75A682DF0E6)]
+   interface IAccountsSettingsPaneInterop 
+   {
+       Object GetForWindow(
+           UInt64 appWindow,
```
jeffstall, line 10: I think all of these should be "IntPtr hwnd" instead of "UInt64 appWindow"

src/Projections/Windows/ComInteropHelpers.cs
```diff
@@ -0,0 +1,227 @@
...
+using WinRT;
+using Windows.UI.Input;
+using Windows.UI.Input.Core;
+using Windows.UI.Input.Spatial;
+using Windows.Media;
+using Windows.Security.Authentication.Web.Core;
+using Windows.Security.Credentials;
+using System.Runtime.InteropServices;
+
+namespace Windows.Interop
```
jeffstall, line 19: Since none of this is in the system itself, I think this is "Microsoft.Windows.Interop. "Windows" is reserved namespace only for system components.

src/Tests/Interop/Interop.vcxproj
```diff
@@ -0,0 +1,158 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project DefaultTargets="Build" ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="..\..\packages\Microsoft.Windows.CppWinRT.2.0.210403.2\build\native\Microsoft.Windows.CppWinRT.props" Condition="Exists('..\..\packages\Microsoft.Windows.CppWinRT.2.0.210403.2\build\native\Microsoft.Windows.CppWinRT.props')" />
```
jeffstall, line 3: Why is C++ being used for a Cs/WinRT project?

src/Tests/Interop/Interop.idl
```diff
@@ -0,0 +1,154 @@
+// Modern IDL 3.0: https://docs.microsoft.com/en-us/uwp/midl-3/intro
+
+namespace Windows.Interop
```
kennykerr, line 3: These interop interfaces are already defined in winmd format by the [win32metadata](https://github.com/microsoft/win32metadata/) project. You can find the [nuget package with the winmd over here](https://www.nuget.org/packages/Microsoft.Windows.SDK.Win32Metadata/). The C#/WinRT project could simply be updated to parse this additional winmd file and generate these definitions/wrappers automatically as the [Windows crate for Rust](https://github.com/microsoft/windows-rs/) already does. The whole point of language projections is to get out of the business of writing such language bindings by hand. :wink:

Note also that the approach used here of using MIDL3 will only work for those interop interfaces that derive from IInspectable. The win32 metadata supports everything. @sotteson1 @mikebattista

src/Tests/Interop/Interop.idl
Scottj1s: IntPtr is not an IDL type - it's a C# type that requires a lossless IDL representation

src/Projections/Windows/ComInteropHelpers.cs
Scottj1s: Yes, given that interops are tightly coupled to WinRT classes, they could be colocated under a relative interop namespace.  But Kenny's comment about using the Microsoft.Windows.SDK.Win32Metadata is a great idea and sidesteps the namespace question.  That coupled with the extension method suggestion (which is great), should make namespaces mostly irrelevant.

src/Tests/Interop/Interop.idl
kennykerr: This speaks to the challenges of trying to express non-WinRT APIs in MIDL3. `IntPtr` isn't available, but `UInt64` is not the right type either. This is an `HWND`, which is either 64 bits or 32 bits depending on architecture. You would need to define distinct x86 and x64 variants of this interface in MIDL3 or just use [win32metadata](https://github.com/microsoft/win32metadata/).

src/Tests/Interop/Interop.idl
Scottj1s: @angelazhangmsft, sigh, this is my oversight.  I meant to follow up on this as a way to model the interops.  Kenny is right - we should absolutely use the Microsoft.Windows.SDK.Win32Metadata package, with appropriate CsWinRTFilters settings, to project the interops we can (and model the remainder in C#, if necessary).

kennykerr: Note that ILDASM produces IL rather than IDL. [ILSpy](https://github.com/icsharpcode/ILSpy/releases) is also rather handy. 

image

kennykerr: > how does win32metadata handle IUnknown-based interops

The `IAccountsSettingsPaneInterop` example above is an interop interface that derives from `IInspectable`, but win32metadata can also define non-WinRT interfaces. Here's good old `IMemoryBufferByteAccess`, which derives directly from `IUnknown`.

image

Scottj1s:
> * Naming convention (Should the interop APIs live under a "Windows.Interop" namespace?)
using the win32metadata package sidesteps this

> * Projections of interfaces and helper classes are both public. We could consider making the projected interfaces internal to avoid confusion.
We would need to change cswinrt.exe to support 'internal' projections, but I think this is a good idea.  Aside from discoverability (which extension method wrappers would solve), there are too many sharp edges to use the projection directly - ensuring the RIIDs match return type casts, and even knowing what those return types are, etc.

> * Only supporting interop interfaces with UniversalApiContract (e.g. not Desktop extension APIs like IProtectionPolicyManagerInterop)
this makes sense to me - we can always generated extension interop assemblies, if necessary.

> * Versioning by Windows version: could add contract version attributes to IDL definitions, which will project SupportedOSPlatform attribute
Note that the win32metadata package includes a definition for Windows.Win32.Interop.SupportedOSPlatformAttribute, which we can of course project to System.Runtime.Versioning.SupportedOSPlatformAttribute as we already do for ContractVersion.  That's another reason to use the win32metadata package - to avoid special solutions to projection requirements like these.

src/Tests/TestComponentCSharp/TestComponentCSharp.idl
```diff
@@ -356,11 +356,11 @@ namespace TestComponentCSharp
     }
     
     // IInspectable-based interop interface
```
Scottj1s, line 358: cruft (remove this)

src/Tests/Interop/Interop.idl
Scottj1s: Yes, IL (not IDL).  ILSpy is terrific, but I'm seeing exceptions trying to display definitions for interfaces (like IMemoryBufferByteAccess):
  **System.ArgumentException: An item with the same key has already been added.)**

so suggested ildasm as a search technique.

src/Tests/Interop/Interop.idl
Scottj1s: I suspected the HWND modeling was incorrect - thanks

src/Tests/Interop/Interop.idl
kennykerr: @scottj1s are you using version 7? https://github.com/icsharpcode/ILSpy/releases

src/Tests/Interop/Interop.idl
jeffstall: The Interop interfaces are C++ interfaces, often deriving from IUnknown, not MIDL interfaces deriving from IInspectable. IntPtr is the correct data type because the real C++ interfaces use HWND, which is pointer size.

src/Tests/Interop/Interop.idl
Scottj1s: I am now ;-)  yes, problem solved

angelazhangmsft:
Thank you everyone for all the helpful feedback! I am closing this PR for now and we will investigate the win32metadata approach, and adding support in C#/WinRT for IUnknown interfaces. Will open a separate PR using this feedback.

angelazhangmsft:
Re-opening this PR, addressing the prior feedback and made the following updates:

- Changed the namespaces to adding the interop-specific classes to the same namespace as the main object. For example: ```Windows.UI.ApplicationSettings.AccountSettingsPane
Windows.UI.ApplicationSettings.AccountsSettingsPaneInterop.GetForWindow(hwnd)```
- Decided to continue defining interop interfaces in IDL, since the win32metadata package does not have winmd file with "windowsruntime" attributed metadata
- Added support in C#/WinRT for internal-only projection of the interop interfaces, so only the wrapper classes (not the interfaces) are exposed to the developer.
- Added custom mapping of WinRT.Interop.HWND to System.IntPtr
- Versioning: Instead of using SupportedOSPlatform/ContractVersion, we are using build-time filtering of types (conditional compilation) to support different versions of the Windows SDK projection.

src/Tests/Interop/Interop.idl
angelazhangmsft: Upon investigation, decided on defining the APIs in IDL instead of using win32metadata, as their metadata isn't "windows runtime" metadata

src/Projections/Windows/ComInteropHelpers.cs
angelazhangmsft: Updates to the PR define the interop classes in the same namespace as the main object, e.g. `Windows.UI.ApplicationSettings.AccountsSettingsPaneInterop`

Scottj1s:
> * Versioning: Instead of using SupportedOSPlatform/ContractVersion, we are using build-time filtering of types (conditional compilation) to support different versions of the Windows SDK projection.

A couple points here - 
1. The developer selects a version of the Windows SDK projection via TFM (e.g., net5.0-windows10.0.18362.0), so later API additions are not visible.  It makes sense for the interop interfaces to follow suit.  The alternative would be confusing - having interop interfaces that are not available on the target platform version.
2. SupportedOSPlatform does not support interfaces, so there's not good way to inform the developer if an interop interface were not available on the target.

Scottj1s:
@Manodasanw, adding you to review as this will be the pattern that we want to implement in Microsoft.Windows.SDK.NET.dll

src/Tests/UnitTest/ComInteropTests.cs
```diff
@@ -0,0 +1,115 @@
...
+using Windows.UI.ApplicationSettings;
+using Windows.UI.Input;
+using Windows.UI.Input.Core;
+using Windows.UI.Input.Spatial;
+using Windows.UI.ViewManagement;
+using Xunit;
+
+namespace UnitTest
+{
+    public class ComInteropTests
```
Scottj1s, line 19: maybe include a comment that many of these tests cannot easily be executed without significant test setup, so cases are either compile time only (skipped, to validate types) or expect a runtime exception.

Scottj1s (dismissed):
Looks good enough now to provide as a template for the Windows SDK projection (which was the ultimate goal)

src/Tests/Interop/Interop.idl
angelazhangmsft: Update: added custom mapping of WinRT.Interop.HWND to System.IntPtr

Scottj1s:
@manodasanW, @angelazhangmsft - automatic support for Windows SDK projection implemented - please review

Scottj1s (commented):
awaiting others' approval

src/cswinrt/Directory.Build.targets
```diff
@@ -1,5 +1,17 @@
 <Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
-  <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.targets', '$(MSBuildThisFileDirectory)../'))" />
+    <PropertyGroup>
+        <MidlRTCopyWinMDToOutputDirectoryDependsOn>$(MidlRTCopyWinMDToOutputDirectoryDependsOn);CopyWinRTInteropWinMD</MidlRTCopyWinMDToOutputDirectoryDependsOn>
+    </PropertyGroup>
+
+    <!--Workaround for MidlRT nupkg expecting project to publish $(RootNamespace).winmd-->
+    <Target Name="CopyWinRTInteropWinMD" Inputs="$(CsWinRTInteropMetadata)" Outputs="$(MidlRTProjectWinMD)">
+        <Copy UseHardlinksIfPossible="$(MidlRTUseHardlinksIfPossible)"
```
manodasanW, line 9: Looks like this is causing a build break?

nuget/Microsoft.Windows.CsWinRT.targets
```diff
@@ -90,14 +90,20 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       <CsWinRTFilters Condition="'$(CsWinRTFilters)' == ''">
 @(CsWinRTExcludeItems->'-exclude %(Identity)', '&#x0d;&#x0a;')
 @(CsWinRTIncludeItems->'-include %(Identity)', '&#x0d;&#x0a;')
-      </CsWinRTFilters> 
+      </CsWinRTFilters>
+      <CsWinRTInteropMetadata Condition="'$(CsWinRTInteropMetadata)' == ''">$([MSBuild]::NormalizeDirectory('$(MSBuildThisFileDirectory)', '..\metadata\WinRT.Interop.winmd'))</CsWinRTInteropMetadata>
+      <CsWinRTIncludeWinRTInterop Condition="$(CsWinRTFilters.Contains('-include Windows&#x0d;&#x0a;'))">
```
manodasanW, line 95: I assume we are trying to look for a new line there?

src/cswinrt/Directory.Build.props
```diff
@@ -0,0 +1,16 @@
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003" >
+
+    <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.props', '$(MSBuildThisFileDirectory)../'))" />
+
+    <PropertyGroup>
+        <!--Avoid UniversalApiContract type not found error-->
+        <CppWinRTMergeNoValidate>true</CppWinRTMergeNoValidate>
```
manodasanW, line 7: for context, was this needed because of the cppwinrt update or something else?

src/cswinrt/Directory.Build.targets
Scottj1s: yes, path was wrong

nuget/Microsoft.Windows.CsWinRT.targets
Scottj1s: yes, to ensure we don't false-positive match on subnamespaces

src/cswinrt/Directory.Build.props
Scottj1s: originally, which were were going to use UniversalApiContract for SupportedOSPlatform support on interop.  in the end, we went with conditional compilation instead - will remove this

nuget/Microsoft.Windows.CsWinRT.nuspec
```diff
@@ -26,7 +26,7 @@
     <file src="Microsoft.Windows.CsWinRT.Prerelease*.targets" target="build"/>
     <file src="Microsoft.Windows.CsWinRT.Authoring.Transitive.targets" target="build"/>
     <file src="Microsoft.Windows.CsWinRT.Authoring.Transitive.targets" target="buildTransitive"/>
-    <file src="$cswinrt_exe$\..\WinRT.Interop.winmd" target="metadata"/>
+    <file src="$cswinrt_exe$\..\..\obj\merged\WinRT.Interop.winmd" target="metadata"/>
```
manodasanW, line 29: should we instead be picking it up for something like bin rather than obj?

src/Tests/TestComponentCSharp/Directory.Build.props
```diff
@@ -1,10 +1,15 @@
...
-  <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.props', '$(MSBuildThisFileDirectory)../'))" />
+    <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.props', '$(MSBuildThisFileDirectory)../'))" />
 
-  <!--Only need the WinUI winmd files for compiling idl, so stub out Xaml targets-->
-  <Target Name="MarkupCompilePass1" />
-  <Target Name="MarkupCompilePass2" />
-  <Target Name="Prep_ComputeProcessXamlFiles" />
+    <PropertyGroup>
+        <!--Avoid UniversalApiContract type not found error-->
+        <CppWinRTMergeNoValidate>true</CppWinRTMergeNoValidate>
```
manodasanW, line 7: Should this go away too?

Scottj1s: no, this has to remain - we use UniversalApiContract in our tests for SupportedOSPlatform

nuget/Microsoft.Windows.CsWinRT.nuspec
Scottj1s: that's how it was to begin with - it's really the same file

manodasanW (approved):

PR #833, opened 2021-04-30
Fix InvalidCastExceptions from RCW reuse

### LINKED ISSUE #762: Random `System.InvalidCastException` thrown when using `Windows.Graphics.Imaging.BitmapDecoder` and `OcrEngine`
**Describe the bug**
`System.Private.CoreLib.dll` throws `System.InvalidCastException` randomly when using WinRT API  `Windows.Graphics.Imaging.BitmapDecoder` and `Windows.Media.Ocr.OcrEngine`. In average, 90% of the call will success. 10% calls to `BitmapDecoder.CreateAsync(IRandomAccessStream)` or `BitmapDecoder.GetSoftwareBitmapAsync` or `OcrEngine.RecognizeAsync` throws exception.

Typical exception message:
```
Exception thrown: 'System.InvalidCastException' in System.Private.CoreLib.dll
Unable to cast object of type 'Windows.Foundation.AsyncOperationWithProgressCompletedHandler`2[Windows.Storage.Streams.IBuffer,System.UInt32]' to type 'Windows.Graphics.Imaging.BitmapDecoder'.
```

**To Reproduce**

Download the sample project and run. Click the "Test 100 times" button and wait until the sum of three numbers is 100. If the number under the "Information" text block is 100, click the button again. 
[WinRTCallUnstable.zip](https://github.com/microsoft/CsWinRT/files/6100253/WinRTCallUnstable.zip)

When the problem is successfully reproduced, the error count > 0. 
image

**Expected behavior**
`System.InvalidCastException` should not be thrown.

**Version Info**

I don't use CsWinRT nuget directly, but my application call WinRT APIs through `WinRT.Runtime.dll` which is generated by CsWinRT. The target framework is `net5.0-windows10.0.18362.0` .

.NET SDK (reflecting any global.json):
 Version:   5.0.200
 Commit:    70b3e65d53

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.19042
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\5.0.200\

Host (useful for support):
  Version: 5.0.3
  Commit:  c636bbdc8a

Visual Studio 2019 16.9.0

**Additional context**
This problem is probably related to threading or garbage collection. I added `try ... catch` to retry WinRT API calls in my code as workaround. I don't know whether my workaround is safe, because I'm not sure if this bug will cause memory corruption.

manodasanW:
We were seeing InvalidCastExceptions because after the native pointer is released, we are still getting old RCW back from the .NET cache if that same pointer is encountered again due to .NET thinks it is still alive.  This is due to the cache cleanup happens after finalization of the RCW which isn't at the same time as when the object reference in the RCW releases the final ref on the pointer.  In .NET 6, there would be a new API for us to use to clear the cache, but until then we are putting in a mitigation for .NET 5.

IObjectReferences which are associated with an RCW are not immediately finalized but their finalization is deferred twice in an attempt to get its finalization to happen in gen 2 which means the RCW finalization and cache cleanup in gen 2 will be done by that time.

Fixes #762

### DIFF
--- a/src/Tests/UnitTest/TestComponentCSharp_Tests.cs
+++ b/src/Tests/UnitTest/TestComponentCSharp_Tests.cs
@@ -2222,23 +2222,31 @@ namespace UnitTest
             static void TestImports()
             {
                 var (initializeWithWindow, windowNative) = MakeImports();
-                
-                GC.Collect();
-                GC.WaitForPendingFinalizers();
+
+                GCCollect();
 
                 var hwnd = new IntPtr(0x12345678);
                 initializeWithWindow.Initialize(hwnd);
                 Assert.Equal(windowNative.WindowHandle, hwnd);
             }
 
+            static void GCCollect()
+            {
+                // Require multiple GC collects due to
+                // the final release is not done immediately.
+                for(int idx = 0; idx < 3; idx++)
+                {
+                    GC.Collect();
+                    GC.WaitForPendingFinalizers();
+                }
+            }
+
             TestObject();
-            GC.Collect();
-            GC.WaitForPendingFinalizers();
+            GCCollect();
             Assert.Equal(0, ComImports.NumObjects);
 
             TestImports();
-            GC.Collect();
-            GC.WaitForPendingFinalizers();
+            GCCollect();
             Assert.Equal(0, ComImports.NumObjects);
         }
 
--- a/src/Tests/UnitTest/TestComponentCSharp_Tests.cs
+++ b/src/Tests/UnitTest/TestComponentCSharp_Tests.cs
@@ -2201,6 +2201,14 @@ namespace UnitTest
         [Fact]
         unsafe public void TestComImports()
         {
+            TestObject();
+            GCCollect();
+            Assert.Equal(0, ComImports.NumObjects);
+
+            TestImports();
+            GCCollect();
+            Assert.Equal(0, ComImports.NumObjects);
+
             static Object MakeObject()
             {
                 Assert.Equal(0, ComImports.NumObjects);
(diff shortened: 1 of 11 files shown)


src/Tests/UnitTest/TestComponentCSharp_Tests.cs
```diff
@@ -2221,24 +2221,32 @@ static Object MakeObject()
...
+                var (initializeWithWindow, windowNative) = MakeImports();
+
+                GCCollect();
 
                 var hwnd = new IntPtr(0x12345678);
                 initializeWithWindow.Initialize(hwnd);
                 Assert.Equal(windowNative.WindowHandle, hwnd);
             }
 
+            static void GCCollect()
```
AaronRobinsonMSFT, line 2233: Style here is still evolving but I would recommend putting all local functions at the end of the function. Placing them in the middle makes reading the function logical flow cumbersome.

src/WinRT.Runtime/ObjectReference.cs
```diff
@@ -17,7 +17,8 @@ public abstract class IObjectReference : IDisposable
         protected bool disposed;
         private readonly IntPtr _thisPtr;
         private object _disposedLock = new object();
-        private IntPtr _referenceTrackerPtr;
+        private IntPtr _referenceTrackerPtr;
+        private int RCWCleanupCounter = 2;
```
AaronRobinsonMSFT, line 21: Nit on style: ~~`RCWCleanupCounter`~~ -> `_rcwCleanupCounter`

src/WinRT.Runtime/ObjectReference.cs
```diff
@@ -74,6 +75,8 @@ internal unsafe IReferenceTrackerVftbl ReferenceTracker
             }
         }
 
+        internal bool CleanupRCW { get; set; }
```
AaronRobinsonMSFT, line 78: I would avoid creation of a boolean field and instead just use the counter.

For example:
```csharp
internal void MarkCleanupRCW() => _rcwCleanupCounter = 2;
private bool CleanupRCW { get => _rcwCleanupCounter == 0; }
```

Scottj1s (approved):

AaronRobinsonMSFT (approved):

Issue #612, opened 2020-11-30
0x8001010E (RPC_E_WRONG_THREAD) on Windows API call


aemarco:
as recommended in https://github.com/dotnet/runtime/issues/45071#issuecomment-735587045 -->

### Description

Windows Api "SetImageStreamAsync" broke with upgrade to .net5
It now throws an Exception --> 0x8001010E (RPC_E_WRONG_THREAD) after porting from .netcore3.1 to net5.0.

### Reproduce the problem

[https://github.com/aemarco/LockScreenTest.git](url)
contains a sample project with nothing but the failing api call in App.xaml.cs.
It should be as easy as checking the repo out and run.

### Working condition (netcore3.1)

By changing back to .netCore 3.1 as below, the lockscreen will get successfully set 

`<Project Sdk="Microsoft.NET.Sdk.WindowsDesktop">`
`<TargetFramework>netcoreapp3.1</TargetFramework>`
`<PackageReference Include="Microsoft.Windows.SDK.Contracts" Version="10.0.19041.1" />`

### Expectations

By specifying the TargetFramework with the windows version, api´s supposed to be functional. 

### Questions (done.... found concept reading stuff in https://github.com/dotnet/designs/tree/main/accepted/2020/minimum-os-version

Maybe you have a hint were to read up about that "define version number to access windows api" concept.
https://docs.microsoft.com/en-us/dotnet/standard/frameworks
explains that till the -windows, but not the part with the version number and api projections(which i learned through .netconf).

Should it not be possible to target "net5.0-windows", guard against specific api calls, dropping the "feature" when not on supported version, but call them if version >= needed ?

AaronRobinsonMSFT:
@scottj1s @manodasanW I directed @aemarco to this repo since C#/WinRT should be handling thread affinity for WinRT scenarios. Please take a first pass at it and let me know if you feel it is actually an issue within the runtime.

azchohfi:
Any updates on this. Its the only bug keeping all the WCT test suite from passing.

manodasanW:
No updates yet, this is next on my list to start looking at again.

mattleibow:
Just to note, I also saw this with `StorageFile`

getrou:
I am also seeing this issue in Win2D's CanvasBitmap.LoadAsync() when running a Desktop Project Reunion app.

manodasanW:
[...]

PR #836, opened 2021-05-05
Add free threaded marshaling to all managed objects

manodasanW:
With the exception of the ones that implement IMarshal, we should implement the free threaded marshaler on the CCW of all managed objects.  The free threaded marshaler is stored as a ThreadStatic to avoid retrieving it for each IMarshal call.

Contributes to #612

### DIFF
--- a/src/WinRT.Runtime/Interop/IMarshal.cs
+++ b/src/WinRT.Runtime/Interop/IMarshal.cs
@@ -1,0 +1,288 @@
+using System;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using WinRT;
+using WinRT.Interop;
+
+namespace WinRT.Interop
+{
+    internal enum MSHCTX : int { Local = 0, NoSharedMem = 1, DifferentMachine = 2, InProc = 3, CrossCtx = 4 }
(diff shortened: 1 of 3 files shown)


src/WinRT.Runtime/Interop/IMarshal.cs
```diff
@@ -0,0 +1,288 @@
...
+                    MarshalInterface_2 = &Do_Abi_MarshalInterface_2,
+                    UnmarshalInterface_3 = &Do_Abi_UnmarshalInterface_3,
+                    ReleaseMarshalData_4 = &Do_Abi_ReleaseMarshalData_4,
+                    DisconnectObject_5 = &Do_Abi_DisconnectObject_5
+                };
+#endif
+            }
+
+            // This object handles IMarshal calls for us for most scenarios.
+            [ThreadStatic]
```
Scottj1s, line 103: why ThreadStatic?

manodasanW: I wasn't sure whether the free threaded marshaler instance we get was thread safe itself or not and I didn't want to retrieve it each time for each call on a function.  So I followed the pattern we use for the other buffer marshaler in CsWinRT (RoGetBufferMarshaler) where we use ThreadStatic.

Scottj1s (approved):

PR #837, opened 2021-05-05
moved gc pressure support from netstandard (mistake) to net5

Scottj1s:


src/cswinrt/code_writers.h
```diff
@@ -5196,8 +5187,7 @@ _defaultLazy = new Lazy<%>(() => ifc);
 _lazyInterfaces = new Dictionary<Type, object>()
 {%
 };
-%}
-%
+}
```
manodasanW, line 5190: Do we not also want this for .net standard?

Scottj1s: At first, I thought not, but I agree - it should be supported there too

src/cswinrt/code_writers.h
```diff
@@ -5383,6 +5393,22 @@ private % AsInternal(InterfaceTag<%> _) => _default;
             default_interface_name,
             default_interface_name,
             bind<write_lazy_interface_initialization>(type),
+            [&](writer& w)
```
manodasanW, line 5396: Do we care about doing the same in the default constructor (the one a C# caller would new up)?  And how does GCPressureAttribute apply to derived types?  Is it possibly for the base type to just have it or will any derived types also have that attribute?  If not, maybe we want it in the protected derived constructor too?

PR #838, opened 2021-05-05
cache GetHelperType lookup

### LINKED ISSUE #785: Cache results in WinRT.TypeExtensions.FindHelperType to avoid calling Type.GetType().
Based on analysis, if many objects are created this is a hot-path.  One of scenarios I was looking at (which was heavy on WinUI object instantiation), this method accounted for 5.6s out of 31s scenario.  Adding caching to this method could improve perf.

Scottj1s:
fixes #785 

Tested with multiple runs of TestStrings (100 iterations) and saw about a 10% improvement

### DIFF
--- a/src/WinRT.Runtime/TypeExtensions.cs
+++ b/src/WinRT.Runtime/TypeExtensions.cs
@@ -1,35 +1,38 @@
 ﻿using System;
-using System.Collections.Generic;
-using System.Diagnostics;
+using System.Collections.Concurrent;
 using System.Reflection;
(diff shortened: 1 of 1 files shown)


manodasanW (approved):

jkoritzinsky:
@Scottj1s can we also update the code-gen to use attributes to tie a type to it's ABI/ABI.Impl type for new projections and use this cached Type.GetType lookup for older projections? Using an attribute will also help with linkability.

PR #849, opened 2021-05-20
[Perf] Add IID Lookup Optimization

### LINKED ISSUE #686: Performance: Improve performance of GUID lookup for types
guid calculation for collections is expensive.  a source generator could be used to analyze all generic constructions, pre-generate their guids, and hard code them in a lookup table.

j0shuams:
This PR adds guid optimization to all assemblies that use cswinrt. Any time an interop assembly would call for a guid (GetIID, say), we replace that call with a look up in a static map from types to guids. 

Using .targets, we run the patcher on a target assembly and pass along the references that assembly has, and copy out the patched assembly mid-build so that the one copied to the bin dir must be the patched dll. 

The optimizer comes with an "opt out" in case we need to turn it off for any new issues that might arise in it. The opt out is "CsWinRTIIDOptimizerOptOut" and just needs to be set to "true" in the target assembly's project file. 

For testing we know that a set of our unittests cover the codepaths used by the optimizer. The async tests and the storage file/folder tests cover the nontrivial code paths. We have also tested that the symbols files written by the optimizer for the patched dll are usable. 

There is an addition to the TestComponentCSharp that arose from a bug in a tool used by the optimizer. The optimizer uses Mono.Cecil nuget package, and on version 0.11.3 of Cecil there was a bug writing long class names to pdb files. I added a new runtime class with a long, nonsensical name, and a unit test that would call for an event on this class. This reproduces the error we originally saw with 0.11.3, and verifies that the Cecil fix in 0.11.4 works for us. 

Also, there was a version mis-match in the TestWinRT repo that caused local build issues. This was fixed by a PR on TestWinRT, and we reflect that by updating the commit used to checkout TestWinRT in our local .cmd. 

Closes #686


nuget/Microsoft.Windows.CsWinRT.targets
```diff
@@ -147,6 +147,44 @@ $(CsWinRTFilters)
...
+           
+         If we are not in the CsWinRT repo, we can use the package paths we know -->
+
+    <PropertyGroup Condition="'$(CsWinRTInRepo)' == 'true'">
+      <!-- We use _DirectoryBuildTargetsBasePath as it is the only way I found to get "c:/.../cswinrt/src" -->
+      <CsWinRTGuidPatchExePath>$(_DirectoryBuildTargetsBasePath)\$(CsWinRTGuidPatchRepoPath)</CsWinRTGuidPatchExePath>
+      <CsWinRTGuidPatchRuntimeDll>$(_DirectoryBuildTargetsBasePath)\$(CsWinRTGuidPatchRuntimePath)</CsWinRTGuidPatchRuntimeDll> 
+    </PropertyGroup>
+
+    <PropertyGroup Condition="'$(CsWinRTInRepo)' != 'true'">
```
manodasanW, line 167: We probably don't want to do cswinrt repo checks in a target we ship for public consumption if we don't have to.  Another way to handle this is set value it if is not already set and in the cswinrt repo we can set it ahead of time in directory.props.  so something like:

      <CsWinRTGuidPatchExePath Condition="'$(CsWinRTGuidPatchExePath)'==''">$(CsWinRTPath)build\tools\</CsWinRTGuidPatchExePath>

nuget/Microsoft.Windows.CsWinRT.targets
```diff
@@ -147,6 +147,44 @@ $(CsWinRTFilters)
...
+    <PropertyGroup>
+      <CsWinRTGuidPatchInput Condition="'$(CsWinRTGuidPatchInput)' == ''">@(BuiltProjectOutputGroupKeyOutput->'%(Identity)')</CsWinRTGuidPatchInput>
+    </PropertyGroup>
+
+    <Exec Command="$(CsWinRTGuidPatchExePath)GuidPatch.exe $(CsWinRTGuidPatchInput) $(CsWinRTGuidPatchRuntimeDll)" ConsoleToMsBuild="true">
+      <Output TaskParameter="ConsoleOutput" PropertyName="CsWinRTGuidPatchOutput" />
+    </Exec>
+    
+    <ItemGroup>
+      <PatchedFiles Include="$(_DirectoryBuildTargetsBasePath)\GuidPatcherOutput\*.dll" />
```
manodasanW, line 181: Is this always set even you don't use directory.builds.targets?

src/Perf/GuidPatch/Program.cs
```diff
@@ -0,0 +1,43 @@
...
+using System;
+using System.IO;
+
+namespace GuidPatch
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            Console.WriteLine("Hello World!");
```
manodasanW, line 11: Remove hello world?

src/Tests/AuthoringConsumptionTest/AuthoringConsumptionTest.vcxproj
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="..\..\packages\Microsoft.WinUI.3.0.0-preview3.201113.0\build\native\Microsoft.WinUI.props" Condition="Exists('..\..\packages\Microsoft.WinUI.3.0.0-preview3.201113.0\build\native\Microsoft.WinUI.props')" />
+  <Import Project="..\..\packages\Microsoft.WinUI.3.0.0-zmain.210514.0\build\native\Microsoft.WinUI.props" Condition="Exists('..\..\packages\Microsoft.WinUI.3.0.0-zmain.210514.0\build\native\Microsoft.WinUI.props')" />
```
manodasanW, line 3: Ideally we do not want to use internal versions to allow others externally to be able to build this repo.  Is there any version still on the public feed that we can use instead?  Otherwise we would need to move to reunion.

src/Tests/AuthoringWinUITest/AuthoringWinUITest/App.xaml.cpp
```diff
@@ -21,7 +26,7 @@ using namespace AuthoringWinUITest::implementation;
 App::App()
 {
     InitializeComponent();
-    Suspending({ this, &App::OnSuspending });
+    // Suspending({ this, &App::OnSuspending });
```
manodasanW, line 29: intentional?

src/cswinrt/main.cpp
```diff
@@ -119,6 +119,7 @@ Where <spec> is one or more of:
         int result{};
         writer w;
 
+        /* Special case the usage exceptions to print CLI options */
```
manodasanW, line 122: Is this comment correlated to a change?

src/Perf/GuidPatch/GuidPatch.csproj
```diff
@@ -0,0 +1,14 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net5.0</TargetFramework>
+    <Nullable>enable</Nullable>
+    <Platforms>AnyCPU;x64;x86</Platforms>
```
manodasanW, line 7: Do we need all 3 platforms or is anycpu enough?

nuget/Microsoft.Windows.CsWinRT.targets
```diff
@@ -147,6 +147,44 @@ $(CsWinRTFilters)
     <Copy SourceFiles="$(CsWinRTGeneratedFilesDir)\$(AssemblyName).winmd" DestinationFolder="$(TargetDir)" UseHardlinksIfPossible="false" SkipUnchangedFiles="true" />
   </Target>
 
+  <!-- ====================== -->
+  <!-- Call The GuidPatch tool on the projection .dll -->
+  <Target Name="CsWinRTInvokeGuidPatcher" Condition="'$(CsWinRTEnableGuidPatch)' == 'true'" AfterTargets="Compile" BeforeTargets="Link">
```
manodasanW, line 152: I see we are doing this behind CsWinRTEnableGuidPatch being set.  What is the motivation behind that?  Is there any reason we wouldn't always do it?  Is the idea to make it opt in first, see how it goes, and then make it permanent?

nuget/Microsoft.Windows.CsWinRT.targets
manodasanW, line 181: How is the output folder decided on?

nuget/Microsoft.Windows.CsWinRT.nuspec
```diff
@@ -34,5 +34,7 @@
     <file src="$winrt_host_arm$" target ="runtimes\win-arm\native"/>
     <file src="$winrt_host_arm64$" target ="runtimes\win-arm64\native"/>
     <file src="$winrt_shim$" target ="lib\net5.0\"/>
+    <file src="$guid_patch$" target ="build\tools\"/>
+    <file src="$net5_runtime$" target ="build\tools\"/>
```
manodasanW, line 38: Are we copying winrt.runtime also to to the tools folder?  Can we just use the one from the lib\net5.0 folder?

src/cswinrt/main.cpp
j0shuams: No I was reading the cswinrt main and felt like this comment would be useful. Can remove if you want

nuget/Microsoft.Windows.CsWinRT.targets
```diff
@@ -147,6 +147,44 @@ $(CsWinRTFilters)
     <Copy SourceFiles="$(CsWinRTGeneratedFilesDir)\$(AssemblyName).winmd" DestinationFolder="$(TargetDir)" UseHardlinksIfPossible="false" SkipUnchangedFiles="true" />
   </Target>
 
+  <!-- ====================== -->
+  <!-- Call The GuidPatch tool on the projection .dll -->
```
manodasanW, line 151: We should give some thought to the name.  Do we want to call it GuidPatch or something like GuidOptimizer or GuidLookupOptimizer?

src/cswinrt/main.cpp
manodasanW: You can leave it, was just initially confused as I didn't see a correlated change.

src/Tests/AuthoringWinUITest/AuthoringWinUITest/App.xaml.cpp
j0shuams: I think this API changed in the winui changed, and I didn't find the alternative. Will find a better workaround

nuget/Microsoft.Windows.CsWinRT.targets
j0shuams: Motivated by some of our test projects tried to invoke it when they didn't need to... Let me see about making an opt out path for those projects instead of one that everyone needs to set.

nuget/Microsoft.Windows.CsWinRT.targets
j0shuams: https://github.com/microsoft/CsWinRT/blob/d6f5c04fe4ba0364e74b14b22be3cef35a71be78/src/Perf/GuidPatch/Program.cs#L31

j0shuams: Yeah for now I really should follow my pattern of casing on the repo. Will follow up offline

manodasanW: I meant the parent folder this is created in.

nuget/Microsoft.Windows.CsWinRT.targets
j0shuams: Maybe GuidLookupOptimization?

nuget/Microsoft.Windows.CsWinRT.targets
j0shuams: The Exec task runs and it's "WorkingDirectory is the project's current working directory [(source)](https://docs.microsoft.com/en-us/visualstudio/msbuild/exec-task?view=vs-2019)

We could set that property when we call Exec to something, then use that property when we go to Copy

nuget/Microsoft.Windows.CsWinRT.targets
j0shuams: Went with IIDOptimizer

nuget/Microsoft.Windows.CsWinRT.targets
```diff
@@ -147,6 +147,38 @@ $(CsWinRTFilters)
...
+          Condition="'$(CsWinRTIIDOptimizerOptOut)' != 'true'" 
+          AfterTargets="Compile" 
+          BeforeTargets="Link">
+
+    <!-- We condition on CsWinRTIIDOptimizerPath and CsWinRTIIDOptimizerRuntimeDll in case we are doing a local (CsWinRT repo) build -->
+    <PropertyGroup>
+      <CsWinRTIIDOptimizerExePath Condition="'$(CsWinRTIIDOptimizerPath)' != ''">$(CsWinRTIIDOptimizerPath)</CsWinRTIIDOptimizerExePath>
+      <CsWinRTIIDOptimizerRuntimeDll Condition="'$(CsWinRTIIDOptimizerRuntimePath)' != ''">$(CsWinRTIIDOptimizerRuntimePath)</CsWinRTIIDOptimizerRuntimeDll> 
+
+      <CsWinRTIIDOptimizerExePath Condition="'$(CsWinRTIIDOptimizerPath)' == ''">$(CsWinRTPath)build\tools\GuidPatch\</CsWinRTIIDOptimizerExePath>
```
Scottj1s, line 161: this seems overly complicated. why not just: 
<CsWinRTIIDOptimizerPath Condition="'$(CsWinRTIIDOptimizerPath)' == ''">$(CsWinRTPath)build\tools\GuidPatch\</CsWinRTIIDOptimizerPath>
in other words, combine the override with the actual property to use.  same for runtime dll.

src/Projections/Benchmark/Benchmark.csproj
```diff
@@ -5,6 +5,10 @@
     <Platforms>x64;x86</Platforms>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
```
Scottj1s, line 8: does this not support .NET 5?

src/Tests/AuthoringTest/AuthoringTest.csproj
```diff
@@ -7,6 +7,7 @@
     <!-- Enable to diagnose generation issues -->
     <!-- <CsWinRTEnableLogging>true</CsWinRTEnableLogging> -->
     <!-- <CsWinRTKeepGeneratedSources>true</CsWinRTKeepGeneratedSources> -->
+    <CsWinRTIIDOptimizerOptOut>true</CsWinRTIIDOptimizerOptOut>
```
Scottj1s, line 10: why are we opting out?

src/Tests/AuthoringWinUITest/AuthoringWinUITest (Package)/build/Microsoft.WinUI.AppX.targets
```diff
@@ -35,6 +35,7 @@
     <!-- Replaces AppxManifest WinRT registrations for Microsoft.UI.Xaml.dll, post-GenerateAppxManifest. -->
     <!-- Target runs before manifest validation occurs. Requires output file to support incremental builds. -->
     <Target
```
Scottj1s, line 37: this target is a remnant from standalone WinUI 3 and should be removed.

src/Tests/AuthoringWinUITest/AuthoringWinUITest (Package)/build/Microsoft.WinUI.AppX.targets
```diff
@@ -187,6 +205,22 @@
       </ReferenceCopyLocalPaths>
     </ItemGroup>
   </Target>
+  
+  <!-- The CreateWapProjPackageFiles task gets confused by MUI files with the same name but different sub-directory paths,
+       stripping out all but one of them. So we need to add their entries manually. -->
+  <Target Name="_IncludeMuiFilesInWapProj"
```
Scottj1s, line 211: I doubt much of this is needed any more with Reunion - would be best to remove it

src/Tests/UnitTest/UnitTest.csproj
```diff
@@ -9,6 +9,7 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <SimulateCsWinRTNugetReference>true</SimulateCsWinRTNugetReference>
     <CsWinRTEnabled>false</CsWinRTEnabled>
+    <CsWinRTIIDOptimizerOptOut>true</CsWinRTIIDOptimizerOptOut>
```
Scottj1s, line 12: here too - seems like we would want to test this and not opt out

src/Tests/AuthoringTest/AuthoringTest.csproj
j0shuams: There is an invalid cast when using the patcher on the AuthoringTest, I am investigating. I think it may be related to the [invalid cast on ICustomProperty](https://github.com/microsoft/CsWinRT/issues/856)

src/Projections/Benchmark/Benchmark.csproj
j0shuams: Right we only use the patcher on net5 assemblies

manodasanW:
Should we and can we run the IID optimizer on WinRT.Runtime?

Scottj1s:
Per standup, can we centralize assignment of CsWinRTIIDOptimizerOptOut to 'true' if netstandard 2, in common targets file?

j0shuams:
Per discussion, opening a separate issue to look at patching the runtime itself - https://github.com/microsoft/CsWinRT/issues/862

nuget/Microsoft.Windows.CsWinRT.targets
```diff
@@ -119,6 +119,66 @@ $(CsWinRTFilters)
...
+          ConsoleToMsBuild="true"
+          IgnoreExitCode="true">
+      <Output TaskParameter="ConsoleOutput" PropertyName="CsWinRTGuidPatchOutput" />
+      <Output TaskParameter="ExitCode" PropertyName="CsWinRTGuidPatchExitCode"/>
+    </Exec>
+  </Target>
+
+  <Target Name="CsWinRTReplaceForPatchedRuntime" 
+          Condition="$(CsWinRTGuidPatchExitCode) == 0"
+          AfterTargets="CsWinRTInvokeGuidPatcher">
```
manodasanW, line 171: This also needs BeforeTargets="Link" as it isn't guaranteed with AfterTargets.

src/Perf/IIDOptimizer/SignatureEmitter.cs
```diff
@@ -0,0 +1,536 @@
...
+            this.describedType = describedType;
+            this.guidDataGetterMethod = guidDataGetterMethod;
+
+            /*
+             *  Initialize the logger with the OptimizerDir property 
+             */
+            Logger = logger;
+        }
+
+        public void CloseLogger() { Logger.Close(); }
```
j0shuams, line 52: Cut

src/Perf/IIDOptimizer/SignatureEmitter.cs
```diff
@@ -0,0 +1,536 @@
...
+
+            var iid = new Guid(hash[0..16]);
+
+            CecilExtensions.WriteIIDDataGetterBody(guidDataGetterMethod, describedType, iid, guidDataBlockType, implementationDetailsType, readOnlySpanOfByteCtor);
+        }
+
+        private void GenerateGuidFactoryFromComplexSignature(TypeDefinition implementationDetailsType, TypeReference readOnlySpanOfByte, MethodReference readOnlySpanOfBytePtrCtor, TypeDefinition guidGeneratorType)
+        {
+            /*
+            Logger.Log("GenerateGuidFactoryFromComplexSignature");
```
j0shuams, line 156: cut

src/Perf/IIDOptimizer/SignatureEmitter.cs
```diff
@@ -0,0 +1,536 @@
...
+                            // signatureLength += bytes.Length
+                            il.Emit(OpCodes.Ldlen);
+                            il.Emit(OpCodes.Ldloc, fullSignatureLength);
+                            il.Emit(OpCodes.Add_Ovf);
+                            il.Emit(OpCodes.Stloc, fullSignatureLength);
+                        }
+                        break;
+                    case RuntimeCustomSignatureStep(MethodReference customSignatureMethod):
+                        {
+                            /// TODO test this pat h
```
j0shuams, line 279: Testing TODO

src/Perf/IIDOptimizer/GuidPatcher.cs
```diff
@@ -0,0 +1,387 @@
...
+        }
+
+        public int ProcessAssembly()
+        {
+            if (guidGeneratorType is null || guidAttributeType is null)
+            {
+                return 0;
+            }
+
+            Logger.Log($"Processing assembly: {assembly}.");
```
j0shuams, line 129: cut

src/Perf/IIDOptimizer/GuidPatcher.cs
```diff
@@ -0,0 +1,387 @@
...
+                    }
+                    break;
+                case UninstantiatedGeneric gen:
+                    {
+                        emitter.PushGenericParameter(gen.OriginalGenericParameter);
+                    }
+                    break;
+                case CustomSignatureMethod custom:
+                    {
+                        /// TODO test this path
```
j0shuams, line 268: Testing TODO

src/Perf/IIDOptimizer/GuidPatcher.cs
```diff
@@ -0,0 +1,387 @@
...
+                    }
+                    break;
+                case GuidSignature guid:
+                    {
+                        emitter.PushString(guid.IID.ToString("B"));
+                    }
+                    break;
+                case NonGenericDelegateSignature del:
+                    {
+                        /// TODO test this path
```
j0shuams, line 257: Testing TODO

src/Perf/IIDOptimizer/IIDOptimizer.csproj
```diff
@@ -0,0 +1,19 @@
...
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <!-- https://github.com/dotnet/command-line-api#packages -->
+    <RestoreAdditionalProjectSources>https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-libraries/nuget/v3/index.json</RestoreAdditionalProjectSources>
+  </PropertyGroup> 
+  
+  <ItemGroup>
+    <PackageReference Include="Mono.Cecil" Version="0.11.4" />
+    <PackageReference Include="System.CommandLine" Version="2.0.0-beta1.21356.1" />
```
manodasanW, line 16: Any reason to not use the one from nuget?

nuget/NOTICE.txt
```diff
@@ -0,0 +1,42 @@
...
+One Microsoft Way
+Redmond, WA 98052
+USA
+
+Notwithstanding any other terms, you may reverse engineer this software to the extent
+required to debug changes to any libraries licensed under the GNU Lesser General Public License.
+
+---------------------------------------------------------
+
+Mono.Cecil 0.11.3 - MIT
```
manodasanW, line 20: Update version.

src/Perf/IIDOptimizer/Program.cs
```diff
@@ -0,0 +1,74 @@
...
+
+namespace GuidPatch
+{
+    class Program
+    {
+        static readonly Option _targetAssembly =
+            new Option
+            (
+                name: "--targetAssembly",
+                description: "The assembly to perform GUID optimizations on.",
```
manodasanW, line 17: GUID **lookup** optimizations

src/cswinrt.sln
```diff
@@ -33,8 +37,6 @@ Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "TestComponent", "TestWinRT\
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "WinRT.Runtime", "WinRT.Runtime\WinRT.Runtime.csproj", "{25244CED-966E-45F2-9711-1F51E951FF89}"
```
manodasanW, line 38: Is WinUIDesktopSample not included anymore?

j0shuams: Fixed

nuget/Microsoft.Windows.CsWinRT.targets
```diff
@@ -183,7 +186,7 @@ $(GuidPatchTargetAssemblyReferences)
           BeforeTargets="Link">
 
     <ItemGroup>
-      <CsWinRTGuidPatchedFiles Include="$(MSBuildProjectDirectory)\$(IIDOptimizerInterimDir)*.dll" />
+      <CsWinRTGuidPatchedFiles Include="$(IIDOptimizerInterimDir)*.dll;$(IIDOptimizerInterimDir)*.pdb" />
```
manodasanW, line 189: Wondering if we should be specific about the dll we copy assuming we know the assembly name.

src/Benchmarks/Benchmarks.csproj
```diff
@@ -10,7 +10,7 @@
     <UseWinmd Condition="'$(TargetFramework)' == 'netcoreapp3.1' And $(BenchmarkWinmdSupport) == true">true</UseWinmd>
     <ApplicationManifest Condition="$(UseWinmd) == true">Benchmarks.manifest</ApplicationManifest>
     <BenchmarkTargetFramework>$(TargetFramework)</BenchmarkTargetFramework>
-    <BenchmarkTargetFramework Condition="'$(BenchmarkTargetFramework)' == 'netcoreapp2.0'">netstandard2.0</BenchmarkTargetFramework>
+    <BenchmarkTargetFramework Condition="'$(BenchmarkTargetFramework)' != 'net5.0'">netstandard2.0</BenchmarkTargetFramework>
```
manodasanW, line 13: @ujjwalchadha FYI, you might need to address these once these are merged.

manodasanW (approved):

src/Benchmarks/Benchmarks.csproj
ujjwalchadha: @j0shuams why was this change needed?

j0shuams: @ujjwalchadha A few lines below this, we have a <Reference> node for the windows projection, and we use this property in the path. The windows projection only builds for netstandard2.0 and net5.0. So I was getting build failures when BenchmarkTargetFramework was netcoreapp3.1

PR #852, opened 2021-05-26
Ujjwalchadha/perf/eventsourcecodegen

ujjwalchadha:


src/Benchmarks/Benchmarks.csproj
```diff
@@ -16,8 +16,14 @@
     <PackageReference Include="Microsoft.Windows.SDK.Contracts" Version="10.0.18362.2005" Condition="$(UseWinmd) == true"/>
     <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="4.5.3" Condition="$(UseWinmd) == false"/>
   </ItemGroup>
-
+	
   <ItemGroup>
+    <Reference Condition="'$(TargetFramework)' == 'netcoreapp2.0'" Include="$(MSBuildThisFileDirectory)..\Projections\Windows\bin\x64\Release\netstandard2.0\Windows.dll"></Reference>
```
manodasanW, line 21: This can be simplified if you define a new property like BenchmarkTargetFramework and set it to TargetFramework and then check if it is equal to necoreapp2.0 and if so, reset it to netstandard 2.0.

src/Benchmarks/Benchmarks.csproj
```diff
@@ -32,6 +38,7 @@
     <Reference Include="BenchmarkComponent.winmd" Condition="$(UseWinmd) == true">
       <HintPath>$(MSBuildThisFileDirectory)..\_build\$(Platform)\$(Configuration)\BenchmarkComponent\bin\BenchmarkComponent\BenchmarkComponent.winmd</HintPath>
       <IsWinMDFile>true</IsWinMDFile>
+		<Visible>True</Visible>
```
manodasanW, line 41: What does this property control?

src/Benchmarks/ReflectionPerf.cs
```diff
@@ -61,5 +62,21 @@ public object ExecuteMarshalingForCustomObject()
         {
             return instance.NewWrappedClassObject;
         }
+
+        [Benchmark]
+        public void IntEventSource()
+        {
+            ClassWithMarshalingRoutines instance = new ClassWithMarshalingRoutines();
```
manodasanW, line 69: Should this be done in setup?

src/cswinrt/strings/WinRT.cs
```diff
@@ -1,4 +1,5 @@
-﻿using System;
+﻿// This file was generated by cswinrt.exe version 0.0.0-private.0
```
manodasanW, line 1: Is this comment intended?

src/cswinrt/main.cpp
```diff
@@ -347,6 +386,7 @@ Where <spec> is one or more of:
         {
             w.write(" error: %\n", e.what());
             result = 1;
+            throw e;
```
manodasanW, line 389: Do we want to throw here?  The idea is print the error message to console and then return an exit code.

src/cswinrt/code_writers.h
```diff
@@ -453,6 +453,71 @@ namespace cswinrt
...
+        w.write("%",
+            bind<write_parameter_name>(param));
+    }
+
+    void write_event_source_type_name(writer& w, type_semantics const& eventTypeSemantics)
+    {
+        auto eventTypeCode = w.write_temp("%", bind<write_type_name>(eventTypeSemantics, typedef_name_type::Projected, false));
+        
+        std::string eventTypeName = "_EventSource_" + eventTypeCode;
+        std::replace(eventTypeName.begin(), eventTypeName.end(), '<', '_');
```
manodasanW, line 467: Would this be simpler to do with maybe a regex replace if possible or writing a custom write_type_name function?

src/cswinrt/strings/WinRT.cs
```diff
@@ -444,6 +446,28 @@ void _UnsubscribeFromNative()
...
+        {
+        }
+
+        override protected System.Delegate EventInvoke
+        {
+            get
+            {
+                System.EventHandler<T> handler = (System.Object obj, T e) =>
+                {
+                    _event.Invoke(obj, e);
```
manodasanW, line 464: Do we need a null check?

src/cswinrt/strings/WinRT.cs
```diff
@@ -444,6 +446,28 @@ void _UnsubscribeFromNative()
...
+            delegate* unmanaged[Stdcall]<System.IntPtr, System.IntPtr, out WinRT.EventRegistrationToken, int> addHandler,
+            delegate* unmanaged[Stdcall]<System.IntPtr, WinRT.EventRegistrationToken, int> removeHandler) : base(obj, addHandler, removeHandler)
+        {
+        }
+
+        override protected System.Delegate EventInvoke
+        {
+            get
+            {
+                System.EventHandler<T> handler = (System.Object obj, T e) =>
```
manodasanW, line 462: Should we be caching this or creating it each time?

src/cswinrt/strings/WinRT.cs
manodasanW: note you can also do _event?.Invoke(obj, e);

src/Benchmarks/ReflectionPerf.cs
ujjwalchadha: I did it to create a new object each time a test is run to better simulate the real-life scenario of creating an object and then adding an event handler. A real app would not add the event handler to the same object multiple times

src/cswinrt/strings/WinRT.cs
ujjwalchadha: I believe that would not work where the handler needs to return something. As is the case here:
```
 internal unsafe class _EventSource_global__TestComponentCSharp_EventWithReturn : EventSource<global::TestComponentCSharp.EventWithReturn>
        {
            private global::TestComponentCSharp.EventWithReturn handler;

            internal _EventSource_global__TestComponentCSharp_EventWithReturn(IObjectReference obj,
                delegate* unmanaged[Stdcall]<System.IntPtr, System.IntPtr, out WinRT.EventRegistrationToken, int> addHandler,
                delegate* unmanaged[Stdcall]<System.IntPtr, WinRT.EventRegistrationToken, int> removeHandler) : base(obj, addHandler, removeHandler)
            {
            }

            override protected System.Delegate EventInvoke
            {
                get
                {
                    if (handler == null)
                    {
                        handler = (int arg) =>
                        {
                            return _event.Invoke(arg);
                        };
                    }
                    return handler;
                }
            }
        }
```

Since the return type won't be optional, we would not be able to return optional. We cannot even return null for example, if the return type is int. One idea would be to not create any handler in case `_event` is uninitialized (or is null) and simply return `null` instead of a dummy handler that does nothing,

src/cswinrt/strings/WinRT.cs
ujjwalchadha: We can do that, but I'm not sure if that would give any significant perf gain

src/Benchmarks/Benchmarks.csproj
ujjwalchadha: I believe it shows the winmd file in the VS solution explorer in case we want to investigate it

src/cswinrt/strings/WinRT.cs
manodasanW: How does that work today?  And what happens when an event with no delegates added is fired?

ujjwalchadha: Today, we return the default value of the return type of the handler delegate (return 0 in this case) using Expression.Default which works for reflection. To mimic that in code gen, we can write a map that contains the default value for each type and use it to return the correct default. If you think there's a simpler way, do suggest.

src/cswinrt/strings/WinRT.cs
manodasanW: My concern was more about whether each time we marshal EventInvoke, whether we are marshaling a new one or the same one.  Looking more at it, it looks like we only call EventInvoke the first time and then we don't so it might be fine.

src/cswinrt/strings/WinRT.cs
manodasanW: do we know the return type?  if so we can do return default(type).

src/cswinrt/code_writers.h
```diff
@@ -453,6 +454,82 @@ namespace cswinrt
...
+    {
+        w.write("%",
+            bind<write_parameter_name>(param));
+    }
+
+    void write_event_source_type_name(writer& w, type_semantics const& eventTypeSemantics)
+    {
+        auto eventTypeCode = w.write_temp("%", bind<write_type_name>(eventTypeSemantics, typedef_name_type::Projected, false));
+        std::string eventTypeName = "_EventSource_" + eventTypeCode;
+        std::regex re("(\\ |\\:|\\<|\\>|\\,|\\.)");
```
Scottj1s, line 467: consider R"-((\ |\:|\<|\>|\,|\.))-"

src/cswinrt/code_writers.h
```diff
@@ -453,6 +454,82 @@ namespace cswinrt
...
+        w.write("%", std::regex_replace(eventTypeName, re, "_"));
+    }
+
+    void write_event_invoke_params(writer& w, TypeDef const& eventType)
+    {
+        for (auto&& method : eventType.MethodList())
```
Scottj1s, line 473: consider factoring out lookup of the "Invoke" method and binding all your write_event_invoke_* methods to that instead of eventType - simpler and faster code.

src/cswinrt/code_writers.h
```diff
@@ -453,6 +454,82 @@ namespace cswinrt
...
+        {
+            if (method.Name() == "Invoke")
+            {
+                method_signature methodSig(method);
+                if (!methodSig.return_signature())
+                {
+                    return;
+                }
+                auto&& semantics = get_type_semantics(methodSig.return_signature().Type());
+                w.write("default(%)", bind<write_projection_type>(semantics));
```
Scottj1s, line 514: wouldn't you want to return after this? (break the loop)

src/cswinrt/code_writers.h
```diff
@@ -2467,14 +2544,36 @@ db_path.stem().string());
         );
     }
 
+    void write_event_source_generic_args(writer& w, cswinrt::type_semantics eventTypeSemantics);
+
     void write_event_source_ctor(writer& w, Event const& evt)
     {
+        bool isSpecialCase = false;
```
Scottj1s, line 2551: note that for_typedef can return a value, which you could use to short-circuit return (and eliminate isSpecialCase)

src/cswinrt/code_writers.h
```diff
@@ -6340,8 +6439,90 @@ type.TypeNamespace(),
...
+                            return %;
+                        }
+                        %_event.Invoke(%);
+                    };
+                }
+                return handler;
+            }
+        }
+    }
+)", bind<write_event_source_type_name>(eventTypeSemantics),
```
Scottj1s, line 6516: indent args to w.write

src/cswinrt/strings/WinRT.cs
```diff
@@ -444,6 +445,34 @@ void _UnsubscribeFromNative()
...
+
+        internal EventSource__EventHandler(IObjectReference obj,
+            delegate* unmanaged[Stdcall]<System.IntPtr, System.IntPtr, out WinRT.EventRegistrationToken, int> addHandler,
+            delegate* unmanaged[Stdcall]<System.IntPtr, WinRT.EventRegistrationToken, int> removeHandler) : base(obj, addHandler, removeHandler)
+        {
+        }
+
+        override protected System.Delegate EventInvoke
+        {
+            get
```
Scottj1s, line 461: maybe a comment that this is synchronized from the base class

src/cswinrt/strings/WinRT.cs
Scottj1s: like what you're already doing in write_event_invoke_return_default

src/cswinrt/strings/WinRT.cs
Scottj1s: yes, EventInvoke is cached and then conditionally dispatches through to the end delegate accumulator.

src/cswinrt/main.cpp
```diff
@@ -212,6 +217,20 @@ Where <spec> is one or more of:
                                         write_factory_class(w, type);
                                     }
                                 }
+                                // Event source code generation:
```
Scottj1s, line 220: consider moving this into a write_class_event_source_subclass near related code in code_writers.h, and implement it in terms of write_interface_event_source_subclass below

src/cswinrt/main.cpp
```diff
@@ -222,6 +241,13 @@ Where <spec> is one or more of:
                                 break;
                             case category::interface_type:
                                 write_interface(w, type);
+                                // Event source code generation:
```
Scottj1s, line 244: consider factoring into write_interface_event_source_subclass in code_writers.h

src/cswinrt/main.cpp
```diff
@@ -188,6 +192,7 @@ Where <spec> is one or more of:
                                 continue;
                             }
                             auto guard{ w.push_generic_params(type.GenericParam()) };
+                            auto guard1{ helperWriter.push_generic_params(type.GenericParam()) };
```
Scottj1s, line 195: I don't think helperWriter is needed - you're only ever calling write_temp

ujjwalchadha: I believe we do need this because it is tied to the generic params that we set here. The write_temp functions uses the generic_params that we push to the helperWriter to write the type name and so on

src/Benchmarks/Benchmarks.csproj
```diff
@@ -3,24 +3,29 @@
...
     <OutputType>Exe</OutputType>
     <Platforms>x64;x86</Platforms>
-    <TargetFrameworks>netcoreapp2.0;net5.0;netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.0;netcoreapp3.1;net5.0;</TargetFrameworks>
     <UseWinmd>false</UseWinmd>
     <BenchmarkWinmdSupport Condition="'$(BenchmarkWinmdSupport)' == ''">false</BenchmarkWinmdSupport>
     <!-- Controls whether to build using WinMDs or the projection. -->
     <UseWinmd Condition="'$(TargetFramework)' == 'netcoreapp3.1' And $(BenchmarkWinmdSupport) == true">true</UseWinmd>
     <ApplicationManifest Condition="$(UseWinmd) == true">Benchmarks.manifest</ApplicationManifest>
+	<BenchmarkTargetFramework>$(TargetFramework)</BenchmarkTargetFramework>
```
Scottj1s, line 12: spaces, not tabs (here and elsewhere - replace each tab with 4 spaces and change your editor default)

src/cswinrt/main.cpp
Scottj1s: but w is pushing those very same generic params

src/cswinrt/code_writers.h
```diff
@@ -6340,8 +6419,117 @@ type.TypeNamespace(),
...
+            delegate* unmanaged[Stdcall]<System.IntPtr, System.IntPtr, out WinRT.EventRegistrationToken, int> addHandler,
+            delegate* unmanaged[Stdcall]<System.IntPtr, WinRT.EventRegistrationToken, int> removeHandler) : base(obj, addHandler, removeHandler)
+        {
+        }
+
+        override protected System.Delegate EventInvoke
+        {
+            get
+            {
+                if (handler == null)
```
manodasanW, line 6482: I wonder if we should just set handler in the constructor to avoid the null check.  But we can also leave it as is.

src/cswinrt/code_writers.h
```diff
@@ -6340,8 +6419,117 @@ type.TypeNamespace(),
...
+                    bind<write_event_invoke_params>(invokeMethodSig),
+                    bind<write_event_invoke_return_default>(invokeMethodSig),
+                    bind<write_event_invoke_return>(invokeMethodSig),
+                    bind<write_event_invoke_args>(invokeMethodSig));
+                });
+    }
+
+    void write_temp_class_event_source_subclass(writer& w, TypeDef const& classType, concurrency::concurrent_unordered_map<std::string, std::string>& typeNameToDefinitionMap)
+    {
+        for (auto&& ii : classType.InterfaceImpl())
```
manodasanW, line 6512: Can you remind me in which scenario this would add a event source class which isn't already added by write_temp_interface_event_source_subclass based on the interface declaration and is used.  The only scenario I see here is a class is declared which implements an interface from another projection.  But in that case would the event source be instantiated in this projection or would it be in the other projection in the interface ABI helper type?  I might be missing a scenario where this is used.

src/cswinrt/code_writers.h
ujjwalchadha: We just save having to create a handler object when not needed. This code is basically doing exactly what the reflection code was doing.

src/cswinrt/code_writers.h
ujjwalchadha: This is helpful when the class has fixed generic types but interface has variable. The following scenario for example:
```
interface Vector<T>
{
    ...
}

class MyVector<string> implements Vector<string>
{
   ...
}
```
In this case Vector<T> and Vector<string> can have different events

manodasanW (approved):

PR #853, opened 2021-05-27
Add support for calling non agile objects regardless of current context

### LINKED ISSUE #814: ContextCallback should safely handle the various disconnected error codes
Today we handle one of the disconnected error codes, we should safely handle the others there or in the Release when it is used.

https://github.com/microsoft/CsWinRT/blob/ad6adcb997b40cea649cf4156451316e714bc3ea/src/WinRT.Runtime/Interop/IContextCallback.cs#L79

https://github.com/microsoft/CsWinRT/blob/ad6adcb997b40cea649cf4156451316e714bc3ea/src/WinRT.Runtime/ObjectReference.cs#L413

### LINKED ISSUE #844: API calls to non agile objects should execute on right context
Today, we get wrong thread errors if APIs on non agile object are called from a different apartment.  We should marshal the object to the current apartment and then do the call.

manodasanW:
Previously in .NET native, apps were able to call APIs on non agile objects regardless of the context / apartment they were in.  .NET marshaled the ptr to the current context and then used that to call the API via a proxy.  This PR achieves something similar where whenever the ThisPtr is retrieved on an non agile object, we check if we are in the context when it was originally retrieved in and if so we use it, otherwise we switch back to the original context and get an agile reference to the object and then use the agile reference to get a proxy for it in the current context.  To avoid doing this each time, the Ptrs retrieved are cached.  If we fail to switch context, we just use the original ptr given that is the best we can do.

In addition, we previously checked whether an object implements IAgileObject as a way to determine whether it is agile or not.  But not all objects implement that but can still be agile such as if they aggregate the free threaded marshaler.  Now we have a check for that too and those who do are considered agile.  This seems to include WinUI types.

Fixes #844 
Fixes #814

### DIFF
--- a/src/WinRT.Runtime/ObjectReference.cs
+++ b/src/WinRT.Runtime/ObjectReference.cs
@@ -486,9 +486,8 @@ namespace WinRT
             ObjectReference<T> GetForCurrentContext(IntPtr _)
             {
                 var agileReference = _agileReference.Value;
-                // During process termination, we may fail to switch context
-                // and thereby not get an agile reference.  In these cases,
-                // fallback to using the current context.
+                // We may fail to switch context and thereby not get an agile reference.
+                // In these cases, fallback to using the current context.
                 if (agileReference == null)
                 {
                     return null;
--- a/src/WinRT.Runtime/ObjectReference.cs
+++ b/src/WinRT.Runtime/ObjectReference.cs
@@ -454,30 +454,47 @@ namespace WinRT
         private protected override IntPtr GetThisPtrForCurrentContext()
         {
             IntPtr currentContext = Context.GetContextToken();
-            if (_contextCallbackPtr == IntPtr.Zero || currentContext == _contextToken)
+            ObjectReference<T> cachedObjRef;
+            if (_contextCallbackPtr == IntPtr.Zero ||
+                currentContext == _contextToken ||
+                (cachedObjRef = GetCachedContext(currentContext)) == null)
             {
                 return base.GetThisPtrForCurrentContext();
             }
 
-            return GetCachedContext(currentContext).ThisPtr;
+            return cachedObjRef.ThisPtr;
         }
 
         private protected override T GetVftblForCurrentContext()
         {
             IntPtr currentContext = Context.GetContextToken();
-            if (_contextCallbackPtr == IntPtr.Zero || currentContext == _contextToken)
+            ObjectReference<T> cachedObjRef;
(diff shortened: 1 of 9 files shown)


src/WinRT.Runtime/ObjectReference.cs
```diff
@@ -21,6 +21,15 @@ public abstract class IObjectReference : IDisposable
...
         public IntPtr ThisPtr
         {
+            get
+            {
+                ThrowIfDisposed();
+                return GetThisPtrForCurrentContext();
+            }
+        }
+
+        private protected virtual IntPtr ThisPtrFromOriginalContext
```
Scottj1s, line 31: this doesn't appear to need to be virtual

src/WinRT.Runtime/ObjectReference.cs
```diff
@@ -389,57 +407,137 @@ private static unsafe T GetVtable(IntPtr thisPtr)
...
+            if (hr >= 0)
             {
                 if (IsAggregated)
                 {
                     Marshal.Release(thatPtr);
                 }
                 AddRefFromTrackerSource();
 
-                using (var contextCallbackReference = ObjectReference<ABI.WinRT.Interop.IContextCallback.Vftbl>.FromAbi(_contextCallbackPtr))
+                objRef = new ObjectReferenceWithContext<U>(thatPtr, Context.GetContextCallback(), Context.GetContextToken(), _agileReference, iid)
```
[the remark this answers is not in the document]
manodasanW: This changes behavior as the ptr that we get from a QI call is valid on the current context is received on.  It is not necessarily valid on the original context from when we constructed the object which we are doing a QI on.

src/WinRT.Runtime/ObjectReference.cs
```diff
@@ -389,57 +407,137 @@ private static unsafe T GetVtable(IntPtr thisPtr)
...
+        }
+
+        private protected override IntPtr GetThisPtrForCurrentContext()
+        {
+            IntPtr currentContext = Context.GetContextToken();
+            ObjectReference<T> cachedObjRef;
+            if (_contextCallbackPtr == IntPtr.Zero ||
+                currentContext == _contextToken ||
+                (cachedObjRef = GetCachedContext(currentContext)) == null)
+            {
```
Scottj1s, line 461: could move all this shared logic into GetCachedContext

Scottj1s (approved):
:shipit:

PR #855, opened 2021-05-28
Add logic to gather respective WinMDs of dependent projection DLLs.

### LINKED ISSUE #656: Dependent WinMDs of authored component should be automatically detected
Currently they need to be manually specified, but we should look into if we can automate it.

manodasanW:
We don't have an easy way of determine the WinMDs for the projection DLLs or determining which binaries are projection DLLs.  But we do need the WinMDs to pass to CsWinRT and there is an option for developers of authoring components to pass them via a property.  But to bring close to parity to the previous experience, we should also try to automate that.  This change automates it for package references by gathering all the nuget packages in use based on the runtime DLLs and then uses that to gather any WinMDs in those packages.  In case there is any issues with this logic, there is a property which allows it to be disabled.  We probably need to look at doing something similar for project references when we address the WinMD leak issue.

Also did some cleanup and moved some targets for authoring scenarios in the main CsWinRT targets to the authoring targets.

Fixes #656


Scottj1s (approved):
nice!

j0shuams (approved):

PR #857, opened 2021-06-02
Update dotnet SDK to to address VS 16.10 regressions

manodasanW:
- Updated to dotnet 300 series to address pipeline failure after recent VS update
- Removed dependency on older rosyln compilers package as a couple VS version have shipped with the minimum that we require and adding this package forces a downgrade on the compilers package which we use compared to the default one in VS
- Updating versions of source generator package to use release ones from nuget rather than one from dotnet feeds


Scottj1s (approved):
nice tidy up

PR #860, opened 2021-06-05
enum arrays are not blittable - special cased

Scottj1s:


Scottj1s:
reported by customer using Win2D in a Reunion app.  a call to CanvasGradientMesh.CreateCoonsPatch, passing an array of enums, was causing:

Exception thrown at 0x00007FFA71DC4B89 (KernelBase.dll) in Win2DTest.exe: WinRT originate error - 0x80070057 : 'Object contains non-primitive or non-blittable data. (Parameter 'value')'

src/WinRT.Runtime/Marshalers.cs
```diff
@@ -1195,13 +1197,23 @@ static Marshaler()
...
-                    GetAbiArray = (object box) => MarshalBlittable<T>.GetAbiArray(box);
-                    FromAbiArray = (object box) => MarshalBlittable<T>.FromAbiArray(box);
-                    FromManagedArray = (T[] array) => MarshalBlittable<T>.FromManagedArray(array);
-                    CopyManagedArray = (T[] array, IntPtr data) => MarshalBlittable<T>.CopyManagedArray(array, data);
-                    DisposeMarshalerArray = (object box) => MarshalBlittable<T>.DisposeMarshalerArray(box);
-                    DisposeAbiArray = (object box) => MarshalBlittable<T>.DisposeAbiArray(box);
+                    if (type.IsEnum)
+                    {
+                        // For marshaling non-blittable enum arrays via MarshalNonBlittable
+                        unsafe void CopyEnum(object value, IntPtr dest) => *(int*)dest.ToPointer() = (int)Convert.ChangeType(value, typeof(int));
```
manodasanW, line 1203: Will all enum values that we get be convertible to int and back without issues?  For instance max value of uint32 for flags enum?

Scottj1s: yes, good call - flags require special treatment

src/WinRT.Runtime/Marshalers.cs
```diff
@@ -1195,13 +1197,33 @@ static Marshaler()
...
+                            {
+                                *(uint*)dest.ToPointer() = (uint)Convert.ChangeType(value, typeof(uint));
+                            }
+                        }
+                        CopyAbi = (object value, IntPtr dest) => CopyEnum(value, dest);
+                        CopyManaged = (T value, IntPtr dest) => CopyEnum(value, dest);
+                    }
+                    else 
+                    { 
+                        CreateMarshalerArray = (T[] array) => MarshalBlittable<T>.CreateMarshalerArray(array);
```
manodasanW, line 1219: After our earlier conversation, I double checked and we seem to have a couple instances of Marshaler<T>.FromAbiArray in like IReferenceArray and IReadOnlyList and so on.  I feel like in those cases where T is an enum, we are going to end up with these functions no longer defined.  Same applies for some of the other functions.  Thoughts?

manodasanW:
Other than for that 1 comment, looks good otherwise.

src/WinRT.Runtime/Marshalers.cs
Scottj1s: I searched for all Marshaler<T> calls in MarshalNonBlittable and ensure that those were implemented.   So unless the calling code (MarshalNonBlittable) changes to introduce new calls, I don't see how we'd get in trouble.  Still, I'll reinstate the array functions, just in case.

Scottj1s:
taking Mano's last comment as an approval

PR #861, opened 2021-06-08
Added caching of event registration state to support unsubscribes across garbage collections

### LINKED ISSUE #842: Garbage collection of EventSource<T> causes subsequent failure to unsubscribe handlers
Reported by an external customer.  

WinRT.EventSource<T> has a bug in Subscribe/Unsubscribe, which cache event handlers (by combining delegates) rather than subscribing each one directly to the WinRT event.  This allows each handler to be removed without having to track individual event registration tokens.  The bug is that if an EventSource is garbage collected, then the next access in C# code to the event source will produce a new RCW.  The first RCW will leak its outstanding event handler registrations.  Subsequent attempts to unregister a handler on the second RCW will fail on this check in Unsubscribe:

                if (oldEvent is object && _event is null)

This is a common scenario for singleton objects (or long-lived objects) which might access the event source object each time without caching it and preserving the lifetime of the EventSource RCW:

        public SomeObject()
        {
            InitializeComponent();
            Loaded += (object sender, RoutedEventArgs e) =>
            {
                Singleton.Instance().SomeEvent += this.EventHandler;
            };

            Unloaded += (object sender, RoutedEventArgs e) =>
            {
                Singleton.Instance().SomeEvent -= this.EventHandler;
            };
        }

The net effect of the bug is that spurious event handlers are executed.

A workaround exists - cache the RCW to ensure that it does not lose track of its registrations.

Scottj1s:
fixes #842


src/cswinrt.sln
```diff
@@ -76,10 +76,6 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "TestHost.ProbeByClass", "Pr
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "WinRT.SourceGenerator", "Authoring\WinRT.SourceGenerator\WinRT.SourceGenerator.csproj", "{E0C26D3A-504A-4826-BAE2-DE775F865B2A}"
 EndProject
```
manodasanW, line 78: What is the context behind these test removals from the sln?

src/cswinrt/code_writers.h
```diff
@@ -6483,11 +6483,11 @@ bind<write_type_name>(type, typedef_name_type::CCW, true)
                 {
                     handler = (%) =>
```
manodasanW, line 6484: I assume we are intentionally not doing anything similar to _state.eventInvoke below as I assume it is done below for performance reasons only?

src/cswinrt/strings/WinRT.cs
```diff
@@ -415,63 +421,165 @@ protected virtual System.Delegate EventInvoke
...
+                    }
+                }
             }
         }
 
         internal EventSource(IObjectReference obj,
             delegate* unmanaged[Stdcall]<System.IntPtr, System.IntPtr, out WinRT.EventRegistrationToken, int> addHandler,
-            delegate* unmanaged[Stdcall]<System.IntPtr, WinRT.EventRegistrationToken, int> removeHandler)
+            delegate* unmanaged[Stdcall]<System.IntPtr, WinRT.EventRegistrationToken, int> removeHandler,
+            int index = 0)
```
manodasanW, line 540: is index always 0 or am I missing something.  Don't seem to see a place where we pass a value for it.

src/cswinrt/strings/WinRT.cs
```diff
@@ -415,63 +421,165 @@ protected virtual System.Delegate EventInvoke
...
+
+            public static void Create(IObjectReference obj, EventSource<TDelegate> source, int index)
+            {
+                // If event source implements weak reference support, track event registrations so that
+                // unsubscribes will work across garbage collections.  Note that most static/factory classes
+                // do not implement IWeakReferenceSource, so static codegen caching approach is also used.
+                IWeakReference target = null;
+                try
+                {
+                    target = obj.AsInterface<IWeakReferenceSource>().GetWeakReference();
```
manodasanW, line 491: AsInterface has a GetCustomAttribute call which we have in the past found to be expensive and worked on reducing.  Can we avoid it in any way like by using AsType instead maybe or maybe an internal version of an optimized AsInterface method which assumes none of the interfaces it gets is for ComImport?

src/cswinrt.sln
Scottj1s: that was a mistake - thanks

src/cswinrt/code_writers.h
Scottj1s: not sure what you're asking - this is Ujjwal's event source optimization logic

manodasanW: In the reflection version of this EventInvoke that this overrides, there was changes to set _state.EventInvoke and store it as part of the state.  I was wondering whether it is intentional that we aren't doing the same with handler here which is the same as eventInvoke there.  Is it because this is just a function lamda in code while that was an expression that had to be computed again?

src/cswinrt/strings/WinRT.cs
Scottj1s: you're not missing anything - dropped out in a merge - good catch

src/cswinrt/code_writers.h
```diff
@@ -6429,18 +6448,17 @@ bind<write_type_name>(type, typedef_name_type::CCW, true)
...
         }
-        for_typedef(w, eventTypeSemantics, [&](TypeDef const& eventType)
+        for_typedef(w, eventTypeSemantics, [&](TypeDef const&)
             {
                 std::vector<std::string> genericArgs;
-                int i = 0;
-                for (auto&& x: std::get<generic_type_instance>(eventTypeSemantics).generic_args)
+                auto arg_count = std::get<generic_type_instance>(eventTypeSemantics).generic_args.size();
+                for (int i = 0; i < arg_count; i++)
                 {
```
Scottj1s, line 6456: this is removing warnings (which are now treated as errors)

manodasanW (approved):

PR #866, opened 2021-06-10
Fixed special cased event source

ujjwalchadha:
This PR fixes the correct use of special case EventSource i.e. EventSource<EventHandler>


src/cswinrt/code_writers.h
```diff
@@ -2529,11 +2529,10 @@ db_path.stem().string());
     {
         if (for_typedef(w, get_type_semantics(evt.EventType()), [&](TypeDef const& eventType)
             {
-                if (eventType.TypeNamespace() == "System" && eventType.TypeName() == "EventHandler`1")
+                if ((eventType.TypeNamespace() == "Windows.Foundation" || eventType.TypeNamespace() == "System") && eventType.TypeName() == "EventHandler`1")
```
manodasanW, line 2532: Look at whether System.EventHandler`1 actually ever comes here as it is a custom type mapping for Windows.Foundation.EventHandler`1 and I wouldn't expect to see it in metadata I think.

manodasanW (approved):

PR #871, opened 2021-06-11
Clean intermediate dirs in pipeline builds to manage disk space

Scottj1s:
Hosted agents are limited to 10GB of disk space, and we just started running out.  Added a post-build target to delete the obj dir of each project.

### DIFF
--- a/src/Directory.Build.targets
+++ b/src/Directory.Build.targets
@@ -29,7 +29,7 @@
     <Error Condition="!Exists('$(SolutionDir)TestWinRT')" Text="This solution requires the TestWinRT repo, which is missing. Please run build.cmd from a Visual Studio command prompt." />
   </Target>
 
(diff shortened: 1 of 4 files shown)


src/Directory.Build.targets
```diff
@@ -29,4 +29,8 @@
     <Error Condition="!Exists('$(SolutionDir)TestWinRT')" Text="This solution requires the TestWinRT repo, which is missing. Please run build.cmd from a Visual Studio command prompt." />
   </Target>
 
+  <Target Name="CleanIntermediateDirs" Condition="'$(CleanIntermediateDirs)'=='true'" AfterTargets="Build">
```
manodasanW, line 32: Is this set by us like on the pipeline?

manodasanW (approved):

PR #872, opened 2021-06-12
Fix issue with interop winmd path

manodasanW:


Scottj1s (commented):
ah, trailing slash - thanks!

PR #873, opened 2021-06-12
Fix codegen when custom attributes have embedded newlines

### LINKED ISSUE #818: Newlines in custom attributes generate a non-compiling projection
**Describe the bug**
We have a WinRT component with classes that have custom attributes, which have values of type string that have newlines in them.

When these attributes are projected, the resulting C# projection is invalid code (does not build)

image

Diff of what it should be:
```diff
-    [Microsoft.ReactNative.DocString(Content = "`IRedBoxHandler` provides an extension point to allow custom error handling within the React instance.
+    [Microsoft.ReactNative.DocString(Content = @"`IRedBoxHandler` provides an extension point to allow custom error handling within the React instance.

    This can be useful if you have an existing error reporting system that you want React errors to be reported to.
    The default implementation of `RedBoxHandler` shows an error messages in a error screen
    that covers the whole application window.
    
    If you want to maintain the existing `RedBox` behaviors, and also report errors to your own reporting system,
    your implementation can call into the default `RedBoxHandler`, which can be obtained by calling :
    
    ```csharp
    RedBoxHelper::CreateDefaultHandler(Host);
    ```
    
    Sample settings up a `RedBoxHandler` that reports errors to an external system, and displays the default `RedBox`
    experience within the application:
    
-    ```csharp
+    ```csharp")]
```

Note that it's not just the `@` sign at the start of the string (to allow embedded newlines) but also it's missing the closing `")]`.

**To Reproduce**
Project the Microsoft.ReactNative winmd or any winmd with a custom attribute that has embedded newlines

**Expected behavior**
The projection is valid

**Version Info**
CsWinRT 1.2.5
.net 5.0.202
Windows SDK 19041

**Additional context**
<!-- Enter any other applicable info here -->

asklar:
Fixes #818 

When stamping out custom attributes, the codegen was not considering that string-typed custom attribute parameters might contain newlines. This results in invalid C# generated code.
The fix is to prepend the string with the "raw string" C# token `@` so that newlines are embedded in the string as intended.

### DIFF
--- a/src/Tests/TestComponentCSharp/TestComponentCSharp.idl
+++ b/src/Tests/TestComponentCSharp/TestComponentCSharp.idl
@@ -500,4 +500,23 @@ namespace TestComponentCSharp
             //overridable event Windows.Foundation.EventHandler<Int32> WarningOverridableEvent;
         }
     }
+
+    [attributeusage(target_runtimeclass, target_interface, target_struct, target_enum, target_delegate, target_field, target_property, target_method, target_event)]
+    [attributename("attr_string")]
+    attribute MyStringAttribute
+    {
+        String Content;
+    }
+
+    [attr_string(
+"This is line one
+This is line two
+
+And this is another one"
+    )]
+    runtimeclass MultiLineStringAttributeTest
+    {
+      void f();
+    }
+
 }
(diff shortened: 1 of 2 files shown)


asklar:
CC @j0shuams @Scottj1s

Scottj1s (approved):
thanks Alex!

PR #880, opened 2021-06-16
Rework Fast Up To Date checks to correctly capture cs:winmd relationship

### LINKED ISSUE #879: Visual Studio Fast Up To Date builds not correctly implemented
An earlier attempt to implement support for VS Fast Up To Date builds was not correct, as it didn't account for conditional writes of both the cswinrt.rsp response file, and the generated *.cs files.  As a result, any input winmd file newer than either the rsp or, indirectly, *.cs files, would cause the project to appear (and report) "not up to date" and MSBuild would be run again.

This fix attempts to capture the relationship correctly between input winmds and output cs files, using WinRT.cs as a marker to ensure that outputs are always newer than inputs.  This also exposed an MSBuild bug which causes the Test projection to always report "not up to date", triggering a rebuild:
https://github.com/dotnet/project-system/issues/7326

Scottj1s:
fixes #879 

see related issue:
[Visual Studio Fast Up To Date build not working - CopyComplete marker is touched when no files are copied · Issue #7326 · dotnet/project-system (github.com)](https://github.com/dotnet/project-system/issues/7326)

### DIFF
--- a/nuget/Microsoft.Windows.CsWinRT.targets
+++ b/nuget/Microsoft.Windows.CsWinRT.targets
@@ -91,7 +91,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 @(CsWinRTExcludeItems->'-exclude %(Identity)', '&#x0d;&#x0a;')
 @(CsWinRTIncludeItems->'-include %(Identity)', '&#x0d;&#x0a;')
       </CsWinRTFilters>
-      <CsWinRTInteropMetadata Condition="'$(CsWinRTInteropMetadata)' == ''">$([MSBuild]::NormalizeDirectory('$(MSBuildThisFileDirectory)', '..\metadata\WinRT.Interop.winmd'))</CsWinRTInteropMetadata>
+      <CsWinRTInteropMetadata Condition="'$(CsWinRTInteropMetadata)' == ''">$([MSBuild]::NormalizePath('$(MSBuildThisFileDirectory)', '..\metadata\WinRT.Interop.winmd'))</CsWinRTInteropMetadata>
       <CsWinRTIncludeWinRTInterop Condition="$(CsWinRTFilters.Contains('-include Windows&#x0d;&#x0a;'))">
 -input $(CsWinRTInteropMetadata)
 -include WinRT.Interop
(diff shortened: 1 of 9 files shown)


nuget/Microsoft.Windows.CsWinRT.targets
```diff
@@ -110,12 +110,13 @@ $(CsWinRTIncludeWinRTInterop)
     <WriteLinesToFile File="$(CsWinRTResponseFile)" Lines="$(CsWinRTParams)" Overwrite="true" WriteOnlyWhenDifferent="true" />
     <Message Text="$(CsWinRTCommand)" Importance="$(CsWinRTMessageImportance)" />
     <Exec Command="$(CsWinRTCommand)" />
+    <!--To support Visual Studio's FastUpToDate builds, we need to cpature the CsWinRT relationship of *.cs: *.winmd.
+    To ensure that modifications to input winmds are reflected in generated *.cs files, WinRT.cs is used as a build
+    marker and updated to the most recent winmd timestamp. UpToDateCheckBuilt then captures the dependency from
+    winmds to cs files.-->
+    <Touch Files="$(CsWinRTGeneratedFilesDir)/WinRT.cs" Time="%(CsWinRTInputs.ModifiedTime)"/>
```
manodasanW, line 117: So what time gets set?  Does it go through all the inputs and just chooses one of them or sets with each time?  should we be choosing the latest time?

Scottj1s: yes, it goes thru all inputs and conditionally sets it, so it gets the latest time.  we can't set to latest time, or it will always be newer than build outputs and will always rebuild.

manodasanW: I do see that it is going through all the inputs and getting each input's modified time and setting it on WinRT.cs.  But what is the end result?  Does WinRT.cs get the time of the last file processed or some other time?

Scottj1s: It's using msbuild batching so each Touch is executed independently and conditionally sets the time of WinRT.cs if and only if the new time is later than the current time.  Net effect - WinRT.cs inherits the time of the newest input winmd.

nuget/Microsoft.Windows.CsWinRT.targets
```diff
@@ -110,12 +110,13 @@ $(CsWinRTIncludeWinRTInterop)
     <WriteLinesToFile File="$(CsWinRTResponseFile)" Lines="$(CsWinRTParams)" Overwrite="true" WriteOnlyWhenDifferent="true" />
     <Message Text="$(CsWinRTCommand)" Importance="$(CsWinRTMessageImportance)" />
     <Exec Command="$(CsWinRTCommand)" />
+    <!--To support Visual Studio's FastUpToDate builds, we need to cpature the CsWinRT relationship of *.cs: *.winmd.
```
srdjanjovcic, line 113: spelling: 'capture'

srdjanjovcic (approved):

manodasanW (approved):

manodasanW:
Looks like this breaks the build when WinRT.cs doesn't get created.  Specifically the one in authoring is that they are all generated in a temp dir and then fed to a source geneator.  For authoring scenarios, we don't need to do this as it would be handled by the source generator so I think can do a conditional check on it to exclude those scenarios.

PR #881, opened 2021-06-17
Added support for out and array params for event source handler

ujjwalchadha:


src/cswinrt/code_writers.h
```diff
@@ -457,8 +457,23 @@ namespace cswinrt
...
+        {
+        case param_category::in:
+        case param_category::ref:
+            w.write("%", bind<write_parameter_name>(param));
+            break;
+        case param_category::out:
+            w.write("out %", bind<write_parameter_name>(param));
+            break;
+        case param_category::pass_array:
+        case param_category::fill_array:
```
manodasanW, line 470: are these used when passing the parameter or declaring the parameter type?

ujjwalchadha: When passing the arguments.
in `event.Invoke(...)`

manodasanW: so that would generate something like param[] which you might want to confirm whether is right.  When a type is being written for a method parameter, that would make sense like type[].  Look at write_parameter_name_with_modifier vs write_projection_parameter_type.

ujjwalchadha: That makes sense. The square brackets seem unnecessary in that case.

manodasanW (approved):

PR #882, opened 2021-06-17
Address potential races in recent event changes

manodasanW:
- If an event source tries to invoke a handler right after it is added, it won't invoke any registered handlers as we set them up after.  Fixing the ordering for it.
- Previously before the event handler optimizations, we stored the delegate in a local before invoking it to prevent races with it being removed.  Putting that back in to avoid that race.

### DIFF
--- a/src/cswinrt/code_writers.h
+++ b/src/cswinrt/code_writers.h
@@ -6488,11 +6488,12 @@ bind<write_type_name>(type, typedef_name_type::CCW, true)
                 {
                     handler = (%) =>
                     {
-                        if (_state.del == null)
+                        % localDel = _state.del;
+                        if (localDel == null)
                         {
                             return %;
                         }
-                        %_state.del.Invoke(%);
+                        %localDel.Invoke(%);
                     };
                 }
                 return handler;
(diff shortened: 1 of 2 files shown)


src/cswinrt/code_writers.h
```diff
@@ -6488,11 +6488,12 @@ bind<write_type_name>(type, typedef_name_type::CCW, true)
                 {
                     handler = (%) =>
                     {
-                        if (_state.del == null)
+                        % localDel = _state.del;
```
Scottj1s, line 6491: or just 'var'

src/cswinrt/strings/WinRT.cs
```diff
@@ -579,8 +581,9 @@ override protected System.Delegate EventInvoke
                 {
                     handler = (System.Object obj, T e) =>
                     {
-                        if (_state.del != null)
-                            _state.del.Invoke(obj, e);
+                        System.EventHandler<T> delLocal = _state.del;
+                        if (delLocal != null)
+                            delLocal.Invoke(obj, e);
```
Scottj1s, line 586: this truly is a scary race - good fix

Scottj1s (approved):

src/cswinrt/code_writers.h
manodasanW: Strange VS intellisense was complaining before when I used var, but its working now.  Will change it.

PR #884, opened 2021-06-17
added out param default to event source handlers

ujjwalchadha:


src/cswinrt/code_writers.h
```diff
@@ -498,6 +498,18 @@ namespace cswinrt
...
     }
 
+    void write_event_out_defaults(writer& w, method_signature const& methodSig)
+    {
+        for (auto&& param : methodSig.params())
+        {
+            if (get_param_category(param) == param_category::out || get_param_category(param) == param_category::receive_array)
+            {
+                w.write("\n% = default(%);", bind<write_parameter_name>(param), bind<write_projection_type>(get_type_semantics(param.second->Type())));
+                break;
```
manodasanW, line 508: Can there be more than 1 out parameter that we need to do for each rather than breaking?

manodasanW (approved):

PR #894, opened 2021-06-22
Interop interfaces are all pass-by-ref for IID param

### LINKED ISSUE #885: COM interop helpers are crashing in x86
Probably an HWND marshaling issue

Scottj1s:
fixes #885 

fixed interop IDL, and decided to just keep the TestHWND I created earlier (when I thought it was an HWND marshaling issue)

Note that:
for 64-bit calling convention, a 128-bit value ends up being passed by address, so we get lucky
for 32-bit calling convention, a 128-bit value is passed directly on the stack, so we get corruption when the callee attempts to interpret the first 4 bytes of that GUID as a pointer to one.

### DIFF
--- a/src/Tests/TestComponentCSharp/Class.cpp
+++ b/src/Tests/TestComponentCSharp/Class.cpp
@@ -1480,6 +1480,25 @@ namespace winrt::TestComponentCSharp::implementation
         return winrt::make<service_provider>();
     }
 
+    struct __declspec(uuid("15651B9F-6C6B-4CC0-944C-C7D7B0F36F81")) IComInterop : ::IUnknown
+    {
+        virtual HRESULT __stdcall GetForWindow(HWND window, guid riid, int64_t* value) noexcept = 0;
+    };
+
+    WF::IInspectable Class::ComInterop()
+    {
+        struct com_interop : winrt::implements<com_interop, WF::IInspectable, IComInterop>
+        {
+            HRESULT __stdcall GetForWindow(HWND window, guid riid, int64_t* value) noexcept override
+            {
+                *value = riid == winrt::guid_of<IComInterop>() ? (int64_t)window : 0;
+                return 0;
+            }
+        };
+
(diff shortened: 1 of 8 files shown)


manodasanW:
Is there any impact to the Windows SDK projection public API surface / assembly version from this change?  I see all the types that were modified had ProjectionInternal, but want to confirm there is indeed no impact.

Scottj1s:
The binlog shows only Windows SDK and Project Reunion winmds as input, but no ReactNative winmds, which explains the output.

Scottj1s:
> Is there any impact to the Windows SDK projection public API surface / assembly version from this change? I see all the types that were modified had ProjectionInternal, but want to confirm there is indeed no impact.

There shouldn't be any impact - the interop IDL is used solely to generate projection sources for the Windows SDK

src/Tests/TestComponentCSharp/Class.cpp
```diff
@@ -1480,6 +1480,25 @@ namespace winrt::TestComponentCSharp::implementation
         return winrt::make<service_provider>();
     }
 
+    struct __declspec(uuid("15651B9F-6C6B-4CC0-944C-C7D7B0F36F81")) IComInterop : ::IUnknown
```
manodasanW, line 1483: Can we add a test similar to this model that exercises one of the interfaces that we added ref const for?  I see this one is consumed via a ComImport in the test and it would be useful for us to custom implement one of the other interfaces and verify we can call it.  This would allow for a test that runs in the test pass which exercises it.

Scottj1s: yes, implemented IDragDropManager interop interface.  note that this required InternalsVisibleTo from Windows to UnitTest, which in turn created an ambiguity in WinrtModule in the WinUI test (which was doing nothing), so I removed that line

manodasanW (approved):

PR #902, opened 2021-07-07
Ensure generated code has the correct header to be picked up by code analysis.

### LINKED ISSUE #848: Generated file is still analyzed for Code Analysis
Every file in "Generated Files\*.cs" is a generated file, so they should not be analyzed for Code Analysis warnings. Using a projection with CsWinRTIncludes and CsWinRTWindowsMetadata should not add a bunch of warnings, as it does today.
**Describe the bug**

**To Reproduce**
1. Project a winmd using CsWinRTIncludes and CsWinRTWindowsMetadata.
2. There will be dozens of SA1400, SA1127, SA1308, etc...

**Expected behavior**
No warnings for generated code.

**Version Info**
CS WinRT 1.2.6

**Additional context**
Maybe leveraging [GeneratedCodeAttribute](https://docs.microsoft.com/en-us/dotnet/api/system.codedom.compiler.generatedcodeattribute?redirectedfrom=MSDN&view=net-5.0) is all that is needed?

jlaanstra:
Fixes #848


src/cswinrt/main.cpp
```diff
@@ -321,7 +321,15 @@ Where <spec> is one or more of:
                 group.add([&componentActivatableClasses, &projectionFileWritten]
                 {
                     writer wm;
-                    wm.write("// This file was generated by cswinrt.exe version %", VERSION_STRING);
+                    wm.write(R"(//------------------------------------------------------------------------------
+// <auto-generated>
```
manodasanW, line 325: Does the auto-generated tag cause the code analysis tooling to not pick it up?  Is that documented somewhere?

jlaanstra: I can't immediately find the public docs for it, but the Roslyn code is here: https://github.com/dotnet/roslyn/blob/cf55f3a58e47298426fa971d3bd9d8857c746c65/src/Compilers/Core/Portable/InternalUtilities/GeneratedCodeUtilities.cs

src/cswinrt/main.cpp
```diff
@@ -321,7 +321,15 @@ Where <spec> is one or more of:
...
+//     This file was generated by cswinrt.exe version %
+//
+//     Changes to this file may cause incorrect behavior and will be lost if
+//     the code is regenerated.
+// </auto-generated>
+//------------------------------------------------------------------------------
+)", VERSION_STRING);
```
manodasanW, line 332: Can you also add a similar write line below to `writer eventHelperWriter("WinRT");`.  It should probably have one, but seems to be have been missed before.

jlaanstra: Done. Let me know if this is what you meant.

manodasanW: Thanks, that is what I meant.

manodasanW (approved):

manodasanW:
Thanks!

PR #903, opened 2021-07-07
Fix datapackage to datapackageview invalid cast

ujjwalchadha:
This PR adds a fallback while casting the rcw to the managed type using the cache. The fallback is to not use the cache and try creating the rcw again. This takes care of the invalid case where aa different runtime class represents the same object pointer that has been seen before.


src/WinRT.Runtime/ComWrappersSupport.net5.cs
```diff
@@ -83,7 +88,8 @@ public static T CreateRcwForComObject<T>(IntPtr ptr)
             // ComWrappers API surface, so we are achieving it via a thread local.  We unset it after in case
             // there is other calls to it via other means.
             CreateRCWType.Value = typeof(T);
-            var rcw = ComWrappers.GetOrCreateObjectForComInstance(ptr, CreateObjectFlags.TrackerObject);
+
+            var rcw = tryUseCache ? ComWrappers.GetOrCreateObjectForComInstance(ptr, CreateObjectFlags.TrackerObject) : ComWrappers.GetOrCreateObjectForComInstance(ptr, CreateObjectFlags.UniqueInstance);
```
manodasanW, line 92: We should still or TrackerObject to UniqueInstance in the 2nd case.  Consider making it a variable and do the ternary when setting the variable which is then passed to the call.

src/WinRT.Runtime/ComWrappersSupport.net5.cs
```diff
@@ -99,12 +105,22 @@ public static T CreateRcwForComObject<T>(IntPtr ptr)
...
+                    ABI.System.Nullable<Type> nt => (T)(object)nt.Value,
+                    _ => (T)rcw
+                };
+            }
+            catch (InvalidCastException)
+            {
+                if (tryUseCache)
+                    return CreateRcwForComObject<T>(ptr, false);
+                throw;
+            }
```
jkoritzinsky, line 122: I'd suggest using a pattern matching mechanism to avoid having any `InvalidCastException`s in a non-failure path.

```suggestion
            return rcw switch
            {
                ABI.System.Nullable<string> ns => (T)(object)ns.Value,
                ABI.System.Nullable<Type> nt => (T)(object)nt.Value,
                T castRcw => castRcw
                _ when tryUseCache => CreateRcwForComObject<T>(ptr, false),
                _ => throw new InvalidCastException()
            };
```

src/WinRT.Runtime/ComWrappersSupport.net5.cs
```diff
@@ -99,12 +106,15 @@ public static T CreateRcwForComObject<T>(IntPtr ptr)
...
-                ABI.System.Nullable<Type> nt => (T)(object) nt.Value,
-                _ => (T) rcw
-            };
+            return rcw switch
+            {
+                ABI.System.Nullable<string> ns => (T)(object)ns.Value,
+                ABI.System.Nullable<Type> nt => (T)(object)nt.Value,
+                T castRcw => castRcw,
+                _ when tryUseCache => CreateRcwForComObject<T>(ptr, false),
+                _ => throw new InvalidCastException()
```
manodasanW, line 115: I wonder if the last fallback should be `(T) rcw` to allow us to get a proper exception message on what is being casted from what and to what.

jkoritzinsky: It might also be worth using a different exception type with a custom message so the failure case doesn't look like a bug in C#/WinRT.

src/WinRT.Runtime/ComWrappersSupport.net5.cs
```diff
@@ -99,12 +106,15 @@ public static T CreateRcwForComObject<T>(IntPtr ptr)
...
-                ABI.System.Nullable<Type> nt => (T)(object) nt.Value,
-                _ => (T) rcw
-            };
+            return rcw switch
+            {
+                ABI.System.Nullable<string> ns => (T)(object)ns.Value,
+                ABI.System.Nullable<Type> nt => (T)(object)nt.Value,
+                T castRcw => castRcw,
+                _ when tryUseCache => CreateRcwForComObject<T>(ptr, false),
+                _ => throw new ArgumentException(string.Format("Unable to create rcw. The com object {0} has type {1} which cannot be assigned to type {2}", ptr, rcw.GetType(), typeof(T)))
```
manodasanW, line 115:
```suggestion
                _ => throw new ArgumentException(string.Format("Unable to create a wrapper object. The WinRT object {0} has type {1} which cannot be assigned to type {2}", ptr, rcw.GetType(), typeof(T)))

```

manodasanW (approved):

jkoritzinsky (approved):

PR #909, opened 2021-07-12
Remove duplicate file from the CsWinRT NuGet package.

### LINKED ISSUE #908: CsWinRT NuGet includes the buildTransitive folder but doesn't follow the NuGet conventions for it.
**Describe the bug**
NuGet guidance says that if you have a buildTransitive folder, it should match the content of the build folder. However in this case I don't think the targets are intended to be transitive. 

Since the file from the buildTransitive folder is only used for generation of packages and is the same as the file in the build folder, it can probably just be removed.

jlaanstra:
This PR fixes #908 by simply using the Microsoft.Windows.CsWinRT.Authoring.Transitive.targets file from the build folder and removing the one from buildTransitive.

### DIFF
--- a/nuget/Microsoft.Windows.CsWinRT.Authoring.targets
+++ b/nuget/Microsoft.Windows.CsWinRT.Authoring.targets
@@ -169,7 +169,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       </Content> 
 
       <!-- Custom targets that copy DLLs for consumers of the authored component. -->
-      <Content Include="$(CsWinRTPath)buildTransitive\Microsoft.Windows.CsWinRT.Authoring.Transitive.targets">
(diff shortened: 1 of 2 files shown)


j0shuams:
Hi @jlaanstra thanks for your consideration! I believe we need the buildTransitive folder, as it is how package reference users will get the custom build logic. A c++ project with a package reference (could be an entry in packages.config and not PackageReference), will not get any assets from the build folder. 

I built a nupkg on a branch with your changes, and put it through my tests and the consuming project failed to generate a projection. This leads me to think this PR would break scenarios that currently work. 
If you want to do the same tests, let me know and I can give more information. It is based on the "Coords" and "App" projects in my test repo here - https://github.com/j0shuams/HostSupport

j0shuams (approved):

PR #916, opened 2021-07-21
fixed typo in IWebAuthenticationCoreManagerInterop

Scottj1s:


manodasanW:
Looks like this might be an API surface change for the Windows SDK projection?

Scottj1s:
> Looks like this might be an API surface change for the Windows SDK projection?

Yes, I suppose that means an AssemblyVersion bump there

manodasanW (approved):

PR #919, opened 2021-07-22
Update authoring support for project references

### LINKED ISSUE #808: Update targets when SDK1130 is addressed
In our work for a full authoring experience, we hit an issue when adding a project reference to a runtime component. The issue is the `.winmd` made by the component cannot flow into a native project, because of SDK1130. There will be an update to the dotnet target that throws 1130 so that the right `.winmd` can flow properly. 

When this change goes in, we should just need to remove [this line](https://github.com/microsoft/CsWinRT/blob/039332d725ea3bdf4063d8b239953d1849bb2067/nuget/Microsoft.Windows.CsWinRT.Authoring.targets#L69) from our targets authoring targets. Also update the comments for that target 
Validate via test apps (managed and native) that have project/package references to cswinrt components.

### LINKED ISSUE #816: [Authoring]: Consuming a component from a packaged C++ app sets wrong implementation
**Describe the bug**

Building the wapproj causes the implementation dll to be set as `MyComponent.dll` instead of `WinRT.Host.dll`. The generated appxmanifest.xml in the Appx folder has this automatically generated:

```
<Extensions>
    <Extension Category="windows.activatableClass.inProcessServer">
      <InProcessServer>
        <Path>CppConsoleApp\AuthoringDemo.dll</Path>
        <ActivatableClass ActivatableClassId="AuthoringDemo.Example" ThreadingModel="both" />
      </InProcessServer>
    </Extension>
  </Extensions>
```
Investigate which target is causing this (may be a wapproj target as CsWinRT does set the implementation as WinRT.Host.dll explicitly)

**To Reproduce**
Repros - 
https://github.com/j0shuams/CsWinRTTest
https://github.com/angelazhangmsft/AuthoringPackagingScenarios

**Expected behavior**

Implementation should be WinRT.Host.dll

**Version Info**

CsWinRT 1.2.5
VS 16.10 Preview 2

j0shuams:
Today, if C++ developers want to use a C#/WinRT authored component, they need to add a project reference as well as a manual reference to the component's winmd. This PR makes it so that they only need the project reference. 

To accomplish this, we have changed the model for our project outputs a bit. 
C#/WinRT authored components are C# class libraries that set a property so that a .winmd is made from the library's public types. 
This .winmd is essential to C++ consumers, as they need to use C++/WinRT to generate a projection from the .winmd. 
C# consumers need the managed implementation -- the .dll of the same assembly as the .winmd. 
We tried outputting both the .dll and the .winmd but this caused issues for consumers -- mainly because they have the same assembly identity, the build system would lose track of one or the other. 

This PR gives an updated model of how the project system works for authored components.
Components now only "output" their .winmd. Output here refers to the value seen by the target GetTargetPath, which is used by MSBuild to transfer project reference binaries. This provides the desired experience for C++ developers.

I have added another target that sets the component's .dll in an item group that ensures it is distributed to consumers as well. If they only got the .winmd, their app would fail. So we mark the .winmd with a new metadata value -- ManagedImplementation -- which holds the path to the implementation .dll. 
In turn, the DotNet SDK needs to use this value to make sure the implementation .dll is provided at compile time. I have a PR on the DotNet SDK to do this work, and this PR should not be merged until the DotNet SDK has been updated. 

If it is confusing, maybe it will help to think about how C++ users need the .dll at runtime, whereas C# users need it at compile time, which is why we need the DotNet SDK's help. 

dotnet\sdk PR: https://github.com/dotnet/sdk/pull/19218 (net5); https://github.com/dotnet/sdk/pull/20153 (net6)

This PR fixes #808 , fixes #816  and fixes a piece of #663 related to hacking the appxmanifest. 
816 is fixed because there is no more confusion on assemblies because of their identity; packaging projects "only see" the .winmd. 

I have tested this using a basic component consumed by a managed app and a native app. Also tested was packaging scenarios, where a packaging project is added on top of the managed and native apps. I have tested both the project reference method of consumption, as well as the nuget package reference method.


nuget/Microsoft.Windows.CsWinRT.Authoring.targets
```diff
@@ -16,6 +16,11 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <CsWinRTDetectDependentAuthoringWinMDs Condition="'$(CsWinRTDetectDependentAuthoringWinMDs)'==''">true</CsWinRTDetectDependentAuthoringWinMDs>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <CsWinRTAuthoring_Platform Condition="'$(Platform)'=='AnyCPU'">x86</CsWinRTAuthoring_Platform>
```
manodasanW, line 20: Are we relying on x86 can probably be loaded in most scenarios for anycpu?

manodasanW:
What are the implications with the ordering of when both changes go out?  If someone has this, but not the .NET SDK change what happens?  If someone has the .NET SDK change, but not this, what happens?

nuget/Microsoft.Windows.CsWinRT.Authoring.targets
angelazhangmsft: I assume this is because C++/Reunion apps can't be run with AnyCPU? Is this making x86 the default then and not x64?

j0shuams:
> What are the implications with the ordering of when both changes go out? If someone has this, but not the .NET SDK change what happens? If someone has the .NET SDK change, but not this, what happens?

These changes rely on the dotnet/sdk changes; if we merged these without a dotnet/sdk update, then we would break the C# customer experience. This is because we would not be threading the managed implementation through. 

The dotnet/sdk changes can go in anytime, they won't break anything. The reason being that those changes only operate on the ManagedImplementation metadata item, which won't exist until we merge this. I tested it out, and checking for a metadata item that doesn't exist is safe, so dotnet/sdk can safely merge before this.

nuget/Microsoft.Windows.CsWinRT.Authoring.targets
j0shuams: I have always seen AnyCPU / x86 / Win32 used interchangeably, which is why I did this. We have to pick a platform here because we don't make a WinRT.Host.dll for AnyCPU. We are assuming that someone who builds for AnyCPU will be fine using the x86 version of WinRT.Host.dll

manodasanW:
Thanks Joshua, I guess we don't merge this PR until the .NET SDK build gets publicly released with the associated change.

docs/authoring.md
```diff
@@ -25,15 +28,17 @@ To author your component, first create a project using the C# **Class Library (.
       - **net5.0-windows10.0.17763.0**
```
manodasanW, line 28: Do we want to also list 20348?

docs/authoring.md
```diff
@@ -4,6 +4,9 @@
 
 **Note: Authoring Support is still in preview**
 
+
+**Update: C++ customers who use project references to CsWinRT components can now use just a single project reference with .NET 6.0 RC1 -- https://dotnet.microsoft.com/download/dotnet/6.0**
```
angelazhangmsft, line 8:
```suggestion
**Update: C++ customers can reference CsWinRT components with a single project reference starting in [.NET 6.0 RC1](https://dotnet.microsoft.com/download/dotnet/6.0).

```

docs/authoring.md
```diff
@@ -109,6 +114,8 @@ Consuming a C#/WinRT component from a C++/WinRT desktop application is supported
 
 - For project references, you also currently need a reference to the component's generated WinMD. The WinMD can be found in the output directory of the authored component's project. To add both the project and WinMD references, right-click on the native project node, and click **Add** -> **Reference**. Select the C#/WinRT component project under the **Projects** node and the generated WinMD file from the **Browse** node.
```
angelazhangmsft, line 115: Maybe instead of the note below, preface this bullet with "If your authored component is built with .NET 5, you also need a reference to the WinMD from referencing C++ apps?

docs/authoring.md
```diff
@@ -2,7 +2,8 @@
 
 ## Overview
 
-**Note: Authoring Support is still in preview**
+
+**Note**: C#/WinRT authoring and hosting project reference support is available with [.NET 6 RC1](https://dotnet.microsoft.com/download/dotnet/6.0) and later. This means the following scenario is **not supported in .NET 5**: C# app with a project reference to a C#/WinRT Component.
```
manodasanW, line 6:
```suggestion
**Note**: C#/WinRT authoring and hosting project reference support is available with [.NET 6 RC1](https://dotnet.microsoft.com/download/dotnet/6.0) and later. This means the following scenario is **not supported in .NET 5**: C# app with a project reference to a C#/WinRT authored Component.

```

docs/authoring.md
```diff
@@ -119,7 +124,9 @@ For native consumption of C#/WinRT components, you also need to create a manifes
 
 ### Consuming from C# applications
 
-Consuming a C#/WinRT component from C#/.NET 5 is supported by both package reference or project reference. This scenario is equivalent to consuming any ordinary C# class library and does not involve WinRT activation in most cases.
+**Note** Starting with C#/WinRT 1.3.5, this requires .NET 6.
```
manodasanW, line 127: Is this true also for package reference?

j0shuams: No, maybe 

`**Note** Starting with C#/WinRT 1.3.5, project references require .NET 6.`

manodasanW (approved):

PR #922, opened 2021-07-23
Support multiple authored runtime components

### LINKED ISSUE #774: C++/WinRT can only consume 1 authored component nupkg
**Describe the bug**
Only one authored component gets used properly by a C++/WinRT app referencing two authored components. 

**To Reproduce**
1. Make two separate C#/WinRT components
2. Have both generate nupkgs
3. add those nupkgs as references to a C++/WinRT app

**Expected behavior**
C++ projection gets made for both component winmds.

j0shuams:
Fixes #774 

This PR changes the transitive targets file we have to be easier to modify using XmlPeek and XmlPoke, and introduces a target to the Authoring.targets  that uses the Xml functions to create a unique targets file for each authored component.


Scottj1s (approved):
looks good!

PR #931, opened 2021-07-26
Run IIDOptimizer on WinRT.Runtime

### LINKED ISSUE #862: Run IIDPatcher on WinRT.Runtime
We discussed during initial work on the patcher that it would be good to run the tool on winrt.runtime and use that version of winrt.runtime for the later uses of the patcher. Meaning, use the patched runtime when patching projections and tests. 

Work revolves around some custom build steps. Might also be worth investigating if we can cut the dependency of the patcher on winrt.runtime

j0shuams:
Fixes #862 

We patch WinRT.Runtime by using the target assembly for patching to resolve the WinRT types. 
I moved the targets related to the IIDOptimizer to their own file, and now we can use a Directory.Build.targets file in WinRT.Runtime project to import these targets. The `cswinrt.sln` change is to add a build dependency for WinRT.Runtime to IIDOptimizer project.

FYI @jkoritzinsky

### DIFF
--- a/src/Perf/IIDOptimizer/Program.cs
+++ b/src/Perf/IIDOptimizer/Program.cs
@@ -56,12 +56,9 @@ namespace GuidPatch
             await rootCommand.InvokeAsync(args);            
         }
 
-        private static int GuidPatch(string targetAssembly, string outputDirectory, IEnumerable<FileInfo> references)
-        {
-            var resolver = new ReferenceAssemblyResolver(references);
-            try 
-            {
-                var readerParameters = new ReaderParameters(ReadingMode.Deferred)
+        private static ReaderParameters MakeReaderParams(ReferenceAssemblyResolver resolver)
+        { 
+            return new ReaderParameters(ReadingMode.Deferred) 
                 {
                     ReadWrite = true,
                     InMemory = true,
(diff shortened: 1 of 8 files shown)


src/Perf/IIDOptimizer/Program.cs
```diff
@@ -76,13 +74,34 @@ private static int GuidPatch(string targetAssembly, string outputDirectory, IEnu
...
+
+                /// Don't patch twice 
                 if (targetAssemblyDefinition.MainModule.Types.Any(typeDef => typeDef.Name == "<GuidPatcherImplementationDetails>"))
                 {
                     Console.WriteLine("Target assembly has already been patched. Exiting early as there is no work to do.");
                     return -2;
                 }
 
-                var guidPatcher = new GuidPatcher(winRTRuntimeAssembly, targetAssemblyDefinition);
+                var guidPatcher = new GuidPatcher(winRTRuntimeAssembly!, targetAssemblyDefinition);
```
jkoritzinsky, line 104: I think it would be worth handling the `winRTRuntimeAssembly == null` case explicitly like how the "don't patch twice" case is handled.

manodasanW (approved):

PR #932, opened 2021-07-27
Fix value type caching

ujjwalchadha:
This fixes the issue with value types being cached and not garbage collected thus causing a problem when a native pointer is recycled and pointed to a value type before. This PR creates a `ABI.System.Nullable<> wrapper` on the type, thus allowing recreation.


src/WinRT.Runtime/ComWrappersSupport.net5.cs
```diff
@@ -106,10 +106,13 @@ private static T CreateRcwForComObject<T>(IntPtr ptr, bool tryUseCache)
                 winrtObj.Resurrect();
             }
 
+            if (rcw.GetType().FullName.StartsWith("ABI.System.Nullable`1") || rcw.GetType().FullName.StartsWith("ABI.Windows.Foundation.IReferenceArray`1"))
```
manodasanW, line 109: Can we somehow do pattern matching to handle this.  This right now I feel like will have some additional cost for each object creation.

manodasanW: does adding ABI.System.Nullable<T> nullable => nullable.Value work?

ujjwalchadha: ABI.System.Nullable<T> nullable => nullable.Value does not seem to work when T is of object type. Nullable.Value requires the exact type of generic parameter.

manodasanW: I wonder if it makes sense to create a non generic version of ABI.System.Nullable that we wrap the object return around and we detect that here and call .Value on it which returns it as an object and we cast it.  Thoughts?

manodasanW: It could also just really be a new wrapper around the .Value object technically.

src/Tests/UnitTest/TestComponentCSharp_Tests.cs
```diff
@@ -2601,5 +2602,54 @@ private void TestSupportedOSPlatformWarnings()
             WarningStatic.WarningEvent += (object s, Int32 v) => { }; // warning CA1416
         }
 #endif
+
+        [Fact]
+        private void TestPnpProperties()
+        {
+            TestPnpPropertiesAsync().Wait(10000);
+        }
```
jkoritzinsky, line 2610: Make the test `async` instead of doing sync-over-async.

```suggestion
        private async Task TestPnpProperties()
        {
            await TestPnpPropertiesAsync();
        }
```

Scottj1s: tests can't be async - the runner will terminate before they finish.  we have to use this pattern.

jkoritzinsky: xUnit supports `async Task` tests and keeps the runner alive until the tasks terminate. We use this throughout dotnet/runtime.

manodasanW (approved):

PR #933, opened 2021-07-27
Don't use exceptions for flow control when loading dlls.

jlaanstra:
This PR changes the way CsWinRT loads dlls to use the Try* pattern. Currently lookup throws an exception when a specific dll is not found resulting in a lot of first chance exceptions being thrown slowing down debugging.

### DIFF
--- a/src/cswinrt/strings/WinRT.cs
+++ b/src/cswinrt/strings/WinRT.cs
@@ -136,7 +136,7 @@ namespace WinRT
                 return false;
             }
 
-            var getActivationFactory = Platform.GetProcAddress(_moduleHandle, nameof(DllGetActivationFactory));
+            var getActivationFactory = Platform.GetProcAddress(moduleHandle, nameof(DllGetActivationFactory));
             if (getActivationFactory == IntPtr.Zero)
             {
                 module = null;
--- a/src/cswinrt/strings/WinRT.cs
+++ b/src/cswinrt/strings/WinRT.cs
@@ -47,16 +47,6 @@ namespace WinRT
         [DllImport("kernel32.dll", SetLastError = true, BestFitMapping = false)]
         internal static extern IntPtr GetProcAddress(IntPtr moduleHandle, [MarshalAs(UnmanagedType.LPStr)] string functionName);
 
-        internal static T GetProcAddress<T>(IntPtr moduleHandle)
-        {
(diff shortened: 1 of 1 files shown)


jlaanstra:
Internal bug for this: 34604102

src/cswinrt/strings/WinRT.cs
```diff
@@ -112,46 +102,61 @@ internal class DllModule
...
-            if (_moduleHandle == IntPtr.Zero)
+            if (moduleHandle == IntPtr.Zero)
             {
-                Marshal.ThrowExceptionForHR(Marshal.GetHRForLastWin32Error());
+                module = null;
+                return false;
             }
 
-            _GetActivationFactory = Platform.GetProcAddress<DllGetActivationFactory>(_moduleHandle);
+            var getActivationFactory = Platform.GetProcAddress(_moduleHandle, nameof(DllGetActivationFactory));
```
manodasanW, line 139: moduleHandle rather than _moduleHandle?

manodasanW:
Can you locally run build.cmd and confirm everything builds fine and all the tests it runs pass?

jlaanstra:
@manodasanW Build completed locally. Some test didn't run because it couldn't find a test adapter, but the tests that did run passed. Probably wanna do a CI build or a local build to also validate yourself.

src/cswinrt/strings/WinRT.cs
```diff
@@ -47,16 +47,6 @@ internal class Platform
         [DllImport("kernel32.dll", SetLastError = true, BestFitMapping = false)]
         internal static extern IntPtr GetProcAddress(IntPtr moduleHandle, [MarshalAs(UnmanagedType.LPStr)] string functionName);
 
-        internal static T GetProcAddress<T>(IntPtr moduleHandle)
```
manodasanW, line 50: Did a build of your branch.  There are still references to this in other parts of the code base like here: https://github.com/microsoft/CsWinRT/blob/75ffe544984b77c2792f36bde07a3cfd6a5c4656/src/WinRT.Runtime/ExceptionHelpers.cs#L44 .  Might want to restore it.

jlaanstra: I added that back. Maybe I didn't push yet.

manodasanW (approved):

PR #942, opened 2021-07-29
Check that a class correctly implements its interfaces

### LINKED ISSUE #940: Generator crash on class that doesn't implement an interface correctly
The code below is incorrect (Class1 doesn't implement Interface1 correctly since f takes a float instead of a MyEnum).
However, the generator will crash, instead of emitting diagnostics.

```cs
namespace CppComponent
{
    public enum MyEnum
    {
        Value1 = 42,
        Value2 = 99
    }

    public interface Interface1
    {
        void f(int x, MyEnum y);
    }

    sealed public class Class1 : Interface1
    {
        public void f(int x, float y);
    }
}
```

asklar:
Fixes #940


asklar:
@j0shuams correctly pointed out that this is raising an error in the CustomButton test. When the method implementation isn't coming from the class directly, but from one of its base types, the current algorithm still returns null. Will work on this a bit more.

src/Authoring/WinRT.SourceGenerator/WinRTRules.cs
```diff
@@ -172,6 +172,11 @@ private static DiagnosticDescriptor MakeRule(string id, string title, string mes
...
             "Method '{0}' has parameter '{1}' which is not an array, and which has either a "
             + "ReadOnlyArray attribute or a WriteOnlyArray attribute. Windows Runtime does "
-            + "not support marking non-array parameters with ReadOnlyArray or WriteOnlyArray.");
-    } 
+            + "not support marking non-array parameters with ReadOnlyArray or WriteOnlyArray.");
+
+        public static DiagnosticDescriptor UnimplementedInterface = MakeRule(
+            "CsWinRT1027",
+            "Class incorrectly implements an interface",
+            "Class '{0}' does not correctly implement interface '{1}' because member '{2}' is missing or non-public.");
```
j0shuams, line 180: Remove the period at the end; the messages are either multiple sentences (last ending with period) or one sentence (no period) 

https://github.com/dotnet/roslyn-analyzers/blob/main/src/Microsoft.CodeAnalysis.Analyzers/Microsoft.CodeAnalysis.Analyzers.md#rs1032-define-diagnostic-message-correctly

j0shuams (approved):

src/Authoring/WinRT.SourceGenerator/DiagnosticUtils.cs
```diff
@@ -124,7 +124,30 @@ private void CheckDeclarations()
                     CheckProperties(props, classId);
 
                     // check types -- todo: check for !valid types
-                    CheckMethods(publicMethods, classId);
+                    CheckMethods(publicMethods, classId);
+
+                    var iWinRTObject = model.Compilation.GetTypeByMetadataName("WinRT.IWinRTObject");
```
j0shuams, line 129: would you consider moving this (the new logic) to a helper function and calling that here?

asklar: sure. also made the thing about 'comment this code' automatic now for tests, should save us from wasting time when we forget to comment things out : )

j0shuams: Great, thank you!

PR #954, opened 2021-08-08
Optimize main

AtariDreams:
Throwing is extremely expensive and should be done for exceptional cases, not things like help

### DIFF
--- a/src/cswinrt/main.cpp
+++ b/src/cswinrt/main.cpp
@@ -11,6 +11,9 @@ namespace cswinrt
     using namespace std::literals;
     using namespace std::experimental::filesystem;
     using namespace winmd::reader;
+    
+    int result;
+    writer w;
 
     inline auto get_start_time()
     {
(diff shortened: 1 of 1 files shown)


manodasanW:
@AtariDreams Thanks for your contribution.  I do agree that exceptions are expensive but I do wonder if it makes a difference in this context of a source generator tool and the scenario this exception is thrown.  But regardless this can be considered an improvement.  I tried building your changes and seem to be running into a build error,  Can you run [build.cmd](https://github.com/microsoft/CsWinRT/blob/master/CONTRIBUTING.md#building-the-cwinrt-repo) and confirm that you are able to build this change without issues?

Here is what I ran into:

`
"S:\CsWinRT\src\cswinrt.sln" (default target) (1) ->
"S:\CsWinRT\src\Tests\TestComponentCSharp\TestComponentCSharp.vcxproj.metaproj" (default target) (2) ->
"S:\CsWinRT\src\cswinrt\cswinrt.vcxproj.metaproj" (default target) (3) ->
"S:\CsWinRT\src\cswinrt\cswinrt.vcxproj" (default target) (7) ->
(ClCompile target) ->
  S:\CsWinRT\src\cswinrt\main.cpp(188,1): error C2220: the following warning is treated as an error [S:\CsWinRT\src\cswinrt\cswinrt.vcxproj]
  S:\CsWinRT\src\cswinrt\text_writer.h(439,17): error C2064: term does not evaluate to a function taking 2 arguments [S:\CsWinRT\src\cswinrt\cswinrt.vcxpro
j]
  S:\CsWinRT\src\cswinrt\text_writer.h(278,42): error C2664: 'void cswinrt::writer_base<T>::write(const uint64_t)': cannot convert argument 1 from 'const F
irst' to 'const uint64_t' [S:\CsWinRT\src\cswinrt\cswinrt.vcxproj]
`

manodasanW:
Closing old PR with unresolved comments.  Feel free to reactivate once they are addressed.

Issue #182, opened 2020-04-23
Perf: Minimize QueryInterface calls, implement FastABI (Consumption)


Scottj1s:
### Background
WinRT API designs are class-based, but interfaces are used to collect the API surface into versioned units. For mature APIs, such as WinUI and IXP, that means objects will implement many interfaces. Use of those interfaces creates overhead in several ways:

1. Runtime performance overhead due to having to QI and Release for the interfaces.  
2. Increased memory consumption for caching interface pointers.
3. Binary growth due to code generation for QI-ing for multiple interfaces.

FastABI addresses these problems by generating a "flattened" default interface, appending methods for all non-default exclusive-to interfaces (e.g., IButtonBase*) across the inheritance hierarchy for the runtime class, through a well-defined algorithm.  This flattened interface is a superset of the default interface and is thus backward compatible with it for unenlightened client code.  The class implementation is still responsible for responding to QI calls on non-default exclusive-to interfaces for unenlightened clients. These QIs are typically implemented as fast, lightweight COM tear-offs using the C++/WinRT fast forwarder library.  Client code that is built with FastABI-enabled language projections look for a "fastabi" attribute attached to a runtime class in metadata, using it to generate the corresponding flattened vtable layout, enabling FastABI gains in client code.  So, FastABI benefits only accrue when both server (class implementation) and client are FastABI-enabled.

### Consumption
For C#/WinRT, the primary benefit for FastABI is on the client (consumption) side, complementing the support in native APIs like WinUI.  This issue is about implementing C#/WinRT FastABI support on the client side, as described above, comprised of these steps:

1. Generating the flattened default interface based on the FastABI algorithm.
2. Dismantling the lazy querying and caching of non-default exclusive-to interfaces in generated class code.
3. Adding FastABI-enabled classes to test components (TestWinRT preferably, as this is a language-general feature).
4. Testing FastABI-enabled classes in the unit test project.

### Authoring
As described, FastABI provides the biggest returns for complex class hierarchies with multiple versions of exclusive-to interfaces.  A conscious tradeoff is made for implementing these interfaces as COM tear-offs instead, "lazily" allocating them on demand.  Whether this applies to C#/WinRT authored classes, and whether tear-offs can be as efficiently implemented, is an open question.  It seems unlikely that C#/WinRT authored types would ever approach the complexity of WinUI, for example.  We may conclude that it makes little sense to implement FastABI authoring support in C#/WinRT.

angelazhangmsft:
[...]

PR #998, opened 2021-09-21
Fast abi consumption

ujjwalchadha:
This PR enables fast abi consumption in cswinrt as per #182 

For quick reference, here is the test idl and the [corresponding generated code](https://microsoft-my.sharepoint-df.com/:f:/p/ujchadha/EgNS2pH2VvxAvRjPmolAVN0BfrXc1poXiEQoKyOM2QkmUQ?e=tryW7h)

Note: Although this builds and passes the current tests, this request is only for **code design review** at the moment and should not be merged before adding some more testing.


src/WinRT.Runtime/ApiCompatBaseline.net5.0.txt
```diff
@@ -1,4 +1,5 @@
 Compat issues with assembly WinRT.Runtime:
 MembersMustExist : Member 'public WinRT.MarshalString.HStringHeader WinRT.MarshalString.HStringHeader WinRT.MarshalString._header' does not exist in the implementation but it does exist in the contract.
 TypesMustExist : Type 'WinRT.MarshalString.HStringHeader' does not exist in the implementation but it does exist in the contract.
-Total Issues: 2
+MembersMustExist : Member 'public void WinRT.SingleInterfaceOptimizedObject..ctor(System.Type, WinRT.IObjectReference)' does not exist in the implementation but it does exist in the contract.
```
manodasanW, line 4: This is API compat breaking, we should look at adding multiple constructors instead rather than modifying existing one.

src/WinRT.Runtime/ObjectReference.cs
```diff
@@ -170,12 +170,22 @@ public virtual unsafe int TryAs<T>(Guid iid, out ObjectReference<T> objRef)
...
+            objRef.ReferenceTrackerPtr = ReferenceTrackerPtr;
+            return objRef;
+        }
+
+    // Used only as part of the GetInterface implementation where the
+    // result is an reference passed across the ABI and doesn't need to
+    // be tracked as an internal reference.  This is separate to handle
+    // tear off aggregate scenario where releasing an reference can end up
+    // deleting the tear off interface.
+    public virtual unsafe int TryAs(Guid iid, out IntPtr ppv)
```
manodasanW, line 188: indenting is off.

manodasanW:
It would be interesting to see if we can get some benchmarks to see what the diff we see as is with and without the fastabi attribute for essentially the same class.

src/cswinrt/code_writers.h
```diff
@@ -5499,20 +5581,65 @@ private % AsInternal(InterfaceTag<%> _) => _default;
...
+                    {
+                        auto default_interface = get_default_interface(type);
+                        auto default_iface_method_count = 0;
+                        for_typedef(w, get_type_semantics(default_interface), [&](TypeDef iface)
+                        {
+                            default_iface_method_count = distance(iface.MethodList());
+                        });
+
+                        w.write(R"(
+protected unsafe % List<Func<IntPtr>> GetDefaultFactories()
```
manodasanW, line 5597: This is called multiple times by each parent class.  We should look at making it only constructed once.  We should also consider making this a readonly array or a readonly fixed capacity list as we know ahead of time the number of entries.

src/cswinrt/code_writers.h
```diff
@@ -5499,20 +5581,65 @@ private % AsInternal(InterfaceTag<%> _) => _default;
...
+var list = new List<Func<IntPtr>>();
+%
+return list;
+})",
+                        hierarchy_index == 0 ? "virtual" : "override",
+                        bind([&](writer& w)
+                        {
+                            for (auto i = 0; i < hierarchy_index; i++)
+                            {
+                                w.write("list.Add(() => (*(delegate* unmanaged[Stdcall]<IntPtr, IntPtr>**)_inner.ThisPtr)[%](_inner.ThisPtr));\n", 6 + default_iface_method_count + i);
```
manodasanW, line 5608: Would it be better if this was instead an array of ints and we used NativeObject to get the thisptr and then use the int we get from this list to do the offset.  That would mean we eliminate the need for the funcs.

src/cswinrt/helpers.h
```diff
@@ -994,4 +994,202 @@ namespace cswinrt
...
+        {
+            // compare relative contracts
+            auto relativeContractValueIface = -1 * get_number_of_attributes(iface, "Windows.Foundation.Metadata"sv, "PreviousContractVersionAttribute"sv);
+            auto relativeContractValueOtherIface = -1 * get_number_of_attributes(otherIface, "Windows.Foundation.Metadata"sv, "PreviousContractVersionAttribute"sv);
+            if (relativeContractValueIface != relativeContractValueOtherIface)
+                return relativeContractValueIface < relativeContractValueOtherIface;
+
+            //compare contract versions if they exist
+            auto contractVersionIface = get_contract_version(iface);
+            auto contractVersionOtherIface = get_contract_version(iface);
```
manodasanW, line 1109: otheriface here and below.

src/cswinrt/helpers.h
```diff
@@ -994,4 +994,202 @@ namespace cswinrt
...
+            if (contractVersionIface.has_value() && contractVersionOtherIface.has_value() && contractVersionIface.value() != contractVersionOtherIface.value())
+                return contractVersionIface.value() < contractVersionOtherIface.value();
+
+            //compare versions
+            auto versionIface = get_version(iface);
+            auto versionOtherIface = get_version(iface);
+            if (versionIface.has_value() && versionOtherIface.has_value() && versionIface.value() != versionOtherIface.value())
```
manodasanW, line 1116: is this has value check guaranteed to return true for both.

src/cswinrt/code_writers.h
```diff
@@ -2346,8 +2350,13 @@ remove => %.ErrorsChanged -= value;
                 auto interface_abi_name = write_type_name_temp(w, interface_type, "%", typedef_name_type::ABI);
 
                 auto is_default_interface = has_attribute(ii, "Windows.Foundation.Metadata", "DefaultAttribute");
-                auto target = wrapper_type ? write_type_name_temp(w, interface_type, "((%) _comp)") :  (is_default_interface ? "_default" : write_type_name_temp(w, interface_type, "AsInternal(new InterfaceTag<%>())"));
-                if (!is_default_interface && !wrapper_type)
+                auto is_fast_abi = has_attribute(type, "Windows.Foundation.Metadata"sv, "FastAbiAttribute"sv);
```
Scottj1s, line 2353: use your new is_fast_abi_class?

src/cswinrt/code_writers.h
```diff
@@ -2346,8 +2350,13 @@ remove => %.ErrorsChanged -= value;
...
+                auto sem = get_type_semantics(ii.Interface());
+                auto is_exclusive_interface = std::holds_alternative<type_definition>(sem) && has_attribute(std::get<type_definition>(sem), "Windows.Foundation.Metadata"sv, "ExclusiveToAttribute"sv);
```
Scottj1s, line 2355: and is_exclusive_to here

src/cswinrt/helpers.h
```diff
@@ -994,4 +994,202 @@ namespace cswinrt
...
+            }
+        }
+        return exclusive_ifaces;
+    }
+
+    void sort_fast_abi_ifaces(std::vector<TypeDef>& fast_abi_ifaces)
+    {
+        std::sort(fast_abi_ifaces.begin(), fast_abi_ifaces.end(), [](TypeDef const& iface, TypeDef const& otherIface)
+        {
+            // compare relative contracts
```
Scottj1s, line 1101: I don't see the corresponding logic in cppwinrt - no mention of PreviousContractVersion here:
[https://github.com/microsoft/cppwinrt/blob/3393e79d5c405ffee25d468f61f921c6216d21fe/cppwinrt/helpers.h#L451](https://github.com/microsoft/cppwinrt/blob/3393e79d5c405ffee25d468f61f921c6216d21fe/cppwinrt/helpers.h#L451)

src/cswinrt/helpers.h
```diff
@@ -994,4 +994,202 @@ namespace cswinrt
...
+        }
+        return get_attribute_value<uint32_t>(get_attribute(type, "Windows.Foundation.Metadata"sv, "VersionAttribute"sv), 0);
+    }
+
+    bool is_fast_abi_class(TypeDef const& type)
+    {
+        return has_attribute(type, "Windows.Foundation.Metadata"sv, "FastAbiAttribute"sv);
+    }
+
+    std::optional<TypeDef> find_fast_abi_class_type(TypeDef const& iface)
```
Scottj1s, line 1059: cppwinrt caches this - fastabi_cache - should we do the same?

src/cswinrt/helpers.h
```diff
@@ -994,4 +994,202 @@ namespace cswinrt
...
+            if (contractVersionIface.has_value() && contractVersionOtherIface.has_value() && contractVersionIface.value() != contractVersionOtherIface.value())
+                return contractVersionIface.value() < contractVersionOtherIface.value();
+
+            //compare versions
+            auto versionIface = get_version(iface);
+            auto versionOtherIface = get_version(iface);
+            if (versionIface.has_value() && versionOtherIface.has_value() && versionIface.value() != versionOtherIface.value())
+                return versionIface.value() < versionOtherIface.value();
+
+            //compare strings
```
Scottj1s, line 1119: // compare type names

src/cswinrt/cswinrt.vcxproj
```diff
@@ -47,7 +47,7 @@
       <AdditionalDependencies>kernel32.lib;user32.lib;%(AdditionalDependencies);windowsapp.lib;advapi32.lib;shlwapi.lib</AdditionalDependencies>
     </Link>
     <ClCompile>
-      <TreatWarningAsError>true</TreatWarningAsError>
+      <TreatWarningAsError>false</TreatWarningAsError>
```
Scottj1s, line 50: what warning?

src/Tests/UnitTest/TestComponent_Tests.cs
```diff
@@ -845,12 +845,46 @@ public void TestVectorGetMany()
...
+        [Fact]
+        public void Fast_Abi_Composition()
+        {
+            var compositor = new test_component_fast.Composition.Compositor();
+            var sv = compositor.CreateSpriteVisual();
+            sv.Offset = 10;
+            sv.StartAnimationGroup();
+            Assert.Equal(10, sv.Offset);
+            Assert.Equal(10, sv.Pad);
+        }
```
Scottj1s, line 879: what about the rest of the cppwinrt TEST_CASE("Composition") test logic?

src/Tests/UnitTest/TestComponent_Tests.cs
```diff
@@ -845,12 +845,46 @@ public void TestVectorGetMany()
             Assert.True(composableObjects.GetRange(1, 3).SequenceEqual(interfaceSubset));
         }
 
+#if NET5_0
+        [Fact]
+        public void Fast_Abi_Simple()
+        {
+            var simple = new test_component_fast.Simple();
+            Assert.Equal("Method1", simple.Method1());
```
Scottj1s, line 853: especially for this change, which rearranges vtable layouts, it would be great to have validation logic on the test_component_fast side, as well (see my comment on that PR).

Scottj1s:
oops - lost code_writers.h on iteration 7

src/cswinrt/code_writers.h
```diff
@@ -6371,6 +6410,36 @@ private struct InterfaceTag<I>{};
...
+                    {
+                        auto default_interface = get_default_interface(type);
+                        auto default_iface_method_count = 0;
+                        for_typedef(w, get_type_semantics(default_interface), [&](TypeDef iface)
+                            {
+                                default_iface_method_count = distance(iface.MethodList());
+                            });
+
+                        w.write(R"(
+protected unsafe % List<Func<IObjectReference>> GetDefaultFactories()
```
Scottj1s, line 6426: Given that this function is returning a list of accessors to the fastabi-flattened default interfaces for the composition hierarchy, and not a list of factories to create them, a better name might be GetDefaultInterfaceAccessors or just GetDefaultInterfaces

src/Tests/UnitTest/TestComponent_Tests.cs
ujjwalchadha: As per offline discussion, a new issue has been created for this.

src/Tests/UnitTest/TestComponent_Tests.cs
ujjwalchadha: This is a general composition test and not a fast_abi_test and was not added in this PR.
The fast abi tests added in this PR are: Fast_Abi_Simple and Fast_Abi_Composition

src/cswinrt/helpers.h
```diff
@@ -995,4 +996,238 @@ namespace cswinrt
...
+    {
+        auto sem = get_type_semantics(classType.Extends());
+        if (std::holds_alternative<type_definition>(sem))
+        {
+            return get_class_hierarchy_index(std::get<type_definition>(sem)) + 1;
+        }
+        return 0;
+    }
+
+    bool is_interfaces_equal(TypeDef const& interface1, TypeDef const& interface2)
```
Scottj1s, line 1026: rename: are_interfaces_equal

src/cswinrt/cswinrt.vcxproj
manodasanW: Was this resolved?  It is still false.

src/WinRT.Runtime/ObjectReference.cs
```diff
@@ -177,6 +177,16 @@ public virtual unsafe int TryAs<T>(Guid iid, out ObjectReference<T> objRef)
             return hr;
         }
 
+        public virtual unsafe ObjectReference<IUnknownVftbl> AsKnownPtr(IntPtr ptr)
```
manodasanW, line 180: Should we rename this to AsFastAbiPtr?

src/WinRT.Runtime/ObjectReference.cs
```diff
@@ -177,6 +177,16 @@ public virtual unsafe int TryAs<T>(Guid iid, out ObjectReference<T> objRef)
             return hr;
         }
 
+        public virtual unsafe ObjectReference<IUnknownVftbl> AsKnownPtr(IntPtr ptr)
+        {
+            AddRef(true);
+            var objRef = ObjectReference<IUnknownVftbl>.Attach(ref ptr);
```
manodasanW, line 183: Should we have an override for this in ObjectReferenceForContext to handle the case where this might be a non agile object?

src/cswinrt/helpers.h
manodasanW: Did we confirm this?

src/cswinrt/helpers.h
```diff
@@ -995,4 +996,236 @@ namespace cswinrt
...
+            cache[iface] = exclusiveToClass;
+            return exclusiveToClass;
+        }
+        return {};
+    }
+
+    std::pair<TypeDef, std::vector<TypeDef>> get_default_and_exclusive_interfaces(TypeDef const& classType)
+    {
+        std::pair<TypeDef, std::vector<TypeDef>> exclusive_ifaces;
+        std::vector<TypeDef> exclusive_interfaces;
```
manodasanW, line 1091: is exclusive_interfaces used?

src/cswinrt/code_writers.h
```diff
@@ -5896,14 +5913,36 @@ public static Guid PIID = Vftbl.PIID;
...
         is_exclusive_to(iface) ? "internal" : internal_if_embedded(), 
         bind<write_type_name>(iface, typedef_name_type::StaticAbiClass, false), 
-        bind<write_static_abi_class_members>(iface));
+        [&](writer& w) {
+            if (!fast_abi_class_val.has_value()) {
+                write_static_abi_class_members(w, iface, INSPECTABLE_METHOD_COUNT);
+                return;
+            }
+            auto abi_methods_start_index = INSPECTABLE_METHOD_COUNT;
+            write_static_abi_class_members(w, fast_abi_class_val.value().default_interface, abi_methods_start_index);
```
manodasanW, line 5938: Is this supposed to be the default interface?  And how are the static and factory interface methods ordering offset with the other exclusive interface methods?  And do we have any fast abi tests for static methods and factory methods?

src/WinRT.Runtime/ObjectReference.cs
ujjwalchadha: I think 'As' method suggests a cast operation when the ptr to cast to is unknown. So 'AsKnownPtr' suggests cast to this object reference but do not find the ptr as it is already known. 
So I think this is not just specific to fast abi but for any scenario where the ptr is already known or precalculated and we just need an object reference pointing to that ptr.
What do you think?

manodasanW: I agree with that, but I was thinking about how to make it clear the addref impacts the objectref that this function is being called on.  Maybe it being an instance method does that.

src/cswinrt/helpers.h
ujjwalchadha: Yes, the winmd would either have version on both or none.

manodasanW (approved):

PR #1001, opened 2021-09-30
Perf improvements

### LINKED ISSUE #825: Don't initialize lazy dictionary used for QI cache when not needed
We have lazy dictionary for caching QI calls some of which are initialized with no entries and will not have any entries ever.  Those should not need to be initialized in the first place.

### LINKED ISSUE #987: Investigate optimizing Marshaler constructor
Today in the Marshaler constructor and specifically MarshalGeneric, we initialize all the various helper functions for a given type the first time it is accessed.  But in practical terms, not all the helper functions are used by each projection and in every scenario.  Due to this, we should consider initializing them lazily to reduce the reflection code that we need to execute and could end up reducing the expense of an app.

Note during perf traces with VS profiler, this showed up as something it claimed as expensive compared to everything else in some scenarios.

### LINKED ISSUE #988: Remove need for object creation in DisposeABI
See https://github.com/microsoft/CsWinRT/blob/e372646d2c0d9a1e614c30df57bb95e580f11c4c/src/WinRT.Runtime/Marshalers.cs#L882

Per the comment, we should probably look at moving this to a vftbl call rather than constructing an ObjectReference just to call dispose.  In a couple of scenarios measured we were seeing DisposeABI take long, but it was due to constructing the object was ending up triggering GC.  If we reduce the number of objects we create like here, we might not trigger GC as often.

### LINKED ISSUE #1000: Activation factories gets created twice for both statics and factory.

manodasanW:
This PR makes several changes to improve the perf of CsWinRT projections based on what is seen in profiling:

- Avoid constructing an ObjectReference when doing a release
- Make all the MarshalGeneric reflection lambdas lazy as we do not use all the helper types in every scenario but rather a couple in most cases.  This reduces JIT cost too.
- Make all the concurrent dictionaries used as part of IDIC support lazy initialized as we don't use the caches in all scenarios and we are seeing a large # of empty ones on the finalizer.
-  Fix issue where activation factories were initialized multiple times for factory and statics when we needed only one instance
- Remove creating lazy interface dictionary when there are no members and setting size in constructor

Fixes #987 
Fixes #988 
Fixes #825 
Fixes #1000

### DIFF
--- a/src/WinRT.Runtime/ApiCompatBaseline.net5.0.txt
+++ b/src/WinRT.Runtime/ApiCompatBaseline.net5.0.txt
@@ -1,4 +1,8 @@
 Compat issues with assembly WinRT.Runtime:
+CannotRemoveAttribute : Attribute 'System.Runtime.CompilerServices.CompilerGeneratedAttribute' exists on 'WinRT.IInspectable.WinRT.IWinRTObject.get_AdditionalTypeData()' in the contract but not the implementation.
+CannotRemoveAttribute : Attribute 'System.Runtime.CompilerServices.CompilerGeneratedAttribute' exists on 'WinRT.IInspectable.WinRT.IWinRTObject.get_QueryInterfaceCache()' in the contract but not the implementation.
 MembersMustExist : Member 'public WinRT.MarshalString.HStringHeader WinRT.MarshalString.HStringHeader WinRT.MarshalString._header' does not exist in the implementation but it does exist in the contract.
 TypesMustExist : Type 'WinRT.MarshalString.HStringHeader' does not exist in the implementation but it does exist in the contract.
-Total Issues: 2
+CannotRemoveAttribute : Attribute 'System.Runtime.CompilerServices.CompilerGeneratedAttribute' exists on 'WinRT.SingleInterfaceOptimizedObject.WinRT.IWinRTObject.get_AdditionalTypeData()' in the contract but not the implementation.
+CannotRemoveAttribute : Attribute 'System.Runtime.CompilerServices.CompilerGeneratedAttribute' exists on 'WinRT.SingleInterfaceOptimizedObject.WinRT.IWinRTObject.get_QueryInterfaceCache()' in the contract but not the implementation.
+Total Issues: 6
--- a/src/WinRT.Runtime/IInspectable.net5.cs
+++ b/src/WinRT.Runtime/IInspectable.net5.cs
@@ -8,8 +8,10 @@ namespace WinRT
         IObjectReference IWinRTObject.NativeObject => _obj;
         bool IWinRTObject.HasUnwrappableNativeObject => true;
 
-        ConcurrentDictionary<RuntimeTypeHandle, IObjectReference> IWinRTObject.QueryInterfaceCache { get; } = new();
-        ConcurrentDictionary<RuntimeTypeHandle, object> IWinRTObject.AdditionalTypeData { get; } = new();
+        private Lazy<ConcurrentDictionary<RuntimeTypeHandle, IObjectReference>> _lazyQueryInterfaceCache = new();
+        ConcurrentDictionary<RuntimeTypeHandle, IObjectReference> IWinRTObject.QueryInterfaceCache => _lazyQueryInterfaceCache.Value;
+        private Lazy<ConcurrentDictionary<RuntimeTypeHandle, object>> _lazyAdditionalTypeData = new();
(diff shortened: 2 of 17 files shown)


src/WinRT.Runtime/ApiCompatBaseline.net5.0.txt
```diff
@@ -1,4 +1,8 @@
 Compat issues with assembly WinRT.Runtime:
+CannotRemoveAttribute : Attribute 'System.Runtime.CompilerServices.CompilerGeneratedAttribute' exists on 'WinRT.IInspectable.WinRT.IWinRTObject.get_AdditionalTypeData()' in the contract but not the implementation.
```
manodasanW, line 2: This is just an attribute change so I do not think this is breaking or requires an assembly version change.

Scottj1s (approved):

src/WinRT.Runtime/IInspectable.net5.cs
```diff
@@ -6,10 +6,12 @@ namespace WinRT
...
     {
         IObjectReference IWinRTObject.NativeObject => _obj;
-        bool IWinRTObject.HasUnwrappableNativeObject => true;
-
-        ConcurrentDictionary<RuntimeTypeHandle, IObjectReference> IWinRTObject.QueryInterfaceCache { get; } = new();
-        ConcurrentDictionary<RuntimeTypeHandle, object> IWinRTObject.AdditionalTypeData { get; } = new();
+        bool IWinRTObject.HasUnwrappableNativeObject => true;
+
+        private Lazy<ConcurrentDictionary<RuntimeTypeHandle, IObjectReference>> _lazyQueryInterfaceCache = new();
+        ConcurrentDictionary<RuntimeTypeHandle, IObjectReference> IWinRTObject.QueryInterfaceCache => _lazyQueryInterfaceCache.Value;
```
jkotas, line 12:
```suggestion
        private volatile ConcurrentDictionary<RuntimeTypeHandle, IObjectReference> = null;
        ConcurrentDictionary<RuntimeTypeHandle, IObjectReference> IWinRTObject.QueryInterfaceCache
        {
            get
            {
                var ret = _lazyQueryInterfaceCache;
                if (ret == null)
                {
                    Interlocked.CompareExchange(ref _lazyQueryInterfaceCache, new ConcurrentDictionary<RuntimeTypeHandle, IObjectReference>(), null);
                    ret = _lazyQueryInterfaceCache;
                }
                return ret;
            }
        }
```
As I have mentioned in the email conversation, `Lazy<T>` is not free and it is much cheaper to use lazy initialization using Interlocked.CompareExchange instead.

The same pattern should be used in all places that are doing the lazy initialization.

manodasanW: We plan on doing that across all Lazy uses as a separate upcoming PR.

src/WinRT.Runtime/Marshalers.cs
```diff
@@ -360,23 +360,24 @@ public class MarshalGeneric<T>
...
+            DisposeAbi = (object box) => DisposeAbiLazy.Value(box);
+            CreateMarshalerArray = (T[] array) => CreateMarshalerArrayLazy.Value(array);
+            GetAbiArray = (object box) => GetAbiArrayLazy.Value(box);
+            FromAbiArray = (object box) => FromAbiArrayLazy.Value(box);
+            FromManagedArray = (T[] array) => FromManagedArrayLazy.Value(array);
+            DisposeMarshalerArray = (object box) => DisposeMarshalerArrayLazy.Value(box);
+            DisposeAbiArray = (object box) => DisposeAbiArrayLazy.Value(box);
         }
 
         public static readonly Func<T, object> CreateMarshaler;
```
jkotas, line 379:
```suggestion
        public static readonly Func<T, object> CreateMarshaler = (T value) => CreateMarshalerLazy.Value(value);
```
It is better to initialize statics inline like this. It creates before-field-init constructor that is cheaper than the explicit constructor that you have there now.

Change all fields to use this pattern and delete `static MarshalGeneric()` above.

manodasanW: Thanks addressed.

src/WinRT.Runtime/IInspectable.net5.cs
manodasanW: Tracked with https://github.com/microsoft/CsWinRT/issues/1003

PR #1002, opened 2021-09-30
build breaks: fixed a few assumptions regarding nuget cache

Scottj1s:
AuthoringTest was not expressing it's dependencies properly
Object lifetime test was not supporting custom nuget cache folder, and not failing build


manodasanW:
I assume the authoring tests continue to pass after these changes?

manodasanW (approved):

PR #1005, opened 2021-10-01
Cache IsWindowsRuntimeType, which is expensive to read custom attributes

Scottj1s:


src/WinRT.Runtime/ComWrappersSupport.cs
```diff
@@ -395,9 +395,10 @@ internal static string GetRuntimeClassForTypeCreation(IInspectable inspectable,
             return runtimeClassName;
         }
 
-        private static bool ShouldProvideIReference(Type type)
+        private readonly static ConcurrentDictionary<Type, bool> IsWindowsRuntimeTypeCache = new ConcurrentDictionary<Type, bool>();
+        private static bool IsWindowsRuntimeType(Type type)
```
manodasanW, line 399: This is really IsIReferenceType as it excludes classes.

src/WinRT.Runtime/ComWrappersSupport.cs
```diff
@@ -395,9 +395,10 @@ internal static string GetRuntimeClassForTypeCreation(IInspectable inspectable,
             return runtimeClassName;
         }
 
-        private static bool ShouldProvideIReference(Type type)
+        private readonly static ConcurrentDictionary<Type, bool> IsWindowsRuntimeTypeCache = new ConcurrentDictionary<Type, bool>();
```
manodasanW, line 398: Should the caching be done at Projections.IsTypeWindowsRuntimeType instead?

src/WinRT.Runtime/ComWrappersSupport.cs
```diff
@@ -395,9 +395,10 @@ internal static string GetRuntimeClassForTypeCreation(IInspectable inspectable,
...
+        private static bool IsWindowsRuntimeType(Type type)
         {
-            static bool IsWindowsRuntimeType(Type type)
+            static bool IsWindowsRuntimeTypeHelper(Type type)
             {
                 if ((type.GetCustomAttribute<WindowsRuntimeTypeAttribute>() is object) ||
```
manodasanW, line 403: I feel like this is equivalent to Projections.IsTypeWindowsRuntimeType, is it not?

src/WinRT.Runtime/ComWrappersSupport.cs
Scottj1s: it includes strings, types, and delegates - all reference types

src/WinRT.Runtime/ComWrappersSupport.cs
Scottj1s: or in addition to, since they're both expensive and independent.  will do that.

src/WinRT.Runtime/ComWrappersSupport.cs
Scottj1s: I'd love to consolidate into a single cache, but they have different purposes and implementations.  To avoid risk of regression, we can maintain two caches for now.

manodasanW (approved):

PR #1007, opened 2021-10-04
Remove unused lock in custom projection type lookup.

manodasanW:
RegisterCustomAbiTypeMapping which was the API taking the write lock is actually not used anywhere starting before when cswinrt was first released.  Removing the lock along with all the read locks we were taking.

### DIFF
--- a/src/WinRT.Runtime/Projections.cs
+++ b/src/WinRT.Runtime/Projections.cs
@@ -6,7 +6,6 @@ using System.ComponentModel;
 using System.Linq;
 using System.Numerics;
 using System.Reflection;
-using System.Threading;
 using System.Windows.Input;
 using Windows.Foundation.Collections;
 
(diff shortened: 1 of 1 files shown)


Scottj1s (approved):

src/WinRT.Runtime/Projections.cs
```diff
@@ -77,17 +75,10 @@ static Projections()
             CustomTypeToAbiTypeNameMappings.Add(typeof(System.Type), "Windows.UI.Xaml.Interop.TypeName");
         }
 
+        // Deprecated
         public static void RegisterCustomAbiTypeMapping(Type publicType, Type abiType, string winrtTypeName, bool isRuntimeClass = false)
```
BenJKuhn, line 79: Just looking for clarification: Can we not remove this because it's a public / exposed method?

manodasanW: Yes, this API is part of the public API surface, so if we were to remove it and someone was calling it, their code would break.  In this case we can potentially do a breaking API compat change given this API was only called in CsWinRT projections built before CsWinRT 1.0 or earlier it seems based on my analysis.  But I am still considering the possibility there might be a future use for this API with some of the generic changes for specialized versions we are considering.

BenJKuhn (approved):

raghavthind2005 (approved):

PR #1009, opened 2021-10-04
Embedded sources: cswinrt

j0shuams:
This includes changes needed to cswinrt for embedded support. 

We now generate internal sources, based on a new argument to cswinrt.exe. 
And we update the strings/ sources, reflect the embedded changes needed for WinRT.Runtime.   

Note that the EMBED constant will never be defined (it is coming in a later PR), so these changes are inert.


manodasanW:
Looks good, assuming the comment regarding RegisterProjectionAssembly is addressed.

PR #1018, opened 2021-10-11
Convert activation and error info delegates to function pointers

jkotas:
Avoids overhead of creating the marshaled delegates


jkotas:
@manodasanW PTLA

PR #1036, opened 2021-10-29
A few optimizations around string marshaling

Scottj1s:
Notes:
- For generic/array/struct input strings, the object-based MarshalString must remain.
- For singleton input strings, MarshalString.Pinnnable now uses a fixed statement to pin, vs a GCHandle allocation
- Introduced a pinnable pattern we can implement for other blittable input types (structs and arrays)
- The "using var" with Dispose() pattern introduces nested try/finally blocks, one per var.  The generated IL is larger than our existing codegen with combined finally blocks.  
- There is no advantage to "ref struct" here - the fixed statement only required GetPinnableReference.


src/WinRT.Runtime/Marshalers.cs
```diff
@@ -157,15 +207,12 @@ public static unsafe MarshalerArray CreateMarshalerArray(string[] array)
                     m._marshalers[i] = MarshalString.CreateMarshaler(array[i]);
                     elements[i] = MarshalString.GetAbi(m._marshalers[i]);
                 };
-                success = true;
                 return m;
             }
-            finally
+            catch (Exception) when (Dispose(m))
```
jkotas, line 212: I have mentioned it in other PR. Using filters (ie `when` in C#) for cleanup is very atypical. The existing pattern with `finally` is better.

src/WinRT.Runtime/Marshalers.cs
```diff
@@ -25,55 +26,105 @@ public static void Dispose(this GCHandle handle)
...
+                    _value.Length,
+                    (IntPtr*)Unsafe.AsPointer(ref _header),
+                    &hstring));
+                return hstring;
+            }
+        }
+
+        public static Pinnable CreatePinnable(string value) => new(value);
+
+        public static IntPtr GetAbi(Pinnable p) => p.GetAbi();
```
jkotas, line 87: This is not right. `Pinnable` is passed by value here. The resulting IntPtr is going to be dangling pointer.

src/WinRT.Runtime/Marshalers.cs
```diff
@@ -25,55 +26,105 @@ public static void Dispose(this GCHandle handle)
...
+                var value = (string)_gchandle.Target;
+                fixed (char* chars = value)
+                { 
+                    IntPtr hstring;
                     Marshal.ThrowExceptionForHR(Platform.WindowsCreateStringReference(
-                        (char*)chars, value.Length, (IntPtr*)m._header, (IntPtr*)handle));
-                };
-                success = true;
-                return m;
+                        chars, value.Length, (IntPtr*)Unsafe.AsPointer(ref _header), &hstring));
```
jkotas, line 114: What guarantees that `_header` is pinned here? You can use `Unsafe.AsPointer` only if the item is pinned by some other means.

src/WinRT.Runtime/Marshalers.cs
```diff
@@ -25,55 +26,105 @@ public static void Dispose(this GCHandle handle)
                 handle.Free();
             }
         }
+
+#if !NET
+        public static unsafe ref readonly char GetPinnableReference(this string str)
+        {
+            return ref *(char*)((*(byte**)Unsafe.AsPointer(ref str)) + RuntimeHelpers.OffsetToStringData);
```
jkotas, line 33: This has GC hole that will lead to intermittent crashes during GC. It is not valid to convert between `ref` and unmanaged pointers like this.

src/WinRT.Runtime/Marshalers.cs
```diff
@@ -25,55 +26,105 @@ public static void Dispose(this GCHandle handle)
...
+                _value = value ?? "";
+                _header = default;
+            }
+
+            public ref readonly char GetPinnableReference()
+            {
+                return ref _value.GetPinnableReference();
+            }
+
+            public unsafe IntPtr GetAbi()
```
jkotas, line 69: Would it make sense to have debug-only check that this is only called after `GetPinnableReference` was called? (My POC had a check like that.)

src/WinRT.Runtime/ApiCompatBaseline.net5.0.txt
```diff
@@ -2,7 +2,10 @@ Compat issues with assembly WinRT.Runtime:
 CannotRemoveAttribute : Attribute 'System.Runtime.CompilerServices.CompilerGeneratedAttribute' exists on 'WinRT.IInspectable.WinRT.IWinRTObject.get_AdditionalTypeData()' in the contract but not the implementation.
 CannotRemoveAttribute : Attribute 'System.Runtime.CompilerServices.CompilerGeneratedAttribute' exists on 'WinRT.IInspectable.WinRT.IWinRTObject.get_QueryInterfaceCache()' in the contract but not the implementation.
 MembersMustExist : Member 'public WinRT.MarshalString.HStringHeader WinRT.MarshalString.HStringHeader WinRT.MarshalString._header' does not exist in the implementation but it does exist in the contract.
+MembersMustExist : Member 'public System.IntPtr System.IntPtr WinRT.MarshalString._handle' does not exist in the implementation but it does exist in the contract.
```
manodasanW, line 5: I presume there are no references to these members in generated projections and won't be an API compat breaking change.

src/WinRT.Runtime/Marshalers.cs
```diff
@@ -25,55 +26,105 @@ public static void Dispose(this GCHandle handle)
...
-            _header = IntPtr.Zero;
         }
 
-        public static unsafe MarshalString CreateMarshaler(string value)
+        public static MarshalString CreateMarshaler(string value)
         {
-            if (value == null) return null;
-
-            var m = new MarshalString();
+            return string.IsNullOrEmpty(value) ? null : new MarshalString(value);
```
manodasanW, line 101: Would this be considered a change in behavior for empty strings where it now passes IntPtr.Zero whereas previously that was only for null and empty strings got an empty HSTRING.

src/WinRT.Runtime/Marshalers.cs
```diff
@@ -25,55 +26,105 @@ public static void Dispose(this GCHandle handle)
...
+            int reserved2;
+            int reserved3;
+            int reserved4;
+            int reserved5;
+        }
+
+        private HSTRING_HEADER _header;
+        private GCHandle _gchandle;
+
+        public struct Pinnable
```
manodasanW, line 53: Should we update our uses of MarshalString in WinRT.Runtime to use the new pinnable behavior where possible?

src/WinRT.Runtime/Marshalers.cs
Scottj1s: Thanks Jan - I measured the two approaches, and finally adds 50% overhead, since it always checks a success boolean, whereas the when clause only executes in the presence of an exception.

src/WinRT.Runtime/Marshalers.cs
Scottj1s: I considered that, but thought since this is generated code, there would be little need for checks.  But you're right, probably better to have the assert - will restore it.

src/WinRT.Runtime/Marshalers.cs
Scottj1s: is there an easy way to bridge the gap for netstandard support?  I was hoping to isolate to a GetPinnableReference extension method.

src/WinRT.Runtime/Marshalers.cs
Scottj1s: ah, good catch - I made that change last minute to keep the codegen consistent with object-based marshalers

src/WinRT.Runtime/Marshalers.cs
Scottj1s: @jkotas, this was taken from your example.  _header is a struct member of the Pinnable struct.  assuming we pass by ref as you point out above, all instances are on the stack, yes?

src/WinRT.Runtime/ApiCompatBaseline.net5.0.txt
Scottj1s: correct, these members should not have been public to begin with.  our marshaling pattern is constrained to the functional interface (CreateMarshaler, GetAbi, DisposeMarshaler, FromAbi, DisposeAbi, etc)

src/WinRT.Runtime/Marshalers.cs
Scottj1s: doh!  yes, of course we should (meant to do that)

src/WinRT.Runtime/Marshalers.cs
jkotas: What is the benchmark that showed the 50% overhead? I assume that it was artificial micro-benchmark that did nothing inside the try. Once you do some actual work inside the try, the overhead should be neglible,

Using filters makes diagnostic worse. The exceptions throw inside filters are ignored, so you won't notice if anything goes wrong during Dispose.

FWIW, we are using the `bool` pattern in number of places in dotnet/runtime and we would not consider switching it to filters.

src/WinRT.Runtime/Marshalers.cs
jkotas:
```
public static unsafe ref readonly char GetPinnableReference(this string str)
{
    fixed (char * p = str) { return ref *p; }
}
```

jkotas: This may work, assuming that the runtime has no bugs.

src/WinRT.Runtime/Marshalers.cs
jkotas: > _header is a struct member of the Pinnable struct

If I am reading the code correctly, `_header` is a field on MarshalString class here. I do not see what would keep MarshalString pinned.

src/WinRT.Runtime/Marshalers.cs
Scottj1s: yes, micro-benchmark with trivial body.  ok, I'll revert

src/WinRT.Runtime/Marshalers.cs
Scottj1s: sorry, I was looking at the wrong code.  Oh, this is unfortunate - I was hoping to eliminate allocating the header on the heap, but now I don't see a way around that.

src/WinRT.Runtime/Marshalers.cs
```diff
@@ -41,37 +114,39 @@ public void Dispose()
...
+                { 
+                    IntPtr hstring;
                     Marshal.ThrowExceptionForHR(Platform.WindowsCreateStringReference(
-                        (char*)chars, value.Length, (IntPtr*)m._header, (IntPtr*)handle));
-                };
+                        chars, value.Length, (IntPtr*)_header, &hstring));
+                    success = true;
+                    return hstring;
+                }
                 success = true;
```
jkotas, line 136:
```suggestion
```
Unreachable code

src/WinRT.Runtime/Marshalers.cs
```diff
@@ -41,37 +114,39 @@ public void Dispose()
...
-            if (value == null) return null;
-
-            var m = new MarshalString();
+            return string.IsNullOrEmpty(value) ? null : new MarshalString(value);
+        }
 
+        public unsafe IntPtr GetAbi()
+        {
             bool success = false;
             try
```
jkotas, line 125: Do you actually need try/finally in this method? I would expect the cleanup to be taken care of by the calling code if something goes wrong here.

jkotas: If something went wrong and the cleanup in this method actually kicked in, I think you would actually ended with double-free of the GCHandle due to buggy implementation of `MarshalString.Dispose`. `MarshalString.Dispose` disposes the GCHandle, but it does not clear the GCHandle value, so calling `MarshalString.Dispose` twice is going to free the GCHandle twice.

src/WinRT.Runtime/Marshalers.cs
```diff
@@ -18,21 +20,92 @@ public static void Dispose(this GCHandle handle)
...
+        }
+
+        public static Pinnable CreatePinnable(string value) => new(value);
+
+        public static IntPtr GetAbi(ref Pinnable p) => p.GetAbi();
+
+        public MarshalString(string value)
+        {
+            _gchandle = GCHandle.Alloc(value, GCHandleType.Pinned);
+            _header = Marshal.AllocHGlobal(Unsafe.SizeOf<HSTRING_HEADER>());
```
jkotas, line 107: Nit: If this `Marshal.AllocHGlobal` fails, you will end up with leaking GCHandle. There is not going to be anything to free the GCHandle. 

Just pointing it out - feel free to won't fix. Memory leaks like this are not a problem in practice. If the process fails to allocate 20 bytes of memory, it is guaranteed to crash very soon.

src/WinRT.Runtime/Marshalers.cs
Scottj1s: right, in this case it's redundant

src/WinRT.Runtime/Marshalers.cs
Scottj1s: no reason to alloc in the ctor - if I delay it to GetAbi, then the ctor can only fail if the GCHandle isn't allocated.

src/WinRT.Runtime/Marshalers.cs
```diff
@@ -18,60 +20,121 @@ public static void Dispose(this GCHandle handle)
...
+
+        public MarshalString(string value)
+        {
+            _gchandle = GCHandle.Alloc(value, GCHandleType.Pinned);
+            _header = IntPtr.Zero;
+        }
+
+        public void Dispose()
+        {
+            _gchandle.Dispose();
```
jkotas, line 112:
```suggestion
            _gchandle.Dispose();
            _gchandle = default;
```

src/WinRT.Runtime/Marshalers.cs
```diff
@@ -18,60 +20,121 @@ public static void Dispose(this GCHandle handle)
...
+            return string.IsNullOrEmpty(value) ? null : new MarshalString(value);
+        }
+
+        public unsafe IntPtr GetAbi()
+        {
+            var value = (string)_gchandle.Target;
+            fixed (char* chars = value)
+            { 
+                IntPtr hstring;
+                _header = Marshal.AllocHGlobal(Unsafe.SizeOf<HSTRING_HEADER>());
```
jkotas, line 128:
```suggestion
                Debug.Assert(_header == IntPtr.Zero);
                _header = Marshal.AllocHGlobal(Unsafe.SizeOf<HSTRING_HEADER>());
```

jkotas (approved):
LGTM otherwise

PR #1038, opened 2021-11-01
Fix writing out strings with embedded quote marks in custom attributes

### LINKED ISSUE #1037: Embedded quotes in custom attributes break the projection
**Describe the bug**
Follow up from #818 - quotes in a custom attribute string value will produce invalid C# code

**To Reproduce**
add a custom attribute like:
```cs
[doc_string("the \\\"null\\\" string.\n\n")]
class C { }
```

this produces the C# code:
```cs
doc_string[@"the \"null\" string.\n\n"]
class C { }
```

which breaks as quotes in @ strings must use double quotes

**Expected behavior**
double quoted strings

**Version Info**
1.3.0

**Additional context**

asklar:
Fixes #1037

### DIFF
--- a/src/cswinrt/code_writers.h
+++ b/src/cswinrt/code_writers.h
@@ -1471,7 +1471,7 @@ remove => %;
                     }
                 },
                 [&](std::string_view type_name)
-                  {
+                {
                     std::string sanitized_type_name;
                     sanitized_type_name.reserve(type_name.length() * 2);
                     for (const auto& c : type_name)
@@ -1479,7 +1479,7 @@ remove => %;
                         sanitized_type_name += c;
                         if (c == '"')
                         {
-                          sanitized_type_name += c;
+                            sanitized_type_name += c;
                         }
                     }
                     w.write("^@\"%\"", sanitized_type_name);
--- a/src/cswinrt/code_writers.h
+++ b/src/cswinrt/code_writers.h
@@ -1476,11 +1476,11 @@ remove => %;
                     sanitized_type_name.reserve(type_name.length() * 2);
                     for (const auto& c : type_name)
                     {
-                      sanitized_type_name += c;
-                      if (c == '"')
-                      {
                         sanitized_type_name += c;
-                      }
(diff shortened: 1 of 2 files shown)


BenJKuhn (approved):

j0shuams:
@manodasanW Seems the bot failed, I cut a branch mirroring Alex's changes and spun a build here - https://microsoft.visualstudio.com/DefaultCollection/Dart/_build/results?buildId=40989703&view=results

manodasanW:
@j0shuams the bot actually worked, the error message above is just wrong / incorrect.

https://microsoft.visualstudio.com/Dart/_build/results?buildId=40973018&view=results

manodasanW (approved):

src/cswinrt/code_writers.h
```diff
@@ -1472,7 +1472,17 @@ remove => %;
                 },
                 [&](std::string_view type_name)
                 {
-                    w.write("^@\"%\"", type_name);
+                    std::string sanitized_type_name;
```
manodasanW, line 1475: It looks like the result of the above test class is `    [global::TestComponentCSharp.MyString(Content = @"a string with embedded quotes, like the string \""quotes\""")]
`.  I am wondering whether we should keep the \ in addition to double quote or whether the \ should be excluded when replacing it with a double quote.  Thoughts?  The change as-is is technically fine in that it no longer will result in a compiler error.

asklar: the `\` is coming from the string in the original IDL, so we shouldn't touch it. IIRC there are requirements around how strings with embedded quotes are processed by the C preprocessor before passing it to MIDL that require the use of the backslash

manodasanW: Right, to escape a quote in C / MIDL, you would use a / like /".  Now we are projecting that into C# which it really is the equivalent of "" when using verbatim strings with @ .   But I can see trying to do that running into weird issues with edge cases and given this is only in attributes, I am fine with leaving it as how you have it.

PR #1042, opened 2021-11-09
Improve the performance of invoking events

manodasanW:
- Reduced the # of allocations on the event invoke code path.
- Move the custom type mapping logic in FromABI to CreateObject which allows for better caching of both the RCW for the ptr and the factory function used to create the RCW.
- Removed the use of Action via MarshalDelegateInvoke in the event invoke code path on .NET 5.
- Moved to using ThreadStatic rather than ThreadLocal for the variable which stores the RCW type as it looks to be more performant according to benchmarks.
- Avoid boxing when checking if it is a reference to a managed object and also avoid the need to create a IObjectReference to do so.
- Reduce the # of IObjectReferences created by creating a specialized version of CreateCCW which gives you the ObjectReference with the right vftbl and does a QI.
- Added a set of benchmarks that helped evaluate the improvements

### DIFF
--- a/src/Benchmarks/EventPerf.cs
+++ b/src/Benchmarks/EventPerf.cs
@@ -103,20 +103,5 @@ namespace Benchmarks
             return instance;
             GC.KeepAlive(s);
         }
-
-//        [Benchmark]
-        public void AddIntEvent()
-        {
-            int z;
-            instance.IntPropertyChanged += (object sender, int value) => z = value;
-        }
-
-//        [Benchmark]
-        public void AddAndInvokeIntEvent()
-        {
-            int z;
-            instance.IntPropertyChanged += (object sender, int value) => z = value;
-            instance.RaiseIntChanged();
-        }
     }
 }
--- a/src/Benchmarks/EventPerf.cs
+++ b/src/Benchmarks/EventPerf.cs
@@ -1,0 +1,122 @@
+﻿using BenchmarkComponent;
+using BenchmarkDotNet.Attributes;
+using System;
+
+namespace Benchmarks
+{
+    [MemoryDiagnoser]
+    public class EventPerf
+    {
+        ClassWithMarshalingRoutines instance;
+        int z2;
+        ClassWithMarshalingRoutines instance2;
+        int z4;
+
+        [GlobalSetup]
+        public void Setup()
+        {
+            instance = new ClassWithMarshalingRoutines();
+            System.EventHandler<int> s = (object sender, int value) => z2 = value;
+            instance.IntPropertyChanged += s;
+
+            instance2 = new ClassWithMarshalingRoutines();
+            System.EventHandler<int> s2 = (object sender, int value) =>
+            {
+                if (sender == this)
+                    z4 = value;
+                else
+                    z4 = value * 3;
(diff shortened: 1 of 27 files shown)


src/Benchmarks/EventPerf.cs
```diff
@@ -0,0 +1,107 @@
...
+        [Benchmark]
+        public object AddAndInvokeIntEventOnNewEventSource()
+        {
+            ClassWithMarshalingRoutines instance = new ClassWithMarshalingRoutines();
+            int z;
+            System.EventHandler<int> s = (object sender, int value) => z = value;
+            instance.IntPropertyChanged += s;
+            instance.RaiseIntChanged();
+            return instance;
+            GC.KeepAlive(s);
```
Scottj1s, line 78: why the KeepAlives here and above?  s is owned by instance

src/Benchmarks/EventPerf.cs
```diff
@@ -0,0 +1,107 @@
...
+                if (sender == this)
+                    z4 = value;
+                else
+                    z4 = value * 3;
+            };
+            instance2.IntPropertyChanged += s2;
+        }
+
+        [Benchmark]
+        public object IntEventOverhead()
```
Scottj1s, line 34: this is measuring construction time of ClassWithMarshalingRoutines and System.EventHandler - how come?

src/WinRT.Runtime/ComWrappersSupport.cs
```diff
@@ -315,6 +315,19 @@ private static bool IsIReferenceArray(Type implementationType)
...
+
+        private static Func<IInspectable, object> CreateCustomTypeMappingFactory(Type customTypeHelperType)
+        {
+            var fromAbiMethod = customTypeHelperType.GetMethod("FromAbi", BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static);
+            if (fromAbiMethod is null)
+            {
+                throw new MissingMethodException();
+            }
+
+            var parms = new[] { Expression.Parameter(typeof(IInspectable), "obj") };
```
Scottj1s, line 328: if you're looking to reduce allocations, this could be shared (global) and combined with a few other places in ComWrappers code

src/Benchmarks/EventPerf.cs
manodasanW: This was so that I can actually calculate the time it takes to add an event.  The diff between AddIntEventToNewEventSource and this would be that.  This essentially calculates the overhead in that other benchmark.

Scottj1s (approved):
thanks!

src/WinRT.Runtime/ComWrappersSupport.cs
manodasanW: These methods are cached but it can still end up overall reducing memory allocations.  I will look at evaluating the perf diff and address that in a future PR if I see differences.

src/Benchmarks/EventPerf.cs
manodasanW: I originally did that to keep the overhead consistent with the overhead benchmark, but I think removing it would make no diff so I will remove it.

PR #1058, opened 2021-11-29
Make DispatcherQueueSynchronizationContext faster and zero-alloc

Sergio0694:
## Overview

The current `DispatcherQueueSynchronizationContext` implementation is rather inefficient, for two reasons:
- It's creating a delegate with closure every single time (!), capturing both the callback and the state
- There's also additional overhead in general due to how the actual callback is then scheduled internally

Here's the specific line that's troubling:

https://github.com/microsoft/CsWinRT/blob/31a191d0e9b557e3732f93d86cf626defbb8d009/src/Projections/Reunion/Microsoft.UI.Dispatching.DispatcherQueueSynchronizationContext.cs#L25

This type is used extremely frequently in applications, especially for those that eg. rely on portable libraries for things such as MVVM support etc. that also deal with proper dispatching. The way that works is by capturing the synchronization context during setup and then dispatch through that, meaning this `Post` method will get invoked every time.

## The proposed solution

This PR optimizes the `Post` implementation by making it faster and zero-alloc.
There are no public API changes.
Opening this as draft as I'm not entirely sure of the procedure to add tests for this for this specific repo.
Pinging @jkoritzinsky as we discussed this feature idea previously on Discord as well 😄

I run some benchmarks for this a while back, got some nice numbers:

|                 Method |     Mean |     Error |    StdDev | Ratio |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|----------------------- |---------:|----------:|----------:|------:|-------:|------:|------:|----------:|
|  TryEnqueueWithClosure | 7.062 us | 0.1353 us | 0.1852 us |  1.00 | 0.0687 |     - |     - |     312 B |
|    **TryEnqueueWithState** | **1.139 us** | 0.0229 us | 0.0676 us |  **0.15** |      - |     - |     - |         **-** |

In this scenario dispatching with this system was about 7x faster, and also completely allocation free.
I'd argue it'd be nice to have such a basic building block like this one to be as efficient as possible.

I'm sure it might be a good idea to setup up some new ones in this repo as well, though I'd need guidance on that 😅

Opening as draft for now because of this, can switch this once everyone's happy with coverage and overall setup.

## Related issues/PRs

- https://github.com/CommunityToolkit/WindowsCommunityToolkit/pull/4097
- https://github.com/microsoft/microsoft-ui-xaml/issues/3321

### DIFF
--- a/src/Projections/Reunion/Microsoft.UI.Dispatching.DispatcherQueueSynchronizationContext.DispatcherQueueProxyHandler.cs
+++ b/src/Projections/Reunion/Microsoft.UI.Dispatching.DispatcherQueueSynchronizationContext.DispatcherQueueProxyHandler.cs
@@ -1,0 +1,243 @@
+#if NET5_0_OR_GREATER
+
+using System;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using System.Threading;
+using WinRT;
+
+#nullable enable
+#pragma warning disable CA1416
+
+namespace Microsoft.System
+{
+    /// <inheritdoc/>
+    public partial class DispatcherQueueSynchronizationContext
+    {
+        /// <summary>
+        /// A custom <c>IDispatcherQueueHandler</c> object, that internally stores a captured <see cref="SendOrPostCallback"/> instance and the
+        /// input captured state. This allows consumers to enqueue a state and a cached stateless delegate without any managed allocations.
+        /// </summary>
+        private unsafe struct DispatcherQueueProxyHandler
+        {
+            /// <summary>
+            /// The shared vtable pointer for <see cref="DispatcherQueueProxyHandler"/> instances.
+            /// </summary>
+            private static readonly void** Vtbl = InitVtbl();
+
+            /// <summary>
+            /// Setups the vtable pointer for <see cref="DispatcherQueueProxyHandler"/>.
+            /// </summary>
+            /// <returns>The initialized vtable pointer for <see cref="DispatcherQueueProxyHandler"/>.</returns>
+            /// <remarks>
+            /// The vtable itself is allocated with <see cref="RuntimeHelpers.AllocateTypeAssociatedMemory(Type, int)"/>,
+            /// which allocates memory in the high frequency heap associated with the input runtime type. This will be
+            /// automatically cleaned up when the type is unloaded, so there is no need to ever manually free this memory.
+            /// </remarks>
+            private static void** InitVtbl()
+            {
+                void** lpVtbl = (void**)RuntimeHelpers.AllocateTypeAssociatedMemory(typeof(DispatcherQueueProxyHandler), sizeof(void*) * 4);
+
+                lpVtbl[0] = (delegate* unmanaged<DispatcherQueueProxyHandler*, Guid*, void**, int>)&Impl.QueryInterface;
+                lpVtbl[1] = (delegate* unmanaged<DispatcherQueueProxyHandler*, uint>)&Impl.AddRef;
+                lpVtbl[2] = (delegate* unmanaged<DispatcherQueueProxyHandler*, uint>)&Impl.Release;
+                lpVtbl[3] = (delegate* unmanaged<DispatcherQueueProxyHandler*, int>)&Impl.Invoke;
+
+                return lpVtbl;
+            }
+
+            /// <summary>
+            /// The vtable pointer for the current instance.
+            /// </summary>
+            private void** lpVtbl;
+
+            /// <summary>
+            /// The <see cref="GCHandle"/> to the captured <see cref="SendOrPostCallback"/>.
+            /// </summary>
+            private GCHandle callbackHandle;
+
+            /// <summary>
+            /// The <see cref="GCHandle"/> to the captured state (if present, or a <see langword="null"/> handle otherwise).
+            /// </summary>
+            private GCHandle stateHandle;
+
+            /// <summary>
+            /// The current reference count for the object (from <c>IUnknown</c>).
+            /// </summary>
+            private volatile uint referenceCount;
+
+            /// <summary>
+            /// Creates a new <see cref="DispatcherQueueProxyHandler"/> instance for the input callback and state.
+            /// </summary>
+            /// <param name="handler">The input <see cref="SendOrPostCallback"/> callback to enqueue.</param>
+            /// <param name="state">The input state to capture and pass to the callback.</param>
+            /// <returns>A pointer to the newly initialized <see cref="DispatcherQueueProxyHandler"/> instance.</returns>
+            public static DispatcherQueueProxyHandler* Create(SendOrPostCallback handler, object? state)
+            {
+#if NET6_0
+                DispatcherQueueProxyHandler* @this = (DispatcherQueueProxyHandler*)NativeMemory.Alloc((nuint)sizeof(DispatcherQueueProxyHandler));
+#else
+                DispatcherQueueProxyHandler* @this = (DispatcherQueueProxyHandler*)Marshal.AllocHGlobal(sizeof(DispatcherQueueProxyHandler));
+#endif
+
+                @this->lpVtbl = Vtbl;
+                @this->callbackHandle = GCHandle.Alloc(handler);
+                @this->stateHandle = state is not null ? GCHandle.Alloc(state) : default;
+                @this->referenceCount = 1;
+
+                return @this;
+            }
+
+            /// <summary>
+            /// Devirtualized API for <c>IUnknown.AddRef()</c>.
+            /// </summary>
+            /// <returns>The updated reference count for the current instance.</returns>
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            public uint AddRef()
+            {
+                return Interlocked.Increment(ref referenceCount);
+            }
+
+            /// <summary>
+            /// Devirtualized API for <c>IUnknown.Release()</c>.
+            /// </summary>
+            /// <returns>The updated reference count for the current instance.</returns>
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            public uint Release()
+            {
+                uint referenceCount = Interlocked.Decrement(ref this.referenceCount);
+
+                if (referenceCount == 0)
+                {
+                    callbackHandle.Free();
+                    
+                    if (stateHandle.IsAllocated)
+                    {
+                        stateHandle.Free();
+                    }
+
+#if NET6_0
+                    NativeMemory.Free(Unsafe.AsPointer(ref this));
+#else
+                    Marshal.FreeHGlobal((IntPtr)Unsafe.AsPointer(ref this));
+#endif
+                }
+
+                return referenceCount;
+            }
+
+            /// <summary>
+            /// A private type with the implementation of the unmanaged methods for <see cref="DispatcherQueueProxyHandler"/>.
+            /// These methods will be set into the shared vtable and invoked by WinRT from the object passed to it as an interface.
+            /// </summary>
+            private static class Impl
+            {
+                /// <summary>
+                /// The HRESULT for a successful operation.
+                /// </summary>
+                private const int S_OK = 0;
+
+                /// <summary>
+                /// The HRESULT for an invalid cast from <c>IUnknown.QueryInterface</c>.
+                /// </summary>
+                private const int E_NOINTERFACE = unchecked((int)0x80004002);
+
+                /// <summary>
+                /// The GUID for the <c>IUnknown</c> COM interface.
+                /// </summary>
+                private static readonly Guid GuidOfIUnknown = new(0x00000000, 0x0000, 0x0000, 0xC0, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x46);
+
+                /// <summary>
+                /// The GUID for the <c>IAgileObject</c> WinRT interface.
+                /// </summary>
+                private static readonly Guid GuidOfIAgileObject = new(0x94EA2B94, 0xE9CC, 0x49E0, 0xC0, 0xFF, 0xEE, 0x64, 0xCA, 0x8F, 0x5B, 0x90);
+
+                /// <summary>
+                /// The GUID for the <c>IDispatcherQueueHandler</c> WinRT interface.
+                /// </summary>
+                private static readonly Guid GuidOfIDispatcherQueueHandler = new(0x2E0872A9, 0x4E29, 0x5F14, 0xB6, 0x88, 0xFB, 0x96, 0xD5, 0xF9, 0xD5, 0xF8);
+
+                /// <summary>
+                /// Implements <c>IUnknown.QueryInterface(REFIID, void**)</c>.
+                /// </summary>
+                [UnmanagedCallersOnly]
+                public static int QueryInterface(DispatcherQueueProxyHandler* @this, Guid* riid, void** ppvObject)
+                {
+                    if (riid->Equals(GuidOfIUnknown) ||
+                        riid->Equals(GuidOfIAgileObject) ||
+                        riid->Equals(GuidOfIDispatcherQueueHandler))
+                    {
+                        @this->AddRef();
+
+                        *ppvObject = @this;
+
+                        return S_OK;
+                    }
+
+                    return E_NOINTERFACE;
+                }
+
+                /// <summary>
+                /// Implements <c>IUnknown.AddRef()</c>.
+                /// </summary>
+                [UnmanagedCallersOnly]
+                public static uint AddRef(DispatcherQueueProxyHandler* @this)
+                {
+                    return Interlocked.Increment(ref @this->referenceCount);
+                }
+
+                /// <summary>
+                /// Implements <c>IUnknown.Release()</c>.
+                /// </summary>
+                [UnmanagedCallersOnly]
+                public static uint Release(DispatcherQueueProxyHandler* @this)
+                {
+                    uint referenceCount = Interlocked.Decrement(ref @this->referenceCount);
+
+                    if (referenceCount == 0)
+                    {
+                        @this->callbackHandle.Free();
+
+                        if (@this->stateHandle.IsAllocated)
+                        {
+                            @this->stateHandle.Free();
+                        }
+
+#if NET6_0
+                        NativeMemory.Free(@this);
+#else
+                        Marshal.FreeHGlobal((IntPtr)@this);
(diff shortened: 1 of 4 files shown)


manodasanW:
Thanks @Sergio0694 for this contribution.  A ~7x improvement with reduced memory pressure is definitely great to see in something that can be frequently used.  The version of this type that you have edited here is actually a copy of the one in WinAppSDK and is here for test purposes.  Adding @jeffstall and @ionvasilian who own this type and should be able to consider and update the actual one in WinAppSDK.

I looked over the change and it looks good to me.  It basically looks like you use a natively allocated proxy instance to hold onto the callback and state and make it look like a delegate by implementing Invoke and then fetch them later when invoked to make the call.  The only thing I wonder is if the delegate proxy should also implement IReferenceTrackerTarget.  I am not sure if XAML reference tracking is used for this type or not, but Jeff or Ion might be able to comment on that.

Sergio0694:
Hi @manodasanW, thank you for taking a look, really glad to hear this is something that could be interesting to have! 🎉
I wasn't aware this type was just a copy and not the real one (whoops!), but then again the main point of this PR was to show a proof of concept, share some numbers and start a conversation about this. Happy to help migrate this if needed! 😄

> "I looked over the change and it looks good to me. It basically looks like you use a natively allocated proxy instance to hold onto the callback and state and make it look like a delegate by implementing Invoke and then fetch them later when invoked to make the call."

Yup, that's correct! The underlying WinRT API for `IDispatcherQueue::TryEnqueue` is just this one:

```cpp
int TryEnqueue(IDispatcherQueueHandler* callback, byte* result);
```

Where that `callback` is this an object inheriting from `IDispatcherQueueHandler`, which is:

```cpp
class IDispatcherQueueHandler : IUnknown
{
    int Invoke();
}
```

So in this PR I'm just creating my own implementation instead of letting CsWinRT create one wrapping that managed delegate/closure (we're essentially defining a C# struct that implements an object derived from `IDispatcherQueueHandler`). This means I can then also make it blittable and just store a couple GC handles in it to track the managed objects to invoke later (delegate and optional state), without extra allocations. And the `Invoke()` method which is then invoked by the WinRT runtime when the callback is dispatched simply grabs the target callback and state back and invokes them, and then ensures exceptions are tracked correctly, if any. That `IReferenceTrackerTarget` interface hasn't come up while investigating this so far (as in, `IDispatcherQueueHandler` didn't inherit from it, and when testing this and checking with a breakpoint on `QueryInterface`, that interface was never queried), but for sure, waiting for Jeff to confirm this or suggest any changes 😊

src/Projections/Reunion/Microsoft.UI.Dispatching.DispatcherQueueSynchronizationContext.DispatcherQueueProxyHandler.cs
```diff
@@ -0,0 +1,260 @@
...
+
+                /// <summary>
+                /// The GUID for the <c>IUnknown</c> COM interface.
+                /// </summary>
+                private static ref readonly Guid GuidOfIUnknown
+                {
+                    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+                    get
+                    {
+                        ReadOnlySpan<byte> data = new byte[] { 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0xC0, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x46 };
```
jkotas, line 143: What does guarantee that the memory is properly aligned?

Multiple people just spent a week tracking a bug in dotnet/runtime where the C/C++ compiler was assuming proper memory alignment, and producing wrong code due to not it being the case: https://github.com/dotnet/runtime/issues/61671

jkotas: This project is using standard readonly Guids constants in many other places so using this hack in a few places does not really move the needle. And the key reason why terrafx is using this hack (single type with thousands of Guid constants, most of them never used) does not apply here.

Sergio0694: This trick originally came up while investigating some performance issues in Paint.NET, as @rickbrew was noticing a big startup slowdown due to a static constructor initializing a few thousand GUIDs. @tannergooding thought about using this to fix that, and it's now part of `TerraFX.Interop.Windows` as well (which is then used by projects such as Paint.NET, ComputeSharp, etc.). IIRC we had a very similar conversation before (other folks raised the point about unaligned reads potentially being a problem, especially on ARM64), and going back to older messages on Discord:

> "Btw does windows support arm CPUs without unaligned reads"

> Tanner: "Don't believe so. Arm64 (like x64) always supports unaligned reads (well outside a special switch which I don't think most apps will work with) and windows doesn't support standalone arm32, just the arm32 subsystem of arm64 which is like v8, rather than v7)"

I think I recall the conclusion being that doing this in this case (and x86/64 doesn't care anyway), and also in a scenario where you only need to run on Windows (such as CsWinRT) and therefore having the same hardware requirements that Windows has), this is fine. That said if that's not the case I'm happy to revert this, and I guess then we would also have a problem in TerraFX then though. Either way it seems we'll (or at least, I'll) learn something new today 😄

jkotas: The problem in dotnet/runtime#61671 was that the ABI cares, nothing to do with whether the CPU cares.

tannergooding: > What does guarantee that the memory is properly aligned?

@jkotas, if the compiler (C#, JIT, Crossgen, etc) isn't ensuring that RVA statics are at least 4-byte aligned already then we likely already have issues with `RuntimeHelpers.InitializeArray` and will run into issues with recently approved `RuntimeHelpers.CreateSpan`.

Likewise, if the underlying ABI expects `Guid` to be more than 4-byte aligned then .NET already has issues because that's the natural alignment we have for `Guid`, not `8` and not `16`.

That isn't necessarily a good reason to introduce risk here, particularly if there are only a handful of `guids` that are trivial to initialize; but it should mean that this pattern is "safe" for up to `int`

tannergooding: > If that's not the case

Looks like the C# compiler tightly packs the data, so if you have:
```
public ReadOnlySpan<byte> M() {
    return new byte[] { 1, 2, 3 };
}

public uint[] N() {
    return new uint[] { 1, 2, 3 };
}
```

Then the backing RVA for `new uint[]` (the one currently passed to `InitializeArray` and that will be passed to `CreateSpan` in the future) may be unaligned.

jkotas: > Then the backing RVA for new uint[] (the one currently passed to InitializeArray and that will be passed to CreateSpan in the future) may be unaligned.

Good catch. It needs to be taken care of as part the part of the Roslyn enablement of `RuntimeHelpers.CreateSpan`. The block has to have same alignment as the element. But I do not think it makes sense to make Roslyn to apply 4 or 8 byte alignment to ReadOnlySpan<byte> constants.

FWIW, InitializeArray does not require the block to be aligned. It is fine since it is one-time operation. The (currently unused) big endian path for InitializeArray is using unaligned reads for this reason: https://github.com/dotnet/runtime/blob/3efb476e2e22f7dd5ac96968203ef931704d34b8/src/coreclr/classlibnative/bcltype/arraynative.cpp#L1134-L1142

> As I mentioned I have no issue reverting this, to me it was more a matter of: is this guaranteed to be safe or not?

I would qualify it as fragile code. The spec does not guarantee that it will work. It works today, but it may stop working as things evolve.

Sergio0694: I'll go revert the change then, we can always change this in another PR later down the line anyway. This whole conversation was rally interesting though, so definitely worth the time, especially if we apparently also found a potential issue with Roslyn and the way we're currently relying on this while investigating this. Thanks everyone for chiming in! 😄

Sergio0694:
Publishing this PR as the issue raised by @jkotas has been resolved, and there doesn't seem to be other outstanding ones. Chatted with @jkoritzinsky on Discord and he said it seems all obvious concerns have been addressed. @manodasanW would these changes still also need to be merged here to mirror the actual type in WASDK? @jeffstall, @ionvasilian happy to help eventually migrating these changes to the right place in WASDK if these changes are accepted 😊

src/Projections/Reunion/Microsoft.UI.Dispatching.DispatcherQueueSynchronizationContext.cs
```diff
@@ -1,14 +1,17 @@
 using System;
 using System.Threading;
-using Microsoft.System;
+using WinRT;
+
+#nullable enable
+#pragma warning disable CA1416
```
benwestprog, line 6: >CA1416 [](http://example.com/codeflow?start=24&length=6)

Could you add a comment documenting why this is disabled?

src/Projections/Reunion/Microsoft.UI.Dispatching.DispatcherQueueSynchronizationContext.cs
```diff
@@ -17,19 +20,57 @@ public DispatcherQueueSynchronizationContext(DispatcherQueue dispatcherQueue)
             m_dispatcherQueue = dispatcherQueue;
         }
 
-        public override void Post(SendOrPostCallback d, object state)
+        /// <inheritdoc/>
+        public override unsafe void Post(SendOrPostCallback d, object? state)
         {
-            if (d == null)
+#if NET6_0
```
benwestprog, line 26: For simplicity, could this whole if/else/endif block be

if (d is null)
{
    ThrowArgumentNullException(nameof(d));
}

Looks like we get the same code but much easier to read.

src/Projections/Reunion/Microsoft.UI.Dispatching.DispatcherQueueSynchronizationContext.IDispatcherQueue.cs
```diff
@@ -0,0 +1,38 @@
...
+
+            /// <summary>
+            /// Native API for <see cref="DispatcherQueue.TryEnqueue(DispatcherQueueHandler)"/>.
+            /// </summary>
+            /// <param name="callback">A pointer to an <c>IDispatcherQueueHandler</c> object.</param>
+            /// <param name="result">The result of the operation (the <see cref="bool"/> WinRT retval).</param>
+            /// <returns>The <c>HRESULT</c> for the operation.</returns>
+            /// <remarks>The <paramref name="callback"/> parameter is assumed to be a pointer to an <c>IDispatcherQueueHandler</c> object.</remarks>
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            public int TryEnqueue(void* callback, byte* result)
```
benwestprog, line 30: >byte [](http://example.com/codeflow?start=50&length=4)

Can this be bool to match the native definition more closely?

src/Projections/Reunion/Microsoft.UI.Dispatching.DispatcherQueueSynchronizationContext.IDispatcherQueue.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        /// <summary>
+        /// A struct mapping the native WinRT <c>IDispatcherQueue</c> interface.
+        /// </summary>
+        internal unsafe struct IDispatcherQueue
+        {
+            /// <summary>
+            /// The vtable pointer for the current instance.
+            /// </summary>
+            private readonly void** lpVtbl;
```
benwestprog, line 20: >lpVtbl [](http://example.com/codeflow?start=36&length=6)

Not sure about coding conventions around here, but can we remove all the hungarian notation from these files?

src/Projections/Reunion/Microsoft.UI.Dispatching.DispatcherQueueSynchronizationContext.DispatcherQueueProxyHandler.cs
```diff
@@ -0,0 +1,233 @@
...
+            /// </summary>
+            /// <param name="handler">The input <see cref="SendOrPostCallback"/> callback to enqueue.</param>
+            /// <param name="state">The input state to capture and pass to the callback.</param>
+            /// <returns>A pointer to the newly initialized <see cref="DispatcherQueueProxyHandler"/> instance.</returns>
+            public static DispatcherQueueProxyHandler* Create(SendOrPostCallback handler, object? state)
+            {
+#if NET6_0
+                DispatcherQueueProxyHandler* @this = (DispatcherQueueProxyHandler*)NativeMemory.Alloc((nuint)sizeof(DispatcherQueueProxyHandler));
+#else
+                DispatcherQueueProxyHandler* @this = (DispatcherQueueProxyHandler*)Marshal.AllocHGlobal(sizeof(DispatcherQueueProxyHandler));
```
benwestprog, line 80: What's the advantage of making a distinction here? Do we need two possible implementations?

benwestprog (approved):
:shipit:

src/Projections/Reunion/Microsoft.UI.Dispatching.DispatcherQueueSynchronizationContext.cs
Sergio0694: Done in 52dd2ddb3a6a036bbb9d8f7f829bde9a85d89039 👍

src/Projections/Reunion/Microsoft.UI.Dispatching.DispatcherQueueSynchronizationContext.IDispatcherQueue.cs
Sergio0694: We can't use `bool*` in the actual function pointer signature, or it will cause issues on .NET Standard 2.0 ([see here](https://github.com/dotnet/coreclr/pull/23899#issuecomment-551371899)). I've changed the parameter to `bool*` though as request, and just changed it to `byte*` internally before invoking the API, which is fine 🙂
Change done in a71a63de4a89435deb2a11ba9e3b9d2859f5307d.

src/Projections/Reunion/Microsoft.UI.Dispatching.DispatcherQueueSynchronizationContext.IDispatcherQueue.cs
Sergio0694: Sorry, force of habit with the vtable pointers 😄
Fixed in 17ebd48dc3ecb2fb01b1235135a23bf19074f354.

src/Projections/Reunion/Microsoft.UI.Dispatching.DispatcherQueueSynchronizationContext.DispatcherQueueProxyHandler.cs
Sergio0694: `Marshal.AllocHGlobal` is particularly slow and generally just not recommended for use anymore ([see here](https://github.com/dotnet/runtime/issues/33244#issuecomment-828619859)). `NativeMemory` has been introduced in .NET 6 and it's just a lightweigt set of wrappers for the relative C APIs. Once .NET 5 goes out of support (which is like in less than a month anyway) and WASDK drops it, we can just remove all of those extra paths anyway and just leave the ones for .NS2.0 (the default one) and the .NET 6 one. Does that make sense? 🙂

src/Projections/Reunion/Microsoft.UI.Dispatching.DispatcherQueueSynchronizationContext.cs
Sergio0694: Done in 91f5a0881857d8e262e47fd6c47aa8d9cc6f73f9 🙂

src/Projections/Reunion/Microsoft.UI.Dispatching.DispatcherQueueSynchronizationContext.cs
```diff
@@ -17,19 +23,53 @@ public DispatcherQueueSynchronizationContext(DispatcherQueue dispatcherQueue)
...
-            if (d == null)
-                throw new ArgumentNullException(nameof(d));
+            if (d is null)
+            {
+                static void ThrowArgumentNullException()
+                {
+                    throw new ArgumentNullException(nameof(d));
+                }
+
+                ThrowArgumentNullException();
```
jkotas, line 36:
```suggestion
                throw new ArgumentNullException(nameof(d));
```
This throw helper micro-optimization is not worth it given everything else that this method does.

Sergio0694: Updated in e95e80e3b39ce94bc0b25e8d0692cb8625e42e10 🙂

AtariDreams:
Any updates on this?

benwestprog:
We ultimately decided not to port these changes into WinAppSDK. They add considerable code complexity, and we could not justify the added risk.

---
In reply to: [1285755982](https://github.com/microsoft/CsWinRT/pull/1058#issuecomment-1285755982) [](http://example.com/codeflow?ancestors=1285755982,1285755982)

mqudsi:
Thanks @Sergio0694 for trying, anyway! It would have been nice to have!

(Too bad GH doesn't have a sad-face emoji to express sadness without disapproval. A 👎 or 😕do not convey the same meaning!)

Sergio0694:
@mqudsi Thank you! 😊

For the record, we've been using this implementation in the Microsoft Store for months already (just slightly different, as the Store uses UWP and .NET Native, not CsWinRT, but it's effectively identical minus a different GUID), and it's something I do want us to open source eventually. If CsWinRT is not the right path, we might look into open sourcing it in the Windows Community Toolkit at some point, maybe packaged in some higher-level API (like the dispatcher service we showed at //BUILD 2022) that people could use like we're doing, we'll see. Of course, you can also just copy this implementation for yourself in the meantime 🙂

PR #1063, opened 2021-12-10
Improvements to nullable / boxing

### LINKED ISSUE #87: Boxing performance analysis
Investigate boxing performance. Currently boxing is being done by both .NET and WinRT, investigate this and other things we can do in the projection.

manodasanW:
Created specialized versions of some of the common nullable types to make use of function pointers rather than delegates and avoid generic marshalers.

Fixed issue with nullable delegates where we can't associate them with 2 different pointers (delegate ptr and nullable ptr) and as part of that also improved how we return back wrapped nullable objects which were used for types and strings and now also used for delegates.

Fixes #87

### DIFF
--- a/src/Tests/TestComponentCSharp/Class.cpp
+++ b/src/Tests/TestComponentCSharp/Class.cpp
@@ -1407,6 +1407,11 @@ namespace winrt::TestComponentCSharp::implementation
         return winrt::unbox_value<hstring>(obj);
     }
 
+    winrt::TestComponentCSharp::ProvideInt Class::UnboxDelegate(WF::IInspectable const& obj)
+    {
+        return winrt::unbox_value<TestComponentCSharp::ProvideInt>(obj);
+    }
+
     com_array<int32_t> Class::UnboxInt32Array(WF::IInspectable const& obj)
     {
         return obj.as<IReferenceArray<int32_t>>().Value();
--- a/src/Tests/TestComponentCSharp/Class.cpp
+++ b/src/Tests/TestComponentCSharp/Class.cpp
@@ -1467,6 +1467,12 @@ namespace winrt::TestComponentCSharp::implementation
         return winrt::box_value(hstring{});
     }
 
+    WF::IInspectable Class::BoxedDelegate()
+    {
+        TestComponentCSharp::ProvideUri handler = [] { return Windows::Foundation::Uri(L"http://microsoft.com"); };
+        return winrt::box_value(handler);
+    }
+
     hstring Class::Catch(hstring const& /*params*/, hstring& /*lock*/)
     {
         // Compile-only test for keyword escaping
--- a/src/Tests/UnitTest/TestComponentCSharp_Tests.cs
+++ b/src/Tests/UnitTest/TestComponentCSharp_Tests.cs
@@ -2227,6 +2227,15 @@ namespace UnitTest
             Assert.Equal(string.Empty, (string)str2);
         }
 
+        [Fact]
+        public void TestDelegateUnboxing()
+        {
+            var del = Class.BoxedDelegate;
+            Assert.IsType<ProvideUri>(del);
+            var provideUriDel = (ProvideUri) del;
+            Assert.Equal(new Uri("http://microsoft.com"), provideUriDel());
+        }
+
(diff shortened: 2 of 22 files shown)


src/WinRT.Runtime/Marshalers.cs
```diff
@@ -1255,6 +1255,23 @@ public static IObjectReference CreateMarshaler(object o, Guid delegateIID, bool
 
             return ComWrappersSupport.CreateCCWForObject<global::WinRT.Interop.IDelegateVftbl>(o, delegateIID);
         }
+
+        public static T FromAbi<T>(IntPtr nativeDelegate, Func<IntPtr, T> createCallback)
```
manodasanW, line 1259: Evaluating whether to keep callback or call create object here.

src/WinRT.Runtime/ComWrappersSupport.cs
```diff
@@ -369,6 +378,10 @@ private static bool IsIReferenceArray(Type implementationType)
                     return CreateReferenceCachingFactory(CreateNullableTFactory(typeof(System.Nullable<>).MakeGenericType(implementationType)));
                 }
             }
+            else if(IsAbiNullableDelegate(implementationType))
```
j0shuams, line 381: nit: space

src/WinRT.Runtime/Marshalers.cs
manodasanW: Replaced callback with CreateRCW function that gets called by CreateObject.  Benchmark shows that in the existing delegate scenario that it brings better perf compared to the callback because we don't need to create an RCW each time even though there is an existing one.  And for the new delegate scenario, there is minimal differences (~50 ns).

src/Tests/UnitTest/TestComponentCSharp_Tests.cs
```diff
@@ -2219,6 +2222,15 @@ public void TestStringUnboxing()
             Assert.IsType<string>(str2);
             Assert.Equal(string.Empty, (string)str1);
             Assert.Equal(string.Empty, (string)str2);
+        }
+
+        [Fact]
+        public void TestDelegateUnboxing()
```
ujjwalchadha, line 2228: Should we add any tests for all the new nullable types we added? Or are all of them being tested already?

src/Tests/TestComponentCSharp/Class.cpp
```diff
@@ -1407,6 +1407,11 @@ namespace winrt::TestComponentCSharp::implementation
         return winrt::unbox_value<hstring>(obj);
     }
 
+    winrt::TestComponentCSharp::ProvideInt Class::UnboxDelegate(WF::IInspectable const& obj)
```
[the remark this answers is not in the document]
manodasanW: Both are actually slightly different.  One is performing an action of unboxing a given object in the parameter.  The other is returning a boxed object itself.

src/WinRT.Runtime/ComWrappersSupport.cs
```diff
@@ -281,21 +287,36 @@ private static bool IsIReferenceArray(Type implementationType)
             return Expression.Lambda<Func<IInspectable, object>>(
                 Expression.Call(type.GetHelperType().GetMethod("CreateRcw", BindingFlags.Public | BindingFlags.Static), 
                     parms), parms).Compile();
+        }
+
+        internal static Func<IntPtr, object> CreateDelegateFactory(Type type)
+        {
+            return DelegateFactoryCache.GetOrAdd(type, (type) =>
```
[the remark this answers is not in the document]
manodasanW: It could be, but there isn't too many types in WinRT.Runtime which can be added here.  Most are in the generated projections.  There looks to be only 3 of which only like 2 can be added here.  Right now looking at traces, this function doesn't seem to be expensive, but if either of those 2 do end up being expensive, we can add them in the future.

j0shuams (approved):

src/Tests/UnitTest/TestComponentCSharp_Tests.cs
manodasanW: Added tests which actually also ended up catching an existing issue with IReference<char>.

PR #1075, opened 2022-01-12
Reduce allocations for generated constructors

j0shuams:
For generated types, we often have a public constructor and then internal/base constructors. We reduce the space used when calling one of these public constructors by refactoring the relevant ObjectReference code. Instead of creating and immediately invoking a lambda to give us an object reference, we inline the code in the public constructor. 
 
Also included is a refactoring of the generated "_lazyInterfaces" dictionary. We saw before that it was being initialized in those internal constructors, but without any references to local variables. We move them to the global level and use the Interlocked.CompareExchange pattern of lazy initialization, rather than the .NET Lazy<T> type. 

I have run some benchmarks and have two scenarios. 

For the change to constructors and making/invoking a lambda, we only make this change for net5.0 code. The code to assign _inner is different for netstandard compatibility, making our optimization untenable. However, the change to the lazy interface initializations provides improvements to all runtimes (net5, net3, netstd2).  

**Net5.0**

| Codepath | Master - time | Topic - time | Master - memory | Topic - memory |
| ----------- | --------------  | ------------ | -------------------- | ----------------- |
| Constructors | 3.89 micro | 3.74 micro | 584 B | 392 B |
| Lazy Interface | 11.57 micro | 10.53 micro | 2752 B | 1864 B |

**NetStandard2.0**

| Codepath | Master - time | Topic - time | Master - memory | Topic - memory |
| ----------- | --------------  | ------------ | -------------------- | ----------------- |
| Constructors | 17.59 micro | 17.86 micro | 3481 B | 3453 B |
| Lazy Interface | 10.98 micro | 10.07 micro | 3000 B | 2024 B |


j0shuams:
Example of difference in generated code, taken from `src\Projections\Windows\GeneratedFiles\net5.0\Windows.Storage.Streams.cs`

**Code generated on master branch**

```csharp
public sealed class DataReader : IDataReader, ...
{
        private IntPtr ThisPtr => _inner == null ? (((IWinRTObject)this).NativeObject).ThisPtr : _inner.ThisPtr;

        private IObjectReference _inner = null;

        private readonly Dictionary<Type, object> _lazyInterfaces;
        
        ...
        
        internal sealed class _IDataReaderFactory { ... }
        
        public DataReader(IInputStream inputStream) : this(((Func<IObjectReference>)(() => {
            IntPtr ptr = (_IDataReaderFactory.Instance.CreateDataReader(inputStream));
            try
            {
                return (ComWrappersSupport.GetObjectReferenceForInterface(ptr));
            }
            finally
            {
                MarshalInspectable<object>.DisposeAbi(ptr);
            }
        }))())
        {
            ComWrappersSupport.RegisterObjectForInterface(this, ThisPtr);
            ComWrappersHelper.Init(_inner, false);
        }
        
        ...
        
        internal DataReader(IObjectReference objRef)
        {
            _inner = objRef.As(GuidGenerator.GetIID(typeof(IDataReader).GetHelperType()));


            _lazyInterfaces = new Dictionary<Type, object>(1)
            {
                {typeof(global::System.IDisposable), new Lazy<global::System.IDisposable>(() => (global::System.IDisposable)(object)new SingleInterfaceOptimizedObject(typeof(global::System.IDisposable), _inner ?? ((IWinRTObject)this).NativeObject))},
            };

        }
        
        ...
        
        private global::System.IDisposable AsInternal(InterfaceTag<global::System.IDisposable> _) =>  ((Lazy<global::System.IDisposable>)_lazyInterfaces[typeof(global::System.IDisposable)]).Value;


}
```

**Code generated on topic branch jlarkin/objref-func**

```csharp
public sealed class DataReader : IDataReader, ...
{
        private IntPtr ThisPtr => _inner == null ? (((IWinRTObject)this).NativeObject).ThisPtr : _inner.ThisPtr;

        private IObjectReference _inner = null;
        
        private volatile global::System.IDisposable _lazy_global__System_IDisposable;

        private global::System.IDisposable Make__lazy_global__System_IDisposable()
        {
            global::System.Threading.Interlocked.CompareExchange(ref _lazy_global__System_IDisposable, (global::System.IDisposable)(object)new SingleInterfaceOptimizedObject(typeof(global::System.IDisposable), _inner ?? ((IWinRTObject)this).NativeObject), null);
            return _lazy_global__System_IDisposable;
        }

        ...
        
        internal sealed class _IDataReaderFactory { ... }
        
        public DataReader(IInputStream inputStream) 
        { 
        IntPtr ptr = (_IDataReaderFactory.Instance.CreateDataReader(inputStream)); 
        try 
        { 
        _inner = ComWrappersSupport.GetObjectReferenceForInterface(ptr); 
        } 
        finally 
        { 
        MarshalInspectable<object>.DisposeAbi(ptr); 
        }

        ComWrappersSupport.RegisterObjectForInterface(this, ThisPtr);
        ComWrappersHelper.Init(_inner, false);
        }
        
        ...
        
        internal DataReader(IObjectReference objRef)
        {
            _inner = objRef.As(GuidGenerator.GetIID(typeof(IDataReader).GetHelperType()));

        }
        
        ... 
        
        private global::System.IDisposable AsInternal(InterfaceTag<global::System.IDisposable> _) =>  Make__lazy_global__System_IDisposable();

```

src/cswinrt/code_writers.h
```diff
@@ -2650,22 +2688,20 @@ remove => %.ErrorsChanged -= value;
...
                             interface_name,
-                            interface_abi_name,
-                            interface_name);
+                            bind<write_lazy_interface_type_name>(interface_type));
                     }
                     else if (is_manually_generated_iface(interface_type))
                     {
                         w.write(R"(
-private % AsInternal(InterfaceTag<%> _) =>  ((Lazy<%>)_lazyInterfaces[typeof(%)]).Value;
+private % AsInternal(InterfaceTag<%> _) =>  Make_%();
```
manodasanW, line 2700: Here and above, you want to first check if the variable already has a value and if not, then make it like we do with the other versions of interlockedexchange.

j0shuams: Done

src/cswinrt/code_writers.h
```diff
@@ -2112,8 +2151,7 @@ Marshal.Release(inner);
                             {
                                 w.write("_defaultLazy = new Lazy<%>(() => (%)new SingleInterfaceOptimizedObject(typeof(%), _inner));", default_interface_name, default_interface_name, default_interface_name);
```
manodasanW, line 2152: We should also look at the scenarios where we use _defaultLazy as I think it might be similarly impacted like _lazyInterfaces where the internal constructor was the one which sets it.

j0shuams: Addressed by adding this back to the public constructor for the same edge case

src/cswinrt/code_writers.h
```diff
@@ -1936,8 +1939,15 @@ MarshalInspectable<object>.DisposeAbi(ptr);
...
                         cache_object,
                         method.Name(),
-                        bind_list<write_parameter_name_with_modifier>(", ", signature.params())
-                    );
+                        bind_list<write_parameter_name_with_modifier>(", ", signature.params()),
+                        bind([&](writer& w)
+                        {
+                            if (is_manually_gen_default_interface)
+                            {
+                                auto default_interface_name_non_ABI = get_default_interface_name(w, class_type, false);
```
manodasanW, line 1947: consider projected_default_interfance_name as projected is the naming convention used for non_abi.

src/cswinrt/code_writers.h
```diff
@@ -2650,22 +2698,22 @@ remove => %.ErrorsChanged -= value;
                     if (settings.netstandard_compat)
```
manodasanW, line 2698: Are both if the same, if so consider merging it into one if?

manodasanW (approved):

PR #1078, opened 2022-01-14
Support DispatcherQueueSynchronizationContext.Send.

jlaanstra:
In various places in .NET, the runtime calls SynchronizationContext.Send. For example in the SystemEvents class Send is used: https://github.com/dotnet/runtime/blob/6de7147b9266d7730b0d73ba67632b0c198cb11e/src/libraries/Microsoft.Win32.SystemEvents/src/Microsoft/Win32/SystemEvents.cs#L1326

This means that subscribing to events on the SystemEvents class from the UI thread in a WinUI3 app works, but the event is never fired, because Send throws an exception. There are likely other places where Send is used and where this causes problems for WinUI3 apps.

This PR adds support for Send in a similar to how Xamarin Android does this for the Android main thread: https://github.com/xamarin/xamarin-android/blob/main/src/Mono.Android/Android.App/SyncContext.cs and MAUI for GtkSynchronizationContext: https://github.com/dotnet/maui/blob/18e0f4ebbcca7c904ccdb67e6439cb72382e2a40/src/Compatibility/Core/src/GTK/GtkSynchronizationContext.cs

### DIFF
--- a/src/Projections/Reunion/Microsoft.UI.Dispatching.DispatcherQueueSynchronizationContext.cs
+++ b/src/Projections/Reunion/Microsoft.UI.Dispatching.DispatcherQueueSynchronizationContext.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Runtime.ExceptionServices;
 using System.Threading;
 using Microsoft.System;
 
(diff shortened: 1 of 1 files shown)


hez2010:
@manodasanW Any chance to get this PR merged?

src/Projections/Reunion/Microsoft.UI.Dispatching.DispatcherQueueSynchronizationContext.cs
```diff
@@ -1,38 +1,68 @@
...
+                    catch (Exception ex)
+                    {
+                        edi = ExceptionDispatchInfo.Capture(ex);
+                    }
+                    finally
+                    {
+                        m.Set();
+                    }
+                });
+                m.WaitOne();
```
teo-tsirpanis, line 55: I'm not sure the `Send` method is allowed to block until the callback runs; it can lead to deadlocks.

Sergio0694:
Closing this, it's not a change we're considering for 2.x at this point.
We could potentially revisit this for 3.0 in the future if we can validate it wouldn't cause issues 🙂

PR #1079, opened 2022-01-18
Generic CreateMarshaler

j0shuams:
This PR adds a new public API to MarshalInspectable<T>, which allows calls to CreateMarshaler to pass in a VTable as generic param and an IID as a function param. This reduces .As() call and prevents unnecessary allocations. 

In the table below, "Unwrap" refers to the case where CreateMarshaler returns early, and "CreateCCW" is the case where CreateMarshaler calls the CreateCCW function. Time is measured in microseconds (us) and memory is measured in bytes (B).

**Net5.0**

| Codepath | Master - time | Topic - time | Master - memory | Topic - memory |
| ----------- | --------------  | ------------ | -------------------- | ----------------- |
| Unwrap | 5.243 us | 4.947 us | 464 B | 344 B |
| CreateCCW | 10.784 us | 8.52 us | 768 B | 648 B |

**NetStandard2.0**

| Codepath | Master - time | Topic - time | Master - memory | Topic - memory |
| ----------- | --------------  | ------------ | -------------------- | ----------------- |
| Unwrap | 8.997 us | 8.288 us | 1560 B | 1440 B |
| CreateCCW | 69.445 us | 67.744 us | 4400 B | 4241 B |

While saving only a microsecond or two, we consistently use at least 120 B less of memory everytime we call CreateMarshaler.

Will provide a diff of generated code 

Below is an example of a common pattern in projected code for structs used with marshaling, we provide a new (generic) function that handles the implementation that does not need the same subsequent method calls.

```csharp

namespace ABI.Windows.System.Profile
{
    [global::System.ComponentModel.EditorBrowsable(global::System.ComponentModel.EditorBrowsableState.Never)]
    public struct AnalyticsVersionInfo
    {
        public static IObjectReference CreateMarshaler(global::Windows.System.Profile.AnalyticsVersionInfo obj) => 
          obj is null ? 
            null : 
            // new 
             MarshalInspectable<global::Windows.System.Profile.AnalyticsVersionInfo>.CreateMarshaler<IUnknownVftbl>(obj, GuidGenerator.GetIID(typeof(global::Windows.System.Profile.IAnalyticsVersionInfo).GetHelperType())
            /* old 
             MarshalInspectable<global::Windows.System.Profile.AnalyticsVersionInfo>.CreateMarshaler(obj).As<IUnknownVftbl>(GuidGenerator.GetIID(typeof(global::Windows.System.Profile.IAnalyticsVersionInfo).GetHelperType()) */
           );
    }
}

### CODE REVIEW

File: src/WinRT.Runtime/Marshalers.cs [hunk 1]
```diff
@@ -1140,6 +1140,31 @@ public static unsafe void CopyManaged(T value, IntPtr dest)
...
+                var createMarshaler = Expression.Lambda<Func<object, IObjectReference>>(
+                    Expression.Call(helperType.GetMethod("CreateMarshaler", BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static),
+                        new[] { Expression.Convert(parms[0], publicType) }), parms).Compile();
+                return createMarshaler(o);
+            }
+
+            return ComWrappersSupport.CreateCCWForObject<V>(o, iid);
+        }
+
         public static IObjectReference CreateMarshaler(T o, bool unwrapObject = true)
```
manodasanW, line 1168: Consider reimplementing this function in terms of the first as it reduces duplication of code.

src/cswinrt/code_writers.h
```diff
@@ -6224,23 +6224,14 @@ public static unsafe void DisposeAbiArray(object box) => MarshalInspectable<obje
...
-                            }
-                            else
-                            {
-                                w.write(
-                                    ".As<IUnknownVftbl>(GuidGenerator.GetIID(typeof(%).GetHelperType()))",
-                                    bind<write_type_name>(get_type_semantics(get_default_interface(type)), typedef_name_type::CCW, false));
-                            }
-                        }));
+                        is_generic ? w.write_temp("%.Vftbl", default_interface_abi_name) : w.write_temp("IUnknownVftbl"),
+                        is_generic ? w.write_temp("GuidGenerator.GetIID(%.Vftbl)", default_interface_abi_name)
```
manodasanW, line 6233: Does the IID optimizer optimize this scenario where GetHelperType isn't there?

j0shuams: While I didn't find an example that calls this specifically (using the Vftbl), I did confirm that we do patch calls of the form `GuidGenerator.GetIID(_)`. Note the lack of `.GetHelperType()`.

manodasanW (approved):

PR #1083, opened 2022-01-21
Control ADO-Maestro job using variable

j0shuams:
Feeding a parameter thru was not working when the queue-time variable was set. To fix this we use the coalesce pattern and change the condition on the Maestro task.

### DIFF
--- a/build/AzurePipelineTemplates/CsWinRT-Variables.yml
+++ b/build/AzurePipelineTemplates/CsWinRT-Variables.yml
@@ -10,3 +10,7 @@ variables:
   Net6.SDK.Version: '6.0.100-rc.2.21505.57'
   NoSamples: 'false'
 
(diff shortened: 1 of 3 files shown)


build/AzurePipelineTemplates/CsWinRT-Variables.yml
```diff
@@ -10,3 +10,7 @@ variables:
   Net6.SDK.Version: '6.0.100-rc.2.21505.57'
   NoSamples: 'false'
 
+  # This 'coalesce' pattern allows the yml to define a default value for a variable but allows the value to be overridden at queue time.
+  # E.g. '_isRelease' defaults to empty string, but if 'isRelease' is set at queue time that value will be used.
+
+  _isRelease: $[coalesce(variables.isRelease, '')]
```
manodasanW, line 16: Consider IsRelease (captial I) to stay consistent with others.

manodasanW (approved):

PR #1087, opened 2022-01-25
More queue-time variables

j0shuams:
Allows us to set the version dependencies communicated to Maestro at queue-time. 
Note that the WindowsSdkPackageVersion is no longer a full windows version, but just the suffix used to get the latest projection -- e.g. "10.0.17763.22" -> "22"

### DIFF
--- a/build/AzurePipelineTemplates/CsWinRT-PublishToMaestro-Stage.yml
+++ b/build/AzurePipelineTemplates/CsWinRT-PublishToMaestro-Stage.yml
@@ -4,9 +4,6 @@ stages:
   jobs:
   - job: TriggerMaestroPublish
     condition: eq(variables['_IsRelease'],'true')
-    variables:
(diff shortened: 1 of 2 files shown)


j0shuams:
I confirmed setting this vars at queue time gives the desired effect by queueing with fake numbers - .NET Runtime = 5.2.0 and WindowsSdkPackage = 30. We can verify by seeing these values used in the post to maestro: 

{"id":1239,"commit":"0308655ff66c5e481c66fca1eccf066042de31ad", ...{"id":1565,"name":"**CsWinRT.Dependency.DotNetCoreRuntime","version":"5.2.0"**,"buildId":1239,"nonShipping":false,"locations":[]},{"id":1566,"name":**"CsWinRT.Dependency.WindowsSdkPackage","version":"30"**,"buildId":1239,"nonShipping":false,"locations":[]} ...

manodasanW (approved):

PR #1092, opened 2022-01-27
Change return type of CanExecuteChanged only for MUX.Input.ICommand

### LINKED ISSUE #597: Projection of XamlUICommand.CanExecuteChanged (system XAML) loses generic type
When projecting [Windows.UI.Xaml.Input.XamlUICommand](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.input.xamluicommand), the [CanExecuteChanged](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.input.xamluicommand.canexecutechanged) event is projected as `EventHandler` instead of `EventHandler<T>` causing a compile break ("`XamlUICommand` cannot implement `ICommand` because the type of `CanExecuteChanged` is different" or something like that)

Using the latest cswinrt nuget: 1.1.0-prerelease.201111.6

mcooley:
Fixes #597

Microsoft.UI.Xaml.Input.ICommand is mapped to System.Windows.Input.ICommand. The CanExecuteChanged events on those two types aren't exactly the same, so there is a special case in the code generation to fix up the type mapping. This was introduced in #112.

The check for this special case is currently too broad: it will apply to _any_ event named "CanExecuteChanged" of type "EventHandler<object>", even if that event is part of some unknown-to-cswinrt type that is not mapped to System.Windows.Input.ICommand. (I hit this with the WUX.Input.ICommand type in system XAML--since system XAML is not supported, its types should be treated as unknown-to-cswinrt everywhere.)

This PR makes the check more specific, so it affects only the mapped MUX type.


asklar:
thanks @mcooley - with this change, what does CanExecuteChanged get projected to for **Windows**.UI.Xaml.Input.ICommand?

mcooley:
@asklar on Windows.UI.Xaml.Input.ICommand, CanExecuteChanged will be `EventHandler<object>` instead of `EventHandler`.

asklar (approved):

asklar:
@manodasanW could you take a look? LGTM.

manodasanW (approved):

Issue #1097, opened 2022-02-02
MSB3271 mismatch between processor architecture when building projection


angelazhangmsft:
**Describe the bug**

An MSB3271 error/warning results when building a projection for AnyCPU while referencing a native C++/WinRT component.

Error message: 
*Error MSB3271 There was a mismatch between the processor architecture of the project being built "MSIL" and the processor architecture, "x86", of the implementation file "C:\Users\angzhang\repos\CsWinRT\src\Samples\NetProjectionSample\_build\Win32\Release\SimpleMathComponent\bin\SimpleMathComponent\SimpleMathComponent.dll" for "C:\Users\angzhang\repos\CsWinRT\src\Samples\NetProjectionSample\_build\Win32\Release\SimpleMathComponent\bin\SimpleMathComponent\SimpleMathComponent.winmd". This mismatch may cause runtime failures. Please consider changing the targeted processor architecture of your project through the Configuration Manager so as to align the processor architectures between your project and implementation file, or choose a winmd file with an implementation file that has a processor architecture which matches the targeted processor architecture of your project.*

**To Reproduce**

See the [NetProjectionSample.](https://github.com/microsoft/CsWinRT/tree/master/src/Samples/NetProjectionSample)

**Version Info**

CsWinRT 1.4.1, .NET 6

**Workaround**
 Add the following property to the library project: 
```xml
<ResolveAssemblyWarnOrErrorOnTargetArchitectureMismatch>None</ResolveAssemblyWarnOrErrorOnTargetArchitectureMismatch>
```

PR #1098, opened 2022-02-02
Update sample to build projection for AnyCPU

angelazhangmsft:
The existing guidance for generating C#/WinRT projections from a C++/WinRT component resulted in architecture-specific projection assemblies being built. 

We are updating the sample to target AnyCPU when building the projection. Note there is a workaround for MSB3271 error that results from targeting AnyCPU (tracking the issue in https://github.com/microsoft/CsWinRT/issues/1097).


src/Samples/NetProjectionSample/SimpleMathProjection/SimpleMathProjection.csproj
```diff
@@ -5,6 +5,10 @@
     <TargetFramework>net6.0-windows10.0.19041.0</TargetFramework>
     <TargetPlatformMinVersion>10.0.17763.0</TargetPlatformMinVersion>
     <Platforms>x64;x86;ARM64</Platforms>
```
angelazhangmsft, line 7: Should we just set Platform to AnyCPU instead of using PlatformTarget? This seems to work as well

manodasanW: If it works with the different configurations, I agree that is probably better, I think as it represents what the binary actually is.

src/Samples/NetProjectionSample/ConsoleAppSample.sln
```diff
@@ -1,20 +1,38 @@
...
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.30509.190
+# Visual Studio Version 17
+VisualStudioVersion = 17.0.32014.148
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ConsoleAppSample", "ConsoleAppSample\ConsoleAppSample.csproj", "{9623756E-90CE-449C-8BD8-36A2F8419DBF}"
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Any CPU = Debug|Any CPU
```
manodasanW, line 10: Does anycpu make sense in the app as it consumes a WinRT component which is arch based.

src/Samples/NetProjectionSample/SimpleMathProjection/SimpleMathProjection.csproj
angelazhangmsft: verified that this works, and we can set the Configuration Manager so that SimpleMathProjection project always builds with Platform AnyCPU for each solution architecture
image

src/Samples/NetProjectionSample/ConsoleAppSample.sln
angelazhangmsft: Will remove AnyCPU

manodasanW (approved):

Issue #1056, opened 2021-11-23
VS designer build errors with NetProjectionSample


angelazhangmsft:
**Describe the bug**

When opening the NetProjectionSample (CppWinRTComponentProjectionSample.sln) in Visual Studio, the following errors appear in VS:

> (1) Error		Path 'C:\Users\deads\Desktop\Dev\CsWinRT-master\src\Samples\Net5ProjectionSample\_build\x64\Debug\SimpleMathComponent\bin\SimpleMathComponent\SimpleMathComponent.winmd' is not a file or directory	SimpleMathProjection	C:\Users\deads\Desktop\Dev\CsWinRT-master\src\Samples\Net5ProjectionSample\SimpleMathProjection\EXEC	1

> (2)  Error	MSB3073	The command ""C:\Users\deads\.nuget\packages\microsoft.windows.cswinrt\1.4.1\cswinrt.exe" @"C:\Users\deads\Desktop\Dev\CsWinRT-master\src\Samples\Net5ProjectionSample\_build\x64\Debug\SimpleMathProjection\bin\cswinrt.rsp"" exited with code 1.	SimpleMathProjection	C:\Users\deads\.nuget\packages\microsoft.windows.cswinrt\1.4.1\build\Microsoft.Windows.CsWinRT.targets	121

These errors don't prevent the projection from building successfully, but they still show up in VS Error List after building. The errors disappear if you close/reopen the solution after building, probably since the designer build happens once upon opening the solution in VS.

**To Reproduce**

1. Open NetProjectionSample (CppWinRTComponentProjectionSample.sln) in VS 2022. Make sure to clear any previous output
2. Errors appear
3. Build the solution, and the errors still show.

**Expected behavior**

No errors

**Version Info**

CsWinRT 1.4.1 (also see the errors with 1.3.5)
.NET SDK 6.0.100
VS 2022
CppWinRT 2.0.210806.1 and 2.0.211028.7

**Additional context**

Related to this target: https://github.com/microsoft/CsWinRT/blob/b216a3854af21aa003cd3ddc19bc4b982e843b40/nuget/Microsoft.Windows.CsWinRT.targets#L121

angelazhangmsft:
Using latest internal build I still see the following warnings before building:

<img width="829" alt="image" src="https://user-images.githubusercontent.com/69016571/156273932-5495e37e-dd62-402a-b6f8-23b12cb1e5d7.png">

Issue #963, opened 2021-08-16
When CsWinRT fails during DesignTimeBuild, the project is skipped as it won't have any outputs.


jlaanstra:
**Describe the bug**
When CsWinRT fails during DesignTimeBuild, the project is skipped as it won't have any outputs.

**To Reproduce**
Reference a WinMD with public types from another WinMD that is not referenced.

**Expected behavior**
CsWinRT fails to generate the projection but it doesn't fail the design time build.

**Version Info**
1.3.3

PR #1105, opened 2022-02-09
Skip projection generation during design-time build if inputs are missing

evelynwu-msft:
Although it is generally desirable for CsWinRT projection generation to explicitly fail if input `.winmd`s are missing, we *don't* want this to happen during a design-time build. Due to Visual Studio's architecture, failures in a project's design-time build can leave the results of dependent projects' design-time builds in an inconsistent state (this was determined to be one of the primary root causes of https://task.ms/37142724). An example of this scenario would be a CsWinRT project that depends on a C++/CX Windows Runtime Component project: the latter produces its output `.winmd` during the `Link` target and so the `.winmd` does not exist for the former to consume until a normal build has completed.

This change modifies the `CsWinRTGenerateProjection` target to skip the `cswinrt.exe` command if it is determined that (a) it is currently running in the context of a design-time build, and (b) one or more of the input `.winmd`s does not exist. This allows the design-time build to succeed while still allowing for the possibility of a legitimate failure during a normal build.

### DIFF
--- a/nuget/Microsoft.Windows.CsWinRT.targets
+++ b/nuget/Microsoft.Windows.CsWinRT.targets
@@ -67,12 +67,30 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     </PropertyGroup>
   </Target>
 
+  <!-- 
+      During a design-time build, input WinMDs may not exist yet if a full build has not been completed.
+      If that is the case, generation of the CsWinRT projection will fail. To avoid causing a design-time
+      build break (which can put the design-time build results of dependent projects into a bad state),
+      we want to skip projection generation if a) this is a design-time build, and b) any of the inputs
+      do not exist.
+  -->
+  <Target Name="CsWinRTCalculateSkipDesignTimeGenerateProjection" 
+          DependsOnTargets="CsWinRTRemoveWinMDReferences">
+    <ItemGroup>
+      <_CsWinRTNonexistentInputs Include="@(CsWinRTInputs)" Condition="!Exists('%(CsWinRTInputs.Identity)')" />
+    </ItemGroup>
+
+    <PropertyGroup>
+      <CsWinRTSkipDesignTimeGenerateProjection Condition="('$(DesignTimeBuild)' == 'true' OR '$(BuildingProject)' != 'true') and '@(_CsWinRTNonexistentInputs)'!=''">true</CsWinRTSkipDesignTimeGenerateProjection>
+    </PropertyGroup>
+  </Target>
+
   <Target Name="CsWinRTCleanGenerateProjectionOutputs">
     <Delete Files="$(CsWinRTGeneratedFilesDir)cswinrt.rsp" />
   </Target>
 
   <Target Name="CsWinRTGenerateProjection"
-          DependsOnTargets="CsWinRTPrepareProjection;CsWinRTRemoveWinMDReferences"
+          DependsOnTargets="CsWinRTPrepareProjection;CsWinRTRemoveWinMDReferences;CsWinRTCalculateSkipDesignTimeGenerateProjection"
           Condition="'$(CsWinRTGenerateProjection)' == 'true'"
           Inputs="$(MSBuildAllProjects);@(CsWinRTInputs);$(CsWinRTExcludes);$(CsWinRTIncludes);$(CsWinRTFilters);$(CsWinRTWindowsMetadata)"
           Outputs="$(CsWinRTGeneratedFilesDir)cswinrt.rsp">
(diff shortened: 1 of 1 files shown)


nuget/Microsoft.Windows.CsWinRT.targets
```diff
@@ -67,12 +67,30 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+      do not exist.
+  -->
+  <Target Name="CsWinRTCalculateSkipDesignTimeGenerateProjection" 
+          DependsOnTargets="CsWinRTRemoveWinMDReferences">
+    <ItemGroup>
+      <_CsWinRTNonexistentInputs Include="@(CsWinRTInputs)" Condition="!Exists('%(CsWinRTInputs.Identity)')" />
+    </ItemGroup>
+
+    <PropertyGroup>
+      <CsWinRTSkipDesignTimeGenerateProjection Condition="('$(DesignTimeBuild)' == 'true' OR '$(BuildingProject)' != 'true') and '@(_CsWinRTNonexistentInputs)'!=''">true</CsWinRTSkipDesignTimeGenerateProjection>
```
evelynwu-msft, line 84: >CsWinRTSkipDesignTimeGenerateProjection [](http://example.com/codeflow?start=7&length=39)

Ideally we'd just override `$(CsWinRTGenerateProjection)` and not have to change the `CsWinRTGenerateProjection` target at all, but unfortunately MSBuild appears to evaluate target `Condition`s *before* any target gets a chance to run.

jlaanstra:
> Although it is generally desirable for CsWinRT projection generation to explicitly fail if input .winmds are missing, we don't want this to happen during a design-time build. Due to Visual Studio's architecture, failures in a project's design-time build can leave the results of dependent projects' design-time builds in an inconsistent state

This problem can also happen when a WinMD is present but it's out of date.

> This change modifies the CsWinRTGenerateProjection target to skip the cswinrt.exe command if it is determined that (a) it is currently running in the context of a design-time build, and (b) one or more of the input .winmds does not exist. This allows the design-time build to succeed while still allowing for the possibility of a legitimate failure during a normal build.

I think we can solve this with a much simpler change, by simply setting COntinueOnError on the CsWinRT Exec task as follows:
```<Exec Command="$(CsWinRTCommand)" ContinueOnError="'$(DesignTimeBuild)' == 'true' OR '$(BuildingProject)' != 'true'" />```

Thoughts?

angelazhangmsft:
@evelynwu-msft does this fix #1056 and/or #963?

evelynwu-msft:
> I think we can solve this with a much simpler change, by simply setting COntinueOnError on the CsWinRT Exec task as follows:
> 
> ```
>         <PropertyGroup>
>             <CsWinRTContinueOnError>false</CsWinRTContinueOnError>
>             <CsWinRTContinueOnError Condition="'$(DesignTimeBuild)' == 'true' OR '$(BuildingProject)' != 'true'">true</CsWinRTContinueOnError>
>         </PropertyGroup>
> 
>         <Exec Command="$(CsWinRTCommand)" ContinueOnError="$(CsWinRTContinueOnError)" />
> ```
> 
> Thoughts?

My preference is to skip the scenario entirely if we know in advance that it's going to generate an error. It minimizes the amount of unnecessary work that is done, and avoids the possibility of cswinrt.exe, even in an error situation, leaving behind cruft that pollutes the build outputs.

(It's also not clear to me from the [MSDN documentation](https://docs.microsoft.com/en-us/visualstudio/msbuild/how-to-ignore-errors-in-tasks?view=vs-2022) if `<Task ContinueOnError="ErrorAndContinue" />` will trigger an `<OnError />` element; if it doesn't then we'd still have to skip creation of the `.rsp` file.)

evelynwu-msft:
> @evelynwu-msft does this fix #1056 and/or #963?

@angelazhangmsft Yes, I believe it would address both of them.

jlaanstra:
> My preference is to skip the scenario entirely if we know in advance that it's going to generate an error. It minimizes the amount of unnecessary work that is done, and avoids the possibility of cswinrt.exe, even in an error situation, leaving behind cruft that pollutes the build outputs.

Skipping the scenario makes the IDE experience worse as you get no intellisense until you do a build. That doesn't seem like the right way forward.

jlaanstra:
> (It's also not clear to me from the [MSDN documentation](https://docs.microsoft.com/en-us/visualstudio/msbuild/how-to-ignore-errors-in-tasks?view=vs-2022) if <Task ContinueOnError="ErrorAndContinue" /> will trigger an <OnError /> element; if it doesn't then we'd still have to skip creation of the .rsp file.)

Per docs here https://docs.microsoft.com/en-us/visualstudio/msbuild/onerror-element-msbuild?view=vs-2022#remarks OnError will be skipped but this is fine. If the inputs don't change, the target wouldn't create any other output either way.

evelynwu-msft:
> Skipping the scenario makes the IDE experience worse as you get no intellisense until you do a build. That doesn't seem like the right way forward.

Wouldn't Intellisense information be missing regardless because `cswinrt.exe` errors out during projection generation?

jlaanstra:
> Wouldn't Intellisense information be missing regardless because `cswinrt.exe` errors out during projection generation?

From observation and experience it still generates the partial projection up to the point it hits the error, so you'd get some intellisense.

jlaanstra:
> > @evelynwu-msft does this fix #1056 and/or #963?
> 
> @angelazhangmsft Yes, I believe it would address both of them.

Note that the fix as proposed in this PR doesn't completely address #693. The designtime build can still fail for a reason other than a missing output file, for example a missing project reference to a WinMD containing types used by a different WinMD. In that case there are no missing files as far as the logic is concerned and so the Designtime build would still fail, causing all the same problems.

evelynwu-msft:
@jlaanstra Your suggestion worked! However, I'm worried by the fact that the `OnError` target no longer runs and cleans up the .rsp; I suspect this would cause MSBuild to later incorrectly think that the target doesn't need to be incrementally built, but I'm not familiar enough with CsWinRT scenarios to determine how that might manifest. Do you have any suggestions?

manodasanW (approved):

PR #1109, opened 2022-02-16
Move user-facing strings to resource files

j0shuams:
This PR introduces resx and rc files for user-facing strings in the C#/WinRT toolchain. These are error messages from various tools we use - WinRT.Host, WinRT Authoring Diagnostics, and WinRT.Runtime. Included in these changes is logic in the respective projects to load resource strings from these resource files.

We will need to add a TDBuild task to our CI YML templates, and properly package the resulting localized files. Some of the code used for loading will need to be updated to find the language/region specific versions of the resource files. 

TDBuild onboarding deliverable: https://globalcollaborationservice.visualstudio.com/Global%20Collaboration%20Service%20Project/_workitems/edit/106432


src/Authoring/WinRT.Host/WinRT.Host.cpp
```diff
@@ -43,52 +46,68 @@ static load_assembly_and_get_function_pointer_fn load_assembly_and_get_function_
...
-        { FrameworkCompatRetry, L"A previously processed framework reference was reprocessed." },
-        { BundleExtractionFailure, L"Error extracting single-file apphost bundle." },
-        { BundleExtractionIOError, L"Error reading or writing files during single-file apphost bundle extraction." },
-        { LibHostDuplicateProperty, L"The .runtimeconfig.json contains a runtime property which is also produced by the hosting layer." },
-        { HostApiUnsupportedVersion, L"Feature which requires certain version of the hosting layer binaries was used on a version which doesn't support it." },
-        { HostInvalidState, L"The current state is incompatible with the requested operation." },
-        { HostPropertyNotFound, L"A property requested by hostfxr_get_runtime_property_value doesn't exist." },
-        { CoreHostIncompatibleConfig, L"The component being initialized requires framework which is not available" },
-        { HostApiUnsupportedScenario, L"hostfxr doesn't currently support requesting the given delegate type using the given context." },
+        { InvalidArgFailure, 106 },
```
BenJKuhn, line 53: The overloading of the constants with different values is very confusing. At a glance I thought this table was mapping from "106 -> 106" for example. Consider differentiating the resource names with a prefix / suffix in resource.h. e.g.

InvalidArgFailure - > InvalidArgFailureStr

Since they're 100% consistent with the original error names, you could macro-ize the mapping or otherwise simplify the code.

@define CSWINR_RESSTR(x) { x, x##Str },

j0shuams:
Per discussion with Ben, I'm looking into using .mui files instead of the satellite dll pattern for winrt.host errors

src/Authoring/WinRT.Host/Directory.Build.targets
```diff
@@ -0,0 +1,23 @@
+<!--
+***********************************************************************************************
+Copyright (C) Microsoft Corporation. All rights reserved.
+***********************************************************************************************
+-->
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+    <Target Name="SplitWinRTHostIntoMUI"
+            Inputs="$(OutDir)WinRT.Host.dll"
+            Outputs="$(OutDir)WinRT.Host.dll.mui"
```
manodasanW, line 10: Should WinRT.Host.dll also be an output here given we replace it or does just having the mui file as the output work out?

src/Authoring/WinRT.Host/WinRT.Host.cpp
```diff
@@ -43,55 +52,85 @@ static load_assembly_and_get_function_pointer_fn load_assembly_and_get_function_
...
+        {
+            // Error if we fail to load WinRT.Host resource file 
+            swprintf_s(error_string_buffer, SUFFICIENTLY_LARGE_ERROR_BUFFER, L"Error: Unable to load the resource container module, last error = %d", GetLastError());
+            MessageBoxW(NULL, error_string_buffer, L"WinRT.Host ERROR!", MB_OK | MB_ICONERROR);
+        } 
+        else if (LoadStringW(resources_handle, string_table_entry, error_string, SUFFICIENTLY_LARGE_STRING_BUFFER) == 0)
+        {
+            // Error if we fail to load the string
+            swprintf_s(error_string_buffer, SUFFICIENTLY_LARGE_ERROR_BUFFER, L"Error: Unable to load the resource string, last error = %d.", GetLastError());
+            MessageBoxW(NULL, error_string_buffer, L"WinRT.Host ERROR!", MB_OK | MB_ICONERROR);
```
manodasanW, line 121: Should these be message box or should we just throw hresult_error or something else?

src/Authoring/WinRT.Host/WinRT.Host.cpp
```diff
@@ -43,55 +52,85 @@ static load_assembly_and_get_function_pointer_fn load_assembly_and_get_function_
...
+        {
+            // Everything loaded fine, load the retrieved string and finish
+            swprintf_s(error_string_buffer, SUFFICIENTLY_LARGE_STRING_BUFFER, L"%s", error_string);
+            FreeLibrary(resources_handle);
+            throw hresult_error(result, error_string_buffer);
         }
     }
     
-    winrt::throw_hresult(result);
+    //winrt::throw_hresult(result);
```
manodasanW: remove comment?

src/Authoring/WinRT.Host/WinRTHostErrorStrings.rc
```diff
@@ -0,0 +1,131 @@
...
+    InvalidArgFailureStr    "One of the specified arguments for the operation is invalid."
+    CoreHostLibLoadFailureStr 
+                            "There was a failure loading a dependent library."
+    CoreHostLibMissingFailureStr "One of the dependent libraries is missing."
+    CoreHostEntryPointFailureStr 
+                            "One of the dependent libraries is missing a required entry point."
+    CoreHostCurHostFindFailureStr 
+                            "Could not deduce installation location from current module."
+    CoreClrResolveFailureStr "The coreclr library could not be found."
+    CoreClrBindFailureStr   "The loaded coreclr library doesn't have one of the required entry points."""
```
manodasanW, line 64: Triple quotes?

src/Authoring/WinRT.SourceGenerator/CsWinRTDiagnosticStrings.resx
```diff
@@ -0,0 +1,330 @@
...
+  <!-- 
+    Microsoft ResX Schema 
+    
+    Version 2.0
+    
+    The primary goals of this format is to allow a simple XML format 
+    that is mostly human readable. The generation and parsing of the 
+    various data types are done through the TypeConverter classes 
+    associated with the data types.
+
```
manodasanW, line 12: Is example needed?

src/WinRT.Runtime/WinRT.Runtime.csproj
```diff
@@ -57,6 +57,21 @@
   <ItemGroup Condition="'$(TargetFramework)' == 'net5.0'">
     <PackageReference Include="Microsoft.DotNet.ApiCompat" Version="5.0.0-beta.20560.3" PrivateAssets="All" />
     <PackageDownload Include="Microsoft.Windows.CsWinRT" Version="[$(CsWinRTApiCompatVersion)]" />
+  </ItemGroup> 
+  
+  <ItemGroup>
+    <Compile Update="WinRTRuntimeWinRTRuntimeErrorStrings.Designer.cs">
```
manodasanW, line 63: WinRTRuntimeErrorStrings?  or is it really repeated twice?

src/WinRT.Runtime/WinRT.Runtime.csproj
```diff
@@ -57,6 +57,21 @@
...
+      <DesignTime>True</DesignTime>
+      <AutoGen>True</AutoGen>
+      <DependentUpon>WinRTRuntimeWinRTRuntimeErrorStrings.resx</DependentUpon>
+    </Compile>
+  </ItemGroup> 
+  
+  <ItemGroup>
+    <EmbeddedResource Update="WinRTRuntimeWinRTRuntimeErrorStrings.resx">
```
manodasanW, line 71: same here

nuget/Microsoft.Windows.CsWinRT.Authoring.Transitive.targets
```diff
@@ -22,6 +22,7 @@
     <HostingAssets Include="$(MSBuildThisFileDirectory)..\build\native\WinRT.Host.runtimeconfig.json"/>
     <!-- Get the proper WinRT.Host.dll -->
     <HostingAssets Include="$(MSBuildThisFileDirectory)..\runtimes\win-$(_NormalizedPlatform)\native\WinRT.Host.dll" />
+    <HostingAssets Include="$(MSBuildThisFileDirectory)..\runtimes\win-$(_NormalizedPlatform)\native\en-us\WinRT.Host.dll.mui" />
```
manodasanW, line 25: Whats the story here for localized ones if en-us is the main one copied  Is it up the component owners to pack the others based on which ones they want?  Same for below where you use en-us.

src/Authoring/WinRT.Host/Directory.Build.targets
j0shuams: Just having the mui seems to work just fine, but we can add it as an output

src/Authoring/WinRT.Host/WinRT.Host.cpp
j0shuams: Thought message box would be better so it's clear what happened instead of error thrown in output. But can change to hresult_error if you rather us stick with that

nuget/Microsoft.Windows.CsWinRT.Authoring.Transitive.targets
j0shuams: Yeah I think automating minimally is better in this case so we don't always output many binaries in their project. The different culture/regions will be available in the nupkg, so they can distribute the whole lot if they want. But we will just handle en-us for them, for now. If we get feedback from folks that they prefer a different model we can adjust accordingly.

src/Authoring/WinRT.Host/Directory.Build.targets
```diff
@@ -0,0 +1,28 @@
...
+            Outputs="$(OutDir)WinRT.Host.dll;$(OutDir)WinRT.Host.dll.mui"
+            AfterTargets="CopyFilesToOutputDirectory">
+
+        <!-- Invoke the muicrt tool to split Winrt.Host into language neutral parts and localizable resources
+            Then invoke the tool again to apply the checksum from the neutral part to the mui
+            Finally, replace the initially built WinRT.Host with the language neutral version -->
+        <PropertyGroup> 
+            <WinRTHostMuirctCommand>Muirct.exe -q "$(MSBuildThisFileDirectory)WinRTHost.rcconfig.xml" -v 2 -x 0x0409 -g 0x0409 "$(OutDir)WinRT.Host.dll" "$(OutDir)WinRT.Host_LN.dll" "$(OutDir)WinRT.Host.dll.mui"
+muirct.exe -c "$(OutDir)WinRT.Host_LN.dll" -e "$(OutDir)WinRT.Host.dll.mui"
+copy /y "$(OutDir)WinRT.Host.dll" "$(OutDir)WinRT.Host.Original.exe"
```
manodasanW, line 19: original.dll?  but it probably also doesn't matter.

src/Authoring/WinRT.Host/WinRT.Host.cpp
```diff
@@ -43,51 +52,81 @@ static load_assembly_and_get_function_pointer_fn load_assembly_and_get_function_
...
 
+    auto string_table_entry = 0;
     for (auto&& elem : hostfxr_status_messages)
-    {
+    {   
         if (elem.status == result)
         {
-            throw hresult_error(result, elem.message);
+            string_table_entry = elem.resourceID;
+        }
```
manodasanW, line 101: break when found?

src/Authoring/WinRT.Host/WinRT.Host.cpp
manodasanW: Well the folks who are going to see the error popup is going to be customers of apps which use the authoring component feature.  The app calling this may have its own way of handling errors which the popup can get in the way of and in the current state, it doesn't seem the developer will get crash dumps either due to it isn't an exception.

src/Authoring/WinRT.Host/WinRTHostErrorStrings.rc
```diff
@@ -0,0 +1,131 @@
...
+STRINGTABLE
+BEGIN
+    ResolverInitFailureStr  "Initialization of the hostpolicy dependency resolver failed."
+    ResolverResolveFailureStr 
+                            "Resolution of dependencies in hostpolicy failed."
+    LibHostCurExeFindFailureStr 
+                            "Failure to determine the location of the current executable."
+    LibHostInitFailureStr   "Initialization of the hostpolicy library failed."
+    LibHostSdkFindFailureStr 
+                            "Failure to find the requested SDK. This happens in the hostfxr when an SDK (also called CLI) command is used with dotnet. In this case the hosting layer tries to find an installed. NET SDK to run the command on. The search is based on deduced install locationand on the requested version from potential global. json file. If either no matching SDK version can be found, or that version exists, but it's missing the dotnet. dll file, this error code is returned."""
```
manodasanW, line 78: Triple quotes intentional?

src/Tests/AuthoringWinUITest/AuthoringWinUITest/Directory.Build.targets
```diff
@@ -1,16 +1,15 @@
...
+
   <PropertyGroup>
     <PrepareForRunDependsOn>CopyTestAssets;$(PrepareForRunDependsOn)</PrepareForRunDependsOn>
   </PropertyGroup>
 
-  <!-- temporary until nuget support -->
   <Target Name="CopyTestAssets">
-    <Copy SourceFiles="$(BuildOutDir)WinRT.Host\bin\WinRT.Host.dll"
-          DestinationFiles="$(OutDir)WinRT.Host.dll" 
+    <Copy SourceFiles="$(BuildOutDir)WinRT.Host\bin\WinRT.Host.dll.mui"
```
manodasanW, line 10: DLL is automatically copied?

src/Authoring/WinRT.Host/Directory.Build.targets
j0shuams: Just a temp name, we're going to replace the output WinRT.Host with the language-neutral version anyway

src/Tests/AuthoringWinUITest/AuthoringWinUITest/Directory.Build.targets
j0shuams: Looks like we have to manually place in both the AuthoringWinUITest and in AuthoringWinUITest (Package) projects. Restored and added mui to (Package) wapproj

manodasanW (approved):

Issue #1088, opened 2022-01-25
Question: How to use IBasicVideoEffect?


HermanEldering:
I've tried implementing IBasicVideoEffect with CsWinRT but I can't get it to activate with `new VideoEffectDefinition("IBasicVideoEffectSample.MyVideoEffect");`. Any help will be appreciated!

Just using the nuget reference I get `System.Runtime.InteropServices.COMException: 'Class not registered (0x80040154 (REGDB_E_CLASSNOTREG))'`.

I've tried adding a reference in the app manifest (in many variations). Pointing to my dll gives `System.Runtime.InteropServices.COMException: 'Error in the DLL (0x800401F9 (CO_E_ERRORINDLL))'` and pointing to WinRT.Host.dll like this

```
<file name="runtimes\win-x64\native\WinRT.Host.dll">
    <activatableClass
        name="IBasicVideoEffectSample.MyVideoEffect"
        threadingModel="both"
        xmlns="urn:schemas-microsoft-com:winrt.v1" />
</file>
```

gives: `Exception thrown at 0x00007FF980664F69 (KernelBase.dll) in VideoEffectConsoleApp.exe: WinRT originate error - 0x80008093 : 'The .runtimeconfig.json file is invalid.'.`

Below is my runtime.config. Note that I have to change it manually each time after building, since the template content is added inside the runtimeOptions and activatableClasses should be in the root according to [this](https://github.com/microsoft/CsWinRT/blob/master/docs/hosting.md). But the error is always there.

```
{
  "runtimeOptions": {
    "tfm": "net6.0",
    "framework": {
      "name": "Microsoft.NETCore.App",
      "version": "6.0.0"
    },
    "configProperties": {
      "System.Runtime.Serialization.EnableUnsafeBinaryFormatterSerialization": false
    }
  },
  "activatableClasses": {
    "IBasicVideoEffectSample.MyVideoEffect": "IBasicVideoEffectSample.dll"
  }
}
```

My testing code can be found here: https://github.com/HermanEldering/IBasicVideoEffectSample

HermanEldering:
I have managed to fix the problem. Using procmon I found out it was looking for `runtimes\win-x64\native\WinRT.Host.runtimeconfig.json`.

[...]

Issue #694, opened 2021-01-27
Authoring guide - integration with WinUI packaged desktop apps


asklar:
The authoring guide talks about MSIX or non-MSIX, but it doesn't mention whether e.g. the winrt.host.runtimeconfig.json and the manifest are to be added to the WinUI desktop app project or to the WinUI desktop app WAP. It would be great to have specific instructions for integrating cswinrt usage in WinUI apps.

PR #1111, opened 2022-02-18
CsWinRT/WinUI Authoring Support

j0shuams:
This PR includes changes to our authoring targets that fix gaps in our WinUI support. The issue we were seeing was with native apps referencing cswinrt components; the managed projection binaries the component depended on (e.g. Microsoft.WinUI.dll) were not being copied to the output folder of the native app. We leverage the item group `AllItemsFullPathWithTargetPath` to place the binaries correctly. This also removes some of the logic we were using to place CsWinRT binaries like WinRT.Runtime, as the new target will handle that for us. 

Because we ship a WinRT.Runtime.dll and the .NET SDK also has its own version, we explicitly pick the dll with the latest version for copying.

This has been tested using Windows App SDK 1.1 (experimental build) and VS2022.


j0shuams:
Should review/revise the `authoring.md` docs

nuget/Microsoft.Windows.CsWinRT.Authoring.Transitive.targets
```diff
@@ -34,6 +35,14 @@
       </Reference>
     </ItemGroup>
 
+
+    <ItemGroup> 
+      <Reference Include="$(MSBuildThisFileDirectory)..\lib\net6.0*\winmd\*.winmd">
```
manodasanW, line 40: Did we want to hard code this to net6?

nuget/Microsoft.Windows.CsWinRT.Authoring.targets
```diff
@@ -80,8 +71,59 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+  <!-- Any managed dependencies of the WinRT component won't automatically be brought 
+        in by a referencing C++ app, so we manually set them to be copied here -->
+  <Target Name="CsWinRTAuthoring_AddManagedDependencies"
+          BeforeTargets="GetCopyToOutputDirectoryItems"
+          Returns="@(AllItemsFullPathWithTargetPath)">
+
+    <!-- Isolate the two versions of WinRT.Runtime.dll -->
+    <ItemGroup>
+
+      <_CsWinRTDeDupRuntime_NETSDK Include="@(ReferenceCopyLocalPaths)" Condition="'%(ReferenceCopyLocalPaths.NuGetPackageId)' == 'Microsoft.Windows.SDK.NET.Ref' AND '%(ReferenceCopyLocalPaths.Filename)' == 'WinRT.Runtime'"/>
```
manodasanW, line 83: This always has NugetPackageId right in the SDK case?

j0shuams:
This should address https://github.com/microsoft/CsWinRT/issues/694 

@asklar  Take a look if you get a chance and think there's more useful clarifications to be made!

docs/authoring.md
```diff
@@ -162,12 +159,33 @@ Consuming C#/WinRT components from MSIX-packaged applications is supported with
 
 C#/WinRT supports authoring out-of-process components that can be consumed by Windows Runtime compatible languages. Currently, consuming an out-of-process component is supported for managed C# apps with the use of a packaging project. This scenario requires creating a `WinRT.Host.runtimeconfig.json` file, as demonstrated in the [background task component sample](https://github.com/microsoft/CsWinRT/tree/master/src/Samples/BgTaskComponent). Native C++ consumption of out-of-process C# components is not fully supported yet.
 
+### Using Windows App SDK
+
+Authoring WinUI types with C#/WinRT is in development and is supported with some exceptions.
+Component authors can implement WinUI types using the Windows App SDK class library template. 
+Components are consumed by projects made from the various other Windows App SDK templates.
```
angelazhangmsft, line 166: How about "Components can be consumed by application projects that reference the Windows App SDK"?  

Someone could have an existing project (e.g. WPF) that references the Windows App SDK, not necessarily created from the templates

docs/authoring.md
```diff
@@ -162,12 +159,33 @@ Consuming C#/WinRT components from MSIX-packaged applications is supported with
 
 C#/WinRT supports authoring out-of-process components that can be consumed by Windows Runtime compatible languages. Currently, consuming an out-of-process component is supported for managed C# apps with the use of a packaging project. This scenario requires creating a `WinRT.Host.runtimeconfig.json` file, as demonstrated in the [background task component sample](https://github.com/microsoft/CsWinRT/tree/master/src/Samples/BgTaskComponent). Native C++ consumption of out-of-process C# components is not fully supported yet.
 
+### Using Windows App SDK
+
+Authoring WinUI types with C#/WinRT is in development and is supported with some exceptions.
+Component authors can implement WinUI types using the Windows App SDK class library template. 
+Components are consumed by projects made from the various other Windows App SDK templates.
+
+When using a project reference, projects made from the following templates are not supported currently: C++ app with WAP, C++ Unpackaged, and C# Unpackaged.
```
angelazhangmsft, line 168:
```suggestion
When referencing a C# authored component as a project reference, the following application configurations are not supported currently: 
- C++/WinRT applications using a separate Windows Application Packaging Project
- C++ Unpackaged applications
- C# Unpackaged applications 

```

docs/authoring.md
```diff
@@ -162,12 +159,33 @@ Consuming C#/WinRT components from MSIX-packaged applications is supported with
...
 
+### Using Windows App SDK
+
+Authoring WinUI types with C#/WinRT is in development and is supported with some exceptions.
+Component authors can implement WinUI types using the Windows App SDK class library template. 
+Components are consumed by projects made from the various other Windows App SDK templates.
+
+When using a project reference, projects made from the following templates are not supported currently: C++ app with WAP, C++ Unpackaged, and C# Unpackaged. 
+
+Using a package reference to the authored component will work unless authoring a user control, which is not supported yet.
```
angelazhangmsft, line 170: Do we need to call out that Unpackaged apps and/or apps using the Windows Application Packaging Project are not supported using a package reference also?

docs/authoring.md
```diff
@@ -162,12 +159,33 @@ Consuming C#/WinRT components from MSIX-packaged applications is supported with
...
+
+When authoring a custom user control, consuming the component via NuGet package is not supported currently.
```
angelazhangmsft, line 172: is this the same as the sentence before?

docs/authoring.md
```diff
@@ -162,12 +159,33 @@ Consuming C#/WinRT components from MSIX-packaged applications is supported with
...
+If a C++ app is consuming a component that provides a custom user control, an additional entry needs to be made in the `pch.h` header file and in the manifest. 
+
+For the `pch.h` file, add `#include <winrt/MyAuthoredComponent.MyAuthoredComponent_XamlTypeInfo.h>`.
+For the manifest, add an additional `ActivatableClass` entry for `MyAuthoredComponent.MyAuthoredComponent_XamlTypeInfo.XamlMetaDataProvider`.
+
+When using a template for a WAP-style project, the manifest entries should be added to the `Package.appxmanifest`. 
+Otherwise, follow the steps above for adding a `Foo.exe.manifest` -- if you get an error in the linker (LNK) 
+typically this is due to multiple manifests being found, and `app.manifest` can be removed to resolve the issue.
+
+There's no inherent need to create a `runtimeconfig.json`; typically the one automatically created suffices.
```
angelazhangmsft, line 183: I don't think we need to explicitly say this unless there are specific scenarios to call out where you need to create this file?

docs/authoring.md
```diff
@@ -90,6 +90,7 @@ This section describes the steps needed to consume a C#/WinRT component for the
 - [C# desktop applications](#consuming-from-c-applications) (.NET 5 or later)
 - [Packaged applications](#Consuming-from-packaged-applications)
 - [Out of process components](#Consuming-an-out-of-process-component)
+- [Windows App SDK Support](#Using-Windows-App-SDK)
```
angelazhangmsft, line 93: How about "Windows App SDK applications"? 

We may also want to note that these are C# and C++ desktop applications/distinguish these from the first 2 scenarios.

docs/authoring.md
```diff
@@ -162,12 +159,33 @@ Consuming C#/WinRT components from MSIX-packaged applications is supported with
...
+Using a package reference to the authored component will work unless authoring a user control, which is not supported yet.
+
+When authoring a custom user control, consuming the component via NuGet package is not supported currently.
+
+If a C++ app is consuming a component that provides a custom user control, an additional entry needs to be made in the `pch.h` header file and in the manifest. 
+
+For the `pch.h` file, add `#include <winrt/MyAuthoredComponent.MyAuthoredComponent_XamlTypeInfo.h>`.
+For the manifest, add an additional `ActivatableClass` entry for `MyAuthoredComponent.MyAuthoredComponent_XamlTypeInfo.XamlMetaDataProvider`.
+
+When using a template for a WAP-style project, the manifest entries should be added to the `Package.appxmanifest`.
```
angelazhangmsft, line 179:
```suggestion
When consuming the component from a packaged application, the manifest entries should be added to the `Package.appxmanifest` file. You will need to right click on the file and select **Open With** -> **XML (Text Editor)** in order to edit the file.
```

angelazhangmsft: I used Package.appxmanifest for both packaged (non-WAP) and packaged (with WAP)

docs/authoring.md
```diff
@@ -162,12 +159,33 @@ Consuming C#/WinRT components from MSIX-packaged applications is supported with
...
+When authoring a custom user control, consuming the component via NuGet package is not supported currently.
+
+If a C++ app is consuming a component that provides a custom user control, an additional entry needs to be made in the `pch.h` header file and in the manifest. 
+
+For the `pch.h` file, add `#include <winrt/MyAuthoredComponent.MyAuthoredComponent_XamlTypeInfo.h>`.
+For the manifest, add an additional `ActivatableClass` entry for `MyAuthoredComponent.MyAuthoredComponent_XamlTypeInfo.XamlMetaDataProvider`.
+
+When using a template for a WAP-style project, the manifest entries should be added to the `Package.appxmanifest`. 
+Otherwise, follow the steps above for adding a `Foo.exe.manifest` -- if you get an error in the linker (LNK) 
+typically this is due to multiple manifests being found, and `app.manifest` can be removed to resolve the issue.
```
angelazhangmsft, line 181: Can we add this specific error to the troubleshooting section?

docs/authoring.md
```diff
@@ -162,12 +159,33 @@ Consuming C#/WinRT components from MSIX-packaged applications is supported with
...
+When using a project reference, projects made from the following templates are not supported currently: C++ app with WAP, C++ Unpackaged, and C# Unpackaged. 
+
+Using a package reference to the authored component will work unless authoring a user control, which is not supported yet.
+
+When authoring a custom user control, consuming the component via NuGet package is not supported currently.
+
+If a C++ app is consuming a component that provides a custom user control, an additional entry needs to be made in the `pch.h` header file and in the manifest. 
+
+For the `pch.h` file, add `#include <winrt/MyAuthoredComponent.MyAuthoredComponent_XamlTypeInfo.h>`.
+For the manifest, add an additional `ActivatableClass` entry for `MyAuthoredComponent.MyAuthoredComponent_XamlTypeInfo.XamlMetaDataProvider`.
```
angelazhangmsft, line 177: Probably add the code snippet here so folks can copy/paste?

asklar:
thanks @j0shuams ! 
> Because we ship a WinRT.Runtime.dll and the .NET SDK also has its own version, we explicitly pick the dll with the latest version for copying.

Is there an expectation that there are no ABI changes between different versions of this binary? IOW is it always safe to swap one WinRT.Runtime for another?

docs/authoring.md
j0shuams: I agree this might not be necessary, I added it to try to add clarification after the discussion in this post - https://github.com/microsoft/CsWinRT/issues/1088

docs/authoring.md
```diff
@@ -108,7 +109,7 @@ Follow these steps to add an application manifest file to your project:
 
 1. In Visual Studio, right click on the project node under **Solution Explorer** and click **Add -> New Item**. Search for the **Text File** template and name your file `YourNativeApp.exe.manifest`.
 
-2. Add your activatable class registrations to the manifest file. Below is an example with a C#/WinRT component `MyAuthoredComponent` and activatable classes `WinRTComponent.Class*`. Note that `WinRT.Host.dll` must be specified as the file name.
+2. Add your activatable class registrations to the manifest file. Below is an example with a C#/WinRT component `MyAuthoredComponent` and activatable classes `WinRTComponent.Class*`. Note that `WinRT.Host.dll` must be specified as the file name, without a relative path.
 
       ```xml
```
asklar, line 114: btw / for future reference that this is automatable if the types exist in a winmd (see the MT task I have in the targets file for the nuget package called `Unpackaged`)

docs/authoring.md
```diff
@@ -162,14 +159,39 @@ Consuming C#/WinRT components from MSIX-packaged applications is supported with
 
 C#/WinRT supports authoring out-of-process components that can be consumed by Windows Runtime compatible languages. Currently, consuming an out-of-process component is supported for managed C# apps with the use of a packaging project. This scenario requires creating a `WinRT.Host.runtimeconfig.json` file, as demonstrated in the [background task component sample](https://github.com/microsoft/CsWinRT/tree/master/src/Samples/BgTaskComponent). Native C++ consumption of out-of-process C# components is not fully supported yet.
 
+### Windows App SDK Applications
+
+Authoring WinUI types with C#/WinRT is in development and is supported with some exceptions.
+Component authors can implement WinUI types using the Windows App SDK class library template. 
+Components can be consumed by application projects that reference the Windows App SDK.
+
+When referencing a C# authored component via project reference, the following application types are not supported currently:
```
asklar, line 168: from reading this I know what _not_ to do but not what to do. Can we also list which configuration _is_ supported?

docs/authoring.md
asklar: You can link to the existing instructions here: https://docs.microsoft.com/en-us/windows/uwp/cpp-and-winrt-apis/troubleshooting#symptoms-and-remedies
See the last entry in that list

docs/authoring.md
asklar: see https://github.com/asklar/xaml-islands/blob/main/build/Unpackaged.targets#L8-L38

j0shuams:
> thanks @j0shuams !
> 
> > Because we ship a WinRT.Runtime.dll and the .NET SDK also has its own version, we explicitly pick the dll with the latest version for copying.
> 
> Is there an expectation that there are no ABI changes between different versions of this binary? IOW is it always safe to swap one WinRT.Runtime for another?

I think that's true about ABI changes, and the only time the version matters would be if a feature from the latest release was being used. With that in mind, we typically suggest the latest version. @manodasanW might have some other thoughts on this

docs/authoring.md
j0shuams: thanks! sure, added

docs/authoring.md
j0shuams: great!! I think that would help us with https://github.com/microsoft/CsWinRT/issues/789

docs/authoring.md
angelazhangmsft: shoudl we list what is supported first?

docs/authoring.md
j0shuams: @asklar From what I see, this is needed for activating XAML types; do you have additional info on why we need this?

asklar: as far as the xamltype info header, this is necessary for things like XAML `Binding` to work; @RealTommyKlein can probably say better words :D
the activatable class entry is necessary to tell winrt "this is the library that knows how to create this type"

docs/authoring.md
j0shuams: Sure!

docs/authoring.md
```diff
@@ -86,78 +86,132 @@ To make your component available as a NuGet package, it is important to include
 
 This section describes the steps needed to consume a C#/WinRT component for the following scenarios:
 
-- [C++/WinRT desktop applications (Unpackaged)](#consuming-from-cwinrt-unpackaged)
-- [C# desktop applications](#consuming-from-c-applications) (.NET 5 or later)
-- [Packaged applications](#Consuming-from-packaged-applications)
-- [Out of process components](#Consuming-an-out-of-process-component)
+- [C# desktop applications](#consuming-from-c-applications) 
+- [C++ desktop applications](#consuming-from-c-applications)
```
asklar, line 90: these now point to the same anchor because markdown removes the ++ and # signs

docs/authoring.md
```diff
@@ -86,78 +86,132 @@ To make your component available as a NuGet package, it is important to include
...
-- [Out of process components](#Consuming-an-out-of-process-component)
+- [C# desktop applications](#consuming-from-c-applications) 
+- [C++ desktop applications](#consuming-from-c-applications)
+- [Windows App SDK applications](#windows-app-sdk-applications)
+- [Out of process components](#consuming-an-out-of-process-component)
 
-### Consuming from C++/WinRT (Unpackaged)
+### Consuming from C# applications
+
+Consuming a C#/WinRT component from C#/.NET apps is supported by both package reference or project reference. This scenario is equivalent to consuming any ordinary C# class library and does not involve WinRT activation in most cases. This scenario is fully supported on .NET 6 and later.
```
asklar, line 96: in most cases -> can we be more precise? what are the scenarios where it does involve winrt activation?

docs/authoring.md
```diff
@@ -86,78 +86,132 @@ To make your component available as a NuGet package, it is important to include
...
 
-**Application Manifest file**
+#### Manifest Class Registration
 
-For consumption of C#/WinRT components from unpackaged applications, you need to create an [application manifest file](https://docs.microsoft.com/windows/win32/sbscs/application-manifests) named `YourNativeApp.exe.manifest`. The application manifest file must specify activatable Windows Runtime classes in the DLL that the application will be making use of. At runtime, this directs activation of the component’s classes. See the [sample manifest file](https://github.com/microsoft/CsWinRT/tree/master/src/Samples/AuthoringDemo/CppConsoleApp/CppConsoleApp.exe.manifest) for an example.
+Consuming a C#/WinRT component from a C++/WinRT desktop application requires manifest class registrations. The specified activatable Windows Runtime class registrations will direct activation of the component’s classes at runtime. There are two different ways to do this depending on if the application is a packaged (with MSIX) or unpackaged C++ app:
```
asklar, line 111:
```suggestion
Consuming a C#/WinRT component from a C++/WinRT desktop application requires adding manifest class registrations to the application. The specified activatable Windows Runtime class registrations will direct activation of the component’s classes at runtime. There are two different ways to do this depending on whether the application is a packaged (with MSIX) or unpackaged C++ app:
```

docs/authoring.md
```diff
@@ -86,78 +86,132 @@ To make your component available as a NuGet package, it is important to include
...
 
-### Consuming from C# applications
+      1. In Visual Studio, right click on the project node under **Solution Explorer** and click **Add -> New Item**. Search for the **Text File** template and name your file `<MyNativeAppName>.exe.manifest`.
 
-Consuming a C#/WinRT component from C#/.NET apps is supported by both package reference or project reference. This scenario is equivalent to consuming any ordinary C# class library and does not involve WinRT activation in most cases.
+      2. Add your activatable class registrations to the manifest file. Below is an example with a C#/WinRT component called `MyAuthoredComponent` and activatable classes `WinRTComponent.Class*`. Note that `WinRT.Host.dll` must be specified as the file name, without a relative path.
 
-**Note:** With C#/WinRT 1.3.5 or later, project references to a C#/WinRT component require .NET 6.
+            ```xml
+            <?xml version="1.0" encoding="utf-8"?>
```
asklar, line 138: does the manifest not need a `<supportedOS>` entry for unpackaged?

docs/authoring.md
```diff
@@ -86,78 +86,132 @@ To make your component available as a NuGet package, it is important to include
...
 
-Consuming C#/WinRT components from MSIX-packaged applications is supported with a few modifications.
+The following table outlines the Windows App SDK application types that are supported when consuming C#/WinRT authored components implementing WinUI types:
+
+- ✅ means the scenario works
+- 🟨 means the scenario partially works (i.e. with specific limitations)
+- 🟥 means the scenario does not work
+- Note: "WAP" refers to the usage of a separate Windows Application Packaging Project in packaged apps.
+
+|Application Type|Project or Package Reference|Supported?|
```
asklar, line 169: I like the table approach though I'd recommend doing it as follows:

### C++

|  | Project Reference | Package Reference |
|:--|:-:|:-:|
| Packaged | ✅ | 🟨 |
| Packaged with WAP | 🟥 | 🟥 | 
| Unpackaged | 🟥 | 🟨 |


then another table for C# following it

docs/authoring.md
```diff
@@ -86,78 +86,132 @@ To make your component available as a NuGet package, it is important to include
...
+Components can be consumed by application projects that reference the Windows App SDK.
 
-### Consuming from packaged applications
+#### Support Matrix
 
-Consuming C#/WinRT components from MSIX-packaged applications is supported with a few modifications.
+The following table outlines the Windows App SDK application types that are supported when consuming C#/WinRT authored components implementing WinUI types:
+
+- ✅ means the scenario works
+- 🟨 means the scenario partially works (i.e. with specific limitations)
```
asklar, line 165: where can I learn more about the specific limitations? as a developer I'd like to be able to assess whether what I want to do will work, so having the limitations documented/called out seems important.

docs/authoring.md
j0shuams: IIUC, We are referring to out of process support, a few sections down, where you need a runtimeconfig.json. There are no scenarios where you need a manifest file for c#. Correct me if I'm wrong @angelazhangmsft  :)

I'll make the clarification in the docs

docs/authoring.md
j0shuams: I'm not aware of this, perhaps thats why the unpackged scenario doesn't work? When I made the unpackaged scenario I followed the [docs here](https://docs.microsoft.com/en-us/windows/apps/winui/winui3/create-your-first-winui3-app?pivots=winui3-unpackaged-cpp#unpackaged-create-a-new-project-for-an-unpackaged-c-or-c-winui-3-desktop-app).

asklar: also @stevewhims for context

docs/authoring.md
j0shuams: Addressed!

docs/authoring.md
```diff
@@ -86,78 +86,156 @@ To make your component available as a NuGet package, it is important to include
 
 This section describes the steps needed to consume a C#/WinRT component for the following scenarios:
 
-- [C++/WinRT desktop applications (Unpackaged)](#consuming-from-cwinrt-unpackaged)
-- [C# desktop applications](#consuming-from-c-applications) (.NET 5 or later)
-- [Packaged applications](#Consuming-from-packaged-applications)
-- [Out of process components](#Consuming-an-out-of-process-component)
+- [C# desktop applications](#consuming-from-c-applications) 
+- [C++ desktop applications](#consuming-from-c++-applications)
```
asklar:
```suggestion
- [C++ desktop applications](#consuming-from-c-applications-1)
```

asklar (approved):

angelazhangmsft (approved):

docs/authoring.md
```diff
@@ -86,78 +86,140 @@ To make your component available as a NuGet package, it is important to include
...
-- [Out of process components](#Consuming-an-out-of-process-component)
+- [C# desktop applications](#consuming-from-c-applications) 
+- [C++ desktop applications](#consuming-from-c-applications-1)
+- [Windows App SDK applications](#windows-app-sdk-applications)
+- [Out of process components](#consuming-an-out-of-process-component)
 
-### Consuming from C++/WinRT (Unpackaged)
+### Consuming from C# applications
+
+Consuming a C#/WinRT component from C#/.NET apps is supported by both package reference or project reference.
```
manodasanW, line 96: Perf offline discussion, might want to reword this as both lines seem to say the same.  Also since you are saying both, might want to use and rather than or.

manodasanW, line 96: Perf offline discussion, might want to reword this as both lines seem to say the same.  Also since you are saying both, might want to use and rather than or.

docs/authoring.md
```diff
@@ -86,78 +86,140 @@ To make your component available as a NuGet package, it is important to include
...
-      </file>
-      </assembly>
-      ```
+  - For **unpackaged C++ apps**: The app consumer needs to create an [application manifest file](https://docs.microsoft.com/windows/win32/sbscs/application-manifests) (also known as a "fusion manifest") that follows the naming: `<MyNativeAppName>.exe.manifest`. See this [sample manifest file](https://github.com/microsoft/CsWinRT/tree/master/src/Samples/AuthoringDemo/CppConsoleApp/CppConsoleApp.exe.manifest) for an example.
 
-3. Modify the project to include the manifest file in the output when deploying the project. Right-click on the file in **Solution Explorer**, select **Properties**, and set the **Content** property to **True** using the drop-down arrow on the right.
+      Follow these steps to add an application manifest file to your project:
 
-### Consuming from C# applications
+      1. In Visual Studio, right click on the project node under **Solution Explorer** and click **Add -> New Item**. Search for the **Text File** template and name your file `<MyNativeAppName>.exe.manifest`.
```
manodasanW, line 138: Wasn't there also a manifest file template?  Is that not better?

manodasanW, line 138: Wasn't there also a manifest file template?  Is that not better?

docs/authoring.md
```diff
@@ -166,10 +228,14 @@ C#/WinRT supports authoring out-of-process components that can be consumed by Wi
 
 - Project reference support (from C++ and C# apps) for C#/WinRT authored components is available with .NET 6. Referencing a C#/WinRT authored component by project reference in a C# app is **not supported with .NET 5**.
 
-- Hosting WinUI3 components is not fully supported yet.
-
 - Authoring issues are [tagged under the *authoring* label](https://github.com/microsoft/CsWinRT/issues?q=is%3Aopen+is%3Aissue+label%3Aauthoring). Feel free to [file an issue](https://github.com/microsoft/CsWinRT/issues/new/choose) tagged with the *authoring* label if you encounter any new issues!
 
+- The docs [here](https://docs.microsoft.com/en-us/windows/uwp/cpp-and-winrt-apis/troubleshooting#symptoms-and-remedies) have useful troubleshooting tips for working with C++/WinRT.
```
manodasanW, line 233: en-us

manodasanW, line 233: en-us

docs/authoring.md
```diff
@@ -86,78 +86,140 @@ To make your component available as a NuGet package, it is important to include
...
+|C# Unpackaged|Project Reference|🟥| [#1116](https://github.com/microsoft/CsWinRT/issues/1116) |
+|C# Packaged|Package Reference| 🟨 | [#1118](https://github.com/microsoft/CsWinRT/issues/1118) |
+|C# Packaged with WAP |Package Reference| 🟨 | [#1118](https://github.com/microsoft/CsWinRT/issues/1118) |
+|C# Unpackaged|Package Reference|🟥 | [#1116](https://github.com/microsoft/CsWinRT/issues/1116) |
+
+#### Consumption of WinUI controls from C++ apps
+
+If a C++ app is consuming a component that provides a custom user control, there are a few extra configuration steps:
+
+1. In the app's `pch.h` header file, add the following line:
```
manodasanW, line 198: Make note somewhere MyAuthoredComponent should be renamed to their component name.

manodasanW, line 198: Make note somewhere MyAuthoredComponent should be renamed to their component name.

docs/authoring.md
```diff
@@ -86,78 +86,140 @@ To make your component available as a NuGet package, it is important to include
...
+Component authors can implement WinUI types using the Windows App SDK C# class library project template. 
+Components can be consumed by application projects that reference the Windows App SDK.
 
-### Consuming from packaged applications
+#### Support Matrix
 
-Consuming C#/WinRT components from MSIX-packaged applications is supported with a few modifications.
+The following table outlines the Windows App SDK application types that are supported when consuming C#/WinRT authored components implementing WinUI types:
+
+- ✅ means the scenario works
```
manodasanW, line 169: Consider removing _means the_ and just say Scenario works.

manodasanW, line 169: Consider removing _means the_ and just say Scenario works.

docs/authoring.md
```diff
@@ -86,78 +86,140 @@ To make your component available as a NuGet package, it is important to include
...
+- 🟨 means the scenario partially works (i.e. with specific limitations)
+- 🟥 means the scenario does not work
+- Note: "WAP" refers to the usage of a separate Windows Application Packaging Project in packaged apps.
```
manodasanW, line 172: May also want to say Packaged without WAP refers to the WinUI preview MSIX support for it.

manodasanW, line 172: May also want to say Packaged without WAP refers to the WinUI preview MSIX support for it.

manodasanW (approved):

docs/authoring.md
angelazhangmsft: Packaged without WAP is not a preview feature anymore I believe - https://docs.microsoft.com/windows/apps/windows-app-sdk/single-project-msix?tabs=csharp

docs/authoring.md
j0shuams: Don't think so, this is how we've always done it

asklar: there is a manifest item template, but only for managed projects, for some reason....
image

PR #1115, opened 2022-02-25
Move object marshalers to be value type based

manodasanW:
Today, when marshaling a reference type, we allocate an IObjectReference to hold onto the ptr which is a heap allocation and needs to also be garbage collected later.  This moves to instead using a value type to hold onto it as we use the ptr and dispose of it immediately in marshaling scenarios and don't need to change threads / contexts.

- Achieved by adding a new CreateMarshaler2 function which returns an ObjectReferenceValue
- Made further use of it in FromManaged and CopyManaged scenarios where the IObjectReference we allocated was even more temporary as they are used to return a ptr to the ABI.
- This change doesn't change generic marshalers which I will look at if it is possible to address in a future PR.


src/WinRT.Runtime/ObjectReference.cs
```diff
@@ -613,4 +631,61 @@ public override unsafe int TryAs<U>(Guid iid, out ObjectReference<U> objRef)
...
+            this.ptr = ptr;
+            this.referenceTracker = referenceTracker;
+            this.preventReleaseOnDispose = preventReleaseOnDispose;
+            this.objRef = objRef;
+
+        }
+
+        public readonly IntPtr GetAbi() => ptr;
+
+        public unsafe readonly IntPtr DetachRef()
```
manodasanW, line 661: Might actually rename this to Detach.

src/WinRT.Runtime/Marshalers.cs
```diff
@@ -830,10 +840,42 @@ public static unsafe MarshalerArray CreateMarshalerArray(T[] array, Func<T, IObj
...
+            {
+                return m;
+            }
+            bool success = false;
+            try
+            {
+                int length = array.Length;
+                var byte_length = length * IntPtr.Size;
+                m._array = Marshal.AllocCoTaskMem(byte_length);
+                m._objectReferenceValues = new ObjectReferenceValue[length];
```
manodasanW, line 856: Consider refactoring to share code.

src/cswinrt/code_writers.h
```diff
@@ -7477,14 +7510,14 @@ delegate* unmanaged[Stdcall]<System.IntPtr, WinRT.EventRegistrationToken, int> r
...
 }
 
-protected override IObjectReference CreateMarshaler(% del) =>
-del is null ? null : %.CreateMarshaler(del);
+protected override ObjectReferenceValue CreateMarshaler(% del) =>
+%.CreateMarshaler2(del);
 
-protected override void DisposeMarshaler(IObjectReference marshaler) =>
-%.DisposeMarshaler(marshaler);
+protected override void DisposeMarshaler(ObjectReferenceValue marshaler) =>
```
manodasanW, line 7516: Might not be need anymore.

ujjwalchadha (approved):

PR #1127, opened 2022-03-08
Fix out of space disk issue in pipelines

manodasanW:
Discovered that pdbs were taking a good amount of space in the output.  And this was from them being copied for each project reference.  Making use of AllowedReferenceRelatedFileExtensions in our pipelines to avoid copying pdbs for each project reference such as in our tests.


build/AzurePipelineTemplates/CsWinRT-BuildAndTest-Stage.yml
```diff
@@ -169,7 +169,8 @@ stages:
 
           set cswinrt_echo=on
           set cswinrt_build_only=true
-          set cswinrt_build_params=/bl:$(Build.SourcesDirectory)\cswinrt.binlog /p:CleanIntermediateDirs=true
+          rem AllowedReferenceRelatedFileExtensions includes all the default extensions other than pdb to address space issues resulting from pdbs being copied for each project reference.
```
j0shuams, line 172: The comment is a bit confusing, it seems to suggest we are taking up more space by including more binaries. I like the description you have in this PR, we can reword this comment?

manodasanW: Addressed.

manodasanW: Addressed.

PR #1130, opened 2022-03-09
add property for authoring project references

angelazhangmsft:
`<WindowsAppContainer>true</WindowsAppContainer>` is required for project references from C++ to C#.


j0shuams (approved):
Thanks Angela!

PR #1154, opened 2022-03-30
Pipeline for localization

j0shuams:
This PR adds a new pipeline yml for localization, which uses a new template for doing a simple build of cswinrt. We only build for Release|x64 because the resource files are platform agnostics, and only building for a specific flavor is faster. Also included is a template for calling TouchdownBuild to localize our resources.


build/AzurePipelineTemplates/CsWinRT-BuildOnly-Stage.yml
```diff
@@ -0,0 +1,380 @@
...
+    
+# Clone TestWinRT
+    - task: CmdLine@2
+      displayName: Clone TestWinRT
+      enabled: false
+      inputs: 
+        script: get_testwinrt.cmd
+        workingDirectory: $(Build.SourcesDirectory)
+
+# Download procdump
```
manodasanW, line 29: Do you think we should reuse the common stuff by refactoring it to a separate file?

build/AzurePipelineTemplates/CsWinRT-BuildOnly-Stage.yml
```diff
@@ -0,0 +1,380 @@
...
+  - job: Build
+    pool:
+      vmImage: windows-2019
+    timeoutInMinutes: 90
+    # https://docs.microsoft.com/en-us/azure/devops/pipelines/process/phases?view=azure-devops&tabs=yaml#multi-job-configuration
+    strategy:
+      maxParallel: 10
+      matrix:
+       x64_Release:
+          BuildPlatform: 'x64'
```
manodasanW, line 14: x64 or x86, I see the checks below are for x86.

manodasanW: Those checks were actually the ones from before that are common.

build/AzurePipelineTemplates/CsWinRT-BuildOnly-Stage.yml
manodasanW: like CsWinRT-Build-Steps or so?

manodasanW (approved):

PR #1166, opened 2022-04-06
Refactor embedded tests for better automation

### LINKED ISSUE #1085: Add embedded tests to pipeline
The embedded sample + tests only build locally, we should add them to our CI.

j0shuams:
Today, the embedded samples and tests use the official CsWinRT NuGet to test embedded support. This required devs to update the reference with an experimental package to do testing. This PR creates a better workflow for testing changes against embedded test collateral by not requiring any dev overhead. 

Resolves #1085


src/Samples/TestEmbedded/Directory.Build.targets
```diff
@@ -1,2 +1,121 @@
...
+ 
+  <Target Name="CsWinRTAddEmbeddedRuntime_AnySources"
+          Condition="'$(CsWinRTEmbedded)' == 'true'"
+          DependsOnTargets="CsWinRTPrepareProjection"
+          AfterTargets="CsWinRTGenerateProjection"
+          BeforeTargets="CsWinRTIncludeProjection"> 
+  
+    <PropertyGroup>
+      <EmbeddedAnyFiles>
+        $(WinRTRuntimeSourcesDir)TypeNameSupport.cs;
```
manodasanW, line 25: Any reason why we can't make this *.cs similar to the cswinrt targets or make use of the cswinrt targets logic here?

j0shuams: Yeah the issue is * will grab net5 and netstandard2 versions of files. I tried some fanciness that grabbed them all and then took them out, and I tried string functions in msbuild to ignore net5/netstandard2 filenames, but neither was satisfactory. 

This is definitely something I want a better solution for. If you add a new file to WinRT.Runtime, you need to add it here too.

manodasanW: The thing you mention about having to add it here was my concern as it is likely to be forgotten.  What was the issue when you tried out the first approach (i.e. get all winrtruntime sources - your net5 list - your netstandard list)?  Also any reason why we can't just get the CsWinRT nuget targets to do what we want here and just update the property it looks under to be like a new property like WinRTRuntimeSourcesDir and in the case of nuget reference we set it to the current value and in the case of it is building here, we override it similar to CsWinRTPath that you see above.

manodasanW (approved):

PR #1167, opened 2022-04-11
Correctly resolve WinRT.Runtime.dll in Authoring targets

### LINKED ISSUE #1156: CsWinRT authoring targets fails in CsWinRTAuthoring_AddManagedDependencies
.nuget\packages\microsoft.windows.cswinrt\1.6.1\build\Microsoft.Windows.CsWinRT.Authoring.targets(100,7): error MSB4184: The expression "[MSBuild]::VersionGreaterthanOrEquals('', 1.6.1.38391)" cannot be evaluated. Version string was not in a correct format.

Authoring component has CsWinRT reference and ```<WindowsSdkPackageVersion>10.0.18362.24</WindowsSdkPackageVersion>```

j0shuams:
We introduced some de-duplication logic in our authoring targets that work to isolate the latest version of WinRt.Runtime and use it when building components. In this we assumed CsWinRT nupkg would always be available, so the WinRT.Runtime from the lib folder would be comparable. But we hit a WinUI failure where this wasn't the case, so we remedy by adding a check to our de-dup logic. 

Fixes #1156 

After we get this in, we will need to remove the [workaroud ](https://microsoft.visualstudio.com/WinUI/_git/microsoft-ui-xaml-lift/commit/b1d9e0d6d126c5eba44ac02a25de405ab1a287fb?refName=refs%2Fheads%2Fuser%2Fjnordi%2FIxpIngestion_0402) in WinUI.


nuget/Microsoft.Windows.CsWinRT.Authoring.targets
```diff
@@ -95,10 +95,11 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <PropertyGroup>
 
       <_CsWinRTDeDupRuntime_CsWinRT_Version_IsLatest Condition="'@(_CsWinRTDeDupRuntime_NETSDK)' == ''">true</_CsWinRTDeDupRuntime_CsWinRT_Version_IsLatest>
+      <_CsWinRTDeDupRuntime_NoCsWinRT_Reference Condition="'@(_CsWinRTDeDupRuntime_CsWinRT)' == ''">true</_CsWinRTDeDupRuntime_NoCsWinRT_Reference>
 
     </PropertyGroup>
 
-    <PropertyGroup Condition="'$(_CsWinRTDeDupRuntime_CsWinRT_Version_IsLatest)' != 'true'">
+    <PropertyGroup Condition="'$(_CsWinRTDeDupRuntime_CsWinRT_Version_IsLatest)' != 'true' and '$(_CsWinRTDeDupRuntime_NoCsWinRT_Reference)' != 'true'">
```
manodasanW, line 102: What is the scenario where both these conditions are met?

manodasanW (approved):

PR #1176, opened 2022-04-22
Resolve authoring issues with AnyCPU

j0shuams:
This PR changes how customers will work with authoring support when using cswinrt components with native apps (via project reference). 

Context: 
* For native apps to activate/use types authored in a CsWinRT component, must (at runtime) have a copy of WinRT.Host.dll (+ MUI) and WinRT.Host.Shim.dll 
* On project references, we assume the component will build for the same platform as the native app. Then, using that assumption, we copy the platform-specific hosting DLLs to the output directory. 

The issue today: 
* Components are advised to only target AnyCPU, and when they do this our build logic assumes they mean x86. 
This leads to an issue where a native app building for x64, referencing a component building for AnyCPU, will get the wrong version of the hosting DLL WinRT.Host.dll

Proposed solution:
* Make the automation triggered by a property instead of automatic. This PR introduces a new property "CsWinRTHostPlatform". Component authors will specify which platform the native app will build for, so the right WinRT.Host.dll is used. 
*  Arrived at this solution after not being able to use CsWinRT targets during the native app's build that read the native apps platform. 
* There is a drawback: if wanting to build for different platforms, the property must be changed in the component's .csproj 

Something else we considered:
* Providing a basic Directory.Build.targets that customers could include in their native app that copies the hosting DLLs. 
* Decided against it because locating the CsWinRT nuget package location (from some pre-written targets) isn't possible in the native app (since it doesn't have a reference to CsWinRT, just the component project)

### DIFF
--- a/nuget/Microsoft.Windows.CsWinRT.Authoring.targets
+++ b/nuget/Microsoft.Windows.CsWinRT.Authoring.targets
@@ -15,11 +15,6 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <GetCopyToOutputDirectoryItemsDependsOn>$(GetCopyToOutputDirectoryItemsDependsOn);CsWinRTAuthoring_AddManagedDependencies</GetCopyToOutputDirectoryItemsDependsOn>
   </PropertyGroup>
 
-  <PropertyGroup>
-    <CsWinRTAuthoring_Platform Condition="'$(Platform)'=='AnyCPU'">x86</CsWinRTAuthoring_Platform>
-    <CsWinRTAuthoring_Platform Condition="'$(Platform)'!='AnyCPU'">$(Platform)</CsWinRTAuthoring_Platform>
-  </PropertyGroup>
-
   <ItemGroup>
     <CompilerVisibleProperty Include="AssemblyName" />
     <CompilerVisibleProperty Include="AssemblyVersion" />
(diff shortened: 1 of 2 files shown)


j0shuams:
@rainersigwald @dsplaisted Would y'all share your thoughts on this proposed solution? We are thinking there may be another drawback when customers want to build for multiple platforms. 

Maybe they could write a pre-build step customers could write?

dsplaisted:
@j0shuams Could you use [SetPlatform negotiation](https://github.com/dotnet/msbuild/blob/main/documentation/ProjectReference-Protocol.md#setplatform-negotiation) for this?

rainersigwald:
> @j0shuams Could you use [SetPlatform negotiation](https://github.com/dotnet/msbuild/blob/main/documentation/ProjectReference-Protocol.md#setplatform-negotiation) for this?

cc @BenVillalobos

nuget/Microsoft.Windows.CsWinRT.Authoring.targets
```diff
@@ -55,22 +50,20 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
       <TargetPath>WinRT.Host.Shim.dll</TargetPath>
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
 
-    <None Condition="Exists('$(CsWinRTPath)hosting\$(CsWinRTAuthoring_Platform)\native\WinRT.Host.dll')" 
-          Include="$(CsWinRTPath)hosting\$(CsWinRTAuthoring_Platform)\native\WinRT.Host.dll">
+    <None Include="$(CsWinRTPath)hosting\$(CsWinRTHostPlatform)\native\WinRT.Host.dll">
       <TargetPath>WinRT.Host.dll</TargetPath>
-      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
```
jlaanstra, line 63: This breaks incremental builds.

j0shuams:
I've investigated the SetPlatform feature of MSBuild, but it is not made for builds within VS. Therefore it won't help us. With no other options, we will update the docs to express that manually "binplace" logic will need to be done by the component author (project reference scenarios only).

PR #1181, opened 2022-04-29
Add support for components defining their own Equals and GetHashCode

### LINKED ISSUE #864: error CS0111: Type 'SomeType' already defines a member called 'Equals' with the same parameter types
**Describe the bug**
CsWinRT generates duplicate Equals methods if the underlying type in the WInMD already has an Equals() method with itself as the type parameter

**To Reproduce**
Create a C++WinRT component with a type that has an Equals method with itself as the equator. For example:

```
    runtimeclass SomeType
    {
        SomeType();

        Boolean Equals(SomeType someType);
    };
```

Pass the WinMD to CsWinRT.

**Expected behavior**
No error.

**Version Info**
CsWinRT 1.2.6

manodasanW:
- If a component doesn't provide one, CsWinRT provides one as it did before.
- If a component does provide one, then CsWinRT calls the provided one instead.
- For Equal(Class..), if the return type doesn't match with IEquatable, then CsWinRT generates its own explicitly implemented version for the IEquatable interface and also defines one to call the component provided one.

Fixes #864

### DIFF
--- a/src/cswinrt/helpers.h
+++ b/src/cswinrt/helpers.h
@@ -463,9 +463,9 @@ namespace cswinrt
     bool is_param_expected_type(TypeSig const& param, T const& expected_type)
     {
         auto semantics = get_type_semantics(param);
-        if (auto ft = std::get_if<T>(&semantics))
+        if (auto variant_type = std::get_if<T>(&semantics))
         {
-            return *ft == expected_type;
+            return *variant_type == expected_type;
         }
 
         return false;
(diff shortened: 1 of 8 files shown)


src/cswinrt/helpers.h
```diff
@@ -459,6 +459,180 @@ namespace cswinrt
         return {};
     }
 
+    template<typename T>
+    bool is_param_expected_type(TypeSig const& param, T const& expected_type)
+    {
+        auto semantics = get_type_semantics(param);
+        if (auto ft = std::get_if<T>(&semantics))
```
j0shuams, line 466: nit: might consider a dif var name than `ft`

j0shuams (approved):

PR #1182, opened 2022-04-29
Fix escaping of embedded quotes.

### LINKED ISSUE #1040: Escaped quotes in custom attribute string values
Followup from #1037 - a string from IDL like:
```cs
[custom("hello \"there\"")]
```

ends up wrong in the resulting cs since the generated code C# "@ string" requires escaped quotes to be double quotes

manodasanW:
Followed the conventions for @ in C#, where quotes (\") are embedded as double quotes ("") and ' and \ are embedded as is without the escape character.

Fixes #1040


j0shuams (approved):

asklar:
thanks @manodasanW !!

PR #1186, opened 2022-05-06
Pack project PRI next to WinMD

### LINKED ISSUE #1118: Cannot used authored user controls via authored nupkg
**Describe the bug**
An app referencing a C#/WinRT Component NuPkg that has a custom user control will fail at runtime with something like 

`Exception thrown at 0x00007FF8D848358C (KernelBase.dll) in WinUICppApp.exe: WinRT originate error - 0x80004005 : 'Cannot locate resource from 'ms-appx:///SampleFormControl.xaml'.'.`

Where `SampleFormControl` is the custom user control.

**To Reproduce**
1. Make a C#/WinRT component 
2. Add a custom user control to the component
3. Make a nupkg of the component
4. Make a WinAppSDK app (Packaged or packaged with wap, either C# or C++) 
5. add a reference to the nupkg from (3) in (4)
6. Build and run

**Expected behavior**
App runs normally

**Version Info**
Windows App SDK - 1.1.0-20220215.0-CI-experimental

**Additional context**
The issue is with how PriIndexName is generated, and could be related to [this](https://microsoft.visualstudio.com/OS/_workitems/edit/36925415).

j0shuams:
Today, C++ apps that try to use XAML/WinUI types authored with C#/WinRT fail at runtime because they can't find the xaml resources. 

E.g.
`WinRT originate error - 0x80004005 : 'Cannot locate resource from 'ms-appx:///WinUIComponentCs/NameReporter.xaml'.'.`

Evelyn and I found that the component's PRI file needs to live in same folder as the WinMD. 
This PR updates our build logic to reflect this requirement. 

Fixes #1118 

We're taking advantage of msbuild here actually. Since the pri file has the same name as the winmd and the winmd is a reference assembly, msbuild will grab the pri. It's not just for PRI, PRI is one of the allowed extensions for this behavior.

### DIFF
--- a/nuget/Microsoft.Windows.CsWinRT.Authoring.targets
+++ b/nuget/Microsoft.Windows.CsWinRT.Authoring.targets
@@ -281,6 +281,12 @@ Copyright (C) Microsoft Corporation. All rights reserved.
         <PackagePath>lib\$(TargetFramework)</PackagePath>
       </Content>
 
+      <!-- Pack any XAML resource PRI files next to the .winmd -->
+      <Content Include="$(ProjectPriFullPath)">
+        <Pack>true</Pack>
+        <PackagePath>lib\$(TargetFramework)\winmd</PackagePath>
+      </Content>    
+
       <!-- Pack the WinMD we generated in its own folder, under the managed TFM folder -->
       <Content Include="$(TargetDir)$(AssemblyName).winmd">
         <Pack>true</Pack> 


evelynwu-msft (approved):
:shipit:

j0shuams:
Confirmed that both C++ and C# apps that use nuget references to authored components work with custom user controls.

manodasanW (approved):

PR #1187, opened 2022-05-10
Add public option to embedded support for enums

j0shuams:
In trying to enable PowerShell's AppX module, we found that some of the commands shipped within the module have options. These command line options are based on projected enum values, and the existing embedded support made those enums private to the module. There's no harm in exposing enums since they're simple objects, so we support public enums now.

### DIFF
--- a/nuget/Microsoft.Windows.CsWinRT.targets
+++ b/nuget/Microsoft.Windows.CsWinRT.targets
@@ -140,8 +140,8 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
       <CsWinRTInternalProjection Condition="'$(CsWinRTPrivateProjection)' == 'true'">-internal</CsWinRTInternalProjection>
       <CsWinRTEmbeddedProjection Condition="'$(CsWinRTEmbedded)' == 'true'">-embedded</CsWinRTEmbeddedProjection>
-	    <CsWinRTEmbeddedEnums Condition="'$(CsWinRTEmbeddedPublicEnums)' == 'true'">-public_enums</CsWinRTEmbeddedEnums>
-	    <CsWinRTExeTFM Condition="$([MSBuild]::GetTargetFrameworkVersion('$(TargetFramework)')) == 5">net5.0</CsWinRTExeTFM>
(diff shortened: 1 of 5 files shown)


nuget/Microsoft.Windows.CsWinRT.targets
```diff
@@ -140,7 +140,8 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
       <CsWinRTInternalProjection Condition="'$(CsWinRTPrivateProjection)' == 'true'">-internal</CsWinRTInternalProjection>
       <CsWinRTEmbeddedProjection Condition="'$(CsWinRTEmbedded)' == 'true'">-embedded</CsWinRTEmbeddedProjection>
-      <CsWinRTExeTFM Condition="$([MSBuild]::GetTargetFrameworkVersion('$(TargetFramework)')) == 5">net5.0</CsWinRTExeTFM>
+	    <CsWinRTEmbeddedEnums Condition="'$(CsWinRTEmbeddedPublicEnums)' == 'true'">-public_enums</CsWinRTEmbeddedEnums>
```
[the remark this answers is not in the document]
j0shuams: It showed the inverse bad indenting when I previewed, not sure how this happened. Fixing

src/Samples/TestEmbedded/TestEmbeddedLibrary/TestEmbeddedLibrary.csproj
```diff
@@ -11,6 +11,8 @@
     <CsWinRTWindowsMetadata>10.0.19041.0</CsWinRTWindowsMetadata>
     <SimulateCsWinRTNugetReference>true</SimulateCsWinRTNugetReference>
     <CsWinRTGenerateProjection>true</CsWinRTGenerateProjection>
+    <!-- Will expose enums as public in the projection
```
manodasanW, line 14: Do we want a test or something for this option?

manodasanW (approved):

src/Samples/TestEmbedded/TestEmbeddedLibrary/TestEmbeddedLibrary.csproj
j0shuams: I think it's fine to leave testing this particular case to manual testing. We'd need another library for the different projection, plus the bells and whistles. I think the impact of these changes is low enough to not warrant such additional testing infrastructure.

PR #1190, opened 2022-05-11
Port cswinmd tooling

asklar:
This is the first PR to bring cswinmd over into this repo.

### Description
CSWinMD is a tool to generate WinMD files from C#/WinRT .cs files (technically not valid C#, but C#-ish).
This can be used as a replacement to MIDL for producing WinMD files; instead of writing .idl, you'd write this C# syntax in your C++ project, reference the cswinmd nuget, and it would run cswinmd to generate the winmd from your C# files, and then pass the winmd to cppwinrt for it to generate component headers/sources.

### Usage
```xml
  <Import Project="..\packages\CsWinMD.0.0.2-prerelease\build\cswinmd.targets" 
          Condition="Exists('..\packages\CsWinMD.0.0.2-prerelease\build\cswinmd.targets')" />
  <!-- ... -->
  <ItemGroup>
    <CSWinMDComponent Include="CSWinMDComponent.cs" />
  </ItemGroup>
```

### Current state and future work
This is not yet a complete implementation (it suffers from the same expressivity limitations that C#/WinRT currently has, like no support for unsealed classes, etc.) however it starts laying the groundwork for a more complete solution.
Included in this:
- CsWinMD generator
- targets file which will used in a new CsWinMD NuGet package
- Test component project that exercises the flow.

For more context (internal to msft only) see my talk https://aka.ms/cswinmd/talk_2021


sylveon:
This sounds like a great idea! IDL for authoring components is not the greatest development experience (both in terms of IDE and tooling support), so hopefully this can make it easier for C++/WinRT devs.

One thing that's probably worth exploring for a future revision is mixing with IDL, since some concepts aren't representable in C#, and to allow progressive migration to cswinmd.

Apparently, currently everything runs before cppwinrt, so it should allow IDL to use types from cswinmd, but not cswinmd to use types from IDL. If we run cswinmd after, then the reverse applies (cswinmd can consume types from IDL, but IDL can't consume types from cswinmd). So maybe an option to determine when in the build order a C# file gets compiled would be nice. Even better would be going for a two-pass build , so that both worlds can consume from each other. But that's much harder to implement, most likely.

src/Tests/CppComponent/CppComponent.vcxproj
```diff
@@ -0,0 +1,163 @@
...
+      <PreprocessorDefinitions>_DEBUG;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+    </ClCompile>
+  </ItemDefinitionGroup>
+  <ItemDefinitionGroup Condition="'$(Configuration)'=='Release'">
+    <ClCompile>
+      <PreprocessorDefinitions>NDEBUG;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <RuntimeLibrary Condition="'$(Configuration)|$(Platform)'=='Release|ARM'">MultiThreadedDLL</RuntimeLibrary>
+      <RuntimeLibrary Condition="'$(Configuration)|$(Platform)'=='Release|ARM64'">MultiThreadedDLL</RuntimeLibrary>
+      <RuntimeLibrary Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">MultiThreadedDLL</RuntimeLibrary>
+      <RuntimeLibrary Condition="'$(Configuration)|$(Platform)'=='Release|x64'">MultiThreadedDLL</RuntimeLibrary>
```
sylveon, line 114: Collapse all those into one and remove the condition?

src/Authoring/cswinmd/Program.cs
```diff
@@ -0,0 +1,139 @@
...
+            }
+        }
+
+        private static bool IsVersion(string v)
+        {
+            return Version.TryParse(Path.GetFileName(v), out var _);
+        }
+        private static string GetWindowsWinMdPath(string? sdkVersion)
+        {
+            using (var hklm = RegistryKey.OpenBaseKey(RegistryHive.LocalMachine, RegistryView.Registry32))
```
sylveon, line 122: Reduce nesting by using C# 8 using declarations?

src/Authoring/cswinmd/Program.cs
```diff
@@ -0,0 +1,137 @@
...
```
sylveon, line 122: We can reduce nesting even further:

```cs
using var hklm = RegistryKey.OpenBaseKey(RegistryHive.LocalMachine, RegistryView.Registry32);
using var roots = hklm.OpenSubKey(@"SOFTWARE\Microsoft\Windows Kits\Installed Roots");

var kitsRoot10 = (string)roots.GetValue("KitsRoot10");
var unionMetadata = Path.Combine(kitsRoot10, "UnionMetadata");
if (sdkVersion == null)
{
    var dirs = Directory.EnumerateDirectories(unionMetadata);
    sdkVersion = Path.GetFileName(dirs.Where(IsVersion).Last());
}

var path = Path.Combine(kitsRoot10, "UnionMetadata", sdkVersion, "Windows.winmd");
return path;
```

Not sure why the throw is here, it shouldn't be hit because there is a return right before it.

riverar:
For others: I was confused to see a C++ focused PR writeup for C# projections, so I asked Alexander for additional details to double-check my understanding.

The gist is that developers prefer idiomatic experiences, especially when it comes to component authoring and metadata generation. That is, rather than write awkward IDL and pass it to MIDL(RT), you can write classes in C#, Rust, etc. and a ECMA-335 compliant blob pops out the other end.

It is his belief that the C# authoring experience trumps the C++ IDL experience and provides a sample above. Shameless plug: Developers interested in a Rust idiomatic experience can also [follow similar development here](https://github.com/microsoft/windows-rs/issues/1093).

src/Authoring/cswinmd/build/cswinmd.targets
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project DefaultTargets="Build" ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Target Name="ProduceWinMDFromCSharpComponent" BeforeTargets="InsertComponentWinMD">
```
manodasanW, line 3: Consider prefixing the targets with like CsWinRT or CsWinMD to avoid conflicts with other potential targets.  This is pretty specific not to be an issue, but InsertComponentWinMD can end up being used elsewhere too.

src/Tests/CppComponent/Class1.cpp
```diff
@@ -0,0 +1,12 @@
+#pragma once
```
manodasanW, line 1: Probably want to rename CppComponent to like CsWinMDComponent so it is clear this is a CsWinMD test or moving it under a CsWinMD folder.

src/Tests/CppComponent/CppComponent.vcxproj
```diff
@@ -0,0 +1,161 @@
...
+  </ItemGroup>
+  <ItemGroup>
+    <CSharpCppComponent Include="CppComponent.cs" />
+  </ItemGroup>
+  <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
+  <PropertyGroup Label="CSWinMD overrides">
+    <CsWinMDPath>..\..\Authoring\cswinmd\bin\$(Configuration)\net5.0\CsWinMD.exe</CsWinMDPath>
+  </PropertyGroup>
+  <ImportGroup Label="ExtensionTargets">
+    <!-- <Import Project="..\packages\CsWinMD.0.0.2-prerelease\build\cswinmd.targets" Condition="Exists('..\packages\CsWinMD.0.0.2-prerelease\build\cswinmd.targets')" /> -->
```
manodasanW, line 149: Did you want to keep these commented just to have a way to easily test the nuget version too?

src/Authoring/cswinmd/CsWinMD.csproj
```diff
@@ -0,0 +1,36 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net5.0</TargetFramework>
+    <LangVersion>preview</LangVersion>
+    <Authors>Microsoft Corporation</Authors>
+    <Company>Microsoft Corporation</Company>
+    <Product>C#/WinRT</Product>
+    <PackageId>CsWinMD</PackageId>
```
manodasanW, line 10: Microsoft.Windows.CsWinMD?

manodasanW (approved):

src/Authoring/cswinmd/Program.cs
```diff
@@ -0,0 +1,152 @@
...
+        private static string GetWindowsWinMdPath(string? sdkVersion)
+        {
+            using var hklm = RegistryKey.OpenBaseKey(RegistryHive.LocalMachine, RegistryView.Registry32);
+            using var roots = hklm.OpenSubKey(@"SOFTWARE\Microsoft\Windows Kits\Installed Roots");
+            var kitsRoot10 = (string)roots.GetValue("KitsRoot10");
+            var unionMetadata = Path.Combine(kitsRoot10, "UnionMetadata");
+            if (sdkVersion == null)
+            {
+                var dirs = Directory.EnumerateDirectories(unionMetadata);
+                sdkVersion = Path.GetFileName(dirs.Where(IsVersion).Last());
```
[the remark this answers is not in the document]
asklar: I think it's guaranteed but we can sort it just in case

src/Tests/CppComponent/CppComponent.vcxproj
asklar: yes I use this to test it locally : )

src/Authoring/cswinmd/Program.cs
```diff
@@ -0,0 +1,212 @@
...
+using Microsoft.Win32;
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Diagnostics.CodeAnalysis;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+
+namespace Generator
```
kant2002, line 13: Just a question. Isn't possible for new code use 
```
namespace Generator;
```
to avoid nesting.

asklar: only for C# 10 and later. This builds with .net 5 which defaults to 9 I think... I do manually set the lang version in the project to preview so this _probably_ works, but feels low priority/impact (rest of the code in the repo uses the nesting style).

hez2010: Isn't .NET 5 about to reach its end of life? I think dropping .NET support is definitely doable.

asklar: the whole repo (and others) are built on .net 5; once there is a move to update, it will be done for all projects at the same time (separate from this PR :)) .

PR #1204, opened 2022-06-15
Some QoL improvements for diag tests

### LINKED ISSUE #1202: Build succeeds even when generator fails
**Describe the bug**
The build should fail when the generator encounters an error

**To Reproduce**
```csharp
namespace CSWinRT_AuthoringPlayground
{
    public enum MyEnum
    {
        A=1,
        B
    }

    public struct MyStruct
    {
        public Enum x;
    }
}
```

Build output:
image

**Expected behavior**
Build fails

**Version Info**
CsWinRT 1.6.4

asklar:
- Fixes DiagnosticTests! There were a few non-passing tests so I debugged into it and saw that we're failing because in some codepaths we didn't have an assembly name or version. Passing these requires a context options object.
- DiagnosticTests can now pass a context with options, so we can remove special casing we had for cswinrt generator when running inside a test host
- When tests failed, the inner exception was not being captured, so the error just said "...But  found:" and no additional information about the exception in the generator was provided
- Setting ExitCode upon error, which should cause builds to fail when the generator fails to produce. Fixes #1202 
- Fixes some typos

image

### DIFF
--- a/src/Tests/DiagnosticTests/Helpers.cs
+++ b/src/Tests/DiagnosticTests/Helpers.cs
@@ -34,7 +34,6 @@ namespace DiagnosticTests
                 additionalTexts: ImmutableArray<AdditionalText>.Empty,
                 parseOptions: (CSharpParseOptions)compilation.SyntaxTrees.First().Options,
                 optionsProvider: options); 
-        // todo: pass the CsWinRTComponent config option here so we don't have to comment out the check in the source generator 
 
         /// <summary>
         /// RunGenerators makes a driver and applies the given generators to the compilation, storing diagnostics in an out param
--- a/src/Tests/DiagnosticTests/Helpers.cs
+++ b/src/Tests/DiagnosticTests/Helpers.cs
@@ -4,6 +4,7 @@ using System.Collections.Immutable;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+using Microsoft.CodeAnalysis.Diagnostics;
 
 namespace DiagnosticTests
 {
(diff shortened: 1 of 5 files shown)


src/Tests/DiagnosticTests/Helpers.cs
```diff
@@ -14,25 +15,25 @@ public sealed partial class UnitTesting
...
         /// <returns></returns>
-        private static GeneratorDriver CreateDriver(Compilation compilation, params ISourceGenerator[] generators)
+        private static GeneratorDriver CreateDriver(Compilation compilation, AnalyzerConfigOptionsProvider? options, params ISourceGenerator[] generators)
             => CSharpGeneratorDriver.Create(
                 generators: ImmutableArray.Create(generators),
                 additionalTexts: ImmutableArray<AdditionalText>.Empty,
                 parseOptions: (CSharpParseOptions)compilation.SyntaxTrees.First().Options,
-                optionsProvider: null); 
+                optionsProvider: options); 
         // todo: pass the CsWinRTComponent config option here so we don't have to comment out the check in the source generator
```
manodasanW, line 37: I think you can drop this comment as your change addresses it.

manodasanW (commented):
Thanks @asklar for these changes.

src/Tests/DiagnosticTests/Helpers.cs
j0shuams: +1

j0shuams (approved):
Looks good to me, just needs some feedback on the comments

src/Tests/DiagnosticTests/UnitTesting.cs
```diff
@@ -19,17 +59,13 @@ public sealed partial class UnitTesting
...
             if (WinRTDiagnostics.Any())
             {
-                string foundDiagnostics = "";
-                foreach (var d in WinRTDiagnostics)
-                {
-                    foundDiagnostics += d.Descriptor.Description + "\n";
-                }
-                throw new System.Exception("Expected no diagnostics. But found:" + foundDiagnostics);
+                var foundDiagnostics = string.Join("\n", WinRTDiagnostics.Select(x => x.GetMessage()));
+                throw new SuccessException("Expected no diagnostics. But found:\n" + foundDiagnostics);
```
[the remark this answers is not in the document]
asklar: SuccessException is "Thrown when an assertion failed."; it's the NUnit exception type for when something fails (the name is confusing :) )

j0shuams (approved):

PR #1212, opened 2022-06-29
Custom IWindowNative for ILTrimming

j0shuams:
To support ILTrimming, we need to remove our dependency on ComImport for IWindowNative and IInitializeWithWindow. By replacing them with custom projections, we become trimmable. We should see some perf gains from this change because we do not require IDIC anymore (`.As<IWindowNative>()`), and the ComImport itself probably had some overhead.

### DIFF
--- a/src/cswinrt/strings/ComInteropHelpers.cs
+++ b/src/cswinrt/strings/ComInteropHelpers.cs
@@ -60,12 +60,18 @@ namespace WinRT.Interop
         public static unsafe global::System.IntPtr get_WindowHandle(object _obj)
         {
             var asObjRef = global::WinRT.MarshalInspectable<object>.CreateMarshaler2(_obj, IWindowNativeIID);
-
             var ThisPtr = asObjRef.GetAbi();
-
             global::System.IntPtr __retval = default;
-            global::WinRT.ExceptionHelpers.ThrowExceptionForHR((*(delegate* unmanaged[Stdcall]<IntPtr, out global::System.IntPtr, int>**)ThisPtr)[3](ThisPtr, out __retval));
-            return __retval;
+
+            try
+            {
+                global::WinRT.ExceptionHelpers.ThrowExceptionForHR((*(delegate* unmanaged[Stdcall]<IntPtr, out global::System.IntPtr, int>**)ThisPtr)[3](ThisPtr, out __retval));
+                return __retval;
+            }
+            finally
+            {
+                asObjRef.Dispose();
+            }
         }
     }
 
--- a/src/cswinrt/strings/ComInteropHelpers.cs
+++ b/src/cswinrt/strings/ComInteropHelpers.cs
@@ -53,10 +53,10 @@ using WinRT.Interop;
 
 namespace WinRT.Interop
 {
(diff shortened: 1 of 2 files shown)


src/cswinrt/strings/ComInteropHelpers.cs
```diff
@@ -53,22 +53,35 @@
...
+#else
+    public
 #endif
     static class WindowNative
     {
-        public static IntPtr GetWindowHandle(object target) => target.As<IWindowNative>().WindowHandle;
+        public static IntPtr GetWindowHandle(object target) => IWindowNativeMethods.get_WindowHandle(target);
     }
 
     [ComImport]
```
manodasanW, line 87: You will also need to address this interface as it is a ComImport interface.

src/cswinrt/strings/ComInteropHelpers.cs
```diff
@@ -53,22 +53,35 @@
 
 namespace WinRT.Interop
 {
-    [ComImport]
-    [InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
-    [Guid("EECDBF0E-BAE9-4CB6-A68E-9598E1CB57BB")]
-    internal interface IWindowNative
+ 
+#if EMBED
+    internal
```
manodasanW, line 58: We should just make this always internal.  This doesn't need to be part of the public API surface.

src/cswinrt/strings/ComInteropHelpers.cs
```diff
@@ -53,22 +53,35 @@
...
+    internal 
+#else
+    public
+#endif
+    static class IWindowNativeMethods
     {
-        IntPtr WindowHandle { get; }
+        public static unsafe global::System.IntPtr get_WindowHandle(object _obj)
+        {
+            var asObjRef = global::WinRT.MarshalInspectable<object>.CreateMarshaler2(_obj, System.Guid.Parse("EECDBF0E-BAE9-4CB6-A68E-9598E1CB57BB"));
```
manodasanW, line 66: we should replace this with a hardcoded static GUID like we do in other places rather than a Guid.Parse.

src/cswinrt/strings/ComInteropHelpers.cs
```diff
@@ -53,17 +53,13 @@
...
-#endif
-    static class IWindowNativeMethods
+    internal static readonly Guid IWindowNativeIID = new("EECDBF0E-BAE9-4CB6-A68E-9598E1CB57BB");
+
+    internal static class IWindowNativeMethods
     {
         public static unsafe global::System.IntPtr get_WindowHandle(object _obj)
         {
-            var asObjRef = global::WinRT.MarshalInspectable<object>.CreateMarshaler2(_obj, System.Guid.Parse("EECDBF0E-BAE9-4CB6-A68E-9598E1CB57BB"));
+            var asObjRef = global::WinRT.MarshalInspectable<object>.CreateMarshaler2(_obj, IWindowNativeIID);
```
manodasanW, line 62: You will need to dispose this and probably wrap in a try catch.

src/cswinrt/strings/ComInteropHelpers.cs
j0shuams: addressing separately

manodasanW (approved):

PR #1216, opened 2022-07-08
Remove CCW vtable logic from ABI interfaces which don't need them

manodasanW:
Today we generate the CCW vtable logic for all interfaces.  But we really only need them for non-exclusive interfaces that can be implemented by managed objects, C# authored interfaces, and overridable exclusive interfaces on unsealed types.  For all the other interfaces (exclusive interfaces that aren't overridable), we do not need to generate the CCW vtable logic which can result in DLL size reductions.

### DIFF
--- a/src/cswinrt/code_writers.h
+++ b/src/cswinrt/code_writers.h
@@ -5997,12 +5997,12 @@ public static Guid PIID = Vftbl.PIID;
         XLANG_ASSERT(get_category(type) == category::interface_type);
         auto type_name = write_type_name_temp(w, type, "%", typedef_name_type::ABI);
 
-        // For exclusive interfaces which aren't overriable interfaces that are implemented by unsealed types,
+        // For exclusive interfaces which aren't overridable interfaces that are implemented by unsealed types,
         // we do not need any of the Do_Abi functions or the vtable logic as we will not create CCWs for them.
         // But we are still keeping the interface itself for any helper type lookup that may happen for like GUID lookup.
         if (!is_generic &&
             is_exclusive_to(type) &&
-            // !authored type
+            // check for !authored type
             !(settings.component && settings.filter.includes(type)))
         {
             bool hasOverridableAttribute = false;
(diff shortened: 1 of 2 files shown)


src/cswinrt/code_writers.h
```diff
@@ -5993,11 +5993,51 @@ public static Guid PIID = Vftbl.PIID;
...
         bool is_generic = distance(type.GenericParam()) > 0;
         XLANG_ASSERT(get_category(type) == category::interface_type);
         auto type_name = write_type_name_temp(w, type, "%", typedef_name_type::ABI);
 
+        // For exclusive interfaces which aren't overriable interfaces that are implemented by unsealed types,
+        // we do not need any of the Do_Abi functions or the vtable logic as we will not create CCWs for them.
+        // But we are still keeping the interface itself for any helper type lookup that may happen for like GUID lookup.
+        if (!is_generic &&
+            is_exclusive_to(type) &&
+            // !authored type
```
j0shuams, line 6005: remove?

src/cswinrt/code_writers.h
```diff
@@ -5993,11 +5993,51 @@ public static Guid PIID = Vftbl.PIID;
...
+        {
+            bool hasOverridableAttribute = false;
+            auto exclusive_to_type = get_exclusive_to_type(type);
+            for (auto&& iface : exclusive_to_type.InterfaceImpl())
+            {
+                for_typedef(w, get_type_semantics(iface.Interface()), [&](auto interface_type)
+                {
+                    if (type == interface_type && is_overridable(iface))
+                    {
+                        hasOverridableAttribute = true;
```
j0shuams, line 6016: should we break out of this loop once we find this?

j0shuams (approved):

src/cswinrt/code_writers.h
j0shuams: ahh okay

PR #1217, opened 2022-07-11
Make CsWinRT build with VS 2022

manodasanW:
- As part of bringing up IL trimming support, there is a csproj property needed to be set which is available in the .NET 6 SDK.  So this PR as a first step enables us to build with VS 2022 and the .NET 6 SDK.  Note that this PR does not change anything about what we target (i.e. we still currently target .NET 5 using the .NET 6 SDK).
- Fixed some issues that arose from moving to .NET 6 SDK and VS 2022.
- Moved off of .NET core app 2.0 for our .NET standard testing as .NET Core 2.0 is EOL and using .NET core 3.1 instead.

### DIFF
--- a/build/AzurePipelineTemplates/CsWinRT-Build-Steps.yml
+++ b/build/AzurePipelineTemplates/CsWinRT-Build-Steps.yml
@@ -24,11 +24,11 @@ steps:
 
       Write-Host ##vso[task.setvariable variable=PATH;]${env:Agent_TempDirectory}\procdump;${env:PATH};
 
-# Use .NET Core SDK 2.1
+# Use .NET Core SDK 3.1
 - task: UseDotNet@2
-  displayName: Use .NET Core SDK 2.1
+  displayName: Use .NET Core SDK 3.1
   inputs:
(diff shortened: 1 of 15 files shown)


build/AzurePipelineTemplates/CsWinRT-Build-Steps.yml
```diff
@@ -24,27 +24,14 @@ steps:
...
-  displayName: Use .NET Core SDK 2.1
+  displayName: Use .NET Core SDK 3.1
   inputs:
-    version: 2.1.x
+    version: 3.1.x
     installationPath: C:\Users\VssAdministrator\AppData\Local\Microsoft\dotnet\
     performMultiLevelLookup: true
 
-# Install .NET 5 SDK
-- task: PowerShell@2
```
j0shuams, line 36: Why don't we need this anymore?

manodasanW: The .NET 6 SDK can be used to build .NET 5 TFMs.

Issue #373, opened 2020-08-25
Support IL Trimming for projection assemblies


Scottj1s:
# Proposal: Support for trimmable (ILLink) assemblies

## Summary
[Trimming](https://github.com/dotnet/linker/tree/main/docs) is a great feature introduced with .NET 5, but CsWinRT does not yet support it. This should be fixed, so assembly sizes can be reduced.

## Rationale
Trimming with ILLink provides great results when publishing self contained applications. I have reduced the size of the Microsoft.Windows.SDK.NET assembly by a factor of `45`, taking it from 25mb to 0.55mb! I see the same great result in other projected libraries as well.

Since WinUI is using CoreCLR, trimming will be an important tool to keep assemblies small. Same goes for Xamarin apps using the desktop SDK.

## Important Notes
Issues encountered so far:
- Internal constructors, used by internal factories, are incorrectly trimmed.
- Missing interfaces and types in the ABI namespace. Perhaps an explicit interface implementation could fix this.

As you can expect, the reflection-based factories in the ABI namespace quickly break, however, these types can manually be preserved by whitelisting them using a `TrimmerRootDescriptor`.

ILLink can be enabled by specifying the following properties:
```xml
<PublishTrimmed>true</PublishTrimmed>
<TrimMode>link</TrimMode>
```

To whitelist the Windows SDK for trimming:
```xml
<ItemGroup>
    <ManagedAssemblyToLink Condition="'%(Filename)' == 'Microsoft.Windows.SDK.NET'">
        <IsTrimmable>true</IsTrimmable>
    </ManagedAssemblyToLink>
</ItemGroup>
```

Further reading: https://github.com/mono/linker/blob/main/docs/data-formats.md

angelazhangmsft:
C#/Win32 ask for C#/WinRT (.NET 6 timeframe) - this could remove baggage from the TFM.

mikebattista:
Thanks Angela. As we integrate more support in the TFMs for things like C#/Win32, we don't want apps to then have to carry around baggage for things they may not even be using. Pay for play is definitely the right model here, so leveraging assembly trimming for that would benefit C#/WinRT customers as well as C#/Win32 customers. It would be a great story to say that setting your TFM gets you access to all Windows APIs without any baggage. You only pay for what you use.

jkoritzinsky:
After speaking with the IL linker team, it looks like there's a few things we'll need:

[...]

PR #1224, opened 2022-07-13
First iteration of making CsWinRT compatible with IL trimming

manodasanW:
This PR addresses a good portion of the IL trimming warnings and makes CsWinRT IL trimming compatible.  Some of the major areas addressed are the following:

- Helper type lookup
CsWinRT today uses naming conventions and Type.GetType to find the helper type associated with a given type.  This change adds a new attribute, `WindowsRuntimeHelperType`, which is projected onto all types other than enums.  This attribute has a property which provides access to the helper type.  Given a type, CsWinRT looks for this attribute and uses it to get the helper type without needing to worry about it getting trimmed or using GetType.  The GetType calls are still kept, but warnings are suppressed as we do need a fallback for older projections that haven't been updated yet.
- RCW construction
During FromABI scenarios, CsWinRT creates an RCW instance by using reflection to call an internal constructor on the RCW type being created.  The code has been attributed to specify we dynamically access the non-public constructor.
- Composable types
There can be scenarios where the app receives an object for which it only interacts with it using one of the base classes.  This means the actual type of the object (derived type) can be trimmed.  Added support to handle this by making projections register a dictionary of all their Child -> Base class mappings using `RegisterProjectionTypeBaseTypeMapping` which is consulted in the case that a type to construct cannot be found.  If it is found the child class doesn't exist, but the base class does, then it is used to construct the RCW.
- Marshaling
Helper types have been attributed to specify everything we access from them such as during marshaling.
- Xaml Type lookup
CsWinRT has a custom Type projection which is used by XAML to look up types.  XAML can end up looking up types that have been trimmed as there are no references to it in C# code but is still used in native code.  For these scenarios, we return a fake type instance to represent that type as this is what .NET native used to do for these scenarios.

As part of validating these changes, added a set of functional tests which exercise trimming and validate core functionality.  Working on collecting stats for size improvements but the functional tests do demonstrate types are getting trimmed and the DLLs are much smaller in the basic scenarios.

There is still a bunch of MakeGenericType calls that need to be suppressed but that will be addressed in a future PR as I do not want to suppress the entire function they are used as part of for trim warnings.

#373


jkoritzinsky (approved):
It might be worth creating some internal constants in WinRT.Runtime for some common bitmasks of the DynamicallyAccessedMemberTypes enum to reduce duplication.

Other than that, LGTM!

src/WinRT.Runtime/Marshalers.cs
```diff
@@ -1181,9 +1194,23 @@ public static unsafe void CopyManaged(T value, IntPtr dest)
 #else 
     public
 #endif
-    static class MarshalInspectable<T>
+    static class MarshalInspectable<
+#if NET6_0_OR_GREATER || NET6_0
```
huoyaoyuan, line 1198: Why is `NET6_0` used together with `_OR_GREATER`? It should be included.

j0shuams (approved):

src/WinRT.Runtime/Projections/Type.cs
```diff
@@ -157,5 +182,202 @@ public static string GetGuidSignature()
...
+
+        public override string Namespace => throw new NotImplementedException();
+
+        public override global::System.Type UnderlyingSystemType => throw new NotImplementedException();
+
+        public override string Name => throw new NotImplementedException();
+
+        public override ConstructorInfo[] GetConstructors(BindingFlags bindingAttr)
+        {
+            throw new NotImplementedException();
```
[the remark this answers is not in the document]
manodasanW: Addressed.

src/WinRT.Runtime/Marshalers.cs
manodasanW: Seems to be working fine with the current .NET SDK version.  Addressed.

PR #1228, opened 2022-07-28
Enable marshaling of restricted error info to System.Exception for UnhandledException handler

Scottj1s:
I don't know why we were not using the restricted error info when marshaling System.Exception, but this is necessary for UnhandledException support.


jlaanstra (approved):

manodasanW (approved):

Scottj1s:
#fixes 430

PR #1240, opened 2022-08-13
Make QueryInterface function pointer blittable

manodasanW:
Per https://github.com/microsoft/CsWinRT/discussions/1232#discussioncomment-3370693, the use of ref and out in the function pointers makes it not blittable and can have an impact on perf.  This change addresses that by making them pointers instead.

This change can be considered breaking given it is in the public API surface, but this change doesn't break any existing projections or users of those projections as the use of QueryInterface is limited to callers in WinRT.Runtime and none of our generated projection code directly references it.

There are other function pointers that maybe should be addressed too (i..e within the generated projections) but given they don't seem to be in the public API surface, deferring that to be addressed later.

### DIFF
--- a/src/WinRT.Runtime/ComWrappersSupport.net5.cs
+++ b/src/WinRT.Runtime/ComWrappersSupport.net5.cs
@@ -446,7 +446,7 @@ namespace WinRT
             IUnknownVftblPtr = RuntimeHelpers.AllocateTypeAssociatedMemory(typeof(IUnknownVftbl), sizeof(IUnknownVftbl));
             (*(IUnknownVftbl*)IUnknownVftblPtr) = new IUnknownVftbl
             {
-                QueryInterface = (delegate* unmanaged[Stdcall]<IntPtr, ref Guid, out IntPtr, int>)qi,
+                QueryInterface = (delegate* unmanaged[Stdcall]<IntPtr, Guid*, IntPtr*, int>)qi,
                 AddRef = (delegate* unmanaged[Stdcall]<IntPtr, uint>)addRef,
                 Release = (delegate* unmanaged[Stdcall]<IntPtr, uint>)release,
             };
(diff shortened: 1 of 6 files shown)


AaronRobinsonMSFT (approved):

src/WinRT.Runtime/ComWrappersSupport.net5.cs
```diff
@@ -446,7 +446,7 @@ static unsafe DefaultComWrappers()
             IUnknownVftblPtr = RuntimeHelpers.AllocateTypeAssociatedMemory(typeof(IUnknownVftbl), sizeof(IUnknownVftbl));
             (*(IUnknownVftbl*)IUnknownVftblPtr) = new IUnknownVftbl
             {
-                QueryInterface = (delegate* unmanaged[Stdcall]<IntPtr, ref Guid, out IntPtr, int>)qi,
+                QueryInterface = (delegate* unmanaged[Stdcall]<IntPtr, Guid*, IntPtr*, int>)qi,
```
Sergio0694, line 449: Mostly just curious, is ther a particular reason for using `IntPtr` in these function pointers? It would be "better" to eg. make this just be `<void*, Guid*, void**, int>`, etc. As in, the general recommendation especially from .NET 7 onwards is to stop using `IntPtr` as "a pointer", given it's technically a numeric type 🙂

Sergio0694 (approved):
Just one nit, otherwise this looks awesome!!
Thank you for listening to the suggestion and working on this 😄

src/WinRT.Runtime/ComWrappersSupport.net5.cs
AaronRobinsonMSFT: Can confirm that is the recommendation, `IntPtr` -> `void*`. See the following comment https://github.com/dotnet/runtime/pull/70297#discussion_r890495003.

manodasanW: Thanks for calling this out.  If I were to change this, I would also want to change AddRef and Release for consistency, but those look to be used within the generated projections and thereby would break existing projections if I were to change it.  Due to that, I am deferring on this for now.

PR #1251, opened 2022-08-30
Update localization pipeline

j0shuams:
Update OS used in localization pipeline

### DIFF
--- a/build/AzurePipelineTemplates/CsWinRT-LocalizeResources-Stage.yml
+++ b/build/AzurePipelineTemplates/CsWinRT-LocalizeResources-Stage.yml
@@ -4,7 +4,7 @@ stages:
   jobs:
   - job: Localize
     pool: 
-      vmImage: windows-2019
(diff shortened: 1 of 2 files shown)


build/AzurePipelineTemplates/CsWinRT-LocalizeResources-Stage.yml
```diff
@@ -4,7 +4,7 @@ stages:
   jobs:
   - job: Localize
     pool: 
-      vmImage: windows-2019
+      vmImage: windows-2022
```
manodasanW, line 7: There is also [CsWinRT-LocalizationPipeline.yml](https://github.com/microsoft/CsWinRT/blob/master/build/AzurePipelineTemplates/CsWinRT-LocalizationPipeline.yml) in which it uses CsWinRT-BuildOnly-Stage.yml which should be updated too.

manodasanW (approved):

PR #1261, opened 2022-10-25
Clean up .NET 5 references and update to WinAppSDK projection for testing

### LINKED ISSUE #778: Update repo to build and test against Windows App SDK
- new WinUI projection
- update WinUIDesktopSample to use new WinUI projection
- update objectlifetime test

manodasanW:
- Build pipelines recently started failing due to the x86 version of .NET 5 doesn't seem to be installed by default anymore.  I took that opportunity to clean up the .NET 5 references we had rather than doing a manual install.  This should also help improve our build times given we are building one less TFM.
- One of the things blocking the cleanup of the .NET 5 TFM was we still used the old WinUI packages for our testing and they ran into build issues with .NET 6.  This PR moves us to using a recent version of WinAppSDK for testing which does build without issues and also addresses the breaking changes from the old WinUI packages to WinAppSDK.  We still build a projection for the old version to help with catching any regressions.
- Fixing issue with build.cmd after one of the previous PRs

Fixes #778


j0shuams (approved):
LGTM

PR #1263, opened 2022-10-25
cached marshaled exception

### LINKED ISSUE #1258: App_UnhandledException's UnhandledExceptionEventArgs.Exception can only be fetched once - should be cached
**Describe the bug**
The App.UnhandledException UnhandledExceptionEventArgs.Exception is volatile and can only be retrieved once, whether by user code or debug visualization. Subsequent accesses result in empty properties.  The Exception property should be cached so it can be accessed multiple times.

**To Reproduce**
1. Clone [WinUI3_UnhandledExceptionTestAppd](https://github.com/k563/WinUI3_UnhandledExceptionTestApp)
2. Build, set a BP in the App.UnhandledException handler, and debug
3. Click the button to cause an unhandled exception to be raised
4. When the BP hits, view the UnhandledExceptionEventArgs.Exception and its properties
5. Step through code to fetch the StackTrace property

**Expected behavior**
The StackTrace property should not be empty.  Instead, it's empty, the stowed exception having been fetched once and then removed from the thread.  This is true of all properties - only Hresult and Message are preserved (for some reason).

**Workarounds**
There are a few workarounds:
1. Fetch the exception and store it in a local, so properties are not lost:
```
var ex = e.Exception;
var ts = ex.TargetSite;
var st = ex.StackTrace;
var ie = ex.InnerException;
var m = ex.Message;
var hr = ex.HResult;
var s = ex.Source;
var message = String.Format(
@"Message: {0}
HResult: 0x{1:X8}
Source: {2}
TargetSite: {3}
Stack Trace: 
{4}", m, hr, s, ts, st);

```
2. Inspect **$exception** (if only viewing properties).

### LINKED ISSUE #1259: Do_Abi_* event/callback handlers should wrap invocations with RoReportUnhandledError
**Describe the bug**
The legacy C# WinRT projection support for UWP apps wrapped all synchronous calls back into C# code with a try/catch, calling RoOriginateLanguageException and then RoReportUnhandledError on any unhandled exceptions, ensuring they were then routed to the App.UnhandledException handler.  This behavior is missing from C#/WinRT projections, causing regressions in Microsoft.Windows.SDK.NET.dll (e.g., Window.Activated) and Microsoft.WinUI.dll (most everything else).  Thus, fixes might be required across several repos.

**To Reproduce**
For an override callback:
1. Clone [WinUI3_UnhandledExceptionTestAppd](https://github.com/k563/WinUI3_UnhandledExceptionTestApp)
2. Build, set a BP in OnLaunched and in the App.UnhandledException handler, and debug
3. When the OnLaunched BP hits, set next statement to throw the exception

For an event:
1. Clone [WinUI3_UnhandledExceptionTestAppd](https://github.com/k563/WinUI3_UnhandledExceptionTestApp)
2. Build, set a BP in the MainWindow's Activated handler and in the App.UnhandledException handler, and debug
3. When the Activated handler BP hits, set next statement to throw the exception

**Expected behavior**
In either case above, callback or event, the App.UnhandledException handler should be invoked as it is in a UWP app.

**Workarounds**
There are a couple workarounds:
1. User callbacks can be marked 'async' (whether or not they contain any awaits).  This causes the CLR to wrap invocations in a try/catch, propagating any caught exceptions through the DispatcherQueueSynchronizationContext.
`this.Activated += **async** (s, e) => { ... };`
2. User callback code can itself be wrapped in a try/catch to propagate errors:
```
this.Activated += (s, e) => {
  try
  {
    // ...
  }
  catch (System.Exception e)
  {
    ExceptionHelpers.ReportUnhandledError(e);
  }
}

```

Scottj1s:
fixes #1258


Scottj1s:
See also:
https://github.com/microsoft/microsoft-ui-xaml/issues/5221

jlaanstra (approved):

src/cswinrt/code_writers.h
```diff
@@ -5257,7 +5257,7 @@ try
 }
 catch (Exception __exception__)
 {
-global::WinRT.ExceptionHelpers.SetErrorInfo(__exception__);
```
manodasanW, line 5260: Is the changes in this file basically calling ReportUnhandledError in all our Do_ABI functions?  If that is right, that would mean we are saying that any error crossing the ABI boundary is an unhandled error.  Is that what we want or did we want it such that any errors throw in handlers are reported as unhandled errors.

Scottj1s: Yes, this is reporting all unhandled errors (uncaught managed exceptions) at the ABI and passing them along to RoReportUnhandledError, which is what the WRL wrappers do, and what the UWP behavior is.

Scottj1s: With UWP, Windows.UI.Xaml.dll relied on twinapi.appcore.dll's CoreApplication implementation, which routed all app events thru WRL's event implementation.  WRL's InvokeDelegates in turn would catch non-zero HRESULTs and invoke RoReportFailedDelegate.  

But with WinAppSDK, Microsoft.UI.Xaml.dll invokes marshaled event handlers directly and just propagates non-zero HRESULTs up the stack to Application::Start (where the debugger then presents an unhelpful 0xc000027b message) or simply ignores them (e.g., CFTMEventSource::Raise). 

We could change Microsoft.UI.Xaml.dll to try and catch all callback/event invocation non-zero HRESULTs and pass them to RoReportUnhandledError (or RoReportFailedDelegate), but we'd still want to call SetErrorInfo to get the language exception (stack, message, etc).  That being the case, it's easiest to code-gen that call alongside SetErrorInfo.

Scottj1s: @manodasanW I've removed all error reporting changes, and implemented them in Microsoft.UI.Xaml.dll instead, where they apply to both C# and C++ apps.

src/WinRT.Runtime/ExceptionHelpers.cs
```diff
@@ -99,15 +118,12 @@ private static Exception GetExceptionForHR(int hr, bool useGlobalErrorState, out
...
 
-            if (useGlobalErrorState && getRestrictedErrorInfo != null)
+            if (useGlobalErrorState)
             {
-                Marshal.ThrowExceptionForHR(getRestrictedErrorInfo(out IntPtr restrictedErrorInfoPtr));
-
-                if (restrictedErrorInfoPtr != IntPtr.Zero)
-                {
-                    IObjectReference restrictedErrorInfoRef = ObjectReference<ABI.WinRT.Interop.IRestrictedErrorInfo.Vftbl>.Attach(ref restrictedErrorInfoPtr);
+                var restrictedErrorInfoRef = BorrowRestrictedErrorInfo();
```
manodasanW, line 123: consider **_using_** var restrictedErrorInfoRef = .. here and below to make it do the cleanup here rather than during GC thread.

manodasanW (approved):

PR #1272, opened 2022-11-15
Address couple trim warnings

### LINKED ISSUE #1269: Spurious trimmer warning is not suppressed
**Describe the bug**

@Sergio0694 and I found that the following trim warnings are emitted by a default WinUI 3 project (built with "trimmer-safe" CsWinRT projections):

```
1>ILLink : Trim analysis warning IL2026: WinRT.Projections..cctor(): Using member 'System.MulticastDelegate.MulticastDelegate(Object, String)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. The target method might be removed.
1>ILLink : Trim analysis warning IL2026: WinRT.Projections..cctor(): Using member 'System.Delegate.Delegate(Object, String)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. The target method might be removed.
1>ILLink : Trim analysis warning IL2026: WinRT.Projections..cctor(): Using member 'System.Delegate.CreateDelegate(Type, Object, String)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. The target method might be removed.
1>ILLink : Trim analysis warning IL2026: WinRT.Projections..cctor(): Using member 'System.Delegate.CreateDelegate(Type, Object, String, Boolean, Boolean)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. The target method might be removed.
1>ILLink : Trim analysis warning IL2026: WinRT.Projections..cctor(): Using member 'System.Delegate.CreateDelegate(Type, Object, String, Boolean)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. The target method might be removed.
```

These warnings are emitted because `Projections.RegisterCustomAbiTypeMappingNoLock` has `DynamicallyAccessedMembers(PublicMethods)` on the type parameter, and a delegate type `typeof(EventHandler<>)` is passed in. As a result, the trimmer warns about these methods with `RequiresUnreferencedCodeAttribute` since it can't be sure you don't access these methods through reflection. These warnings aren't emitted by the trimmer analyzer due to https://github.com/dotnet/linker/issues/3111.

These warnings should be suppressed because these methods are never invoked via reflection based on the usage.

**To Reproduce**

1. Create a default WinUI 3 project
2. Set `<PublishTrimmed>true</PublishTrimmed>`
3. Publish the project.

**Expected behavior**

No trimmer warnings from WinRT.Runtime.dll

**Version Info**

VS2022.4

**Additional context**

@Sergio0694 has a repro.

manodasanW:
- The delegate in IDisposable seems to cause trim warnings when used in RegisterCustomAbiTypeMappingNoLock.  It doesn't seem to be used, so removed it and confirmed the warnings went away.
- Improved trim warning for C#/WinRT authoring support.

Fix #1269

### DIFF
--- a/src/WinRT.Runtime/Projections/IDisposable.net5.cs
+++ b/src/WinRT.Runtime/Projections/IDisposable.net5.cs
@@ -21,7 +21,6 @@ namespace ABI.System
         [Guid("30D5A829-7FA4-4026-83BB-D75BAE4EA99E")]
         public struct Vftbl
         {
-            private delegate int CloseDelegate(IntPtr thisPtr);
             internal IInspectable.Vftbl IInspectableVftbl;
             private void* _Close_0;
             public delegate* unmanaged[Stdcall]<IntPtr, int> Close_0 { get => (delegate* unmanaged[Stdcall]<IntPtr, int>)_Close_0; set => _Close_0 = value; }
@@ -35,9 +34,7 @@ namespace ABI.System
                 AbiToProjectionVftable = new Vftbl
                 {
                     IInspectableVftbl = global::WinRT.IInspectable.Vftbl.AbiToProjectionVftable,
-
                     _Close_0 = (delegate* unmanaged<IntPtr, int>)&Do_Abi_Close_0
-
                 };
                 var nativeVftbl = (IntPtr*)ComWrappersSupport.AllocateVtableMemory(typeof(Vftbl), Marshal.SizeOf<global::WinRT.IInspectable.Vftbl>() + sizeof(IntPtr) * 1);
                 Marshal.StructureToPtr(AbiToProjectionVftable, (IntPtr)nativeVftbl, false);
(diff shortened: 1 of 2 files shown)


j0shuams (approved):

jkoritzinsky:
Are you sure this addresses the trim warning I mentioned? I could have sworn the issue was the `typeof(EventHandler<>)` that was causing the issue.

manodasanW:
> Are you sure this addresses the trim warning I mentioned? I could have sworn the issue was the `typeof(EventHandler<>)` that was causing the issue.

I confirmed with a repro I got and then updated it to `2.0.1-prerelease.221115.11` and no longer saw those warnings.  We can use the same build to confirm with @Sergio0694 's repro.  If you look at it, you will note that `typeof(EventHandler<>)` is passed for `publicType` which has no `DynamicallyAccessedMembers` markings.  It is the `abiType` which does which is getting our ABI version of that type.  The only warning left I see is for `GetAuthoringMetadataType` which I put a better warning for now.

```
        private static void RegisterCustomAbiTypeMappingNoLock(
            Type publicType,
#if NET
            [DynamicallyAccessedMembers(
                DynamicallyAccessedMemberTypes.PublicMethods |
                DynamicallyAccessedMemberTypes.NonPublicMethods |
                DynamicallyAccessedMemberTypes.PublicNestedTypes |
                DynamicallyAccessedMemberTypes.PublicFields)]
#endif
            Type abiType, 
            string winrtTypeName,
            bool isRuntimeClass = false)
```

PR #1299, opened 2023-03-14
Make the lookup cache thread safe for multiple lookups and extend the same caching to IReadOnlyDictionary.

### LINKED ISSUE #1295: Multiple concurrent dictionary lookups via a dictionary wrapped in a ReadOnlyDictionary are causing the concurrent exception.
**Describe the bug**
Multiple concurrent dictionary lookup can result in the internal CSWinRT lookup caching to simultaneously insert and lookup from the same cache causing a concurrent modification exception being thrown.

**To Reproduce**
See task.ms/43496689

**Expected behavior**
No crash.

**Version Info**
CsWinRT 2.0.1

**Additional context**
See task.ms/43496689

jlaanstra:
Fixes #1295

### DIFF
--- a/src/WinRT.Runtime/Projections/IDictionary.net5.cs
+++ b/src/WinRT.Runtime/Projections/IDictionary.net5.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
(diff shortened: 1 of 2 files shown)


jlaanstra:
Note that this caching is internal and has the potential to leak all keys/values in any RCW dictionary. We might need to do cleanup.

jlaanstra:
Superseded by #1353

PR #1300, opened 2023-03-14
Add vtbls for IWeakReferenceSource and IWeakReference.

jlaanstra:
Fixes ADO#41937903

When resolving an IWeakReference from a different context, we resolve as IUnknown and then assume that Resolve is the next method in the vtbl, see https://github.com/microsoft/CsWinRT/blob/418229436353d05edccd4f94fd3f6539735d810a/src/WinRT.Runtime/Interop/IWeakReferenceSource.net5.cs#L256. This turns out to not be true and we end up calling an unrelated COM interface with incorrect parameters that than tries to allocate a very large array on the stack resulting in stack overflow.


jlaanstra:
Superseded by #1301. Thanks @manodasanW

PR #1310, opened 2023-03-31
Scoped change to have enums implement IPropertyValue to be consistent with coreclr and cppwinrt.

jlaanstra:
Need to make sure this doesn't regress #605


jlaanstra:
This indeed breaks #605 again because TypeNameSUpport.GetNameForType returns null for enums.

PR #1313, opened 2023-04-04
Remove IReference cache and add various tests for IPropertyValue and GetRuntimeClassName.

jlaanstra:
Generation of boxed types is now consistent between Vtable generation and GetRuntimeClassName.

### DIFF
--- a/src/Tests/AuthoringTest/Program.cs
+++ b/src/Tests/AuthoringTest/Program.cs
@@ -459,7 +459,6 @@ namespace AuthoringTest
                 new BasicDelegate[] { new BasicDelegate((uint value) => {}) },
                 new [] { new DisposableClass().GetType() , new NonProjectedDisposableClass().GetType() },
                 new Type[] { typeof(TestClass), typeof(DisposableClass) },
-                new PrivateEnum[] { PrivateEnum.PrivateFirst, PrivateEnum.PrivateSecond}
             };
         }
     }
--- a/src/WinRT.Runtime/TypeNameSupport.cs
+++ b/src/WinRT.Runtime/TypeNameSupport.cs
@@ -25,7 +25,7 @@ namespace WinRT
         /// <summary>
         /// Don't output a type name of a custom .NET type. Generate a compatible WinRT type name if needed.
         /// </summary>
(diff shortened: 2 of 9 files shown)


src/WinRT.Runtime/TypeNameSupport.cs
```diff
@@ -425,22 +419,41 @@ private static bool TryAppendTypeName(Type type, StringBuilder builder, TypeName
...
+                    return false;
                 }
-                return true;
+                else
+                {
+                    return false;
+                }
+            }
+
+            if ((flags & TypeNameGenerationFlags.GenerateBoxedName) != 0 && type.ShouldProvideIReference())
```
manodasanW, line 440: For a non-projected enum, if type.ShouldProvideIReference() returns false, then does it still go to TryAppendSimpleTypeName below and have its type name appended?

jlaanstra: Yes in the next if statement.

src/Tests/AuthoringTest/Program.cs
```diff
@@ -459,7 +459,6 @@ public IList<object> GetTypeErasedProjectedArrays()
                 new BasicDelegate[] { new BasicDelegate((uint value) => {}) },
                 new [] { new DisposableClass().GetType() , new NonProjectedDisposableClass().GetType() },
                 new Type[] { typeof(TestClass), typeof(DisposableClass) },
-                new PrivateEnum[] { PrivateEnum.PrivateFirst, PrivateEnum.PrivateSecond}
```
jlaanstra, line 462: This enum is internal and not projected.

src/WinRT.Runtime/TypeNameSupport.cs
```diff
@@ -24,7 +25,7 @@ internal enum TypeNameGenerationFlags
         /// <summary>
         /// Don't output a type name of a custom .NET type. Generate a compatible WinRT type name if needed.
         /// </summary>
-        NoCustomTypeName = 0x2
+        ForGetRuntimeClassName = 0x1,
```
manodasanW, line 28: 0x2?

manodasanW (approved):

PR #1330, opened 2023-06-07
Fix issue with ABI delegates in collections for .NET Standard projection

manodasanW:
Our .NET Standard embedded projection can be used with both .NET core and .NET Framework.  But it seems our generic collection types runs into issues with the latter due to using `Expression.GetDelegateType` to generate custom delegates with `void*` parameters doesn't seem to be supported there.  To address that, in our .NET Standard projection, we are now code generating all the ABI delegates that we can detect as being needed while generating the projection.  This is then used by the collection types for when they need the delegate via a lookup mechanism.  There is a fallback to `Expression.GetDelegateType` which handles the scenario where we couldn't detect a certain ABI delegate is needed and didn't generate it as the WinRT boundary takes the value as an `object`.  .NET Core users would be handled via that fallback, but .NET Framework users would see an error.  In those cases, there is a way for the app to declare their own ABI delegates and register them if needed.

### DIFF
--- a/src/Samples/TestEmbedded/UnitTestEmbedded/TestClass.cs
+++ b/src/Samples/TestEmbedded/UnitTestEmbedded/TestClass.cs
@@ -38,5 +38,11 @@ namespace UnitTestEmbedded
         {
             Assert.Equal(20, TestLib.Test5());
         }
+
+        [Fact]
+        public void Test6()
+        {
+            Assert.Equal(5, TestLib.Test6());
+        }
     }
 }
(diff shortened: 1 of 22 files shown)


src/Samples/TestEmbedded/UnitTestEmbedded/TestClass.cs
```diff
@@ -37,6 +37,12 @@ public void Test4()
         public void Test5()
         {
             Assert.Equal(20, TestLib.Test5());
+        }
+
+        [Fact]
+        public void Test6()
```
jamespik, line 43: Nit: We should improve the naming of these tests in the future

jamespik (approved):

PR #1333, opened 2023-06-14
Fix invalid assembly reference in type lookup

hez2010:
```csharp
CecilExtensions.FindTypeReference(module, "System", "MemoryExtensions", "System.Memory", false);
```
will return a `null` and causes the subsequent emit process failed.
But the exception was caught and ignored.

Instead it should look up the type in `System.Private.CoreLib`.

### DIFF
--- a/src/Perf/IIDOptimizer/SignatureEmitter.cs
+++ b/src/Perf/IIDOptimizer/SignatureEmitter.cs
@@ -453,7 +453,7 @@ namespace GuidPatch
             il.Emit(OpCodes.Pop);
 
             // Fix endianness, bytes
-            var memoryExtensions = CecilExtensions.FindTypeReference(module, "System", "MemoryExtensions", "System.Memory", false);
+            var memoryExtensions = CecilExtensions.FindTypeReference(module, "System", "MemoryExtensions", "System.Private.CoreLib", false);
 
             var reverseMethod_Generic = new MethodReference("Reverse", module.TypeSystem.Void, memoryExtensions) { };
 


manodasanW:
Thanks @hez2010 for this change and catching this issue.  CI pipelines all passed.  I want to take a closer look at which types were impacted by this on Monday and will then merge it.

manodasanW:
@hez2010 I took a look at this, and I am curious in what scenario are you seeing this issue and it returning null (i.e TFM and so on)?  At least with debugging IIDOptimizer running on the Windows SDK projection in this repo, I am seeing both System.Memory and System.Private.CoreLib work and not return null.  And looking at the Windows SDK Projection on NuGet in IlSpy, I see the generated code with the optimizations using reverse and so on for the generic types, but I haven't yet confirmed whether it runs or not.

hez2010:
You will see `<SignatureCache>` being duplicated for every single type because of the exception (happened [here](https://github.com/microsoft/CsWinRT/blob/9374d158e102d120b55274dd06c83301d446f15f/src/Perf/IIDOptimizer/GuidPatcher.cs#L188C25-L188C25)) being caught and leading to the insertion to `methodCache` ([here](https://github.com/microsoft/CsWinRT/blob/9374d158e102d120b55274dd06c83301d446f15f/src/Perf/IIDOptimizer/GuidPatcher.cs#L201)) being skipped.

hez2010:
Never mind, this is caused by the mistake that I didn't pass reference assemblies to IIDOptimizer in the command line.

Closing.

PR #1335, opened 2023-06-19
Create COMException from IRestrictedErrorInfo instead of calling Marshal.GetExceptionForHR.

### LINKED ISSUE #1334: ExceptionHelpers.GetExceptionForHR creates exception with generic error message dropping more useful info.
**Describe the bug**
CsWinRT build the Exception from an hr using Marshal.GetExceptionForHR and passes in the IErrorInfo, but .NET doesn't query for IRestrictedErrorInfo as it considers this a WinRT concept. The result is that the Exception is less useful than it could have been.

**To Reproduce**
1. Throw an exception from a native winrt component with a useful error message.
2. Check the error message of the exception in C# code.

**Expected behavior**
You see the useful error message.

**Version Info**
2.0.3 and earlier.

**Additional context**
Source for GetExceptionForHR with IErrorInfo: https://github.com/dotnet/runtime/blob/07def40a1d8a900aa6e88c548742e547a24c0e8f/src/coreclr/vm/excep.cpp#L2980-L3037

jlaanstra:
Fixes #1334 

This change manually creates the COMException from the IRestrictedErrorInfo. It also updates the message for all exceptions to include the more useful restrictedError as .NET used to do when it had WinRT support.

This change currently doesn't handle any .NET specific HResults which Marshal.GetExceptionForHR does. We could potentially check for IErrorInfo if there is no IRestrictedErrorInfo and still call Marshal.GetExceptionForHR in that case as it's unlikely to be a WinRT call at that point. Thoughts?

### DIFF
--- a/src/WinRT.Runtime/ExceptionHelpers.cs
+++ b/src/WinRT.Runtime/ExceptionHelpers.cs
@@ -30,6 +30,8 @@ namespace WinRT
         internal const int E_ELEMENTNOTENABLED = unchecked((int)0x802B001E);
         internal const int E_ELEMENTNOTAVAILABLE = unchecked((int)0x802B001F);
         internal const int ERROR_INVALID_WINDOW_HANDLE = unchecked((int)0x80070578);
+        private const int E_POINTER = unchecked((int)0x80004003);
+        private const int E_NOTIMPL = unchecked((int)0x80004001);
 
         [DllImport("oleaut32.dll")]
         private static extern int SetErrorInfo(uint dwReserved, IntPtr perrinfo);
@@ -110,7 +112,6 @@ namespace WinRT
         {
             restoredExceptionFromGlobalState = false;
 
-            ObjectReference<ABI.WinRT.Interop.IErrorInfo.Vftbl> iErrorInfo = null;
             IObjectReference restrictedErrorInfoToSave = null;
             Exception ex;
             string description = null;
@@ -118,6 +119,7 @@ namespace WinRT
             string restrictedErrorReference = null;
             string restrictedCapabilitySid = null;
             bool hasOtherLanguageException = false;
+            string errorMessage = string.Empty;
 
             if (useGlobalErrorState)
             {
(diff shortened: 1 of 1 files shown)


manodasanW:
For the most part, this seems fine to me.  @AaronRobinsonMSFT feel free to comment if you see anything `Marshal.GetExceptionForHr` handled before that this change doesn't handle.

My only concern is lets say a DirectoryNotFoundException is thrown in C#.  This I assume gets translated to STG_E_PATHNOTFOUND based on [this](https://github.com/microsoft/CsWinRT/blob/04a590360338092bed7f80355e51a9af9950b022/src/WinRT.Runtime/ExceptionHelpers.cs#LL261C1-L262C1) and [this](https://learn.microsoft.com/en-us/dotnet/api/system.io.directorynotfoundexception?view=net-7.0#remarks).  If this exception was to cross the ABI boundary and then end up in C# code again in the same process that probably would be fine given we will retrieve the managed exception from the error info and rethrow it.  But if this was from a different process, we won't do that today and then when we do the error code translation, I think we are going to end up with a ComException.  I assume before we would have ended up with a DirectoryNotFoundException (needs to be confirmed).  Some of the other .NET errors are probably less common, but I feel like there might be some that are common like this one.

All other calls to [Marshal.GetExceptionForHr](https://github.com/search?q=repo%3Amicrosoft%2FCsWinRT+Marshal.throwexceptionforhr&type=code) probably needs to be updated to go through ExceptionHelpers.GetExceptionForHr, but this PR doesn't need to necessarily address that.

jlaanstra:
> For the most part, this seems fine to me. @AaronRobinsonMSFT feel free to comment if you see anything `Marshal.GetExceptionForHr` handled before that this change doesn't handle.
> 
> My only concern is lets say a DirectoryNotFoundException is thrown in C#. This I assume gets translated to STG_E_PATHNOTFOUND based on [this](https://github.com/microsoft/CsWinRT/blob/04a590360338092bed7f80355e51a9af9950b022/src/WinRT.Runtime/ExceptionHelpers.cs#LL261C1-L262C1) and [this](https://learn.microsoft.com/en-us/dotnet/api/system.io.directorynotfoundexception?view=net-7.0#remarks). If this exception was to cross the ABI boundary and then end up in C# code again in the same process that probably would be fine given we will retrieve the managed exception from the error info and rethrow it. But if this was from a different process, we won't do that today and then when we do the error code translation, I think we are going to end up with a ComException. I assume before we would have ended up with a DirectoryNotFoundException (needs to be confirmed). Some of the other .NET errors are probably less common, but I feel like there might be some that are common like this one.
> 
> All other calls to [Marshal.GetExceptionForHr](https://github.com/search?q=repo%3Amicrosoft%2FCsWinRT+Marshal.throwexceptionforhr&type=code) probably needs to be updated to go through ExceptionHelpers.GetExceptionForHr, but this PR doesn't need to necessarily address that.

What would the IRestrictedErrorInfo look like when the exception is coming from another process?

manodasanW:
> > For the most part, this seems fine to me. @AaronRobinsonMSFT feel free to comment if you see anything `Marshal.GetExceptionForHr` handled before that this change doesn't handle.
> > My only concern is lets say a DirectoryNotFoundException is thrown in C#. This I assume gets translated to STG_E_PATHNOTFOUND based on [this](https://github.com/microsoft/CsWinRT/blob/04a590360338092bed7f80355e51a9af9950b022/src/WinRT.Runtime/ExceptionHelpers.cs#LL261C1-L262C1) and [this](https://learn.microsoft.com/en-us/dotnet/api/system.io.directorynotfoundexception?view=net-7.0#remarks). If this exception was to cross the ABI boundary and then end up in C# code again in the same process that probably would be fine given we will retrieve the managed exception from the error info and rethrow it. But if this was from a different process, we won't do that today and then when we do the error code translation, I think we are going to end up with a ComException. I assume before we would have ended up with a DirectoryNotFoundException (needs to be confirmed). Some of the other .NET errors are probably less common, but I feel like there might be some that are common like this one.
> > All other calls to [Marshal.GetExceptionForHr](https://github.com/search?q=repo%3Amicrosoft%2FCsWinRT+Marshal.throwexceptionforhr&type=code) probably needs to be updated to go through ExceptionHelpers.GetExceptionForHr, but this PR doesn't need to necessarily address that.
> 
> What would the IRestrictedErrorInfo look like when the exception is coming from another process?

I assume it would be some form of the one created [here ](https://github.com/microsoft/CsWinRT/blob/04a590360338092bed7f80355e51a9af9950b022/src/WinRT.Runtime/ExceptionHelpers.cs#L236) with some error information possibly removed during marshaling based on access permissions of the process.  If none were removed, today, that would mean it would have the restricted error and a proxy to the managed exception.  The latter is on my list to look at if there is a better way to handle that to allow reusing the managed exception.  But in both cases, the HRESULT that was originally used for `RoOriginateLanguageException` would be in the `IRestrictedErrorInfo`.

jlaanstra:
Need to map at least the following set of COR_E hresults that are effectively aliases:

#define COR_E_UNAUTHORIZEDACCESS E_ACCESSDENIED
#define COR_E_ARGUMENT E_INVALIDARG
#define COR_E_INVALIDCAST E_NOINTERFACE
#define COR_E_OUTOFMEMORY E_OUTOFMEMORY
#define COR_E_NULLREFERENCE E_POINTER
#define COR_E_ARITHMETIC __HRESULT_FROM_WIN32(ERROR_ARITHMETIC_OVERFLOW)
#define COR_E_PATHTOOLONG __HRESULT_FROM_WIN32(ERROR_FILENAME_EXCED_RANGE)
#define COR_E_FILENOTFOUND __HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND)
#define COR_E_ENDOFSTREAM __HRESULT_FROM_WIN32(ERROR_HANDLE_EOF)
#define COR_E_DIRECTORYNOTFOUND __HRESULT_FROM_WIN32(ERROR_PATH_NOT_FOUND)
#define COR_E_STACKOVERFLOW __HRESULT_FROM_WIN32(ERROR_STACK_OVERFLOW)
#define COR_E_AMBIGUOUSMATCH _HRESULT_TYPEDEF_(0x8000211DL)
#define COR_E_TARGETPARAMCOUNT _HRESULT_TYPEDEF_(0x8002000EL)
#define COR_E_DIVIDEBYZERO _HRESULT_TYPEDEF_(0x80020012L)
#define COR_E_BADIMAGEFORMAT _HRESULT_TYPEDEF_(0x8007000BL)

jlaanstra:
Updated to handle additional hresults.

Left these out as I wasn't sure about the mappings:
#define COR_E_AMBIGUOUSMATCH HRESULT_TYPEDEF(0x8000211DL)
#define COR_E_TARGETPARAMCOUNT HRESULT_TYPEDEF(0x8002000EL)
#define COR_E_DIVIDEBYZERO HRESULT_TYPEDEF(0x80020012L)

I added mappings for E_BOUNDS and ERROR_CANCELED

src/WinRT.Runtime/ExceptionHelpers.cs
```diff
@@ -152,54 +169,119 @@ private static Exception GetExceptionForHR(int hr, bool useGlobalErrorState, out
                     }
 
                     if (hr == hrLocal)
-                    {
-                        restrictedErrorInfoRef.TryAs<ABI.WinRT.Interop.IErrorInfo.Vftbl>(out iErrorInfo);
+                    {
+                        // For cross language WinRT exceptions, general information will be available in the description,
+                        // which is populated from IRestrictedErrorInfo::GetErrorDetails and more specific information will be available
+                        // in the resrictedError which also comes from IRestrictedErrorInfo::GetErrorDetails. If both are available, we
```
manodasanW, line 175:
```suggestion
                        // in the restrictedError which also comes from IRestrictedErrorInfo::GetErrorDetails. If both are available, we

```

manodasanW (approved):

AaronRobinsonMSFT:
This looks good. Sorry for the late reply.

PR #1340, opened 2023-06-26
Fix the tooling incompatible issue with NativeAOT

hez2010:
An RVA field within a generic type is not expressible with C#, and the NativeAOT compiler doesn't support such IL at all.
To resolve this issue, we can simply move IIDDataField out of the generic types instead of nesting the data inside of a generic type.

Unblocks the tooling issue to make sure ilc can compile CsWinRT projection assemblies. 

After this PR is merged, we can finally start investigating what is missing for NativeAOT-compatible. 

Fixes https://github.com/dotnet/runtime/issues/68278

/cc: @manodasanW @jkotas @MichalStrehovsky


manodasanW:
Thanks @hez2010, looks good to me.  Will merge after the CI pipelines are done.

PR #1343, opened 2023-07-31
Fix vftbl nested types finding

dongle-the-gadget:
In some cases (such as NativeAOT compilation), `TypeExtensions.GetAbiToProjectionVftblPtr` will not be able to get the nested `Vftbl` type in the specified helper type, resulting in the method being unable to get the field `AbiToProjectionVftblPtr`, causing the method to fail.

This error is caused due to the `DynamicallyAccessedMemberTypes` specified in the `DynamicallyAccessedMembersAttribute` of `GetAbiToProjectionVftblPtr` being inconsistent with that of `FindVftblType`, causing the `FindVftblType` method to be unable to get the nested `Vftbl` type.

This PR fixes that by adding in `DynamicallyAccessedMemberTypes.PublicNestedTypes` to `GetAbiToProjectionVftblPtr`, allowing `FindVftblType` to find nested types.

### DIFF
--- a/src/WinRT.Runtime/TypeExtensions.cs
+++ b/src/WinRT.Runtime/TypeExtensions.cs
@@ -134,7 +134,7 @@ namespace WinRT
 
         internal static IntPtr GetAbiToProjectionVftblPtr(
 #if NET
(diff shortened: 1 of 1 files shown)


manodasanW:
Thanks @dongle-the-gadget for finding this and for the change.

PR #1346, opened 2023-08-17
Use blittable signatures for Platform P/Invoke

### LINKED ISSUE #1317: Don't use managed signatures in Platform P/Invoke methods
**Describe the bug**

Related to #1315, same rationale, but for P/Invoke-s this time. This issue tracks managed signatures in several `Platform` methods:

https://github.com/microsoft/CsWinRT/blob/e644f9cca17612adea4d726bef02069eb422f9aa/src/cswinrt/strings/WinRT.cs#L29-L84

Specifically, these P/Invoke methods should be updated to use a blittable signature:
- `CoCreateInstance`
- `TryGetProcAddress`
- `TryGetProcAddress`
- `RoGetActivationFactory`
- `WindowsCreateString`
- `RoGetAgileReference`

If we can't use `[LibraryImport]` due to downlevel support, we should just update these manually.

**Expected behavior**

These P/Invoke-s should use blittable signatures.

cc. @jkoritzinsky

dongle-the-gadget:
Closes #1317.

cc @Sergio0694

### DIFF
--- a/src/WinRT.Runtime/ExceptionHelpers.cs
+++ b/src/WinRT.Runtime/ExceptionHelpers.cs
@@ -59,25 +59,19 @@ namespace WinRT
         private static delegate* unmanaged[Stdcall]<int, IntPtr, IntPtr, int> roOriginateLanguageException;
         private static delegate* unmanaged[Stdcall]<IntPtr, int> roReportUnhandledError;
 
-        private static readonly bool initialized =  Initialize();
+        private static readonly bool initialized = Initialize();
 
         private static bool Initialize()
         {
-            fixed (char* moduleName = "api-ms-win-core-winrt-error-l1-1-1.dll")
+            IntPtr winRTErrorModule = Platform.LoadLibraryExW("api-ms-win-core-winrt-error-l1-1-1.dll", IntPtr.Zero, (uint)DllImportSearchPath.System32);
+            if (winRTErrorModule != IntPtr.Zero)
             {
-                IntPtr winRTErrorModule = Platform.LoadLibraryExW((ushort*)moduleName, IntPtr.Zero, (uint)DllImportSearchPath.System32);
-                if (winRTErrorModule != IntPtr.Zero)
-                {
-                    roOriginateLanguageException = (delegate* unmanaged[Stdcall]<int, IntPtr, IntPtr, int>)Platform.GetProcAddress(winRTErrorModule, "RoOriginateLanguageException"u8);
-                    roReportUnhandledError = (delegate* unmanaged[Stdcall]<IntPtr, int>)Platform.GetProcAddress(winRTErrorModule, "RoReportUnhandledError"u8);
-                }
-                else
-                {
-                    fixed (char* fallbackModuleName = "api-ms-win-core-winrt-error-l1-1-0.dll")
-                    {
-                        winRTErrorModule = Platform.LoadLibraryExW((ushort*)fallbackModuleName, IntPtr.Zero, (uint)DllImportSearchPath.System32);
-                    }
-                }
+                roOriginateLanguageException = (delegate* unmanaged[Stdcall]<int, IntPtr, IntPtr, int>)Platform.GetProcAddress(winRTErrorModule, "RoOriginateLanguageException"u8);
+                roReportUnhandledError = (delegate* unmanaged[Stdcall]<IntPtr, int>)Platform.GetProcAddress(winRTErrorModule, "RoReportUnhandledError"u8);
+            }
+            else
+            {
+                winRTErrorModule = Platform.LoadLibraryExW("api-ms-win-core-winrt-error-l1-1-0.dll", IntPtr.Zero, (uint)DllImportSearchPath.System32);
             }
 
             if (winRTErrorModule != IntPtr.Zero)
@@ -227,7 +221,7 @@ namespace WinRT
                     ex = new COMException(
 @"Invalid window handle. (0x80070578)
 Consider WindowNative, InitializeWithWindow
-See https://aka.ms/cswinrt/interop#windows-sdk", 
+See https://aka.ms/cswinrt/interop#windows-sdk",
                         ERROR_INVALID_WINDOW_HANDLE);
                     break;
                 case RO_E_CLOSED:
--- a/src/WinRT.Runtime/ExceptionHelpers.cs
+++ b/src/WinRT.Runtime/ExceptionHelpers.cs
@@ -73,17 +73,20 @@ namespace WinRT
                 }
                 else
                 {
-                    winRTErrorModule = Platform.LoadLibraryExW((ushort*)moduleName, IntPtr.Zero, (uint)DllImportSearchPath.System32);
+                    fixed (char* fallbackModuleName = "api-ms-win-core-winrt-error-l1-1-0.dll")
+                    {
+                        winRTErrorModule = Platform.LoadLibraryExW((ushort*)moduleName, IntPtr.Zero, (uint)DllImportSearchPath.System32);
+                    }
                 }
-
-                if (winRTErrorModule != IntPtr.Zero)
-                {
-                    getRestrictedErrorInfo = (delegate* unmanaged[Stdcall]<IntPtr*, int>)Platform.GetProcAddress(winRTErrorModule, "GetRestrictedErrorInfo"u8);
-                    setRestrictedErrorInfo = (delegate* unmanaged[Stdcall]<IntPtr, int>)Platform.GetProcAddress(winRTErrorModule, "SetRestrictedErrorInfo"u8);
-                }
-
-                return true;
             }
+
+            if (winRTErrorModule != IntPtr.Zero)
+            {
+                getRestrictedErrorInfo = (delegate* unmanaged[Stdcall]<IntPtr*, int>)Platform.GetProcAddress(winRTErrorModule, "GetRestrictedErrorInfo"u8);
+                setRestrictedErrorInfo = (delegate* unmanaged[Stdcall]<IntPtr, int>)Platform.GetProcAddress(winRTErrorModule, "SetRestrictedErrorInfo"u8);
+            }
+
+            return true;
         }
 
         public static void ThrowExceptionForHR(int hr)
--- a/src/WinRT.Runtime/ExceptionHelpers.cs
+++ b/src/WinRT.Runtime/ExceptionHelpers.cs
@@ -85,14 +85,14 @@ namespace WinRT
             if (winRTErrorModule != IntPtr.Zero)
             {
 #if NET7_0_OR_GREATER || CsWinRT_LANG_11_FEATURES
-                ReadOnlySpan<byte> getRestrictedErrorInfo = "GetRestrictedErrorInfo"u8;
-                ReadOnlySpan<byte> setRestrictedErrorInfo = "SetRestrictedErrorInfo"u8;
+                ReadOnlySpan<byte> getRestrictedErrorInfoFuncName = "GetRestrictedErrorInfo"u8;
+                ReadOnlySpan<byte> setRestrictedErrorInfoFuncName = "SetRestrictedErrorInfo"u8;
 #else
-                string getRestrictedErrorInfo = "GetRestrictedErrorInfo";
-                string setRestrictedErrorInfo = "SetRestrictedErrorInfo";
+                string getRestrictedErrorInfoFuncName = "GetRestrictedErrorInfo";
+                string setRestrictedErrorInfoFuncName = "SetRestrictedErrorInfo";
 #endif
-                getRestrictedErrorInfo = (delegate* unmanaged[Stdcall]<IntPtr*, int>)Platform.GetProcAddress(winRTErrorModule, getRestrictedErrorInfo);
-                setRestrictedErrorInfo = (delegate* unmanaged[Stdcall]<IntPtr, int>)Platform.GetProcAddress(winRTErrorModule, setRestrictedErrorInfo);
+                getRestrictedErrorInfo = (delegate* unmanaged[Stdcall]<IntPtr*, int>)Platform.GetProcAddress(winRTErrorModule, getRestrictedErrorInfoFuncName);
+                setRestrictedErrorInfo = (delegate* unmanaged[Stdcall]<IntPtr, int>)Platform.GetProcAddress(winRTErrorModule, setRestrictedErrorInfoFuncName);
             }
 
             return true;
--- a/src/WinRT.Runtime/ExceptionHelpers.cs
+++ b/src/WinRT.Runtime/ExceptionHelpers.cs
@@ -66,8 +66,16 @@ namespace WinRT
             IntPtr winRTErrorModule = Platform.LoadLibraryExW("api-ms-win-core-winrt-error-l1-1-1.dll", IntPtr.Zero, (uint)DllImportSearchPath.System32);
             if (winRTErrorModule != IntPtr.Zero)
             {
-                roOriginateLanguageException = (delegate* unmanaged[Stdcall]<int, IntPtr, IntPtr, int>)Platform.GetProcAddress(winRTErrorModule, "RoOriginateLanguageException"u8);
-                roReportUnhandledError = (delegate* unmanaged[Stdcall]<IntPtr, int>)Platform.GetProcAddress(winRTErrorModule, "RoReportUnhandledError"u8);
+#if NET7_0_OR_GREATER || CsWinRT_LANG_11_FEATURES
+                ReadOnlySpan<byte> langExceptionString = "RoOriginateLanguageException"u8;
+                ReadOnlySpan<byte> reportUnhandledErrorString = "RoReportUnhandledError"u8;
+#else
+                string langExceptionString = "RoOriginateLanguageException";
+                string reportUnhandledErrorString = "RoReportUnhandledError";
+#endif
+
+                roOriginateLanguageException = (delegate* unmanaged[Stdcall]<int, IntPtr, IntPtr, int>)Platform.GetProcAddress(winRTErrorModule, langExceptionString);
+                roReportUnhandledError = (delegate* unmanaged[Stdcall]<IntPtr, int>)Platform.GetProcAddress(winRTErrorModule, reportUnhandledErrorString);
             }
             else
             {
--- a/src/WinRT.Runtime/ExceptionHelpers.cs
+++ b/src/WinRT.Runtime/ExceptionHelpers.cs
@@ -84,8 +84,15 @@ namespace WinRT
 
             if (winRTErrorModule != IntPtr.Zero)
             {
-                getRestrictedErrorInfo = (delegate* unmanaged[Stdcall]<IntPtr*, int>)Platform.GetProcAddress(winRTErrorModule, "GetRestrictedErrorInfo"u8);
-                setRestrictedErrorInfo = (delegate* unmanaged[Stdcall]<IntPtr, int>)Platform.GetProcAddress(winRTErrorModule, "SetRestrictedErrorInfo"u8);
+#if NET7_0_OR_GREATER || CsWinRT_LANG_11_FEATURES
+                ReadOnlySpan<byte> getRestrictedErrorInfo = "GetRestrictedErrorInfo"u8;
+                ReadOnlySpan<byte> setRestrictedErrorInfo = "SetRestrictedErrorInfo"u8;
+#else
+                string getRestrictedErrorInfo = "GetRestrictedErrorInfo";
+                string setRestrictedErrorInfo = "SetRestrictedErrorInfo";
+#endif
+                getRestrictedErrorInfo = (delegate* unmanaged[Stdcall]<IntPtr*, int>)Platform.GetProcAddress(winRTErrorModule, getRestrictedErrorInfo);
+                setRestrictedErrorInfo = (delegate* unmanaged[Stdcall]<IntPtr, int>)Platform.GetProcAddress(winRTErrorModule, setRestrictedErrorInfo);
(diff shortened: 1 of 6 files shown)


src/WinRT.Runtime/ExceptionHelpers.cs
```diff
@@ -63,24 +63,27 @@ static unsafe class ExceptionHelpers
...
-            }
+                IntPtr winRTErrorModule = Platform.LoadLibraryExW((ushort*)moduleName, IntPtr.Zero, (uint)DllImportSearchPath.System32);
+                if (winRTErrorModule != IntPtr.Zero)
+                {
+                    roOriginateLanguageException = (delegate* unmanaged[Stdcall]<int, IntPtr, IntPtr, int>)Platform.GetProcAddress(winRTErrorModule, "RoOriginateLanguageException"u8);
+                    roReportUnhandledError = (delegate* unmanaged[Stdcall]<IntPtr, int>)Platform.GetProcAddress(winRTErrorModule, "RoReportUnhandledError"u8);
+                }
+                else
+                {
+                    winRTErrorModule = Platform.LoadLibraryExW((ushort*)moduleName, IntPtr.Zero, (uint)DllImportSearchPath.System32);
```
Sergio0694, line 76: This doesn't seem correct, you're trying to load the same library again. The previous code was trying to load the `l1-1-1` .dll first, and falling back to `l1-1-0` otherwise. You're using the same .dll name in both cases instead 😅

src/WinRT.Runtime/WinRT.Runtime.csproj
```diff
@@ -4,7 +4,7 @@
     <TargetFrameworks>$(LibBuildTFMs)</TargetFrameworks>
     <RootNamespace>WinRT</RootNamespace>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
-    <LangVersion>9</LangVersion>
+    <LangVersion>11.0</LangVersion>
```
Sergio0694, line 7: This is technically a breaking change for the embedded version of CsWinRT. I think we should consider bumping that but we might need a major release for that (maybe along with the rest of AOT changes?). cc. @manodasanW

Sergio0694 (changes requested):
Did a first quick pass and left some notes 🙂

src/WinRT.Runtime/ExceptionHelpers.cs
dongle-the-gadget: Whoops. Fixed now.

src/WinRT.Runtime/WinRT.Runtime.csproj
dongle-the-gadget: From what I understand, doesn't the embedded version use a compiled version of WinRT.Runtime, unless if we are going out our way to build our own WinRT.Runtime?

Though it might break some tests, yeah. I haven't checked.

src/WinRT.Runtime/ExceptionHelpers.cs
```diff
@@ -51,133 +51,139 @@ static unsafe class ExceptionHelpers
...
+        private static delegate* unmanaged[Stdcall]<IntPtr*, int> getRestrictedErrorInfo;
+        private static delegate* unmanaged[Stdcall]<IntPtr, int> setRestrictedErrorInfo;
+        private static delegate* unmanaged[Stdcall]<int, IntPtr, IntPtr, int> roOriginateLanguageException;
+        private static delegate* unmanaged[Stdcall]<IntPtr, int> roReportUnhandledError;
+
+        private static readonly bool initialized =  Initialize();
+
+        private static bool Initialize()
+        {
+            fixed (char* moduleName = "api-ms-win-core-winrt-error-l1-1-1.dll")
```
Sergio0694, line 66: This is making the code quite messy. I'm thinking we should consider doing what `[LibraryImport]` is doing, just manuall — keep managed signatures for the platform invoke APIs, and just implement them with blittable P/Invoke internally. Same performance and AOT-friendliness improvements, and much less error prone and clunky code elsewhere 🙂

src/WinRT.Runtime/WinRT.Runtime.csproj
Sergio0694: No it's shipping all raw source files in the NuGet package so consumers can build it locally into their assemblies. If it was a compiler version of the runtime it wouldn't really be embedded, as it'd be a separate assembly 😄

src/WinRT.Runtime/ExceptionHelpers.cs
dongle-the-gadget: I've created an abstraction for `LoadLibraryExW`. As for the other methods, I don't think it affects code readability much. Will probably still create abstractions for them in case the user requires access to those methods (even if I'm not sure why).

src/WinRT.Runtime/WinRT.Runtime.csproj
dongle-the-gadget: The safest version would probably be 10.0 (.NET 6 default), since that's the oldest version that C#/WinRT officially supports. However, it lacks the UTF8 string literals feature, which is quite extensively used in this PR.

manodasanW: Incrementing the lang version is a concern for the reasons mentioned.  If UTF8-string literals were why we had to increment it so that we can pass them as pointers, what is the thought on using the pinnable string marshaling from [here](https://github.com/microsoft/CsWinRT/blob/546cffe742100fdd4bf91eb29486fabf5d22d742/src/cswinrt/strings/WinRT.cs#L169-L172).

manodasanW: Basically making the other string parameters to what the WindowsCreateStringReference method signature was before this change.

dongle-the-gadget: We could just pass in the raw bytes of the UTF-8 string or use `Encoding.UTF8`. The first approach is faster but hurts code readability. The second approach is more readable but makes the code slower.

src/WinRT.Runtime/ExceptionHelpers.cs
dongle-the-gadget: I've added the rest of the wrappers (except for `GetProcAddress`, which is being discussed in the other thread). Will flag this as resolved.

src/cswinrt/strings/WinRT.cs
```diff
@@ -29,7 +30,12 @@ public static void DynamicInvokeAbi(this System.Delegate del, object[] invoke_pa
     internal sealed class Platform
     {
         [DllImport("api-ms-win-core-com-l1-1-0.dll")]
-        internal static extern unsafe int CoCreateInstance(ref Guid clsid, IntPtr outer, uint clsContext, ref Guid iid, IntPtr* instance);
+        internal static extern unsafe int CoCreateInstance(Guid* clsid, IntPtr outer, uint clsContext, Guid* iid, IntPtr* instance);
+
+        internal static unsafe int CoCreateInstance(ref Guid clsid, IntPtr outer, uint clsContext, ref Guid iid, IntPtr* instance)
+        {
+            return CoCreateInstance((Guid*)Unsafe.AsPointer(ref clsid), outer, clsContext, (Guid*)Unsafe.AsPointer(ref iid), instance);
```
Sergio0694, line 37: Using `Unsafe.AsPointer` like this is a GC hole, and it's not safe.

Sergio0694 (changes requested):

src/cswinrt/strings/WinRT.cs
dongle-the-gadget: Pinning?

Sergio0694: Yes, pinning would be the correct way to do things here 🙂

dongle-the-gadget: Should be pinned now.

src/WinRT.Runtime/WinRT.Runtime.csproj
manodasanW: Looking at the generated code of the `LibraryImport` source generator, it looks like they are using `Encoding.UTF8`.  I think we use `Encoding.UTF8` here too.  That would also work with all scenarios like `LoadLibrary(_currentModuleDirectory...)`.

dongle-the-gadget: What if we could cross compile with .NET 7 and use `#if NET7_0_OR_GREATER`? Since .NET 7 default is C# 11, which allows UTF-8 string literals.

manodasanW: That is a good question.  The issue is `WinRT.Runtime` is distributed in a targeting package `Microsoft.Windows.SDK.NET.Ref` which gets ingested by the .NET SDK which is the primary way folks get it and unless we start shipping a .NET 6 and .NET 7 version of the package, which I want to avoid at the moment, I don't think I can have 2 versions of the DLL and have the appropriate one included based on TFM.

This would also mean that all projections that are generated with CsWinRT would also need to generate a .NET 7 version for apps to get an AOT compatible version which I want to avoid until there is a real need to.

dongle-the-gadget: > This would also mean that all projections that are generated with CsWinRT would also need to generate a .NET 7 version for apps to get an AOT compatible version which I want to avoid until there is a real need to.

I meant that we would use UTF-8 literals for .NET 7 and `Encoding.UTF8` for .NET 6 and older. Both will still benefit from the AOT-friendly Platform P/Invokes, just that the .NET 6 version will be slower due to encoding overhead.

However, of course, this still means we need to cross target .NET 6 and .NET 7, which as you said is currently undesirable. How about we would currently use `Encoding.UTF8` for both runtimes (and build a single binary for .NET 6) and then reinvestigate using UTF-8 string literals when there are more opportunities for breaking changes?

manodasanW: If `Encoding.UTF8` is the fallback, then I can see doing what you suggested.  Actually, given this is a language feature, I can see also using it for `WinRT.Runtime` and the `Windows SDK projection` as we can continue to target .NET 6 but build using the .NET 7 SDK and have language version set to 11 for our builds while not breaking ones that are doing embedded.

To allow that, for the ifdefs we should do `#if NET7_0_OR_GREATER || CsWinRT_LANG_11_FEATURES` or something along those lines and then define `CsWinRT_LANG_11_FEATURES` in `WinRT.Runtime.csproj` along with `LangVersion 11`.

dongle-the-gadget: Should be done in 25ab8f9ffe47352341c42fd3d29958cd5cd06899

src/cswinrt/strings/WinRT.cs
```diff
@@ -42,34 +54,94 @@ internal sealed class Platform
...
+        {
+            fixed (byte* lpFunctionName = functionName)
+            {
+                return TryGetProcAddress(moduleHandle, (sbyte*)lpFunctionName);
+            }
+        }
+
+        internal static unsafe void* TryGetProcAddress(IntPtr moduleHandle, string functionName)
+        {
+            fixed (byte* lpFunctionName = Encoding.UTF8.GetBytes(functionName))
```
Sergio0694, line 69: This is extremely inefficient, and it will allocate and throw away a buffer for the name every time.

src/cswinrt/strings/WinRT.cs
```diff
@@ -42,34 +54,94 @@ internal sealed class Platform
...
+
+        internal static unsafe void* TryGetProcAddress(IntPtr moduleHandle, ReadOnlySpan<byte> functionName)
+        {
+            fixed (byte* lpFunctionName = functionName)
+            {
+                return TryGetProcAddress(moduleHandle, (sbyte*)lpFunctionName);
+            }
+        }
+
+        internal static unsafe void* TryGetProcAddress(IntPtr moduleHandle, string functionName)
```
Sergio0694, line 67: If this overload is only for downlevel, we should conditionally only include this when the other one is not.
We shouldn't just always include both of these, as only one of them is used at a time.

Sergio0694 (changes requested):

src/cswinrt/strings/WinRT.cs
dongle-the-gadget: So we should have like two P/Invoke declarations for `GetProcAddress`? One blittable and the other not?

Sergio0694: I think we should evaluate with @manodasanW what approach to take here with the constraints we have, and possibly reevaluate some of them. For instance, I'm not sure the split for C# 11 is worth the complexity. If the embedded source files have to support downlevel to .NET Standard 2.0, then their current C# 9 language level is already officially not supported anyway. So at that point why not just bump that to C# 11 directly and say if you want to embed CsWinRT, you need to set your language version to C# 11. They'd have to do the same anyway for C# 9, and both are equally "not supported". Doing this would at least let us stop worrying about the language, and just check the TFM.

src/cswinrt/strings/WinRT.cs
dongle-the-gadget: What do you think would be a better way to do this? From looking at `LibraryImport`, it seems to stack allocate the UTF-8 string if it's small and heap allocate it otherwise.

src/cswinrt/strings/WinRT.cs
dongle-the-gadget: Well the TFM check isn't really needed (after all, UTF-8 string literals are only language-dependent rather than runtime-dependent). It's mostly to make sure that we are on the "correct" language version for the feature (since .NET 7 uses C#11 by default). Actually checking for the C# version is... not a very easy task to do in MSBuild.

src/cswinrt/strings/WinRT.cs
Sergio0694: That's a reasonable approach, since most method names would be small enough anyway. Could pick a given threshold (maybe 0x100 like `LibraryImport`) and use that with a `stackalloc` if the name is small enough, or just allocate a new array if not, which would realistically never happen.

src/cswinrt/strings/WinRT.cs
Sergio0694: Yes but my point is even if we kept the limit of C# 9, that'd still be technically unsupported already. So what I'm saying is maybe we might as well just say you need C# 11 for embed support, and make the code simpler and more consistent. I can't make a decision on this though, I'm just suggesting what I think a possible approach could be 🙂

src/cswinrt/strings/WinRT.cs
dongle-the-gadget: Does this seem reasonable?
```cs
internal static unsafe void* TryGetProcAddress(IntPtr moduleHandle, string functionName)
{
    bool allocated = false;
    Span<byte> buffer = stackalloc byte[0x100];
    if (functionName.Length * 3 >= 0x100) // Maximum of 3 bytes per UTF-8 character, stack allocation limit of 256 bytes (including the null terminator)
    {
        // Calculate accurate byte count when the provided stack-allocated buffer is not sufficient
        int exactByteCount = checked(Encoding.UTF8.GetByteCount(functionName) + 1); // + 1 for null terminator
        if (exactByteCount > 0x100)
        {
#if NET6_0_OR_GREATER
            buffer = new((byte*)NativeMemory.Alloc((nuint)exactByteCount), exactByteCount);
#else
            buffer = (byte*)Marshal.AllocHGlobal(exactByteCount);
#endif
            allocated = true;
        }
    }

    var rawByte = (byte*)Unsafe.AsPointer(ref MemoryMarshal.GetReference(buffer));

    int byteCount = Encoding.UTF8.GetBytes(functionName, buffer);
    buffer[byteCount] = 0;

    void* functionPtr = TryGetProcAddress(moduleHandle, (sbyte*)rawByte);

    if (allocated)
#if NET6_0_OR_GREATER
        NativeMemory.Free(rawByte);
#else
        Marshal.FreeHGlobal((IntPtr)rawByte);
#endif

    return functionPtr;
}
```

src/cswinrt/strings/WinRT.cs
manodasanW: We have folks with embedded projections and regular projections that are going to be on the .NET 6 SDK due to that is the LTS build.  So that is why we will still need the non C# 11 fallback.

src/WinRT.Runtime/ExceptionHelpers.cs
```diff
@@ -51,133 +51,141 @@ static unsafe class ExceptionHelpers
...
+                roReportUnhandledError = (delegate* unmanaged[Stdcall]<IntPtr, int>)Platform.GetProcAddress(winRTErrorModule, reportUnhandledErrorString);
+            }
+            else
+            {
+                winRTErrorModule = Platform.LoadLibraryExW("api-ms-win-core-winrt-error-l1-1-0.dll", IntPtr.Zero, (uint)DllImportSearchPath.System32);
+            }
+
+            if (winRTErrorModule != IntPtr.Zero)
+            {
+                getRestrictedErrorInfo = (delegate* unmanaged[Stdcall]<IntPtr*, int>)Platform.GetProcAddress(winRTErrorModule, "GetRestrictedErrorInfo"u8);
```
manodasanW, line 87: ifdef here too?

src/cswinrt/strings/WinRT.cs
Sergio0694: Is this something we could reevaluate after .NET 8 hits RTM? As in, could we just say consumers of the embedded projections would be expected to be using a language version matching whatever the latest LTS is?

manodasanW: Yea we will revisit this a couple months after .NET 8 comes out.

src/cswinrt/strings/WinRT.cs
manodasanW: This looks reasonable but might want to confirm that the version of `Encoding.UTF8.GetBytes` you are using is accessible in NET Standard 2.0.  It is fine if the .NET Standard version is not blittable / AOT complaint if there is no equivalent.

src/WinRT.Runtime/ExceptionHelpers.cs
dongle-the-gadget: Not sure why it didn't show up in VS Code Search 😅. Will fix that.

src/cswinrt/strings/WinRT.cs
dongle-the-gadget: I think what I'll do is pin the string on .NET Standard 2.0 and use the `GetBytes(Char*, Int32, Byte*, Int32)` function.

src/cswinrt/strings/WinRT.cs
```diff
@@ -42,7 +54,74 @@ internal sealed class Platform
...
+#endif
+                    allocated = true;
+                }
+            }
+
+            var rawByte = (byte*)Unsafe.AsPointer(ref MemoryMarshal.GetReference(buffer));
+
+            int byteCount;
+
+#if NETSTANDARD2_1_OR_GREATER
```
manodasanW, line 90: Based on [this](https://learn.microsoft.com/en-us/dotnet/standard/frameworks#preprocessor-symbols), I am not sure if `NETSTANDARD2_1_OR_GREATER` will be set or not when targeting .NET 6.  To differentiate between our .NET Standard projection and our .NET5+ projection, we typically just do `#if NET` which I think will work here.

dongle-the-gadget: You're right. Fixed now.

manodasanW:
I had ran the CI pipeline and it looks like it ran into the following errors:

```
  src\WinRT.Runtime\ExceptionHelpers.cs(57,67): warning CS0649: Field 'ExceptionHelpers.getRestrictedErrorInfo' is never assigned to, and will always have its default value [src\WinRT.Runtime\WinRT.Runtime.csproj]
  src\WinRT.Runtime\ExceptionHelpers.cs(58,66): warning CS0649: Field 'ExceptionHelpers.setRestrictedErrorInfo' is never assigned to, and will always have its default value [src\WinRT.Runtime\WinRT.Runtime.csproj]

  src\cswinrt\strings\WinRT.cs(80,30): error CS0029: Cannot implicitly convert type 'byte*' to 'System.Span<byte>' [src\WinRT.Runtime\WinRT.Runtime.csproj]
  src\WinRT.Runtime\ExceptionHelpers.cs(94,42): error CS0029: Cannot implicitly convert type 'delegate* unmanaged[Stdcall]<System.IntPtr*, int>' to 'System.ReadOnlySpan<byte>' [src\WinRT.Runtime\WinRT.Runtime.csproj]
  src\WinRT.Runtime\ExceptionHelpers.cs(95,42): error CS0029: Cannot implicitly convert type 'delegate* unmanaged[Stdcall]<System.IntPtr, int>' to 'System.ReadOnlySpan<byte>' [src\WinRT.Runtime\WinRT.Runtime.csproj]

  src\WinRT.Runtime\ExceptionHelpers.cs(94,42): error CS0029: Cannot implicitly convert type 'delegate* unmanaged[Stdcall]<nint*, int>' to 'System.ReadOnlySpan<byte>' [src\WinRT.Runtime\WinRT.Runtime.csproj]
  src\WinRT.Runtime\ExceptionHelpers.cs(95,42): error CS0029: Cannot implicitly convert type 'delegate* unmanaged[Stdcall]<nint, int>' to 'System.ReadOnlySpan<byte>' [src\WinRT.Runtime\WinRT.Runtime.csproj]
```

dongle-the-gadget:
@manodasanW the errors should be fixed now. Can you try running the CI again?

manodasanW:
@dongle-the-gadget the latest CI run passed.  Thanks for your changes and working through the feedback.

PR #1348, opened 2023-08-24
Delete generated files before regenerating projection.

### LINKED ISSUE #1189: CsWinRT doesn't handle removed namespaces or project references.
**Describe the bug**
CsWinRT doesn't handle removed namespaces or project references. When a namespace is removed, the file containing the projection stays behind.

**Expected behavior**
Files should be properly cleaned up.

**Version Info**
1.6.3

jlaanstra:
CsWinRT generates one .cs file per namespace. Once you remove a namespace CsWinRT doesn't remove the file for that namespace but still includes it since it was previously generated.

This change explicitly removes all generated files before generating a new projection.

Fixes #1189

### DIFF
--- a/nuget/Microsoft.Windows.CsWinRT.targets
+++ b/nuget/Microsoft.Windows.CsWinRT.targets
@@ -66,8 +66,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
   <Target Name="CsWinRTPrepareProjection" DependsOnTargets="$(CsWinRTPrepareProjectionDependsOn)" Outputs="$(CsWinRTGeneratedFilesDir)">
     <PropertyGroup>
-      <CsWinRTGeneratedFilesDir Condition="'$(CsWinRTGeneratedFilesDir)' == ''">$(GeneratedFilesDir)</CsWinRTGeneratedFilesDir>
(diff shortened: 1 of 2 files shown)


nuget/Microsoft.Windows.CsWinRT.targets
```diff
@@ -66,8 +66,8 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
   <Target Name="CsWinRTPrepareProjection" DependsOnTargets="$(CsWinRTPrepareProjectionDependsOn)" Outputs="$(CsWinRTGeneratedFilesDir)">
     <PropertyGroup>
-      <CsWinRTGeneratedFilesDir Condition="'$(CsWinRTGeneratedFilesDir)' == ''">$(GeneratedFilesDir)</CsWinRTGeneratedFilesDir>
-      <CsWinRTGeneratedFilesDir Condition="'$(CsWinRTGeneratedFilesDir)' == ''">$([MSBuild]::NormalizeDirectory('$(MSBuildProjectDirectory)', '$(IntermediateOutputPath)', 'Generated Files'))</CsWinRTGeneratedFilesDir>
+        <CsWinRTGeneratedFilesDir Condition="'$(CsWinRTGeneratedFilesDir)' == '' and '$(GeneratedFilesDir)' != ''">$(GeneratedFilesDir)\CsWinRT\</CsWinRTGeneratedFilesDir>
```
manodasanW, line 69: nit: seems like there is extra spacing here.

manodasanW (approved):

Issue #1270, opened 2022-11-14
CopyBuffer() method of SoftwareBitmap to an array crashes when trimming is enabled


lhak:
**Describe the bug**

When using the CopyBuffer() method of a SoftwareBitmap to copy the pixels to a byte array the app crashes with an "Invalid cast" exception if trimming is enabled

**To Reproduce**

1. Run attached repro project

[SoftwareBitmapTrimRepro.zip](https://github.com/microsoft/CsWinRT/files/10005705/SoftwareBitmapTrimRepro.zip)

**Expected behavior**

Should work the same as without trimming enabled

**Version Info**

WindowsAppSDK 1.2

**Additional context**

lhak:
It seems that the issue is caused by the WindowsRuntimeBuffer class (and related extension methods like AsBuffer() for arrays) when trimming is enabled. Workaround: Add a TrimmerRootDescriptor file with the following content:

```xml
<linker>
  <assembly fullname="Microsoft.Windows.SDK.NET">
    <type fullname="System.Runtime.InteropServices.WindowsRuntime.WindowsRuntimeBuffer"/>
  </assembly>
</linker>
```

PR #1352, opened 2023-08-31
Initial changes to make vtable generation trimming safe for AOT

manodasanW:
This PR adds a new attribute `WinRTExposedType` which can be put on C# implemented types that are passed across the ABI.  This attribute takes an array of WinRT interface types which are entries on the CCW vtable for that type.  Previously we were using `DynamicallyAccessedMemberTypes.Interfaces` along with reflection, but given this attribute was placed on a function several levels down and one of which is of type `object` rather than the concrete type, it didn't work as intended when all assemblies were trimmed.  With this new attribute, we are also able to skip some of the WinRT type checks along with covariance type generation at runtime along with not needing to keep all the interfaces on the type that aren't WinRT interfaces if they aren't needed.

This attribute can be added by libraries and apps on their C# types that they use with CsWinRT projected APIs if desired.  But this PR also adds a source generator which will analyze the sources and determine if there are any WinRT interfaces implemented by the class and if so adds the attribute to the class if it is declared partial.  CsWinRT also puts that attribute on `unsealed classes` so that any internal interfaces implemented by such projected classes in override / protected scenarios are also on the vtable if they are extended by another C# class.

```
[global::WinRT.WinRTExposedType(typeof(TestComponentCSharp.IProperties1), typeof(TestComponentCSharp.IUriHandler))]
partial class ManagedProperties
{
}
```

Upcoming todo:

- Currently the source generator only handles classes but in a future PR will also handle structs and delegates to add the IReference interface.  Custom type mapped delegates like `System.EventHandler<int>` may need to be handled differently.
- Diagnostics will also be added in a future PR to do a code fix to add `partial` if the class isn't already marked as such.
- CsWinRT Authored WinRT components will also be updated to have this attribute.


src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
```diff
@@ -0,0 +1,210 @@
...
+                    source.Append($@"}}");
+                }
+
+                string prefix = vtableAttribute.IsGlobalNamespace ? "" : $"{vtableAttribute.Namespace}.";
+                sourceProductionContext.AddSource($"{prefix}{vtableAttribute.ClassName}.WinRTVtable.g.cs", source.ToString());
+            }
+        }
+    }
+
+    internal readonly struct VtableAttribute
```
Sergio0694, line 193: This model is not equatable, so it'll break the incremental pipeline. You'll need two changes:
- Make this a `sealed record` (it's large enough, not worth making it a struct)
- Replace that `string[]` with `EquatableArray<string>` (copy it from [here](https://github.com/CommunityToolkit/dotnet/blob/main/src/CommunityToolkit.Mvvm.SourceGenerators/Helpers/EquatableArray%7BT%7D.cs))

src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
```diff
@@ -0,0 +1,210 @@
...
+                static (n, _) => GetVtableAttributeToAdd(n));
+
+            context.RegisterImplementationSourceOutput(vtableAttributesToAdd, GenerateVtableAttributes);
+        }
+
+        // Restrict to non-projected classes which can be instantiated
+        // and are partial allowing to add attributes.
+        private static bool NeedVtableAttribute(SyntaxNode node)
+        {
+            return node is ClassDeclarationSyntax declaration &&
```
Sergio0694, line 26: @jkoritzinsky do you think using a syntax provider with this filtering is enough to not make the transform resolving symbols be too expensive? I know that at least for now, `RegisterImplementationSourceOutput` isn't yet wired to actually skip work during normal edits, so just wondering whether there might be any concerns here 🤔

src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
```diff
@@ -0,0 +1,210 @@
...
+        {
+            compatibleTypes = null;
+
+            // Out of all the C# interfaces which are valid WinRT interfaces and
+            // support covariance, they all only have one generic parameter,
+            // so scoping to only handle that.
+            if (!(type.IsGenericType &&
+                type.TypeParameters.Length == 1 &&
+                type.TypeParameters[0].Variance == VarianceKind.Out &&
+                !type.TypeArguments[0].IsValueType))
```
Sergio0694, line 95: Nit: consider using a pattern to simplify things?
```suggestion
            if (type is not { IsGenericType: true, TypeParameters: [{ Variance: VarianceKind.Out, IsValueType: false })
```
You can add PolySharp as a package reference to this project to get all the necessary polyfills, then you can just set the language version to C# 11 in the generator and it'll work fine 🙂

src/WinRT.Runtime/ComWrappersSupport.cs
```diff
@@ -216,36 +210,45 @@ internal static object GetRuntimeClassCCWTypeIfAny(object obj)
             else
             {
                 var objType = type.GetRuntimeClassCCWType() ?? type;
-                var interfaces = objType.GetInterfaces();
-                foreach (var iface in interfaces)
+                // Check whether the type itself has the attribute to make sure it is using the new source generator
+                // and just doesn't have an updated dependent projection which is.
+                if (type.GetCustomAttribute(typeof(WinRTExposedTypeAttribute), false) is IWinRTExposedTypeDetails winrtExposedClassAttribute)
```
Sergio0694, line 215: Just so I understand, why the `IWinRTExposedTypeDetails` instead of just accessing the attribute directly? Wouldn't this also add a tiny bit more overhead due to the interface calls and the extra interface on the attribute?

src/WinRT.Runtime/ComWrappersSupport.cs
```diff
@@ -334,6 +337,22 @@ internal static object GetRuntimeClassCCWTypeIfAny(object obj)
             });
 
             return entries;
+
+            void AddInterfaceToVtable(Type iface)
```
Sergio0694, line 341: I know this method is not annotated, but is part is trim-safe yet? I'm seeing eg. some warnings in `FindHelperType` and whatnot, which are just not surfaced here because the method itself is not annotated. Just trying to better understand which parts are "ok" already and which are still work in progress 🙂

Sergio0694 (commented):
Just leaving some initial thoughts 🙂

I assume the analyzer you mentioned would be to warn users if their types are not `partial`, because the generator wouldn't be able to add the necessary attributes to it?

Sergio0694:
For authored WinRT components, will it be an issue that the new attributes takes an array of types? Array of types aren't allowed as attribute parameters in WinRT components as far as I know, and I remember we specifically hit this issue in the Microsoft Store as well. Unless that restriction was just due to .NET Native specifically and not something about WinRT components per se? 🤔

manodasanW:
> I assume the analyzer you mentioned would be to warn users if their types are not `partial`, because the generator wouldn't be able to add the necessary attributes to it?

Right, I believe the type would need to be marked partial for the source generator to add the attribute.  There is always the option for the developer themselves to put the attribute if they want and I plan on updating the source generator to detect that and not put it itself.

manodasanW:
> For authored WinRT components, will it be an issue that the new attributes takes an array of types? Array of types aren't allowed as attribute parameters in WinRT components as far as I know, and I remember we specifically hit this issue in the Microsoft Store as well. Unless that restriction was just due to .NET Native specifically and not something about WinRT components per se? 🤔

Based on this from the docs "The types of a custom attribute's parameters and fields are limited to the WinRT fundamental types, enums, and references to other WinRT types. No other parameter or field type is allowed.", it sounds like array of types might not be allowed on an WinRT attribute.  But in this case, the attribute we are putting will not be emitted to the WinMD to project, it will just be on the actual implementation type as a C# attribute (not a WinRT attribute).

src/WinRT.Runtime/ComWrappersSupport.cs
manodasanW: I expect there to be changes to both FindHelperType and GetAbiToProjectionVftblPtr as I work through my generic changes.

src/WinRT.Runtime/ComWrappersSupport.cs
manodasanW: I may end up removing the interface, but the interface is there atm as for both generic interface scenarios and possibly for structs and delegates, I was thinking about generating unique attributes extending `WinRTExposedTypeAttribute` that have custom constructors that provide more flexibility to get types and so on while knowing the type information without reflection.  I was debating between doing that vs structs and delegates getting their own common unique attribute type.  I most likely will just go with extending instead so I probably would be able to get rid of the interface.

manodasanW:
This should possibly fix #1270, will need to validate.

src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
```diff
@@ -0,0 +1,216 @@
...
+
+        public VtableAttribute(string @namespace, bool isGlobalNamespace, string className, EquatableArray<string> interfaces, bool hasWinRTExposedBaseType)
+        {
+            Namespace = @namespace;
+            IsGlobalNamespace = isGlobalNamespace;
+            ClassName = className;
+            Interfaces = interfaces;
+            HasWinRTExposedBaseType = hasWinRTExposedBaseType;
+        }
+    }
```
Sergio0694, line 215: Nit: now that you have PolySharo, you can just use the record syntax to keep it more compact 🙂
```suggestion
    internal sealed record VtableAttribute(
        string Namespace,
        bool IsGlobalNamespace,
        string ClassName,
        EquatableArray<string> Interfaces,
        bool HasWinRTExposedBaseType);
```

vineeththomasalex (approved):

manodasanW:
Merging these changes for now to unblock my other PRs, but I will probably have an upcoming PR that will slightly change the design here to the below.  This will allow to eliminate some of the remaining reflection to get the vftbl ptr and also be a design that we can eventually easily move to static interfaces from C# 11 when we can.

```
[global::WinRT.WinRTExposedType(typeof(ManagedPropertiesCCW))]
partial class ManagedProperties
{
}

public ManagedPropertiesCCW : IWinRTExposedTypeDetails
{
...
}
```

PR #1354, opened 2023-09-11
Manually generate signatures for some PInvokes to handle SetLastError.

jlaanstra:
Mimicks the generated code from LibraryImport but with handrolled Get/SetLastError as it's not available everywhere.


src/cswinrt/strings/WinRT.cs
```diff
@@ -49,18 +56,50 @@ internal static unsafe int CoCreateInstance(ref Guid clsid, IntPtr outer, uint c
...
+            returnValue = nativeReturnValue != 0;
+            SetLastError(lastError);
+            return returnValue;
+
+            // Local P/Invoke
+            [DllImportAttribute("kernel32.dll", EntryPoint = "FreeLibrary", ExactSpelling = true)]
+            static extern unsafe int PInvoke(IntPtr nativeModuleHandle);
+        }
+
+        internal static unsafe void* TryGetProcAddress(IntPtr moduleHandle, byte* functionName)
```
manodasanW, line 80: From my perspective, converting this signature from `sbyte*` to `byte*` should be fine and that looks to be also what the .NET source generator generates.  In addition, we are doing explicit casts from `byte*` to `sbyte*` in all the cases when passing it.  But to confirm, @dongle-the-gadget / @Sergio0694, is there anything I am missing on why `sbyte*` can be useful here with respect to #1346?

Sergio0694: I'd recommend keeping `sbyte*` (which is also what TerraFX does). `LPCSTR` is defined as `char*`, which isn't signed by default, but most compilers treat it as signed. Therefore, `sbyte*` is "more correct" in this scenario.

dongle-the-gadget:
https://github.com/microsoft/CsWinRT/blob/cc36846eba0342e9f3dd0f457783202df198842f/src/cswinrt/strings/WinRT.cs#L32-L36

Should these signatures translate to `uint`?

src/cswinrt/strings/WinRT.cs
```diff
@@ -29,6 +29,13 @@ public static void DynamicInvokeAbi(this System.Delegate del, object[] invoke_pa
 
     internal sealed class Platform
     {
+        [DllImportAttribute("kernel32.dll", EntryPoint = "SetLastError", ExactSpelling = true)]
+        internal static extern void SetLastError(int errorCode);
```
Sergio0694, line 33: These should use the `Marshal` APIs (on .NET 6+), not manually P/Invoke these

src/cswinrt/strings/WinRT.cs
```diff
@@ -49,18 +56,50 @@ internal static unsafe int CoCreateInstance(ref Guid clsid, IntPtr outer, uint c
...
+            int nativeReturnValue;
+            {
+                SetLastError(0);
+                nativeReturnValue = PInvoke(moduleHandle);
+                lastError = GetLastError();
+            }
+
+            // Unmarshal - Convert native data to managed data.
+            returnValue = nativeReturnValue != 0;
+            SetLastError(lastError);
```
Sergio0694, line 72: This isn't correct, it's literally just setting the same error it just retrieved from the same place.

Sergio0694 (changes requested):
As a general rule: if a library is multi-targeting, unless very extreme cases where not possible, do not make compromises on the latest target by using downlevel APIs everywhere. Instead, add necessary polyfills only on the downlevel target, and use the best approach on the latest one.

Sergio0694:
My two cents: like with previous changes in this file, I'm not really sure this is worth the effort at all right now, as this is not code we should keep. We need to bump the TFM for CsWinRT to .NET 8 (I guess once it goes GA), and then only handle NS2.0 and .NET 8, where on the .NET 8 we'll be able to just use `[LibraryImport]`, which will do the right thing. For downlevel, we can then either not worry about eg. the last system error, or just add polyfills then. But making so many changes now I think is not ideal, given we'll end up rewriting most of this file soon anyway once we finally can bump that TFM 🤔

dongle-the-gadget:
> We need to bump the TFM for CsWinRT to .NET 8 (I guess once it goes GA)

@manodasanW plans on moving away from the .NET 6 TFM?

jlaanstra:
I'll update the PR with multitargeting given some of the concerns raised about doing this correct downlevel. 

I don't see .NET 6 getting dropped soon.

jlaanstra:
> As a general rule: if a library is multi-targeting, unless very extreme cases where not possible, do not make compromises on the latest target by using downlevel APIs everywhere. Instead, add necessary polyfills only on the downlevel target, and use the best approach on the latest one.

Changed to only use new signatures on .NET 6 and newer.

src/cswinrt/strings/WinRT.cs
Sergio0694: .NET also uses `sbyte*` in several places, eg. in the `string(sbyte*)` constructor. And `sbyte*` is just more correct given that Windows treats `char` as signed by default (cc. @tannergooding). I mean practically speaking I agree it doesn't make much difference here, but FWIW I'd use `sbyte*` for correctness.

manodasanW (approved):

manodasanW:
We will need to continue supporting .NET 6 as it is an LTS.  But I am exploring what it means to have both .NET 6 and .NET 8 versions of the Windows SDK projection.  With that said, given we are a targeting pack, it is looking like it is going to require .NET SDK changes first before that can be done so I won't take a dependency on that for now.

PR #1357, opened 2023-09-13
Return the actual type as it makes more sense and avoids an interface dispatch.

jlaanstra:
Should result in a small perf improvement.


jlaanstra:
This causes a build break when the underlying type has a property named Instance.

Issue #1112, opened 2022-02-22
COMException (0x800401FD): Object is not connected to server (CO_E_OBJNOTCONNECTED)


jbe2277:
**Describe the bug**
I get a `COMException (0x800401FD): Object is not connected to server (CO_E_OBJNOTCONNECTED)` when calling `Windows.Storage.Search.QueryOptions..ctor` after application runs about 30 seconds.

When `Windows.Storage.Search.QueryOptions..ctor` is called immediately after the application was started then it works fine.

**To Reproduce**
1. Build the following source with VS2022: https://github.com/jbe2277/musicmanager/releases/tag/v3.0.0
2. Start the application
3. Use the file navigation immediately after application was started -> works fine
4. Wait about 30 seconds
5. Use the file navigation -> the following exception occurs:

```
System.Runtime.InteropServices.COMException (0x800401FD): Objekt ist nicht mit dem Server verbunden. (0x800401FD (CO_E_OBJNOTCONNECTED))
   at WinRT.ExceptionHelpers.<ThrowExceptionForHR>g__Throw|20_0(Int32 hr)
   at Windows.Storage.Search.QueryOptions._IQueryOptionsFactory.CreateCommonFileQuery(CommonFileQuery query, IEnumerable`1 fileTypeFilter)
   at Windows.Storage.Search.QueryOptions.<>c__DisplayClass16_0.<.ctor>b__0()
   at Windows.Storage.Search.QueryOptions..ctor(CommonFileQuery query, IEnumerable`1 fileTypeFilter)
   at Waf.MusicManager.Presentation.Services.FileService.GetFiles(String directory, Boolean deep, String userSearchFilter, String applicationSearchFilter, CancellationToken cancellation)
   at Waf.MusicManager.Applications.Controllers.ManagerController.UpdateMusicFiles(Boolean deep)
   at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__128_0(Object state)
   at System.Windows.Threading.ExceptionWrapper.InternalRealCall(Delegate callback, Object args, Int32 numArgs)
   at System.Windows.Threading.ExceptionWrapper.TryCatchWhen(Object source, Delegate callback, Object args, Int32 numArgs, Delegate catchHandler)
   at System.Windows.Threading.DispatcherOperation.InvokeImpl()
   at MS.Internal.CulturePreservingExecutionContext.CallbackWrapper(Object obj)
[...]

PR #1359, opened 2023-09-13

jlaanstra:
Cache factories and statics per context, similar to coreclr/.NET Native behavior.

Fixes https://github.com/microsoft/CsWinRT/issues/1112

### DIFF
--- a/src/WinRT.Runtime/Context.cs
+++ b/src/WinRT.Runtime/Context.cs
@@ -7,11 +7,8 @@ using WinRT.Interop;
 
 namespace WinRT
 {
-    static class Context
+    static partial class Context
     {
-        [DllImport("api-ms-win-core-com-l1-1-0.dll")]
-        private static extern unsafe int CoGetContextToken(IntPtr* contextToken);
-
         [DllImport("api-ms-win-core-com-l1-1-0.dll")]
         private static extern int CoGetObjectContext(ref Guid riid, out IntPtr ppv);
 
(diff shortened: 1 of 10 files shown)


src/cswinrt/strings/WinRT.cs
```diff
@@ -501,33 +538,158 @@ public unsafe ObjectReference<I> _ActivateInstance<I>()
...
-#if NET
-        [UnconditionalSuppressMessage("ReflectionAnalysis", "IL2091:RequiresUnreferencedCode",
-            Justification = "No members of the generic type are dynamically accessed in this code path.")]
-#endif
-        public ObjectReference<I> _As<I>() => _IActivationFactory.As<I>();
-        public IObjectReference _As(Guid iid) => _IActivationFactory.As<WinRT.Interop.IUnknownVftbl>(iid);
     }
 
-    internal sealed class ActivationFactory<T> : BaseActivationFactory
+    internal class Factory
```
jlaanstra, line 543: Should be able to remove this.

src/WinRT.Runtime/Context.cs
```diff
@@ -7,11 +7,8 @@
...
 namespace WinRT
 {
-    static class Context
+    static partial class Context
     {
-        [DllImport("api-ms-win-core-com-l1-1-0.dll")]
-        private static extern unsafe int CoGetContextToken(IntPtr* contextToken);
-
         [DllImport("api-ms-win-core-com-l1-1-0.dll")]
         private static extern int CoGetObjectContext(ref Guid riid, out IntPtr ppv);
```
Sergio0694, line 13: Can we also fix this signature and drop the runtime marshalling while we're at it? 🙂

PR #1364, opened 2023-09-26
Remove the need for GuidGenerator/IIDOptimizer for all factory and statics interfaces.

jlaanstra:
Scope Guid change down to static and factory interfaces.

### DIFF
--- a/src/cswinrt/code_writers.h
+++ b/src/cswinrt/code_writers.h
@@ -5995,7 +5995,7 @@ public static Guid PIID = Vftbl.PIID;
 
         w.write(R"(% static class %
 {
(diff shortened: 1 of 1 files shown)


src/cswinrt/code_writers.h
```diff
@@ -5964,25 +5995,28 @@ public static Guid PIID = Vftbl.PIID;
 
         w.write(R"(% static class %
 {
+public static global::System.Guid IID { get; } = new Guid(new global::System.ReadOnlySpan<byte>(new byte[] { % }));
```
manodasanW, line 5998: We should declare this internal for now so that that it is not an API surface change for non-exclusive interfaces.  And in your other PR, you can make it public.

jlaanstra: Done.

manodasanW (approved):

PR #1367, opened 2023-10-14

jlaanstra:
Alternative implementation that maintains static classes compared to #1359

### DIFF
--- a/src/cswinrt/code_writers.h
+++ b/src/cswinrt/code_writers.h
@@ -1851,30 +1851,15 @@ remove => %;
 
             auto statics_type_name = staticsType.TypeName();
             w.write(R"(
-private static volatile % __%;
-private static % Make__%()
-{
-    global::System.Threading.Interlocked.CompareExchange(ref __%, new %("%", "%.%", %.IID), null);
-    return __%;
-}
-private static % _% => __% ?? Make__%();
-
+private static % _% = new %("%", "%.%", %.IID);
 )",
                 factory_class_name,
-                statics_type_name,
-                factory_class_name,
-                statics_type_name,
                 statics_type_name,
                 factory_class_name,
                 classType.TypeNamespace(),
                 classType.TypeNamespace(),
                 classType.TypeName(),
-                bind<write_type_name>(staticsType, typedef_name_type::StaticAbiClass, true),
-                statics_type_name,
-                factory_class_name,
-                statics_type_name,
-                statics_type_name,
-                statics_type_name);
+                bind<write_type_name>(staticsType, typedef_name_type::StaticAbiClass, true));
         }
     }
 
@@ -1882,25 +1867,12 @@ private static % _% => __% ?? Make__%();
     {
         auto class_type_name = classType.TypeName();
         w.write(R"(
-private static volatile BaseActivationFactory __%Factory;
-private static BaseActivationFactory Make__%Factory()
-{
-    global::System.Threading.Interlocked.CompareExchange(ref __%Factory, new BaseActivationFactory("%", "%.%"), null);
-    return __%Factory;
-}
-private static BaseActivationFactory _%Factory => __%Factory ?? Make__%Factory();
-
+private static BaseActivationFactory _%Factory = new BaseActivationFactory("%", "%.%");
 )",
             class_type_name,
-            class_type_name,
-            class_type_name,
             classType.TypeNamespace(),
             classType.TypeNamespace(),
-            classType.TypeName(),
-            class_type_name,
-            class_type_name,
-            class_type_name,
-            class_type_name);
+            classType.TypeName());
     }
 
     template<auto method_writer>
--- a/src/cswinrt/code_writers.h
+++ b/src/cswinrt/code_writers.h
@@ -1854,7 +1854,7 @@ remove => %;
 private static volatile % __%;
 private static % Make__%()
 {
-    global::System.Threading.Interlocked.CompareExchange(ref __%, new %(typeof(%.%), %.IID), null);
+    global::System.Threading.Interlocked.CompareExchange(ref __%, new %("%", "%.%", %.IID), null);
     return __%;
 }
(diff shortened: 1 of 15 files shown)


src/cswinrt/strings/WinRT.cs
```diff
@@ -501,33 +547,144 @@ public unsafe ObjectReference<I> _ActivateInstance<I>()
...
+#if NET
+    internal class BaseFactory
+    {
+        private IObjectReference _factory;
+        public IObjectReference Value { get => _factory; }
+#else
+    internal class BaseFactory<I>
     {
-        public ActivationFactory() : base(typeof(T).Namespace, typeof(T).FullName) { }
+        private ObjectReference<I> _factory;
```
Sergio0694, line 587: Shouldn't this also be volatile? It's being written to concurrently with volatile writes in that `cmpxgh` below.
```suggestion
        private volatile ObjectReference<I> _factory;
```

Sergio0694 (commented):
Just leaving a couple more comments 😄

src/cswinrt/strings/WinRT.cs
```diff
@@ -448,31 +491,65 @@ public static unsafe (ObjectReference<IActivationFactoryVftbl> obj, int hr) GetA
...
             Marshal.ThrowExceptionForHR(Platform.CoDecrementMTAUsage(_mtaCookie));
         }
-    }
-
+    }
+
     internal class BaseActivationFactory
     {
-        private readonly ObjectReference<IActivationFactoryVftbl> _IActivationFactory;
+        private Tuple<ObjectReference<IActivationFactoryVftbl>, IntPtr> _IActivationFactory;
```
Sergio0694, line 498: This should still be marked as `volatile`, no? You're reading it here concurrently, but writing with an interlocked op.

src/cswinrt/strings/WinRT.cs
```diff
@@ -502,32 +580,117 @@ public unsafe ObjectReference<I> _ActivateInstance<I>()
...
 #endif
-            I>() => _factory._ActivateInstance<I>();
+            I>() => _instance._ActivateInstance<I>();
+    }
+
+#if NET
+    internal class BaseFactory
+    {
+
+        private Tuple<IObjectReference, IntPtr> _factory;
```
Sergio0694, line 611: Same here for this, shouldn't it also be `volatile`?

src/cswinrt/strings/WinRT.cs
```diff
@@ -502,32 +580,117 @@ public unsafe ObjectReference<I> _ActivateInstance<I>()
...
+        public static ObjectReference<I> As<I>() => _instance._As<I>();
+        public static IObjectReference As(Guid iid) => _instance._As(iid);
 
-#if NET
-        [UnconditionalSuppressMessage("ReflectionAnalysis", "IL2091:RequiresUnreferencedCode", 
-            Justification = "No members of the generic type are dynamically accessed in this code path.")]
-#endif
         public static ObjectReference<I> ActivateInstance<
 #if NET
             [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.None)]
```
Sergio0694, line 602: Is this actually needed? Isn't `I` already not annotated for any dynamic access in this context? 🤔

src/cswinrt/strings/WinRT.cs
```diff
@@ -491,7 +568,8 @@ public BaseActivationFactory(string typeNamespace, string typeFullName)
         public unsafe ObjectReference<I> _ActivateInstance<I>()
```
Sergio0694, line 568: Also double checking: I see you removed the `[UnconditionalSuppressMessage]` below. Is this one still needed or can it also be removed after the other changes you did?

Sergio0694 (commented):
Left a few comments 🙂
Also, Q: I see you removed `INativeGuid`, is that just because it'll all be done in another PR?

src/cswinrt/strings/WinRT.cs
```diff
@@ -448,31 +491,65 @@ public static unsafe (ObjectReference<IActivationFactoryVftbl> obj, int hr) GetA
...
+            get
+            {
+                var existingInstance = _IActivationFactory;
+                if (existingInstance != null && (existingInstance.Item2 == IntPtr.Zero || existingInstance.Item2 == Context.GetContextToken()))
+                {
+                    return existingInstance.Item1;
+                }
+
+                var newFactory = InitializeFactory();
+                _IActivationFactory = newFactory;
```
Sergio0694, line 511: Q: two racing threads might end up returning different instances of the activation factory. Is that a concern?

jlaanstra: That is not a concern. Caching it is opportunistic.

src/cswinrt/code_writers.h
```diff
@@ -1805,87 +1808,89 @@ remove => %;
...
-                cache_vftbl_type);
+            auto statics_type_name = staticsType.TypeName();
             w.write(R"(
-internal sealed class _% : ABI.%.%
-{
-public _%() : base(%()) { }
-%
-internal static % Instance => _instance;
-}
+private static % _% = new %("%.%", %.IID);
```
Sergio0694, line 1854: Is this ever reassigned at runtime? If not, shouldn't this be `readonly` to improve codegen?

src/cswinrt/code_writers.h
```diff
@@ -1805,87 +1808,89 @@ remove => %;
...
+    }
 
-%
-internal static % Instance => (%)_instance;
-}
+    void write_activation_factory_static_cache_definition(writer& w, TypeDef const& classType)
+    {
+        auto class_type_name = classType.TypeName();
+        w.write(R"(
+private static BaseActivationFactory _%Factory = new BaseActivationFactory("%.%");
```
Sergio0694, line 1869: Same for this one, shouldn't it be `readonly`, or could this ever be reassigned?

PR #1375, opened 2023-11-05

jlaanstra:
Alternative implementation compared to https://github.com/microsoft/CsWinRT/pull/1367  that uses a custom IObjectReference type to track context.


src/cswinrt/strings/WinRT.cs
```diff
@@ -450,48 +453,75 @@ public static unsafe (ObjectReference<IActivationFactoryVftbl> obj, int hr) GetA
...
-            Justification = "No members of the generic type are dynamically accessed in this code path.")]
-#endif
-        public unsafe ObjectReference<I> _ActivateInstance<I>()
+        public bool IsObjectInContext()
+        {
+            return _contextToken == IntPtr.Zero || _contextToken == Context.GetContextToken();
+        }
+
+        // If we are free threaded, we do not need to keep track of context.
+        // This can either be if the object implements IAgileObject or the free threaded marshaler.
```
manodasanW, line 507: I assume we are not checking the free threaded marshaler as it is internal to WinRT.Runtime.  We can probably address this in the future, but we should update the comment to reflect that.

src/cswinrt/strings/WinRT.cs
```diff
@@ -450,48 +453,75 @@ public static unsafe (ObjectReference<IActivationFactoryVftbl> obj, int hr) GetA
...
+        }
+    }
+
+    internal static class IActivationFactoryMethods
+    {
+        public static unsafe ObjectReference<I> ActivateInstance<I>(IObjectReference obj)
         {
             IntPtr instancePtr;
-            Marshal.ThrowExceptionForHR(_IActivationFactory.Vftbl.ActivateInstance(_IActivationFactory.ThisPtr, &instancePtr));
+            global::WinRT.ExceptionHelpers.ThrowExceptionForHR((*(delegate* unmanaged[Stdcall]<IntPtr, out IntPtr, int>**)obj.ThisPtr)[6](obj.ThisPtr, out instancePtr));
```
manodasanW, line 524: Can we make this function pointer signature blittable (out IntPtr -> IntPtr*)

src/cswinrt/strings/WinRT.cs
```diff
@@ -500,34 +532,100 @@ public unsafe ObjectReference<I> _ActivateInstance<I>()
...
+            {
 #if NET
-            [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.None)]
-#endif
-            I>() => _factory._ActivateInstance<I>();
+                return factory;
+#else
+                using (factory)
+                {
+                    return factory.As<I>(iid);
```
manodasanW, line 592: Is this else clause actually needed?  Given you are passing IID to GetActivationFactory and using the same I, does this do something more?

manodasanW (approved):

Issue #1369, opened 2023-10-22
WinRTTypeWriter generator fails with NullReferenceException


Sergio0694:
**Describe the bug**

It seems there's a NRE bug in the "WinRTTypeWriter" generator, at least in some scenarios.
Might be fixed already with all the changes in the AOT branch, but figured I'd report just for tracking 🙂

**To Reproduce**

1. Checkout to https://github.com/Sergio0694/ComputeSharp/commit/79ff77b3a1dd10affc7ee6047b9935a98f85a96d
2. Build the `ComputeSharp.NativeLibrary.WinRT` project

It's failing for me both when building from VS and when trying to build with `msbuild` from a developer command prompt.

> CSC : error CS8785: Generator 'SourceGenerator' failed to generate source. It will not contribute to the output and compilation errors may occur as a result. Exception was of type 'NullReferenceException' with message 'Object reference not set to an instance of an object.'.

```
Error	CS8785	Generator 'SourceGenerator' failed to generate source. It will not contribute to the output and compilation errors may occur as a result. Exception was of type 'NullReferenceException' with message 'Object reference not set to an instance of an object.'.
System.NullReferenceException: Object reference not set to an instance of an object.
   at Generator.WinRTTypeWriter.QualifiedName(ISymbol symbol, Boolean includeGenerics)
   at Generator.WinRTTypeWriter.EncodeSymbol(Symbol symbol, SignatureTypeEncoder typeEncoder)
   at Generator.WinRTTypeWriter.EncodeParameters(Parameter[] parameters, ParametersEncoder parametersEncoder)
   at Generator.WinRTTypeWriter.<>c__DisplayClass65_0.<AddMethodReference>b__1(ParametersEncoder parametersEncoder)
   at System.Reflection.Metadata.Ecma335.MethodSignatureEncoder.Parameters(Int32 parameterCount, Action`1 returnType, Action`1 parameters)
   at Generator.WinRTTypeWriter.AddMethodReference(String name, Parameter[] parameters, Symbol returnSymbol, INamedTypeSymbol parentType, Boolean isStatic)
   at Generator.WinRTTypeWriter.AddMethodReference(IMethodSymbol method)
   at Generator.WinRTTypeWriter.AddProjectedType(INamedTypeSymbol type, String projectedTypeOverride)
   at Generator.WinRTTypeWriter.AddType(INamedTypeSymbol type, Boolean treatAsProjectedType)
[...]

Issue #1347, opened 2023-08-18
Building WinRT.Runtime results in hundreds of trim warnings


Sergio0694:
**Describe the bug**

WinRT.Runtime is enabled for trimming, but not all APIs are correctly annotated. There's quite a few that are missing annotations, causing hundreds of warnings when building the project, as well as a not ideal experience for consumers trying to use those unannotated APIs and not getting the right warnings surfaced to them in IntelliSense.

> [!NOTE]
> Partially related to #1319 (this is a superset of that)

**To Reproduce**

1. Clone the repo
2. Build WinRT.Runtime

**Expected behavior**

Builds with no warnings.

**Actual result**

Hundreds of trim warnings.

<details>
    <summary><b>Build output (click to expand):</b></summary>
<br/>

```
\Projections\Bindable.net5.cs(79,23,79,28): warning CA2257: The 'ABI.Microsoft.UI.Xaml.Interop.IBindableIterator.Vftbl' member on the 'ABI.Microsoft.UI.Xaml.Interop.IBindableIterator' type should be marked 'static' as 'ABI.Microsoft.UI.Xaml.Interop.IBindableIterator' has the 'DynamicInterfaceImplementationAttribute' applied
\Projections\Bindable.net5.cs(428,29,428,50): warning CA2257: The 'ABI.System.Collections.IEnumerable.AdaptiveFromAbiHelper' member on the 'ABI.System.Collections.IEnumerable' type should be marked 'static' as 'ABI.System.Collections.IEnumerable' has the 'DynamicInterfaceImplementationAttribute' applied
\Projections\Bindable.net5.cs(447,22,447,35): warning CA2257: The 'ABI.System.Collections.IEnumerable.FromAbiHelper' member on the 'ABI.System.Collections.IEnumerable' type should be marked 'static' as 'ABI.System.Collections.IEnumerable' has the 'DynamicInterfaceImplementationAttribute' applied
\Projections\Bindable.net5.cs(484,29,484,40): warning CA2257: The 'ABI.System.Collections.IEnumerable.ToAbiHelper' member on the 'ABI.System.Collections.IEnumerable' type should be marked 'static' as 'ABI.System.Collections.IEnumerable' has the 'DynamicInterfaceImplementationAttribute' applied
[...]

Issue #1318, opened 2023-04-16
CsWinRT runtime assembly and all projection assembly should disable built-in marshalling


Sergio0694:
## Summary

See https://learn.microsoft.com/en-us/dotnet/standard/native-interop/disabled-marshalling. CsWinRT's runtime assembly and all projection assemblies should use `[assembly: DisableRuntimeMarshalling]`. This would also ensure that no accidental marshalling could be triggered, as seen in the other issues around using managed signatures in function pointers and P/Invoke-s.

## Rationale

* Disabling built-in marshalling makes code more AOT-friendly
* It also ensures that built-in marshalling isn't accidentally used, which is slower

PR #1377, opened 2023-11-09
Addressing issues with generics on AOT

manodasanW:
This PR addresses a couple of different AOT compat issues related to generics:

- Refactors the AOT source generator to not have the `WinRTExposedType` attribute take an array of interfaces but instead take a type which then returns the list of interfaces to put on the vftbl.  This allows for the ability to do initialization of generic interfaces such as initializing the code generated version of the vftbl for the specific generic instantiation before returning the interface array.
- Updating `WindowsRuntimeTypeAttribute` to take the source generated guid signature for structs so that in the scenarios where we can't precompute the IID, we can still compute the IID without using reflection to iterate through all the struct fields.
- Making all the function pointer delegates blittable.
- Making authoring scenarios AOT and trimming compatible by having the generated authoring projection register a lookup table for the authoring metadata types which have the `WinRTExposedType` attribute for vftbl generation.
-   Refactoring the generic collection interfaces and the generated generic interfaces to be AOT compatible.  This is done through a couple of different ways.  First, both the Do_ABI functions and the RCW functions of these generic interfaces are source generated.  The source generated version mainly just handles the managed to ABI or the ABI to managed marshaling of the parameters and return and then calls a common function in the Methods static class for that interface to handle the actual implementation logic.  These source generated versions are registered by calling the Init function in the new Methods static class which takes the ABI type as also a generic.  These new static classes also have a fallback for projections which were not generated using an updated cswinrt version but the app they are used in happen to have an updated version of `WinRT.Runtime`.
- The source generated RCW functions are generated by the cswinrt tool itself based on the projections it is generating.  To avoid generating the same ones multiple times, they are generated at the end and are shared across all the functions in that projection.  This can mean that different projections have similar generated functions but the first one to register them will be the one that is used.
- The source generated Do_ABI vftable functions are generated by the source generator which analyzes the sources to determine the actual C# implemented types implementing WinRT interfaces that are passed to projected functions.  Given there is a fixed number of generic interfaces, these implementation are mostly hand authored with format specifiers for the marshaler to use based on the generic type. 
- Enabling all the functional tests to run for AOT.
- Added missing array marshal functions which should have been there before.
- Adds a new `CsWinRTAotOptimizerEnabled` property which is by default on.  It is mainly there if after we end discovering some scenarios where our source generator behaves incorrectly.  This will allow those folks to unblock themselves by disabling the optimizer especially if they aren't yet concerned about publishing for AOT or trimming.
- In of our previous perf changes, we removed the use of IDIC for our collection generic interfaces via the addition of the Methods static class and the Impl class.  This is now mostly expanded for the remaining custom mapped interfaces and also the code generated ones like IAsyncOperation.


src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
```diff
@@ -160,46 +254,512 @@ private static bool TryGetCompatibleWindowsRuntimeTypesForVariantType(INamedType
...
 
-                source.Append($$"""
-                    [global::WinRT.WinRTExposedType({{vtableEntries}})]
+                    source.AppendLine($$"""
+                    [global::WinRT.WinRTExposedType(typeof({{vtableAttribute.ClassName}}WinRTTypeDetails))]
                     partial class {{vtableAttribute.ClassName}}
                     {
                     }
+
+                    internal sealed class {{vtableAttribute.ClassName}}WinRTTypeDetails : global::WinRT.IWinRTExposedTypeDetails
```
Sergio0694, line 358: Question: right now we have this `WinRTExposedType`, which links to a separate generated type with the details, implementing the interface, which is constructed at runtime using reflection. Could we not do the same as `ComWrappers` is doing instead, that is, make `WinRTExposedType` non-sealed, and then have the generator create a derived type directly implementing the interface, and use that to annotate the projection types? This way CsWinRT can simply look that up and cast directly to the interface type, no need to use `Activator.CreateInstance` and reflection to get an instance to invoke methods on. Basically it'd just generate this:
```csharp
[FooWinRTExposedType]
partial class Foo
{
}

internal sealed class FooWinRTExposedType : WinRTExposedType, IWinRTExposedTypeDetails
{
    public ComWrappers.ComInterfaceEntry[] GetExposedInterfaces() => ...;
}
```
Then CsWinRT can just do `(IWinRTExposedTypeDetails?)type.GetCustomAttribute(typeof(WinRTExposedType))` and use that directly. Should also help startup a tiny bit since there's no more `Activator.CreateInstance` being called. And this also works downlevel just fine, as we don't even need any attribute generics.

What do you think? 🙂

Sergio0694, line 358: As a follow up, should consider checking whether C# 11 is enabled in the project (via `((CSharpCompilation)compilation).LanguageVersion`) and make all of these types `file` accessible instead of `internal`? For context, the `ComWrappers` COM generators also do the same, as these types are only needed for attributes over types and are never directly referenced anywhere else.

src/WinRT.Runtime/ComWrappersSupport.cs
```diff
@@ -23,16 +22,25 @@
...
+            KeyValuePairCCW.TryAdd(keyValuePairType, new ComInterfaceEntry { IID = iid, Vtable = abiToProjectionVftablePtr });
+        }
+    }
+
 #if EMBED
     internal
 #else
     public 
 #endif
     static partial class ComWrappersSupport
```
Sergio0694, line 40: I see a few calls to `GuidGenerator` in `ProvideIReferenceArray` below, for constructed generic types resolved over specific type arguments. Would it be possible in theory to just hardcode those GUIDs instead? 🤔

Sergio0694, line 40: Can't leave a comment for this in that file as it's not in the diff — `Context.CoGetObjectContext` still has a marshalling stub (has a `ref` and an `out` parameter), should we address that as part of this PR? Also related to #1318.

Sergio0694 (commented):
Leaving a few more notes in case it helps 🙂

Tried building this branch locally but I'm still getting about 490 warnings from WinRT.Runtime, most of them being incorrect or missing trim/AOT annotations (related to #1347). Is this expected, and would this just be fixed in a separate PR?

---

All this work is really great to see though, love CsWinRT getting some improvements! 🎉

src/WinRT.Runtime/ComWrappersSupport.cs
manodasanW: Addressed.

manodasanW: These probably can be hard coded.  Will address this in a future PR.

src/WinRT.Runtime/ComWrappersSupport.cs
```diff
@@ -168,72 +176,75 @@ internal static ObjectReference<T> GetObjectReferenceForInterface<T>(IntPtr exte
...
+            var winrtExposedClassAttribute = type.GetCustomAttribute<WinRTExposedTypeAttribute>(false);
+
+            // Handle scenario where it can be an authored type
+            // which means the attribute lives on the authoring metadata type.
+            if (winrtExposedClassAttribute == null)
             {
-                // Delegates have no interfaces that they implement, so adding default WinRT entries.
-                var helperType = type.FindHelperType();
-                if (helperType is object)
+                var authoringMetadaType = type.GetRuntimeClassCCWType();
```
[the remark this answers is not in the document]
manodasanW: This code path is indeed AOT and trimming safe due to with updated projections, the metadata type would be found in the registered lookup table.  So, it won't get to the unsafe part.  But I meant to refactor the unsafe part into a separate function and never did.  Addressed that by moving it to a separate function.

manodasanW:
> Leaving a few more notes in case it helps 🙂
> 
> Tried building this branch locally but I'm still getting about 490 warnings from WinRT.Runtime, most of them being incorrect or missing trim/AOT annotations (related to #1347). Is this expected, and would this just be fixed in a separate PR?
> 
> All this work is really great to see though, love CsWinRT getting some improvements! 🎉

I plan on going through all the remaining warnings and looking at if they are in fallback code paths or are not needed anymore or wrong annotations.  I wanted to first get the changes in that would address all known AOT and trimming compat issues.

src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
manodasanW: I did see the COM generators using file accessible but didn't think to dynamically support it by checking the lang version.  I will look at doing this in an upcoming PR.

manodasanW: This was one of my initial ideas during the refactor.  This might still be preferrable to do, will need to do some comparisons.  The main reason I switched to what I have was I was concerned about generating much more code for each struct, especially blittable ones, where we only generate the struct today and not any marshaling classes.  With the current approach, I was able to use a generic type that I passed to share the implementation for structs.  But after this PR is in, I am willing to try out the other approach again and see how much more it really adds to code size and if it is minimal.  Also I was thinking this approach may make it easier in the future to move to generic attributes when we can, but with that said, not sure which is better perf.

src/build.cmd
```diff
@@ -212,7 +254,7 @@ call :exec vstest.console.exe %this_dir%\Tests\ObjectLifetimeTests\bin\%cswinrt_
 if ErrorLevel 1 (
   echo.
   echo ERROR: Lifetime test failed, skipping NuGet pack
-  exit /b !ErrorLevel!
+  rem exit /b !ErrorLevel!
```
manodasanW, line 257: Undo comment.

vineeththomasalex (approved):

src/WinRT.Runtime/ComWrappersSupport.cs
```diff
@@ -168,72 +176,75 @@ internal static ObjectReference<T> GetObjectReferenceForInterface<T>(IntPtr exte
...
+            else if (!type.IsEnum && GetComInterfaceEntriesForTypeFromLookupTable(type) is var lookupTableEntries && lookupTableEntries != null)
             {
-                var objType = type.GetRuntimeClassCCWType() ?? type;
-                // Check whether the type itself has the attribute to make sure it is using the new source generator
-                // and just doesn't have an updated dependent projection which is.
-                if (type.GetCustomAttribute(typeof(WinRTExposedTypeAttribute), false) is IWinRTExposedTypeDetails winrtExposedClassAttribute)
+                hasWinrtExposedClassAttribute = true;
+                entries.AddRange(lookupTableEntries);
+            }
+            else if (RuntimeFeature.IsDynamicCodeCompiled)
```
dongle-the-gadget, line 223: Is it possible for this path to be-AOT compatible somehow? This change can cause some IIDs to be missing entirely, causing `CreateCCWForObjectForABI` to fail. One affected API is `StorageFile.GetFileFromPathAsync`.

Here's a screenshot of `AsyncOperationCompletedHandler<StorageFile>` in CoreCLR (left) versus NativeAOT (right) (provided by @hez2010):
Screenshot of two Windows Terminal panes. The GUIDs of `AsyncOperationCompletedHandler<StorageFile>` are printed in both CoreCLR and NativeAOT environments

manodasanW: This code path is meant to be the fallback.  It looks like the first guid is the one that is missing with the other extra ones being duplicates.  For this scenario, the WinRT AOT source generator should have generated the necessary vtable entries as part of the lookup table.  Do you see any warnings for `WinRTAotSourceGenerator` failing to run?  If not, are you able to check if there is a generated file called WinRTGlobalVtableLookup.g.cs and inside of it, do you see a reference to the mentioned type?

hez2010: Those missing IIDs are delegates, such as `AsyncOperationCompletedHandler<T>`.

hez2010: I checked the `WinRTGlobalVtableLookup.g.cs`, it did exist but I didn't find the code generated for `Windows.Foundation.AsyncOperationCompletedHandler<Windows.Storage.StorageFile>`.

manodasanW: I added a test case to test this scenario and I discovered a different issue, but not this specific one you are running into.  I think I know why you are seeing this.  Today, you typically add a CsWinRT package reference to generate a projection for a set of WinMD files or for authoring scenarios.  With these changes, you also need to add a CsWinRT package reference to make your library or app which consumes CsWinRT projected APIs `AOT compatible`.  This is because we have a new source generator that we are using to generate some code in order to be AOT compatible specifically in the generic type scenarios.  In the future we may look for over ways to have the source generator run without needing a CsWinRT package reference such as when using the net windows10 TFM.

Can you confirm whether your app or library where this `AsyncOperationCompletedHandler` is being added has a CsWinRT package reference.  If not, can you add one and see if that helps.  I assume you are testing this by manually building my changes, so you might not have a CsWinRT package built.  If so, you might be able to directly add a reference to the source generator as we do in our functional tests [here](https://github.com/microsoft/CsWinRT/blob/2a7b7548b237597e32de9dcfcac0927bb6fc6ff3/src/Tests/FunctionalTests/Async/Async.csproj#L12).  Let me know if that addresses it.

manodasanW: I didn't see your last comment when I responded.  Was the `WinRTGlobalVtableLookup.g.cs` file you found for the projection or for your library / app.

hez2010: It's for the projection.

hez2010: Okay now I tested with the source generator, but seems that it's still failing as `OpenStreamForReadAsync` is not a method from WinRT types, so nothing was generated for ``Windows.Foundation.AsyncOperationCompletedHandler`1[Windows.Storage.Streams.IRandomAccessStream]`` as this type is used by `OpenStreamForReadAsync` which is not available as source code.

```cs
var file = await StorageFile.GetFileFromPathAsync(...);
using var stream = await file.OpenStreamForReadAsync();
using var sw = new StreamReader(stream);
Console.WriteLine(sw.ReadToEnd());
```

```
Unhandled Exception: System.InvalidCastException: Specified cast is not valid.
   at WinRT.ComWrappersSupport.CreateCCWForObjectForABI(Object, Guid) + 0xb0
   at WinRT.MarshalDelegate.CreateMarshaler2(Object, Guid, Boolean) + 0x76
   at ABI.Windows.Foundation.AsyncOperationCompletedHandler`1.CreateMarshaler2(AsyncOperationCompletedHandler`1 managedDelegate) + 0x4d
   at WinRT.GenericTypeInstantiations.Windows_Foundation_IAsyncOperation_1_Windows_Storage_Streams_IRandomAccessStream.put_Completed(IObjectReference, AsyncOperationCompletedHandler`1) + 0x63
   at System.WindowsRuntimeSystemExtensions.AsTask[TResult](IAsyncOperation`1, CancellationToken) + 0x130
   at System.IO.WindowsRuntimeStorageExtensions.<OpenStreamForReadAsyncCore>d__1.MoveNext() + 0x49
```

The generated code by the source generator is:

```cs
namespace WinRT.GenericHelpers
{

// Windows.Foundation.AsyncOperationCompletedHandler<global::Windows.Storage.StorageFile>
internal static class AsyncOperationCompletedHandler_Windows_Storage_StorageFile
{
    private static bool _initialized = Init();
    internal static bool Initialized => _initialized;

    private static unsafe bool Init()
    {
        return global::ABI.Windows.Foundation.AsyncOperationCompletedHandlerMethods<Windows.Storage.StorageFile, IntPtr>.InitCcw(
           &Do_Abi_Invoke
        );
    }

    [UnmanagedCallersOnly(CallConvs = new[] { typeof(CallConvStdcall) })]
    private static unsafe int Do_Abi_Invoke(IntPtr thisPtr, IntPtr asyncInfo, global::Windows.Foundation.AsyncStatus asyncStatus)
    {
        try
        {
            global::ABI.Windows.Foundation.AsyncOperationCompletedHandlerMethods<Windows.Storage.StorageFile, IntPtr>.Abi_Invoke(thisPtr, MarshalInterface<global::Windows.Foundation.IAsyncOperation<Windows.Storage.StorageFile>>.FromAbi(asyncInfo), asyncStatus);
        }
        catch (global::System.Exception __exception__)
        {
            global::WinRT.ExceptionHelpers.SetErrorInfo(__exception__);
            return global::WinRT.ExceptionHelpers.GetHRForException(__exception__);
        }
        return 0;
    }
}
}

namespace WinRT.GenericHelpers
{

    internal static class GlobalVtableLookup
    {

        [System.Runtime.CompilerServices.ModuleInitializer]
        internal static void InitializeGlobalVtableLookup()
        {
            ComWrappersSupport.RegisterTypeComInterfaceEntriesLookup(LookupVtableEntries);
        }

        private static ComWrappers.ComInterfaceEntry[] LookupVtableEntries(Type type)
        {
            if (type == typeof(global::Windows.Foundation.AsyncOperationCompletedHandler<global::Windows.Storage.StorageFile>))
            {
                        _ = global::WinRT.GenericHelpers.AsyncOperationCompletedHandler_Windows_Storage_StorageFile.Initialized;

        var delegateInterface = new global::System.Runtime.InteropServices.ComWrappers.ComInterfaceEntry
        {
            IID = global::WinRT.GuidGenerator.GetIID(typeof(global::Windows.Foundation.AsyncOperationCompletedHandler<global::Windows.Storage.StorageFile>).GetHelperType()),
            Vtable = global::ABI.Windows.Foundation.AsyncOperationCompletedHandler<global::Windows.Storage.StorageFile>.AbiToProjectionVftablePtr
        };

        return global::WinRT.DelegateTypeDetails<Windows.Foundation.AsyncOperationCompletedHandler<global::Windows.Storage.StorageFile>>.GetExposedInterfaces(delegateInterface);

            }
            return default;
        }
    }
}

```

hez2010: Also note that the source generator doesn't work for code like:

```cs
var file = await StorageFile.GetFileFromPathAsync(...);
using var stream = await file.OpenAsync(FileAccessMode.Read).AsTask().ConfigureAwait(continueOnCapturedContext: false);
using var sw = new StreamReader(stream.AsStreamForRead());
Console.WriteLine(sw.ReadToEnd());
```

You have to write it like the below to make the source generator work:

```cs
var file = await StorageFile.GetFileFromPathAsync(...);
using var stream = await file.OpenAsync(FileAccessMode.Read);
using var sw = new StreamReader(stream.AsStreamForRead());
Console.WriteLine(sw.ReadToEnd());
```

manodasanW: Thanks for confirming.  `OpenStreamForReadAsync` should be coming from [here](https://github.com/microsoft/CsWinRT/blob/71169932920500310921b4c3fadba1e5a99c6138/src/cswinrt/strings/additions/Windows.Storage/WindowsRuntimeStorageExtensions.cs#L28).  I will need to enlighten the source generator to handle this scenario.  I will also look into the other scenario you mentioned which it doesn't handle.

hez2010: I think you can generate code for async operations at the invocation instead of the await expression:

i.e. adding the below code under `if (invocationSymbol is IMethodSymbol methodSymbol && GeneratorHelper.IsWinRTType(methodSymbol.ContainingSymbol))` in `GetVtableAttributesToAddOnLookupTable`:
```cs

                    if (GeneratorHelper.IsWinRTType(methodSymbol.ReturnType))
                    {
                        // if (!System.Diagnostics.Debugger.IsAttached) System.Diagnostics.Debugger.Launch();
                        var completedProperty = methodSymbol.ReturnType.GetMembers("Completed").FirstOrDefault() as IPropertySymbol;
                        if (completedProperty?.Type.MetadataName.Contains("`") ?? false)
                        {
                            vtableAttributes.Add(GetVtableAttributeToAdd(completedProperty.Type, GeneratorHelper.IsWinRTType, context.SemanticModel.Compilation.Assembly, false));
                        }
                    }
```

hez2010: I created a PR for you to resolve the above issue: https://github.com/microsoft/CsWinRT/pull/1383

src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
```diff
@@ -130,6 +130,7 @@ class TypeDeclaration
         public string DefaultInterface;
         public string StaticInterface;
         public bool IsSynthesizedInterface;
+        public bool IsComponentType;
```
Sergio0694, line 133: Doesn't have to be fixed in this PR, but just for additional info — I've tried to use the CsWinRT package from this PR (as of https://github.com/microsoft/CsWinRT/pull/1377/commits/2a7b7548b237597e32de9dcfcac0927bb6fc6ff3) and can confirm that https://github.com/microsoft/CsWinRT/issues/1369 still happens (same repro project).

hez2010:
Hit another exception "Write lock may not be acquired with read lock held." 
At WinRT.ComWrappersSupport.RegisterTypeComInterfaceEntriesLookup.

manodasanW:
> Hit another exception "Write lock may not be acquired with read lock held." At WinRT.ComWrappersSupport.RegisterTypeComInterfaceEntriesLookup. This exception will happen when you try to call WASDK. For example, in a WinUI 3 app.

I did also run into this during my validation, working on a fix.

src/WinRT.Runtime/ComWrappersSupport.cs
Sergio0694: Partially unrelated, but I'm very confused by this line in the generated code above:
```csharp
IID = global::WinRT.GuidGenerator.GetIID(typeof(global::Windows.Foundation.AsyncOperationCompletedHandler<global::Windows.Storage.StorageFile>).GetHelperType())
```
Given this is all generated code, can't the generator precompute that IID into some static field as usual (like all new AOT-friendly projections are also doing in this PR) and just use that here, rather than using `GuidGenerator`?

manodasanW: I left this for now as I have plans to try to remove the delegates from the lookup table if possible and move it into the projection itself similar to the rcw instantiations.

src/WinRT.Runtime/Interop/IActivationFactory.cs
```diff
@@ -24,9 +24,25 @@ interface IActivationFactory
...
+#else
+    public
+#endif
+    static class IActivationFactoryMethods
+    {
+#if NET
+        public static global::System.Guid IID { get; } = new Guid(new global::System.ReadOnlySpan<byte>(new byte[] { 0x35, 0, 0, 0, 0, 0, 0, 0, 0xC0, 0, 0, 0, 0, 0, 0, 0x46 }));
+#else
+        public static global::System.Guid IID { get; } = new(0x00000035, 0, 0, 0xC0, 0, 0, 0, 0, 0, 0, 0x46);
+#endif
```
jlaanstra, line 38: I did something similar in InterfaceIIDs. Do we want to remove those?

manodasanW: Possibly, will handle cleanup in separate PR.

PR #1380, opened 2023-11-11
Remove Resurrect logic as it shouldn't be needed for .NET 6 and newer.

jlaanstra:
After talking with @manodasanW we should no longer need Resurrect on .NET 6 and newer.

### DIFF
--- a/src/WinRT.Runtime/ObjectReference.cs
+++ b/src/WinRT.Runtime/ObjectReference.cs
@@ -247,12 +247,9 @@ namespace WinRT
 
         protected void ThrowIfDisposed()
         {
-            if (disposed)
+            lock (_disposedLock)
             {
-                lock (_disposedLock)
-                {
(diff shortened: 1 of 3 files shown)


src/WinRT.Runtime/ObjectReference.cs
Sergio0694, line 1: Might probably want to also cleanup `ObjectReference.ThrowIfDispose()` (line 248). That double check is no longer necessary, because nobody will ever un-set `disposed` now that the resurrection logic is gone.

Sergio0694 (changes requested):

Sergio0694 (approved):
LGTM! :shipit:

manodasanW:
Confirmed using both .NET 6 and .NET 8 with the old repro app that this change runs without issue.

manodasanW (approved):

PR #1381, opened 2023-11-11
Cache StringBuilder per thread and remove Lazy from GetRuntimeClassName.

jlaanstra:
Remove uses of Lazy in InspectableInfo and uses a StringBuilder cache per thread to reduce memory waste for GetRuntimeClassName.

### DIFF
--- a/src/WinRT.Runtime/ComWrappersSupport.cs
+++ b/src/WinRT.Runtime/ComWrappersSupport.cs
@@ -12,6 +12,7 @@ using System.Linq.Expressions;
 using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
+using System.Threading;
 using WinRT.Interop;
 
 #if NET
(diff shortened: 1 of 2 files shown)


src/WinRT.Runtime/TypeNameSupport.cs
```diff
@@ -298,19 +298,29 @@ struct VisitedType
...
         public static string GetNameForType(Type type, TypeNameGenerationFlags flags)
         {
             if (type is null)
             {
                 return string.Empty;
             }
-            StringBuilder nameBuilder = new StringBuilder();
+
+            // Get instance for this thread
+            StringBuilder? nameBuilder = nameForTypeBuilderInstance ?? new StringBuilder();
```
Sergio0694, line 314: Can this not just use `??=` so it doesn't need those two assignments below?

manodasanW (approved):

src/WinRT.Runtime/ComWrappersSupport.cs
```diff
@@ -871,14 +872,23 @@ private static ComInterfaceEntry ProvideIReferenceArray(Type arrayType)
...
+
+            private volatile string _runtimeClassName;
+
+            private string MakeRuntimeClassName()
+            {
+                global::System.Threading.Interlocked.CompareExchange(ref _runtimeClassName, TypeNameSupport.GetNameForType(_type, TypeNameGenerationFlags.GenerateBoxedName | TypeNameGenerationFlags.ForGetRuntimeClassName), null);
+                return _runtimeClassName;
+            }
+
+            public string RuntimeClassName => _runtimeClassName ?? MakeRuntimeClassName();
```
Sergio0694, line 887: Nit: can simplify this by just moving the code for `MakeRuntimeClassName` in the getter (using a body), rather than needing the separate method there, which is only ever used here anyway?

jlaanstra: This is the pattern followed elsewhere so sticking with that.

Sergio0694: But it's... Not a great pattern? And it's also wasteful (more IL size and extra JIT work for no reason), not to mention unnecessarily verbose. We should try to improve the codebase as we find things, rather than just sticking to existing code patterns that are not good in the first place. Of course, this can be a follow up 🙂

jlaanstra: I'm not saying that we shouldn't change it but I'd rather change it everywhere at the same time :).

PR #1383, opened 2023-11-12
Handle async calls correctly in AotOptimizer

hez2010:
Instead of handling async calls from the await expression, we need to handle it from the invocation expression.

Because an async call is not necessarily to be awaited directly.


manodasanW (approved):

manodasanW:
Thanks for fixing this.

PR #1390, opened 2023-11-18
Move ActivationFactory into WinRT.Runtime.dll

jlaanstra:
This enables additional optimizations that couldn't be done in #1375


src/WinRT.Runtime/ActivationFactory.cs
```diff
@@ -0,0 +1,272 @@
...
+                MarshalInspectable<object>.DisposeAbi(instancePtr);
+            }
+        }
+
+        ~DllModule()
+        {
+            System.Diagnostics.Debug.Assert(_CanUnloadNow == null || _CanUnloadNow() == 0); // S_OK
+            lock (_cache)
+            {
+                _cache.Remove(_fileName);
```
Sergio0694, line 131: How can this finalizer ever be triggered, if a loaded `DllModule` instance is stored in the `_cache` dictionary, which contains a strong reference to the `DllModule` itself? I don't see how any of those can ever be collected 🤔

src/WinRT.Runtime/ActivationFactory.cs
```diff
@@ -0,0 +1,272 @@
...
+            }
+            if ((_moduleHandle != IntPtr.Zero) && !Platform.FreeLibrary(_moduleHandle))
+            {
+                Marshal.ThrowExceptionForHR(Marshal.GetHRForLastWin32Error());
+            }
+        }
+    }
+
+    internal sealed class WinrtModule
```
Sergio0694, line 140: Nit: incorrect naming, should be `WinRTModule`?

src/WinRT.Runtime/ActivationFactory.cs
```diff
@@ -0,0 +1,272 @@
...
+using WinRT.Interop;
+
+namespace WinRT
+{
+    internal unsafe sealed class DllModule
+    {
+        readonly string _fileName;
+        readonly IntPtr _moduleHandle;
+        readonly delegate* unmanaged[Stdcall]<IntPtr, IntPtr*, int> _GetActivationFactory;
+        readonly delegate* unmanaged[Stdcall]<int> _CanUnloadNow; // TODO: Eventually periodically call
```
Sergio0694, line 18: Nit: don't skip the accessibility modifier, add `private`?

src/WinRT.Runtime/ActivationFactory.cs
```diff
@@ -0,0 +1,272 @@
...
+using System.Runtime.InteropServices;
+using WinRT.Interop;
+
+namespace WinRT
+{
+    internal unsafe sealed class DllModule
+    {
+        readonly string _fileName;
+        readonly IntPtr _moduleHandle;
+        readonly delegate* unmanaged[Stdcall]<IntPtr, IntPtr*, int> _GetActivationFactory;
```
Sergio0694, line 17: Nit: don't use `IntPtr` instead of actual pointers, when you are in fact dealing with pointers.
```suggestion
        readonly delegate* unmanaged[Stdcall]<void*, void**, int> _GetActivationFactory;

```

src/WinRT.Runtime/ActivationFactory.cs
```diff
@@ -0,0 +1,272 @@
...
+                else if (TryCreate(fileName, out module))
+                {
+                    _cache[fileName] = module;
+                    return true;
+                }
+                return false;
+            }
+        }
+
+        private static unsafe bool TryCreate(string fileName, out DllModule module)
```
Sergio0694, line 41: Nit: redundant `unsafe` modifier, it's already on the type declaration. Same for the other method below.

src/WinRT.Runtime/ActivationFactory.cs
```diff
@@ -0,0 +1,272 @@
...
+            {
+                Marshal.ThrowExceptionForHR(Marshal.GetHRForLastWin32Error());
+            }
+        }
+    }
+
+    internal sealed class WinrtModule
+    {
+        readonly IntPtr _mtaCookie;
+        volatile static WinrtModule _instance;
```
Sergio0694, line 143: Nit: [incorrect modifiers order](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0036), should be `static volatile`. Also, actually, add `private` as well (same in the other fields).

Sergio0694, line 143: Nit: move static fields before instance ones.

Sergio0694 (commented):
Left a few comments 🙂

src/WinRT.Runtime/ActivationFactory.cs
jlaanstra: This was copied from the projection generated code without doing cleanup. It's possible it isn't needed.

jlaanstra: @manodasanW?

manodasanW: I do believe this code path is unused on .NET 6+ as .NET doesn't run down all the finalizers when you exit an app.  But I have seen it get executed in our .NET Standard projection when running on .NET Framework which does.  The .NET GC can detect this circular reference and thereby I believe the finalizer would still run.  But we can probably run into unexpected results accessing the managed object in a finalizer as it probably has been collected.  But it is probably not finalized yet which is why we haven't heard folks running into issues from it.  For this change, I would leave this untouched and open a separate issue to track looking more into this separately as I do think there is some issues here from a perspective of .NET framework.

Sergio0694: > ".NET GC can detect this circular reference"

Wait, yes that is correct, but the `_cache` field is static here 😅

src/WinRT.Runtime/ActivationFactory.cs
```diff
@@ -0,0 +1,272 @@
...
+                    {
+                        return factory;
+                    }
+                }
+            }
+        }
+
+        public static ObjectReference<I> Get<
+#if NET
+        [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.NonPublicConstructors | DynamicallyAccessedMemberTypes.PublicFields)]
```
Sergio0694, line 233: Isn't `PublicFields` unnecessary here?
```suggestion
        [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.NonPublicConstructors)]
```

src/WinRT.Runtime/Context.cs
```diff
@@ -9,6 +9,16 @@ namespace WinRT
 {
     static partial class Context
```
Sergio0694, line 10: Can you explicitly add `internal` here, for clarity?

src/WinRT.Runtime/ActivationFactory.cs
```diff
@@ -0,0 +1,272 @@
...
+        }
+
+        ~WinrtModule()
+        {
+            Marshal.ThrowExceptionForHR(Platform.CoDecrementMTAUsage(_mtaCookie));
+        }
+    }
+
+#if EMBED
+    internal
```
Sergio0694, line 192: Perhaps move this to a standalone file, so we can better separate it? Since this type is public and other two are not

src/WinRT.Runtime/ActivationFactory.cs
```diff
@@ -0,0 +1,272 @@
...
+                    (factory, hr) = module.GetActivationFactory(typeName);
+                    if (factory != null)
+                    {
+                        return factory;
+                    }
+                }
+            }
+        }
+
+        public static ObjectReference<I> Get<
```
Sergio0694, line 231: Can you clarify where is this overload still needed? Is this only for netstandard?

Sergio0694 (changes requested):

Sergio0694:
Just a couple nits but I do love this PR, and the fact `ActivationFactory.Get(string)` is now public! 🎉

src/WinRT.Runtime/ActivationFactory.cs
Sergio0694: Don't think this should be resolved, I'm still not sure this is correct. That field really is static.

src/WinRT.Runtime/ActivationFactory.cs
jlaanstra: Keeping as is as we use IntPtr everywhere and this would result in much more changes unrelated to moving this into WinRT.Runtime.

src/WinRT.Runtime/ActivationFactory.cs
jlaanstra: Let's do cleanup separate from this PR.

src/WinRT.Runtime/ActivationFactory.cs
```diff
@@ -0,0 +1,284 @@
...
+    internal unsafe sealed class DllModule
+    {
+        private readonly string _fileName;
+        private readonly IntPtr _moduleHandle;
+        private readonly delegate* unmanaged[Stdcall]<IntPtr, IntPtr*, int> _GetActivationFactory;
+        private readonly delegate* unmanaged[Stdcall]<int> _CanUnloadNow; // TODO: Eventually periodically call
+
+        private static readonly string _currentModuleDirectory = AppContext.BaseDirectory;
+
+        private static Dictionary<string, DllModule> _cache = new Dictionary<string, DllModule>(StringComparer.Ordinal);
```
Sergio0694, line 22: Shouldn't this be readonly?

src/WinRT.Runtime/ActivationFactory.cs
```diff
@@ -0,0 +1,284 @@
...
+using System.Diagnostics.CodeAnalysis;
+using System.Reflection;
+using System.Runtime.InteropServices;
+using WinRT.Interop;
+
+namespace WinRT
+{
+    internal unsafe sealed class DllModule
+    {
+        private readonly string _fileName;
```
Sergio0694, line 15: Nit: move instance fields below static ones

src/WinRT.Runtime/ActivationFactory.cs
```diff
@@ -0,0 +1,284 @@
...
+        public static WinRTModule Instance => _instance ?? MakeWinRTModule();
+
+        public unsafe WinRTModule()
+        {
+            IntPtr mtaCookie;
+            Marshal.ThrowExceptionForHR(Platform.CoIncrementMTAUsage(&mtaCookie));
+            _mtaCookie = mtaCookie;
+        }
+
+        public static unsafe (ObjectReference<I> obj, int hr) GetActivationFactory<I>(string runtimeClassId, Guid iid)
```
Sergio0694, line 158: This needs `[DAM(NonPublicConstructors)]` on the `I` type parameter

src/WinRT.Runtime/ActivationFactory.cs
```diff
@@ -0,0 +1,284 @@
...
+
+        ~WinRTModule()
+        {
+            Marshal.ThrowExceptionForHR(Platform.CoDecrementMTAUsage(_mtaCookie));
+        }
+    }
+
+    internal static class IActivationFactoryMethods
+    {
+        public static unsafe ObjectReference<I> ActivateInstance<I>(IObjectReference obj)
```
Sergio0694, line 193: Same missing `[DAM]` here on the `I`

src/WinRT.Runtime/ActivationFactory.cs
```diff
@@ -0,0 +1,284 @@
...
+                return ComWrappersSupport.GetObjectReferenceForInterface<I>(instancePtr);
+            }
+            finally
+            {
+                MarshalInspectable<object>.DisposeAbi(instancePtr);
+            }
+        }
+    }
+
+    internal static class ActivationFactory
```
Sergio0694, line 208: Why is this internal if it's been added to the API baseline diff?

Sergio0694 (changes requested):
Done another review pass on the updated diff 🙂

src/WinRT.Runtime/ActivationFactory.cs
jlaanstra: They are very much related. Not much benefit to splitting I'd say.

src/WinRT.Runtime/ActivationFactory.cs
```diff
@@ -0,0 +1,285 @@
...
+                        return factory;
+                    }
+                }
+            }
+        }
+
+#if NET
+        public static IObjectReference Get(string typeName, Guid iid)
+#else
+        public static ObjectReference<I> Get<I>(string typeName, Guid iid)
```
Sergio0694, line 238: Note: this could potentially break if someone tried to run some NS2.0 projections against the .NET 6+ version of WinRT.Runtime, because the generic method would not exist there anymore. Do we care about that scenario?

Sergio0694: If so, I would recommend still leaving the non-generic overload, so the updated projections can use that, but also just leave the "old" generic overload as well for binary compatibility purposes. Maybe mark it obsolete too on .NET 6+.

jlaanstra: I'd imagine that would be problematic in other places as well?

Sergio0694: I brought that up because we've had the same issue in the Toolkit (we made the same mistake). You'd basically get code from old assemblies running against the new runtime throwing `MissingMethodException`. I'd recommend keeping both, but just making the generic one `[EditorBrowsable(Never)]` and also `[Obsolete]` on .NET 6+. If no new projection is runtime code is ever calling this one, it won't matter that it's there. And only NS2.0 will use it.

src/WinRT.Runtime/ActivationFactory.cs
```diff
@@ -0,0 +1,289 @@
...
+        public static WinRTModule Instance => _instance ?? MakeWinRTModule();
+
+        public unsafe WinRTModule()
+        {
+            IntPtr mtaCookie;
+            Marshal.ThrowExceptionForHR(Platform.CoIncrementMTAUsage(&mtaCookie));
+            _mtaCookie = mtaCookie;
+        }
+
+        public static unsafe (ObjectReference<I> obj, int hr) GetActivationFactory<
```
Sergio0694, line 158: Does .NET 6+ also need this to be generic?

src/WinRT.Runtime/ActivationFactory.cs
manodasanW: We don't support our .NET Standard projections on .NET 6+ as there are other things that will probably break like ComWrappers.  But I am fine with going with either approach.

Sergio0694: Oh, well if that's the case then we can also just save code and leave this as is, even better 😄

src/WinRT.Runtime/ActivationFactory.cs
Sergio0694: Discussed offline, resolving this.

Sergio0694 (approved):
SHIP IT! :shipit:

src/WinRT.Runtime/ActivationFactory.cs
```diff
@@ -0,0 +1,289 @@
...
+        private static readonly Dictionary<string, DllModule> _cache = new Dictionary<string, DllModule>(StringComparer.Ordinal);
+
+        private readonly string _fileName;
+        private readonly IntPtr _moduleHandle;
+        private readonly delegate* unmanaged[Stdcall]<IntPtr, IntPtr*, int> _GetActivationFactory;
+        private readonly delegate* unmanaged[Stdcall]<int> _CanUnloadNow; // TODO: Eventually periodically call
+
+        public static bool TryLoad(string fileName, out DllModule module)
+        {
+            lock (_cache)
```
manodasanW, line 26: It is now more likely that we see parallel calls here due to different threads activating different types.  Given these are already cached too, it shouldn't have too much of an impact, but will need to monitor perf traces to see if this shows up and if we need to improve the locking here.  But for now, this is fine.

manodasanW (approved):

PR #1391, opened 2023-11-18
Expose api's for Context on ObjectReference.

jlaanstra:
This enables extra optimizations for factories as part of #1390 and should also allow us to add a Detach method.

This requires #1380 to go in first.

### DIFF
--- a/src/WinRT.Runtime/ComWrappersSupport.cs
+++ b/src/WinRT.Runtime/ComWrappersSupport.cs
@@ -82,25 +82,23 @@ namespace WinRT
                 Marshal.Release(agilePtr);
                 return true;
             }
-            else
+
+            if (Marshal.QueryInterface(iUnknown, ref Unsafe.AsRef(InterfaceIIDs.IMarshal_IID), out var marshalPtr) >= 0)
             {
-                if (Marshal.QueryInterface(iUnknown, ref Unsafe.AsRef(InterfaceIIDs.IMarshal_IID), out var marshalPtr) >= 0)
+                try
                 {
-                    try
+                    Guid iid_IUnknown = InterfaceIIDs.IUnknown_IID;
+                    Guid iid_unmarshalClass;
+                    Marshal.ThrowExceptionForHR((**(ABI.WinRT.Interop.IMarshal.Vftbl**)marshalPtr).GetUnmarshalClass_0(
+                        marshalPtr, &iid_IUnknown, IntPtr.Zero, MSHCTX.InProc, IntPtr.Zero, MSHLFLAGS.Normal, &iid_unmarshalClass));
+                    if (iid_unmarshalClass == ABI.WinRT.Interop.IMarshal.IID_InProcFreeThreadedMarshaler.Value)
                     {
-                        Guid iid_IUnknown = InterfaceIIDs.IUnknown_IID;
-                        Guid iid_unmarshalClass;
-                        Marshal.ThrowExceptionForHR((**(ABI.WinRT.Interop.IMarshal.Vftbl**)marshalPtr).GetUnmarshalClass_0(
-                            marshalPtr, &iid_IUnknown, IntPtr.Zero, MSHCTX.InProc, IntPtr.Zero, MSHLFLAGS.Normal, &iid_unmarshalClass));
-                        if (iid_unmarshalClass == ABI.WinRT.Interop.IMarshal.IID_InProcFreeThreadedMarshaler.Value)
-                        {
-                            return true;
-                        }
+                        return true;
                     }
-                    finally
-                    {
-                        Marshal.Release(marshalPtr);
-                    }
+                }
+                finally
+                {
+                    Marshal.Release(marshalPtr);
                 }
             }
             return false;
--- a/src/WinRT.Runtime/ComWrappersSupport.cs
+++ b/src/WinRT.Runtime/ComWrappersSupport.cs
@@ -75,32 +75,42 @@ namespace WinRT
 
         // If we are free threaded, we do not need to keep track of context.
         // This can either be if the object implements IAgileObject or the free threaded marshaler.
-        internal unsafe static bool IsFreeThreaded(IObjectReference objRef)
+        internal unsafe static bool IsFreeThreaded(IntPtr iUnknown)
         {
-            if (objRef.TryAs(InterfaceIIDs.IAgileObject_IID, out var agilePtr) >= 0)
+            Guid iid_IAgileObject = InterfaceIIDs.IAgileObject_IID;
+            if (Marshal.QueryInterface(iUnknown, ref iid_IAgileObject, out var agilePtr) >= 0)
             {
                 Marshal.Release(agilePtr);
                 return true;
             }
-            else if (objRef.TryAs(InterfaceIIDs.IMarshal_IID, out var marshalPtr) >= 0)
+            else
             {
-                try
+                Guid iid_IMarshal = InterfaceIIDs.IMarshal_IID;
+                if (Marshal.QueryInterface(iUnknown, ref iid_IMarshal, out var marshalPtr) >= 0)
                 {
-                    Guid iid_IUnknown = InterfaceIIDs.IUnknown_IID;
-                    Guid iid_unmarshalClass;
-                    Marshal.ThrowExceptionForHR((**(ABI.WinRT.Interop.IMarshal.Vftbl**)marshalPtr).GetUnmarshalClass_0(
-                        marshalPtr, &iid_IUnknown, IntPtr.Zero, MSHCTX.InProc, IntPtr.Zero, MSHLFLAGS.Normal, &iid_unmarshalClass));
-                    if (iid_unmarshalClass == ABI.WinRT.Interop.IMarshal.IID_InProcFreeThreadedMarshaler.Value)
+                    try
                     {
-                        return true;
+                        Guid iid_IUnknown = InterfaceIIDs.IUnknown_IID;
+                        Guid iid_unmarshalClass;
+                        Marshal.ThrowExceptionForHR((**(ABI.WinRT.Interop.IMarshal.Vftbl**)marshalPtr).GetUnmarshalClass_0(
+                            marshalPtr, &iid_IUnknown, IntPtr.Zero, MSHCTX.InProc, IntPtr.Zero, MSHLFLAGS.Normal, &iid_unmarshalClass));
+                        if (iid_unmarshalClass == ABI.WinRT.Interop.IMarshal.IID_InProcFreeThreadedMarshaler.Value)
+                        {
+                            return true;
+                        }
                     }
-                }
-                finally
-                {
-                    Marshal.Release(marshalPtr);
+                    finally
+                    {
+                        Marshal.Release(marshalPtr);
+                    }
                 }
             }
             return false;
+        }
+
+        internal unsafe static bool IsFreeThreaded(IObjectReference objRef)
+        {
+            return IsFreeThreaded(objRef.ThisPtr);
         }
 
         public static IObjectReference GetObjectReferenceForInterface(IntPtr externalComObject)
@@ -115,21 +125,7 @@ namespace WinRT
                 return null;
             }
 
-            ObjectReference<T> objRef = ObjectReference<T>.FromAbi(externalComObject);
-            if (IsFreeThreaded(objRef))
-            {
-                return objRef;
-            }
-            else
-            {
-                using (objRef)
-                {
-                    return new ObjectReferenceWithContext<T>(
-                        objRef.GetRef(),
-                        Context.GetContextCallback(),
-                        Context.GetContextToken());
-                }
-            }
+            return ObjectReference<T>.FromAbi(externalComObject);
         }
 
         public static ObjectReference<T> GetObjectReferenceForInterface<T>(IntPtr externalComObject, Guid iid)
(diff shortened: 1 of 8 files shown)


src/WinRT.Runtime/ComWrappersSupport.cs
```diff
@@ -68,34 +68,44 @@ public static void MarshalDelegateInvoke<T>(IntPtr thisPtr, Action<T> invoke)
 
         // If we are free threaded, we do not need to keep track of context.
         // This can either be if the object implements IAgileObject or the free threaded marshaler.
-        internal unsafe static bool IsFreeThreaded(IObjectReference objRef)
+        internal unsafe static bool IsFreeThreaded(IntPtr iUnknown)
         {
-            if (objRef.TryAs(InterfaceIIDs.IAgileObject_IID, out var agilePtr) >= 0)
+            Guid iid_IAgileObject = InterfaceIIDs.IAgileObject_IID;
+            if (Marshal.QueryInterface(iUnknown, ref iid_IAgileObject, out var agilePtr) >= 0)
```
Sergio0694, line 74: Can skip the temp copy here (and in all other equivalent cases):
```suggestion
            if (Marshal.QueryInterface(iUnknown, ref Unsafe.AsRef(InterfaceIIDs.IAgileObject_IID), out var agilePtr) >= 0)
```

src/WinRT.Runtime/ComWrappersSupport.cs
```diff
@@ -68,34 +68,44 @@ public static void MarshalDelegateInvoke<T>(IntPtr thisPtr, Action<T> invoke)
...
             {
                 Marshal.Release(agilePtr);
                 return true;
             }
-            else if (objRef.TryAs(InterfaceIIDs.IMarshal_IID, out var marshalPtr) >= 0)
+            else
             {
-                try
+                Guid iid_IMarshal = InterfaceIIDs.IMarshal_IID;
+                if (Marshal.QueryInterface(iUnknown, ref iid_IMarshal, out var marshalPtr) >= 0)
```
Sergio0694, line 82: Same here

src/WinRT.Runtime/ComWrappersSupport.cs
```diff
@@ -68,34 +68,44 @@ public static void MarshalDelegateInvoke<T>(IntPtr thisPtr, Action<T> invoke)
...
-                {
-                    Marshal.Release(marshalPtr);
                 }
             }
             return false;
         }
 
+        internal unsafe static bool IsFreeThreaded(IObjectReference objRef)
+        {
+            return IsFreeThreaded(objRef.ThisPtr);
```
Sergio0694, line 106: Question on this, but really I noticed the same thing throughout CsWinRT. How is this safe without a `GC.KeepAlive` call? The GC could collect and finalize `objRef` while `IsFreeThreaded` is still running, which will cause things to explode (as the object would be deleted while some method is running). Does this just rely on callers keeping the references alive in some other way? Is this provably safe or is this just a "it's more of a theoretical problem but in practice we didn't see any crashes yet"?

src/WinRT.Runtime/ObjectReference.cs
```diff
@@ -36,6 +38,10 @@ public IntPtr ThisPtr
             }
         }
 
+        public bool IsFreeThreaded => _contextPtr == IntPtr.Zero;
```
Sergio0694, line 41: Why does CsWinRT not have almost any XML docs? Should we not have them, at least for public APIs?

src/WinRT.Runtime/ObjectReference.cs
```diff
@@ -100,22 +106,14 @@ protected unsafe IUnknownVftbl VftblIUnknown
...
-        {
-            get
-            {
-                ThrowIfDisposed();
-                return **(IUnknownVftbl**)ThisPtrFromOriginalContext;
-            }
-        }
-
-        protected IObjectReference(IntPtr thisPtr)
+        protected IObjectReference(IntPtr thisPtr, IntPtr contextPtr)
```
Sergio0694, line 109: This is both a source and a binary breaking change. Is this intended? If so, does this mean that we have accepted the idea of making breaking changes in the next CsWinRT release? Because if so, does that mean we can then also go fix more issues that are elsewhere in the library that would also require breaking changes?

src/WinRT.Runtime/ObjectReference.cs
```diff
@@ -333,47 +324,39 @@ internal unsafe bool IsReferenceToManagedObject
         {
             get
             {
-                return VftblIUnknown.Equals(IUnknownVftbl.AbiToProjectionVftbl);
+                ThrowIfDisposed();
```
Sergio0694, line 327: `ThisPtr` already checks for disposal, can we not remove this? I mean, is it not redundant?

src/WinRT.Runtime/ObjectReference.cs
```diff
@@ -333,47 +324,39 @@ internal unsafe bool IsReferenceToManagedObject
...
+                return (**(IUnknownVftbl**)ThisPtr).Equals(IUnknownVftbl.AbiToProjectionVftbl);
```
Sergio0694, line 328: More of a general thing I've noticed in CsWinRT, but all of these methods don't seem thread safe at all. You could have a thread concurrently dispose an instance right after another thread got past the disposal check, but before it actually invoked the native method with the obtained this pointer. Would that not crash? Is this just explicitly not supported? Essentially, consider this sequence of events:
- Thread A calls `ProjectedType.Foo()`
- A calls `ThisPtr`, the dispose check passes and it gets the `IntPtr` value (not incremented)
- Thread B calls `Dispose()`, which releases the object
- A now uses that `IntPtr` to invoke some function pointer
- 💥

Is there some guarantee that makes this not actually possible?

src/WinRT.Runtime/ObjectReference.cs
manodasanW: In theory we should have XML docs and may add them at some point.  But at the same time, most of our public APIs other than the Marshaler APIs are intended for generated projection consumption and not to be called directly by consumers.

src/WinRT.Runtime/ComWrappersSupport.cs
jlaanstra: Will do a pass.

src/WinRT.Runtime/ComWrappersSupport.cs
jlaanstra: This would rely on the caller keeping the reference alive.

src/WinRT.Runtime/ObjectReference.cs
jlaanstra: I don't think the constructors are meant to be called from outside WinRT.Runtime? I believe that for the nativeaot work we are already making breaking changes and this would go as part of that set of changes. /cc @manodasanW

src/WinRT.Runtime/ObjectReference.cs
jlaanstra: Yeah we can remove this since the method is private anyway.

src/WinRT.Runtime/ObjectReference.cs
jlaanstra: This is certainly possible. The ObjectReference owns the ThisPtr and if you disposed that while still in use then all kinds of things would go wrong. This is the same as calling SafeHandle.DangerousGetHandle().

src/WinRT.Runtime/ObjectReference.cs
manodasanW: If the breaking change is contained within `WinRT.Runtime` uses, it is acceptable for the staging branch.  In this case, will projections generated using the new `FactoryObjectReference` added in the other PR break?  If so, we might want to keep the existing protected constructor with one parameter and add the two parameter one as a new one.

src/WinRT.Runtime/ComWrappersSupport.cs
manodasanW: In this case, as mentioned, the caller would probably keep it alive.  For the general problem, we do have a bug to go look at if we should add `GC.KeepAlive` in certain places but so far, we haven't seen any crashes arise yet from not having it.

jlaanstra: Added a GC.KeepAlive here.

src/WinRT.Runtime/ObjectReference.cs
jlaanstra: Added it back.

src/WinRT.Runtime/ComWrappersSupport.cs
```diff
@@ -76,34 +76,45 @@ public static void MarshalDelegateInvoke<T>(IntPtr thisPtr, Action<T> invoke)
...
+            if (Marshal.QueryInterface(iUnknown, ref Unsafe.AsRef(InterfaceIIDs.IAgileObject_IID), out var agilePtr) >= 0)
             {
                 Marshal.Release(agilePtr);
                 return true;
             }
-            else if (objRef.TryAs(InterfaceIIDs.IMarshal_IID, out var marshalPtr) >= 0)
+            else
             {
-                try
+                if (Marshal.QueryInterface(iUnknown, ref Unsafe.AsRef(InterfaceIIDs.IMarshal_IID), out var marshalPtr) >= 0)
```
Sergio0694, line 88: Nit: you can simplify this by removing the outer `else` and one indentation for this whole inner `if`, since the previous block has a `return` statement anyway, so this code will always be executed here if that branch is not taken. I mean:
```csharp
if (...)
{
    return ...;
}
else
{
    if (...)
    {
        return ...;
    }
}
```
To:
```csharp
if (...)
{
    return ...;
}

if (...)
{
    return ...;
}
```

src/WinRT.Runtime/ObjectReference.cs
```diff
@@ -26,6 +26,7 @@ abstract class IObjectReference : IDisposable
         private readonly IntPtr _thisPtr;
         private object _disposedLock = new object();
         private IntPtr _referenceTrackerPtr;
+        protected IntPtr _contextToken;
```
Sergio0694, line 29: How come we've moving this field here rather than leaving it just in `ObjectReferenceWithContext<T>`? This will increase the size of every single object wrapper by `sizeof(nint)` even though they have no context to carry around. Can we not keep the context just in the derived type

Sergio0694, line 29: This field is also defined in `ObjectReferenceWithContext<T>`, so it's duplicated. We need to remove one of them.

Sergio0694 (changes requested):
Did another pass, looks good! Just left a couple notes 🙂

src/WinRT.Runtime/ObjectReference.cs
jlaanstra: Out of scope for this PR.

src/WinRT.Runtime/ObjectReference.cs
jlaanstra: That would require to make the new properties virtual. I suppose we can do that.

jlaanstra: We only need one :)

Sergio0694: Just to recap: didn't we say we should make this internal?

src/WinRT.Runtime/ObjectReference.cs
```diff
@@ -109,13 +114,25 @@ private protected unsafe IUnknownVftbl VftblIUnknownFromOriginalContext
...
         {
             if (thisPtr == IntPtr.Zero)
             {
                 throw new ArgumentNullException(nameof(thisPtr));
             }
             _thisPtr = thisPtr;
+            _contextToken = ComWrappersSupport.IsFreeThreaded(thisPtr) ? IntPtr.Zero : Context.GetContextToken();
+        }
+
+        protected IObjectReference(IntPtr thisPtr, IntPtr contextToken)
```
Sergio0694, line 128: Private protected?

src/WinRT.Runtime/ObjectReference.cs
```diff
@@ -423,37 +440,62 @@ public T Vftbl
             }
         }
 
-        public static ObjectReference<T> Attach(ref IntPtr thisPtr)
+        ObjectReference(IntPtr thisPtr, IntPtr contextToken, T vftblT) :
```
Sergio0694, line 443: Please add the explicit accessibility modifier

src/WinRT.Runtime/ObjectReference.cs
jlaanstra: Being able to inherit IObjectReference outside WinRT.Runtime allows for custom implementations. I Think that's useful to keep.

Sergio0694 (approved):
LGTM! :shipit:

src/WinRT.Runtime/ObjectReference.cs
```diff
@@ -474,6 +474,31 @@ public static ObjectReference<T> Attach(ref IntPtr thisPtr)
             }
         }
 
+        public static ObjectReference<T> Attach(ref IntPtr thisPtr, Guid iid)
```
Sergio0694, line 477: Could we not make the `Attach` overload without iid just call this with `default(Guid)`?

src/WinRT.Runtime/ObjectReference.cs
```diff
@@ -491,12 +516,38 @@ public static unsafe ObjectReference<T> FromAbi(IntPtr thisPtr, T vftblT)
...
                 var obj = new ObjectReferenceWithContext<T>(
                     thisPtr,
+                    vftblT,
                     Context.GetContextCallback(),
                     Context.GetContextToken());
                 return obj;
             }
         }
 
+        public static unsafe ObjectReference<T> FromAbi(IntPtr thisPtr, T vftblT, Guid iid)
```
Sergio0694, line 526: Is this needed? Doesn't seem like anyone is actually calling it?

jlaanstra: Called by other FromAbi methods.

src/WinRT.Runtime/ObjectReference.cs
jlaanstra: Possibly, but let's get this in to unblock. We can make more changes once this goes in.

Sergio0694 (approved):

manodasanW (approved):

PR #1394, opened 2023-11-20
Skip interfaces not publicly accessible in authoring scenarios

### LINKED ISSUE #1369: WinRTTypeWriter generator fails with NullReferenceException

Sergio0694:
### Overview

This PR updates the WinRT authoring generator to skip interfaces that are not publicly accessible.

### DIFF
--- a/src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
+++ b/src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
@@ -1211,7 +1211,7 @@ namespace Generator
             // Ignoring them here makes sure that they're not processed to be part of the .winmd file.
             foreach (var @interface in symbol.AllInterfaces)
             {
-                if (IsPubliclyAccessible(@interface))
+                if (@interface.IsPubliclyAccessible())
                 {
                     _ = interfaces.Add(@interface);
                 }
@@ -2696,28 +2696,6 @@ namespace Generator
                 type is IPropertySymbol property && !property.ExplicitInterfaceImplementations.IsDefaultOrEmpty ||
                 type is IEventSymbol @event && !@event.ExplicitInterfaceImplementations.IsDefaultOrEmpty;
         }
-
-#nullable enable
-        /// <summary>
-        /// Checks whether a given type symbol is publicly accessible (ie. it's public and not nested in any non public type).
-        /// </summary>
-        /// <param name="type">The type symbol to check for public accessibility.</param>
-        /// <returns>Whether <paramref name="type"/> is publicly accessible.</returns>
-        public bool IsPubliclyAccessible(ITypeSymbol type)
-        {
-            for (ITypeSymbol? currentType = type; currentType is not null; currentType = currentType.ContainingType)
-            {
-                // If any type in the type hierarchy is not public, the type is not public.
-                // This makes sure to detect public types nested into eg. a private type.
-                if (currentType.DeclaredAccessibility is not Accessibility.Public)
-                {
-                    return false;
-                }
-            }
-
-            return true;
-        }
-#nullable restore
 
         public void GetNamespaceAndTypename(string qualifiedName, out string @namespace, out string typename)
         {
--- a/src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
+++ b/src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
@@ -2696,12 +2696,19 @@ namespace Generator
             }
         }
 
-        public bool IsPublic(ISymbol type)
+        public bool IsPublic(ISymbol symbol)
         {
-            return type.DeclaredAccessibility == Accessibility.Public ||
-                type is IMethodSymbol method && !method.ExplicitInterfaceImplementations.IsDefaultOrEmpty ||
-                type is IPropertySymbol property && !property.ExplicitInterfaceImplementations.IsDefaultOrEmpty ||
-                type is IEventSymbol @event && !@event.ExplicitInterfaceImplementations.IsDefaultOrEmpty;
+            // Check that the type has either public accessibility, or is an explicit interface implementation
+            if (symbol.DeclaredAccessibility == Accessibility.Public ||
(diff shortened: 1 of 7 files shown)


src/Tests/AuthoringTest/Program.cs
```diff
@@ -1569,4 +1574,140 @@ public partial struct PartialStruct
...
+                lpVtbl->GetNumber = &GetDeviceFromAbi;
+
+                return (IntPtr)lpVtbl;
+            }
+
+            private IUnknownVftbl IUnknownVftbl;
+            private delegate* unmanaged[Stdcall]<void*, int*, int> GetNumber;
+
+            [UnmanagedCallersOnly(CallConvs = new[] { typeof(CallConvStdcall) })]
+            private static int GetDeviceFromAbi(void* thisPtr, int* value)
```
manodasanW, line 1632: nit: was this here and below supposed to be GetNumberFromAbi rather than GetDvice.

src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
```diff
@@ -2683,12 +2707,19 @@ static bool IsWinRTType(ISymbol symbol)
...
-                type is IPropertySymbol property && !property.ExplicitInterfaceImplementations.IsDefaultOrEmpty ||
-                type is IEventSymbol @event && !@event.ExplicitInterfaceImplementations.IsDefaultOrEmpty;
+            // Check that the type has either public accessibility, or is an explicit interface implementation
+            if (symbol.DeclaredAccessibility == Accessibility.Public ||
+                symbol is IMethodSymbol method && !method.ExplicitInterfaceImplementations.IsDefaultOrEmpty ||
+                symbol is IPropertySymbol property && !property.ExplicitInterfaceImplementations.IsDefaultOrEmpty ||
+                symbol is IEventSymbol @event && !@event.ExplicitInterfaceImplementations.IsDefaultOrEmpty)
+            {
+                // If we have a containing type, we also check that it's publicly accessible
+                return symbol.ContainingType is not { } containingType || containingType.IsPubliclyAccessible();
```
manodasanW, line 2719: I am trying to understand this, should this have been an `&&` rather than a `||`  or am I misunderstanding what is not {} does.

manodasanW (approved):

src/Tests/AuthoringTest/Program.cs
Sergio0694: Oops, yes, leftover from ComputeSharp, fixed! 🙂

src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
Sergio0694: So, `is not { }` matches if the expression is not an instance (it's like `is null`, but also declares a local). So:
- If `containingType` is `null`, return `true`
- Otherwise (ie. `containingType != null`), return `true` if it's publicly accessible

If we used `&&`, we'd be trying to access `containingType` when the first expression already matched, but if that's the case, then `containingType` is not defined (because the expression matches on `is not`). In fact, if you changed it to `&&`, the code would just not compile, as `containingType` would be uninitialized in that case 🙂

PR #1395, opened 2023-11-22
Allow dropping native host for NativeAOT WinRT components

### LINKED ISSUE #1387: Drop native host when publishing WinRT components with NativeAOT
# Proposal: drop native host when publishing WinRT components with NativeAOT

Opening this for tracking so we don't forget about it. Follow up from our previous conversation about improving the WinRT component authoring scenario specifically when NativeAOT is used, so that standalone self-contained .dll-s can be produced with no extra stuff.

## Summary

Right now, publishing a WinRT component causes CsWinRT to also bundle the native host which then bootstraps the .NET .dll bit. When publishing with NAOT, this is unnecessary. In this case, CsWinRT should just emit the `DllGetActivationFactory` export (or any other it needs) straight into the C# project (via `[UnmanagedCallersOnly]`) and only produce that standalone .dll with NAOT.

## Rationale

* Allowing authoring WinRT components with NAOT as a single .dll

Sergio0694:
### Overview

This PR adds a new `CsWinRTAotExportsEnabled` MSBuild property that can be set in authoring scenarios (and is automatically set if `PublishAot` is already set). This triggers the generator to emit the native exports for WinRT .dll-s directly into the same binary, with the correct logic, and also skips adding all references and copying all files related to the native host infrastructure.

Keeping as a draft as I'm not entirely sure how to set up unit tests for this just yet 😅


src/Authoring/WinRT.SourceGenerator/Generator.cs
```diff
@@ -1,258 +1,342 @@
...
+                if (e.InnerException != null)
+                {
+                    Logger.Log(e.InnerException.ToString());
+                }
+                Logger.Close();
+                Environment.ExitCode = -2;
+                throw;
+            }
+
+            Logger.Log("Done");
```
Sergio0694, line 185: Note: I normalized whitespaces in this file, **review with whitespace diff off**.
The actual changes are just a few lines of code being added.

src/Authoring/WinRT.SourceGenerator/Generator.cs
```diff
@@ -1,258 +1,342 @@
-﻿using Microsoft.CodeAnalysis;
```
Sergio0694, line 1: Note: I normalized whitespaces in this file, review with whitespace diff off.
The actual changes are just a few lines of code being added.

src/Authoring/WinRT.SourceGenerator/Generator.cs
```diff
@@ -1,258 +1,342 @@
...
+
+                            try
+                            {
+                                IntPtr obj = GetActivationFactory(MarshalString.FromAbi(activatableClassId));
+
+                                if ((void*)obj is null)
+                                {
+                                    *factory = null;
+
+                                    return E_NOINTERFACE;
```
ahmed605, line 235: you should return `CLASS_E_CLASSNOTAVAILABLE` if the class doesn't exist in the dll and **optionally** `E_NOINTERFACE` if the class exists but doesn't implement the factory interface.
ref (C++/WinRT): https://github.com/microsoft/cppwinrt/blob/bf4459b25aeeb7093e8262fb3b29ca70b0ab2e60/cppwinrt/component_writers.h#L200
ref (WRL): https://github.com/dwmcore/WinSDK-Tools/blob/78c9c2464212f8ecf16d6f20f072423b3ab76901/Include/10.0.25967.0/winrt/wrl/module.h#L323-L326

Sergio0694: Yeah I saw that in the docs, but it seems that's not what the host currently does:

https://github.com/microsoft/CsWinRT/blob/71169932920500310921b4c3fadba1e5a99c6138/src/Authoring/WinRT.Host.Shim/Module.cs#L44-L48

So I mostly just tried to keep it consistent. @manodasanW do we want to update this in both places?

src/Authoring/WinRT.SourceGenerator/Generator.cs
```diff
@@ -1,258 +1,342 @@
...
+                        /// Determines whether the DLL that implements this function is in use. If not, the caller can unload the DLL from memory.
+                        /// </summary>
+                        /// <returns>This method always returns <c>S_FALSE</c>.</returns>
+                        /// <seealso href="https://learn.microsoft.com/en-us/windows/win32/api/combaseapi/nf-combaseapi-dllcanunloadnow"/>
+                        [UnmanagedCallersOnly(EntryPoint = nameof(DllCanUnloadNow), CallConvs = new[] { typeof(CallConvStdcall) })]
+                        public static int DllCanUnloadNow()
+                        {
+                            const int S_FALSE = 1;
+
+                            return S_FALSE;
```
ahmed605, line 260: This will prevent the dll from unloading even if it's not used anymore, you should return true if the total ref count of implemented classes is zero, I'm not sure whether CsWinRT already has a ref tracker for this (similar to Module Lock in CppWinRT) or not tho.

Sergio0694: This matches the behavior of the native host. Also, note that NativeAOT doesn't support unloading anyway.

nuget/Microsoft.Windows.CsWinRT.Authoring.targets
```diff
@@ -253,30 +260,30 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <!-- When packing, include all necessary DLLs and the targets file for DLL copying on the native side -->
     <ItemGroup>
 
-      <Content Include="$(CsWinRTPath)lib\net6.0\WinRT.Host.Shim.dll">
+      <Content Condition="'$(CsWinRTAotExportsEnabled)' != 'true'" Include="$(CsWinRTPath)lib\net6.0\WinRT.Host.Shim.dll">
```
hez2010, line 263: You may want to put the condition on the ItemGroup instead of duplicating them everywhere.

hez2010 (changes requested):

src/Authoring/WinRT.SourceGenerator/Generator.cs
hez2010: It is really confusing to see `E_NOINTERFACE` when the class doesn't exist due to trimming or other issues: you have no way to know whether the implementation got trimmed or the class got trimmed.
I would suggest changing here to return `CLASS_E_CLASSNOTAVAILABLE` instead.

nuget/Microsoft.Windows.CsWinRT.Authoring.targets
Sergio0694: This is not added to all items in this group. For instance, the .winmd etc. are always copied. I suppose I could split those to be in a separate group though, to keep things simpler, so we'd have a group for "always copied" and one for "only copied when not publishing with NAOT", that does seem better 😄

src/Authoring/WinRT.SourceGenerator/Generator.cs
```diff
@@ -1,258 +1,342 @@
...
+                                    *factory = null;
+
+                                    return E_NOINTERFACE;
+                                }
+
+                                *factory = (void*)obj;
+
+                                return S_OK;
+                            }
+                            catch (Exception e)
```
hez2010, line 242: typo
```suggestion
                            catch (Exception ex)

```

src/Authoring/WinRT.SourceGenerator/Generator.cs
```diff
@@ -1,258 +1,342 @@
...
+                            const int S_OK = 0;
+
+                            if (activatableClassId is null || factory is null)
+                            {
+                                return E_INVALIDARG;
+                            }
+
+                            try
+                            {
+                                IntPtr obj = GetActivationFactory(MarshalString.FromAbi(activatableClassId));
```
hez2010, line 229:
```suggestion
                                IntPtr obj = GetActivationFactory(MarshalString.FromAbi((nint)activatableClassId));

```

hez2010 (changes requested):

src/Authoring/WinRT.SourceGenerator/Generator.cs
manodasanW: Yea, I think it should be fine to switch both versions to `CLASS_E_CLASSNOTAVAILABLE`.  In the case where the interface is not implemented, we should throw an exception during the `as` which should result in `E_NOINTERFACE `.

Sergio0694: Done in 74cb8a27fbb2d05801f6bb917ba1e2dedbd08bab. I'm returning `REGDB_E_READREGDB` since that matches what [the docs say](https://learn.microsoft.com/en-us/previous-versions/br205771(v=vs.85)#return-value). It's also what the host shim was returning already in case of failure. I've also updated that `E_NOINTERFACE` there as well.

hez2010:
I would like to suggest changing the MSBuild property to something like `CsWinRTSelfContainedExportsEnabled` because it can not only be used by NativeAOT, an out-of-proc WinRT server can also benefit from it.

Sergio0694:
I think "SelfContained" would be confusing, because this only works when the .NET assembly is compiled with NativeAOT, and not just with a self-contained deployment. If you're using CoreCLR and a native host, then you don't need these exports.

manodasanW (approved):

manodasanW:
Just to note in the PR, there is one remaining authoring test failure for AOT (AuthoringTest.CustomTypes).  That would be addressed in a separate PR.

hez2010:
> If you're using CoreCLR and a native host, then you don't need these exports.

We will still need the generated `int GetActivationFactory(void*, void**)` so that we can pass it to `RoRegsiterActivationFactories`.

PR #1396, opened 2023-11-26
Minor trimming optimizations for authoring

Sergio0694:
This PR includes two small tweaks for trimming in authoring scenarios:
- Remove an unnecessary `DynamicallyAccessedMemberTypes.PublicFields` flag in `TryAs<T>`
- Use `IActivationFactoryMethods.IID` to resolve the activation factory from `Make()`

### DIFF
--- a/src/cswinrt/code_writers.h
+++ b/src/cswinrt/code_writers.h
@@ -9647,7 +9647,7 @@ RuntimeHelpers.RunClassConstructor(typeof(%).TypeHandle);
 
 public static IntPtr Make()
 {
-using var marshaler = MarshalInspectable<%>.CreateMarshaler(_factory).As<ABI.WinRT.Interop.IActivationFactory.Vftbl>();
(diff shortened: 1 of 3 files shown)


src/cswinrt/code_writers.h
```diff
@@ -9415,7 +9415,7 @@ RuntimeHelpers.RunClassConstructor(typeof(%).TypeHandle);
 
 public static IntPtr Make()
 {
-using var marshaler = MarshalInspectable<%>.CreateMarshaler(_factory).As<ABI.WinRT.Interop.IActivationFactory.Vftbl>();
+using var marshaler = MarshalInspectable<%>.CreateMarshaler(_factory).As<ABI.WinRT.Interop.IActivationFactory.Vftbl>(IActivationFactoryMethods.IID);
```
Sergio0694, line 9418: While looking into this, I also noticed that the authored component includes a generated `IActivationFactoryVftbl` type, which seems to basically be a duplicate of the vtable types already in WinRT.Runtime anyway? @jlaanstra do you know whether we can just remove (as in, stop generating) that entirely and just switch to the ones in the runtime? Maybe it could be part of your activation factory tweaks?

jlaanstra (approved):

manodasanW (approved):

PR #1397, opened 2023-11-26
Remove generated VftblPtr type

Sergio0694:
This PR removes the generated VftblType, which wasn't even used most of the time. The only code path referencing it has been rewritten to manually reinterpret a pointer instead of using `Marshal`, which also makes it a bit faster while at it.

### DIFF
--- a/src/WinRT.Runtime/Interop/IUnknownVftbl.cs
+++ b/src/WinRT.Runtime/Interop/IUnknownVftbl.cs
@@ -4,14 +4,6 @@
 using System;
 using System.Runtime.InteropServices;
 
-namespace WinRT
-{
-    internal struct VftblPtr
-    {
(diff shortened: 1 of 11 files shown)


src/WinRT.Runtime/Interop/IUnknownVftbl.cs
```diff
@@ -4,6 +4,14 @@
 using System;
 using System.Runtime.InteropServices;
 
+namespace WinRT
+{
+    internal struct VftblPtr
```
jlaanstra, line 9: Do you still need this?

Sergio0694: You're right, we don't. Removed it entirely 😄

manodasanW (approved):

PR #1399, opened 2023-11-29
Make exported DllGetActivationFactory zero-alloc

Sergio0694:
This PR makes the exported `DllGetActivationFactory` zero-alloc, skipping the useless `string runtimeClassId` parameter.
It does so by introducing the following new API to `MarshalString`:

```csharp
public static ReadOnlySpan<char> FromAbiUnsafe(IntPtr value);
```

This allows marshalling an `HSTRING` to a `ReadOnlySpan<char>` value, without allocating a temporary `string`. The "Unsafe" suffix matches the naming convention in the BCL for methods that are not safe when misused. In this case, callers have to make sure not to escape the returned `ReadOnlySpan<char>` value outside of the scope where the `HSTRING` is valid.

Then, it updates the generated `Module` type so that when not on .NET Standard, the `GetActivationFactory` method will actually take a `ReadOnlySpan<char>` parameter (with a second `string` overload just forwarding to it). The new native exports for NativeAOT can then just call this new `ReadOnlySpan<char>` overload using the new `FromAbiUnsafe` marshalling method.

### DIFF
--- a/src/Authoring/WinRT.Host.Shim/Module.cs
+++ b/src/Authoring/WinRT.Host.Shim/Module.cs
@@ -36,7 +36,7 @@ namespace WinRT.Host
                 {
                     return REGDB_E_READREGDB;
                 }
-                var GetActivationFactory = type.GetMethod("GetActivationFactory");
+                var GetActivationFactory = type.GetMethod("GetActivationFactory", new Type[] { typeof(string) });
                 if (GetActivationFactory == null)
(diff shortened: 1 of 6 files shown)


Sergio0694:
Abandoning this for now because we're currently not building the projections for .NET 7+.
This change would need that, as switching on spans is only supported with C# 11+.

的9GIF (2)

PR #1400, opened 2023-11-30
Add MarshalString.FromAbiUnsafe API

Sergio0694:
### Description

This PR adds a new `MarshalString` API:

```csharp
public static ReadOnlySpan<char> FromAbiUnsafe(IntPtr value);
```

This allows marshalling an `HSTRING` to a `ReadOnlySpan<char>` value, without allocating a temporary `string`. The "Unsafe" suffix matches the naming convention in the BCL for methods that are not safe when misused. In this case, callers have to make sure not to escape the returned `ReadOnlySpan<char>` value outside of the scope where the `HSTRING` is valid.

Developers can use this where appropriate, and we can also use this in eg. marshalling stubs in the future. Once we update the projections to require C# 11, we can also restore #1399 and leverage this API there as well.

### DIFF
--- a/src/WinRT.Runtime/Marshalers.cs
+++ b/src/WinRT.Runtime/Marshalers.cs
@@ -171,6 +171,37 @@ namespace WinRT
             return new string(buffer, 0, (int)length);
         }
 
+        /// <summary>
+        /// Marshals an input <c>HSTRING</c> value to a <see cref="ReadOnlySpan{T}"/> value.
+        /// </summary>
+        /// <param name="value">The input <c>HSTRING</c> value to marshal.</param>
+        /// <returns>The resulting <see cref="ReadOnlySpan{T}"/> value.</returns>
+        /// <remarks>
+        /// <para>
+        /// This method is equivalent to <see cref="FromAbi"/>, but it does not create a new <see cref="string"/> instance.
+        /// Doing so makes it zero-allocation, but extra care should be taken by callers to ensure that the returned value
+        /// does not escape the scope where the source <c>HSTRING</c> is valid.
+        /// </para>
+        /// <para>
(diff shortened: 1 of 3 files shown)


Sergio0694:
@manodasanW the test failure in Debug x64 seems unrelated, maybe transient? Could you re-trigger that CI job? 🙂

src/WinRT.Runtime/Marshalers.cs
```diff
@@ -171,6 +171,37 @@ public static unsafe string FromAbi(IntPtr value)
...
+        /// does not escape the scope where the source <c>HSTRING</c> is valid.
+        /// </para>
+        /// <para>
+        /// For instance, if this method is invoked in the scope of a method that receives the <c>HSTRING</c> value as one of
+        /// its parameters, the resulting <see cref="ReadOnlySpan{T}"/> is always valid for the scope of such method. But, if
+        /// the <c>HSTRING</c> was created by reference in a given scope, the resulting <see cref="ReadOnlySpan{T}"/> value
+        /// will also only be valid within such scope, and should not be used outside of it.
+        /// </para>
+        /// </remarks>
+        public static unsafe ReadOnlySpan<char> FromAbiUnsafe(IntPtr value)
```
manodasanW, line 192: Should we consider calling this FromABIAsReadOnlySpan or similar or FromAbiAsUnsafeSpan if we really want unsafe in the name in addition to declaring the method unsafe.  I am not sure whether we need unsafe also in the name when the method is declared unsafe.  Do .NET functions follow similar patterns for such use cases?

manodasanW: Right that makes sense.

manodasanW (approved):

PR #1401, opened 2023-11-30
Make exported DllGetActivationFactory zero-alloc (take 2)

Sergio0694:
Second attempt from #1399 😄

This PR makes the exported `DllGetActivationFactory` zero-alloc, skipping the useless `string runtimeClassId` parameter.
It does so by leveraging the new `MarshalString.FromAbiUnsafe` API added in #1400, and it updates the generated `Module` type so that when not on .NET Standard, the `GetActivationFactory` method will actually take a `ReadOnlySpan<char>` parameter (with a second `string` overload just forwarding to it). The new native exports for NativeAOT can then just call this new `ReadOnlySpan<char>` overload using the new `FromAbiUnsafe` marshalling method.

### DIFF
--- a/src/cswinrt/code_writers.h
+++ b/src/cswinrt/code_writers.h
@@ -9684,23 +9684,25 @@ namespace WinRT
 % static partial class Module
 {
 public static unsafe IntPtr GetActivationFactory(% runtimeClassId)
-{%
-return IntPtr.Zero;
+{
+switch (runtimeClassId)
+{
+%
+default:
+    return IntPtr.Zero;
+}
 }
 %
 }
 }
 )",
     internal_accessibility(),
-    settings.netstandard_compat ? "string" : "ReadOnlySpan<char>",
+    settings.net7_0_or_greater ? "ReadOnlySpan<char>" : "string",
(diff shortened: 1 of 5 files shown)


src/cswinrt/code_writers.h
```diff
@@ -9683,30 +9683,45 @@ namespace WinRT
...
+default:
+    return IntPtr.Zero;
 }
 }
+%
+}
 }
 )",
     internal_accessibility(),
+    settings.net7_0_or_greater ? "ReadOnlySpan<char>" : "string",
```
manodasanW, line 9700: Consider a comment here, that doing a switch on `ReadOnlySpan<char>` it supported on net7 or greater which is what we are taking advantage of.

manodasanW (approved):

src/cswinrt/code_writers.h
Sergio0694: Done in https://github.com/microsoft/CsWinRT/pull/1401/commits/35edf3b2704af993ec6727d8ce6482d5827a8518 🙂

PR #1402, opened 2023-12-01
Use hardcoded IIDs for all well known IReferenceArray<T> types

Sergio0694:
This PR adds precomputed IIDs to `InterfaceIIDs` for all well known `IReferenceArray<T>` types used in `ComWrappersSupport`, and replaces all calls to `GuidGenerator.GetIID` with these constants. Couldn't remove all the additional cruft in `IReferenceArray<T>` yet because there's the generic path that's still using that, but this is a step in the right direction. We'll have to revisit that later anyway given it's 100% not AOT compatible (due to `MakeGenericType`).


src/WinRT.Runtime/ComWrappersSupport.cs
```diff
@@ -776,135 +776,135 @@ private static ComInterfaceEntry ProvideIReferenceArray(Type arrayType)
             {
```
Sergio0694, line 776: There's also some whitespace normalization.

**REVIEW WITH SHOW WHITESPACE OFF**

manodasanW (approved):

PR #1411, opened 2023-12-12
Move generic delegates vtable generation into projections

manodasanW:
As part of the previous AOT changes, the generic delegate vtable generation was handled by the AOT optimizer using a lookup table.  But this turned out that each of the projections got a lookup table when that wasn't the intention as it is supposed to be for consumers of projections.  This lookup table in projections also made those types not trimming friendly.  Given delegates have a known set of interfaces that they implement unlike classes which needs to deal with covariance, this PR moves the delegate vtable generation into the projection.

With this PR, when a delegate is being used by a projection, it would make sure the generic instantiation class for the delegate has been initialized and its vtable entries have been added so that on lookup they are found.  The AOT optimizer will continue to handle cases where delegates are passed as object and the projection isn't aware of that.


src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
```diff
@@ -529,10 +530,13 @@ private static List<VtableAttribute> GetVtableAttributesToAddOnLookupTable(Gener
...
 
-                                // This handles the case where a generic delegate is passed.
+                                // This handles the case where a generic delegate is passed as an object.
+                                // When they are not passed as an object, the marshaler makes sure
+                                // the vtable information is available.
                                 if (argumentClassTypeSymbol.TypeKind == TypeKind.Delegate &&
                                     argumentClassTypeSymbol.MetadataName.Contains("`") &&
-                                    GeneratorHelper.IsWinRTType(argumentClassTypeSymbol))
+                                    GeneratorHelper.IsWinRTType(argumentClassTypeSymbol) &&
+                                    methodSymbol.Parameters[paramsIdx].Type.SpecialType == SpecialType.System_Object)
```
vineeththomasalex, line 539: Not clear why this additional check is required now. Maybe a comment would help.

vineeththomasalex (approved):

src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
manodasanW: The comment I had originally changed was intended to do that.  I adjusted the comment to make it clearer.

PR #1415, opened 2023-12-13
Fix issue with properties implemented across multiple static interfaces

### LINKED ISSUE #1414: 'Could not find property getter interface' when processing I____Statics2
We have a runtimeclass where v1 has a static property "Foo" with only a getter. Then in v2 contract the static "Foo" setter got added.

Running cswinrt over this winmd fails with this error. Mano says:
> Looking at the cswinrt implementation, I think I see the issue.  The logic isn't correctly handling traversing the exclusive interfaces for static classes as it assumes it would have interfaces implemented when they are actually attributes.  I will look at implementing a fix.

Sample IDL would be:
```
    [contract(ReproContract, 1.0)]
    runtimeclass Repro
    {
        static String Current{ get; };
        [contract(ReproContract, 2.0)]
        {
            static String Current{ set; };
        }
    }
```

manodasanW:
Today if you had a static class which had a property implemented across multiple interfaces (i.e. get in one, set in the other), you would get "Could not find property getter interface".  This is because we didn't merge properties across interfaces for static classes like we did for regular classes.  This brings that logic that we had in classes to do merging of properties to static classes.

Fix #1414

### DIFF
--- a/src/Tests/TestComponentCSharp/TestComponentCSharp.idl
+++ b/src/Tests/TestComponentCSharp/TestComponentCSharp.idl
@@ -505,11 +505,13 @@ namespace TestComponentCSharp
         static void Method();
         static Int32 Property;
         static event Windows.Foundation.EventHandler<Int32> Event;
+        static Int32 ReadWriteProperty{ get; };
         [contract(Windows.Foundation.UniversalApiContract, 10)]
         {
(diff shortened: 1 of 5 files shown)


vineeththomasalex (approved):
LGTM!

Issue #1315, opened 2023-04-15
Don't use managed function pointer signatures in generated marshalling stubs


Sergio0694:
**Describe the bug**

CsWinRT is currently generating managed function pointer signatures in generated stubs, eg.:

![](https://user-images.githubusercontent.com/10199417/232229793-7f61bdae-c602-444a-873f-a58804149d9a.png)

These are some generated signatures for Win2D APIs created with CsWinRT 2.0.2. Note those `out` params.
For reference, the IDL signature for these two is the following:

![](https://user-images.githubusercontent.com/10199417/232229931-8cda8706-6b49-4c27-adb2-eb4f650be84c.png)

Using managed function pointers (ie. with any managed parameter/return, such as `out` params here) triggers the creation of runtime-powered marshalling stubs, which is not AOT friendly and it's also bad for performance. Here the generator should instead just manually handle passing this parameter by reference, and use an unmanaged function pointer signature.

For instance, `get_Size` could be something like this:

```csharp
public unsafe static Size get_Size(IObjectReference _obj)
{
	IntPtr thisPtr = _obj.ThisPtr;
	Size result = default(Size);
	ExceptionHelpers.ThrowExceptionForHR(((delegate* unmanaged[Stdcall]<IntPtr, Size*, int>)(*(IntPtr*)((nint)(*(IntPtr*)(void*)thisPtr) + (nint)12 * (nint)sizeof(delegate* unmanaged[Stdcall]<IntPtr, Size*, int>))))(thisPtr, &result));
	return result;
}
```

In theory it'd be better to also use eg. `void*` instead of `IntPtr`, but at least not doing so doesn't affect performance.
Having managed pointers, on the other hand, should definitely be avoided.

**To Reproduce**

1. Define a WinRT API with an `[out]` value parameter, eg. `HRESULT Number([out, retval] int result)`
2. Generate C# projections for its assembly with CsWinRT 2.0.2
3. Observe the generated marshalling stub is using an unmanaged function pointer with a managed signature

**Expected behavior**

The generated stub should use a blittable signature for the unmanaged function pointer being used.

**Version Info**

- CsWinRT 2.0.2

**Additional context**

I noticed this in Win2D, which we currently updated to use WASDK 1.3 and CsWinRT 2.0.2.
This affects quite a lot of APIs from our projections there. In general, this is just very common.
I'd expect fixing this would give some nice perf wins to WASDK as a whole, plus it's "free".

[...]

PR #1416, opened 2023-12-13
Remove out/in parameters from function pointers to avoid generating marshalling stubs.

jlaanstra:
Contributes to #1315 and #1318

### DIFF
--- a/src/cswinrt/code_writers.h
+++ b/src/cswinrt/code_writers.h
@@ -615,20 +615,20 @@ namespace cswinrt
             w.write(", %", bind<write_abi_type>(semantics));
             break;
         case param_category::ref:
-            w.write(settings.netstandard_compat ? ", in %" : ", %*", bind<write_abi_type>(semantics));
+            w.write(", in %", bind<write_abi_type>(semantics));
             break;
         case param_category::out:
-            w.write(settings.netstandard_compat ? ", out %" : ", %*", bind<write_abi_type>(semantics));
+            w.write(", out %", bind<write_abi_type>(semantics));
             break;
         case param_category::pass_array:
         case param_category::fill_array:
             w.write(", int, IntPtr");
             break;
         case param_category::receive_array:
-            w.write(settings.netstandard_compat ? ", out int, out IntPtr" : ", int*, IntPtr*");
+            w.write(", out int, out IntPtr");
             break;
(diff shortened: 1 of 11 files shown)


manodasanW:
Some of these are already addressed in the staging/aot branch.  But these changes might be worth having in master before the aot branch is merged.  We should do a comparison between the changes here and there and make sure they are done in the same way or any that are missing there are ported.

src/cswinrt/code_writers.h
```diff
@@ -3195,19 +3195,23 @@ db_path.stem().string());
     auto get_invoke_info(writer& w, MethodDef const& method, uint32_t const& abi_methods_start_index = INSPECTABLE_METHOD_COUNT)
     {
         TypeDef const& type = method.Parent();
-        if (!settings.netstandard_compat && distance(type.GenericParam()) == 0)
+        bool signature_has_generic_parameters = abi_signature_has_generic_parameters(w, method_signature{ method });
```
jlaanstra, line 3198: @ManodasanW to check this logic that I ported from staging/AOT.

Sergio0694 (commented):
I'm thinking it probably makes sense to wait for #1497 to be merged first, and then change the target branch for this PR to `staging/AOT` and rebase and update it from there 🤔

dongle-the-gadget:
> change the target branch for this PR to staging/AOT

Didn't Mano say that the changes should probably be merged to master before we officially merge the AOT branch?

Sergio0694:
It just feels like a recipe to have a bazillion merge conflicts. I don't really see the point of merging them first. You don't gain much even if you published another patch release from master, just with this, if you don't have all the other improvements from the AOT branch anyway. Just seems like extra (and very annoying) work for no reason. Better ways to spend your time 😅

dongle-the-gadget:
Some of these fixes have been implemented in the AOT branch, so I feel like rebasing this would probably lead to lots of duplicates.

Sergio0694:
You would have to deal with that eventually anyway. My point is, we can at least just do it once, in the AOT branch.

jlaanstra:
The point of this PR would be to bring some of this from staging/AOT to master. Everything should be already in staging/AOT. If something is missing I can see about porting those change to staging/AOT.

PR #1418, opened 2023-12-15
WUX/MUX prototype

jkoritzinsky:
Implement prototype WUX/MUX support for both consumption and authoring.

Tests are still TODO.

The remaining TODO-WuxMux items will provide a better UX or perf but are not required for basic functionality.


src/WinRT.Runtime/Projections.cs
```diff
@@ -25,9 +25,33 @@ namespace WinRT
...
+                {
+                    throw new NotSupportedException("Windows.UI.Xaml is not supported before .NET 5. Please use built-in WinRT interop for Windows.UI.Xaml experiences");
+                }
+#endif
+                return mode;
+            }
+            return UiXamlMode.MicrosoftUiXaml; // We default to MUX for back-compat with existing projects.
+        }
+
+        internal static UiXamlMode UiXamlModeSetting { get; } = GetUIXamlModeSetting();
```
Sergio0694, line 49: Not needed for the initial prototype, but we might want to have the final version also include an `ILLink.Substitutions.xml` that can swap the body of the property getter with the hardcoded value based on the runtime switch configuration from MSBuild. I mean, it's simple enough but it's nice to have 🙂

src/WinRT.Runtime/Projections/IEnumerable.net5.cs
```diff
@@ -998,6 +998,11 @@ public uint GetMany(ref T[] items)
             public bool MoveNext() => _MoveNext();
 
             uint global::Microsoft.UI.Xaml.Interop.IBindableIterator.GetMany(ref object[] items)
+            {
+                // Should not be called.
+                throw new NotImplementedException();
```
Sergio0694, line 1003: Nit (a bit of a side note): should these be `NotSupportedException`? IIRC the recommendation was to never throw `NotImplementedException` from production code.

src/WinRT.Runtime/Projections/INotifyCollectionChanged.net5.cs
```diff
@@ -45,11 +46,21 @@ public static unsafe (Action<global::System.Collections.Specialized.NotifyCollec
...
     [EditorBrowsable(EditorBrowsableState.Never)]
     [Guid("530155E1-28A5-5693-87CE-30724D95A06D")]
+    [WuxMuxProjectedType]
     internal unsafe interface INotifyCollectionChanged : global::System.Collections.Specialized.INotifyCollectionChanged
     {
-        [Guid("530155E1-28A5-5693-87CE-30724D95A06D")]
+        public static string GetGuidSignature()
+            => Projections.UiXamlModeSetting == Projections.UiXamlMode.WindowsUiXaml
+                ? "{CF75D69C-F2F4-486B-B302-BB4C09BAEBFA}"
+                : typeof(INotifyCollectionChanged).GUID.ToString("B");
```
Sergio0694, line 55: Do we want to hardcode the GUID string in both cases (for consistency and marginally better perf)?

Sergio0694 (commented):
Just glanced it so far and left some random notes as I read through the code. Looking really good, and it's cool how the changes are pretty targeted and most code is the same 🚀

src/WinRT.Runtime/Projections.cs
jkoritzinsky: Yes, that is a good idea! I'll see if I can figure it out (since it's not a bool I'll need to check the right way to do it)

src/WinRT.Runtime/Projections/IEnumerable.net5.cs
jkoritzinsky: I think the `E_NOTIMPL` HResult here actually better matches the expected experience on the WinRT consumption side than throwing a `NotSupportedException` would as this method is only used when a WinRT consumer reinterpet_casts this type to `IIterator<object>`, which is technically invalid, but we need to support.

src/WinRT.Runtime/Projections/INotifyCollectionChanged.net5.cs
jkoritzinsky: I can go back and do that.

src/WinRT.Runtime/Projections/IEnumerable.net5.cs
Sergio0694: Oh, I see, that makes sense. Thank you for the additional info! 🙂

jkoritzinsky:
I added some tests here to validate the experience.

dongle-the-gadget:
The `cswinrt.exe` generator had an issue that if it had to generate some classes in a particular namespace, if that namespace is present under `strings\additions` it would always pull those files in. This means that when we try to build a WUX only projection with `Windows.UI.Colors`, it would also pull in `Windows.UI.Color`, conflicting with the SDK-provided projection.

Is there a fix underway for this?

jkoritzinsky:
@dongle-the-gadget that's by design. The idea is that the WUX projection would only project the WUX namespaces and exclude any namespaces that are present in the SDK-provided projection and it would reference the SDK-provided projection of the rest of the OS-built-in Windows SDK.

jkoritzinsky:
Okay looks like I might need some help with getting the INPC test to run in GoogleTest (I need to start-up the WUX engine to get it to work). @Sergio0694 I might need your help on this

Sergio0694:
Ahahah glad to hear you found a workaround! 😄

Also, just a note, as this will get merged after #1395, it will also eventually need to get the same tweaks so that in Release x64 mode the authoring test is published with NAOT and the manifest is skipped (so we can also verify that things work fine without it when there's no native host and the implementation .dll is the same). Just worth keeping in mind in case any of the things done here so far could potentially cause any issues with that (hopefully not, but figured I'd mention that just in case) 🙂

jkoritzinsky:
We'll still need a manifest here as it needs to specify a new-enough version of Windows to enable XAML Islands.

jkoritzinsky:
Marking this as ready for review as I have some tests now that validate activation and consumption scenarios (events end up validating both).

Sergio0694:
> "We'll still need a manifest here as it needs to specify a new-enough version of Windows to enable XAML Islands."

Oh, I mean that's fine, but can we skip all the definitions of the activatabke classes though (perhaps with a different version of the manifest that's only used for the NAOT test)? Just to make sure that we're resting that too for consistency with the WASDK stuff.

Also... Not entirely sure why the CI isn't running 🤔

jkoritzinsky:
It's probably because I'm working from a fork instead of upstream. I'll push to a branch in microsoft/cswinrt and try again. I forgot about that aspect of this repo.

jkoritzinsky:
Closing in favor of #1421 that will get CI

Issue #723, opened 2021-02-05
Support for UWP and System XAML


asklar:
Filing this here to track a dependency on cswinrt from https://github.com/microsoft/react-native-windows/issues/6539
We would like to be able to move away from C# UWP but we still need to target system xaml. So we would like the ability for cswinrt to target UWP apps and system XAML so that we can avoid a whole bunch of issues related to .net native and consuming nugets from C++ apps.

Related: https://github.com/microsoft/react-native-windows/issues/6539

AdamBraden:
This would require custom type mapping for xaml to map onto inbox System XAML. Right now the custom type mappings go against WinUI.

yoshiask:
I'm a little surprised that C#/WinRT doesn't already support system XAML. Just a noob here, but I had assumed that "lifting this projection support out of the compiler and runtime" would also entail decoupling from WinUI. Are there technical reasons why it can't/won't be done?

driver1998:
Also this will unblock the XAML islands on .NET 5+ scenario, without this we'll have to wait for WinUI 3 XAML islands which is "scheduled to a later release".

Jaiganeshkumaran:
I have a component that exposes a few platform XAML types. Before I could just use them from C# but now with C#/WinRT, it blacklists all those types from the projection, unlike every other WinRT projection. Even if you reserve certain mappings to Microsoft.UI.Xaml (eg: INotifyPropertyChanged), Windows.UI.Xaml types must be available even if you have Microsoft.UI.Xaml (from WinUI 3) referenced. C#/WinRT shouldn't be framework dependent.

<img src="https://cdn.discordapp.com/attachments/714838721790476289/925408560274341958/unknown.png">

DontKnowOfTheGaps:
@angelazhangmsft  This issue is hanging on thin air and should be dropped right away.

UWP/CoreWindow was doomed to oblivion a loooong time ago. 
As .NET 5 / .NET 6 / .NET 7 and upwards versions themselves will _never_ support it , CsWinRT can't do very much, can it ?

driver1998:
.NET 5 / 6 / 7 won't be compatible with AppContainer and UAP-only environments anytime soon, so UWP is not actually relevant here.
This is mainly for XAML islands.

DontKnowOfTheGaaps:
so the title be modified to reflect XAML Island only. (Support for ~~UWP and~~ System XAML for XAML Islands.

[...]

PR #1421, opened 2023-12-20
WUX/MUX prototype

jkoritzinsky:
Implement prototype WUX/MUX support for both consumption and authoring.
The remaining TODO-WuxMux items will provide a better UX or perf but are not required for basic functionality.

Replacement for #1418 that will get CI.


jkoritzinsky:
Follow-up notes from @Sergio0694 in the other PR:

We'll need to react to #1395 and have a separate manifest for AOT that only specifies the "latest tested OS version" to allow the Xaml Islands APIs.

Also, we should hook up a linker substitution file to hard-code the UiXamlMode switch when trimming or AOT compiling.

dongle-the-gadget:
Wanted to clarify the comment I made in #1418. The `Windows.UI.Colors` class isn’t available in the SDK projection, as it is a WUX type. However, the issue is due to CsWinRT’s behaviors regarding additions, you can’t project `Windows.UI.Colors` without also projecting `Windows.UI.Color`, which would conflict with the SDK projection and generate a warning

src/Projections/Windows.UI.Xaml/Windows.UI.Xaml.csproj
```diff
@@ -0,0 +1,29 @@
...
+
+  <PropertyGroup>
+    <TargetFrameworks>$(LibBuildTFMsNoNetStandard)</TargetFrameworks>
+    <Platforms>x64;x86</Platforms>
+    <AssemblyName>Microsoft.Windows.UI.Xaml</AssemblyName>
+    <EmitCompilerGeneratedFiles>true</EmitCompilerGeneratedFiles>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="System.Numerics.Vectors" Version="4.5.0" />
```
Sergio0694, line 11: Can we condition this so it's skipped when on TFMs that already have these APIs?

src/WinRT.Runtime/Interop/WuxMuxProjectedInterfaceAttribute.cs
```diff
@@ -0,0 +1,18 @@
...
+using System.Text;
+
+namespace WinRT.Interop
+{
+    /// <summary>
+    /// This type signals that the type it is applied to is projected into .NET from either a Windows.UI.Xaml type or a Microsoft.UI.Xaml type.
+    /// For this type, the GuidAttribute is not used and instead the GetGuidSignature method must be called to get the IID or generic IID signature part of the type.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Struct | AttributeTargets.Interface | AttributeTargets.Delegate, AllowMultiple = false, Inherited = false)]
+    internal sealed class WuxMuxProjectedTypeAttribute(string wuxIID, string muxIID) : Attribute
```
Sergio0694, line 12: Just double checking: are we sure this file is not also being included in the embedded sources, for consumers of that? Because if so, we can't use C# 12 features there, as those also support .NET 6 TFMs, which use C# 10.

src/WinRT.Runtime/Projections/NotifyCollectionChangedEventHandler.cs
```diff
@@ -13,6 +13,7 @@ namespace ABI.System.Collections.Specialized
 
     [global::System.ComponentModel.EditorBrowsable(global::System.ComponentModel.EditorBrowsableState.Never)]
     [Guid("8B0909DC-2005-5D93-BF8A-725F017BAA8D")]
+    [WuxMuxProjectedType(wuxIID: "ca10b37c-f382-4591-8557-5e24965279b0", muxIID: "8B0909DC-2005-5D93-BF8A-725F017BAA8D")]
```
Sergio0694, line 16: Nit: can we always use uppercase for all GUIDs for consistency? Applies to all occurrences in this PR.

Sergio0694:
> "The Windows.UI.Colors class isn’t available in the SDK projection, as it is a WUX type. However, the issue is due to CsWinRT’s behaviors regarding additions, you can’t project Windows.UI.Colors without also projecting Windows.UI.Color, which would conflict with the SDK projection and generate a warning"

Given this is a single, well known case, perhaps we can special case it so that when `Windows.UI.Colors` is projected, we skip also projecting `Windows.UI.Color`, as we can rely on it already being present in the Windows SDK projections assembly? Or, alternatively, at the very least, perhaps we could generate some type-identity nonsense or whatever to avoid the warning?

src/Projections/Windows.UI.Xaml/Windows.UI.Xaml.csproj
jkoritzinsky: This projection is just for testing. I'll take a look at cleaning this up, but I don't see it as high-priority.

src/WinRT.Runtime/Interop/WuxMuxProjectedInterfaceAttribute.cs
jkoritzinsky: I'll change this to use C#10 for the embedding scenario.

jkoritzinsky:
> > "The Windows.UI.Colors class isn’t available in the SDK projection, as it is a WUX type. However, the issue is due to CsWinRT’s behaviors regarding additions, you can’t project Windows.UI.Colors without also projecting Windows.UI.Color, which would conflict with the SDK projection and generate a warning"
> 
> Given this is a single, well known case, perhaps we can special case it so that when `Windows.UI.Colors` is projected, we skip also projecting `Windows.UI.Color`, as we can rely on it already being present in the Windows SDK projections assembly? Or, alternatively, at the very least, perhaps we could generate some type-identity nonsense or whatever to avoid the warning?

Honestly, some manual type identity nonsense isn't the worst idea. I'll try that out in this PR and see what I can do.

src/Perf/IIDOptimizer/SignatureGenerator.cs
```diff
@@ -73,7 +77,7 @@ public SignaturePart GetSignatureParts(TypeReference type)
 
             var typeDef = type.Resolve();
 
-            var helperType = new TypeReference($"ABI.{typeDef.Namespace}", typeDef.Name, typeDef.Module, assembly.MainModule);
+            TypeReference helperType = type.GetCswinrtAbiTypeDefinition(winRTRuntimeAssembly);
```
jkoritzinsky, line 80:
```suggestion
            var helperType = new TypeReference($"ABI.{typeDef.Namespace}", typeDef.Name, typeDef.Module, assembly.MainModule);
```

src/WinRT.Runtime/Projections/INotifyPropertyChanged.net5.cs
```diff
@@ -2,31 +2,36 @@
...
 #if EMBED
     internal
 #else
     public
 #endif
     static class INotifyPropertyChangedMethods
-    {
-        public static global::System.Guid IID { get; } = new Guid(new global::System.ReadOnlySpan<byte>(new byte[] { 0x01, 0x76, 0xB1, 0x90, 0x65, 0xB0, 0x6E, 0x58, 0x83, 0xD9, 0x9A, 0xDC, 0x3A, 0x69, 0x52, 0x84 }));
+    {
+        public static global::System.Guid IID { get; } = GuidGenerator.GetWuxMuxIID(typeof(INotifyPropertyChanged).GetCustomAttribute<WuxMuxProjectedTypeAttribute>());
```
Sergio0694, line 19: I don't really love this bit of performance hit on NativeAOT (I don't think it'll be able to correctly preinitialize everything). Can we replace all of these `IID` initialization with a helper method that simply returns the right value based on the feature switch? Then we can still keep both in RVA fields, and the linker will just drop the one that's unused.

I mean something like:
```csharp
public static global::System.Guid IID { get; } = new(GetIID());

private static ReadOnlySpan<byte> GetIID()
{
    return FeatureSwitches.IsMicrosoftUIXaml
        ? new global::System.ReadOnlySpan<byte>(new byte[] { ... })
        : new global::System.ReadOnlySpan<byte>(new byte[] { ... });
}
```

src/WinRT.Runtime/Projections.cs
```diff
@@ -25,9 +25,33 @@ namespace WinRT
 #endif
     static class Projections
     {
+        internal enum UiXamlMode
+        {
+            MicrosoftUiXaml,
+            WindowsUiXaml
+        }
+
+        private static UiXamlMode GetUIXamlModeSetting()
```
Sergio0694, line 34: You should be able to replace all this with `FeatureSwitches` now that that's in place. Should be very easy also to make everything trimmable, basically just minor changes than eg. #1446.

src/WinRT.Runtime/Projections/INotifyPropertyChanged.net5.cs
jkoritzinsky: I was just trying to avoid typing out the same guids a billion times, but I can manually copy them I guess.

Sergio0694: Where else do we have to copy them, and why can we not just reference this `IID` field from there? Do you just mean the copies as in, the string versions that are literal expressions for the attribute arguments?

jkoritzinsky: Yeah mainly that. There's also some duplication on the delegate types that's harder to get rid of, but that's okay. I'll clean it up.

src/WinRT.Runtime/Configuration/FeatureSwitches.cs
```diff
@@ -77,4 +77,44 @@ private static bool GetConfigurationValue(string propertyName, ref int cachedRes
 
         return isEnabled;
     }
+
+    private static Projections.WuxMuxMode _wuxMuxMode = Projections.WuxMuxMode.Unknown;
+
+    internal static Projections.WuxMuxMode WuxMuxMode => GetWuMuxMode("CsWinRT.WuxMuxMode", ref _wuxMuxMode);
```
Sergio0694, line 83: So, thinking about this some more, I think we can simplify it to not even need any new enum types nor new feature switch helper methods compared to what we had before. We should also follow the same conventions as the other feature switches. Here's how I think we should structure this:
- We add this switch: `CSWINRT_USE_WINDOWS_UI_XAML_PROJECTIONS`
- The MSBuild property will be: `CsWinRTUseWindowsUIXamlProjections` (defaults to `false`)
- Here in `FeatureSwitches`, we just need the `UseWindowsUIXamlProjections` property (a `bool`)
- Everywhere else we simply check and branch on that flag
- We can have a single throw path for below .NET 5, in `Projections.cs` (we already have it anyway)

jkoritzinsky: We need the tri-state in MSBuild so we can detect in the SDK targets that you're trying to use the WUX projections but explicitly set MUX mode. We can use a boolean property for the feature switch though.

src/WinRT.Runtime/Projections/INotifyPropertyChanged.net5.cs
```diff
@@ -67,6 +77,7 @@ private static unsafe int Do_Abi_add_PropertyChanged_0(IntPtr thisPtr, IntPtr ha
                 *token = default;
                 try
                 {
+                    global::System.Diagnostics.Debugger.Launch();
```
Sergio0694, line 80: Leftover `Debugger.Launch()` call

src/Authoring/WinRT.SourceGenerator/TypeMapper.cs
```diff
@@ -0,0 +1,125 @@
...
+                    { "System.Collections.Generic.IEnumerator`1", new MappedType("Windows.Foundation.Collections", "IIterator`1", "Windows.Foundation.FoundationContract") },
+                    { "System.Collections.Generic.KeyValuePair`2", new MappedType("Windows.Foundation.Collections", "IKeyValuePair`2", "Windows.Foundation.FoundationContract") },
+                    { "System.Collections.Generic.IReadOnlyDictionary`2", new MappedType("Windows.Foundation.Collections", "IMapView`2", "Windows.Foundation.FoundationContract") },
+                    { "System.Collections.Generic.IDictionary`2", new MappedType("Windows.Foundation.Collections", "IMap`2", "Windows.Foundation.FoundationContract") },
+                    { "System.Collections.Generic.IReadOnlyList`1", new MappedType("Windows.Foundation.Collections", "IVectorView`1", "Windows.Foundation.FoundationContract") },
+                    { "System.Collections.Generic.IList`1", new MappedType("Windows.Foundation.Collections", "IVector`1", "Windows.Foundation.FoundationContract") },
+                    { "Windows.UI.Color", new MappedType("Windows.UI", "Color", "Windows.Foundation.UniversalApiContract", true, true) },
```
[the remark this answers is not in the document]
jkoritzinsky: Yes, but that applies to the authoring of the projection assembly itself, not the basic Wux/Mux switch (so I'd rather track it separately).

src/Perf/IIDOptimizer/SignatureGenerator.cs
```diff
@@ -170,6 +174,13 @@ public SignaturePart GetSignatureParts(TypeReference type)
                 return new RuntimeClassSignature(type, GetSignatureParts(iface));
             }
 
+            // For types projected from WUX or MUX into .NET, we'll need to do a runtime lookup for the IID.
+            // TODO-WuxMux: We can instead take an option in the IID optimizer to hard-code the lookup for WUX or MUX when specified, which would be more efficient for scenarios where this is possible.
```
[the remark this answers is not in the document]
jkoritzinsky: I think we can do it, but since it's not needed for the MVP of this feature, I'd rather not fight MSBuild and the command line to get it working.

I don't remember if we run the IID optimizer on library assemblies too though, and I think you could build a library that works with both WUX and MUX if you're writing it in C# and it only uses the mapped WUX/MUX types.

src/WinRT.Runtime/Projections/INotifyCollectionChanged.net5.cs
```diff
@@ -35,19 +37,24 @@ public static unsafe (Action<global::System.Collections.Specialized.NotifyCollec
...
+            => FeatureSwitches.IsWuxMode
+                ? new(new byte[] { 0xd5, 0x67, 0xb1, 0x28, 0x31, 0x1a, 0x5b, 0x46, 0x9b, 0x25, 0xd5, 0xc3, 0xae, 0x68, 0x6c, 0x40 })
+                : new(new byte[] { 0xe1, 0x55, 0x1, 0x53, 0xa5, 0x28, 0x93, 0x56, 0x87, 0xce, 0x30, 0x72, 0x4d, 0x95, 0xa0, 0x6d });
+        public static IntPtr AbiToProjectionVftablePtr => INotifyCollectionChanged.Vftbl.AbiToProjectionVftablePtr;
     }
 
     [DynamicInterfaceCastableImplementation]
     [EditorBrowsable(EditorBrowsableState.Never)]
-    [Guid("530155E1-28A5-5693-87CE-30724D95A06D")]
+    [Guid("530155E1-28A5-5693-87CE-30724D95A06D")]
```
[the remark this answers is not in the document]
jkoritzinsky: I'd rather leave it as otherwise we'd get an effectively-randomly-generated guid for `Type.GUID` instead of maintaining existing behavior.

src/WinRT.Runtime/Projections/INotifyCollectionChanged.net5.cs
```diff
@@ -35,19 +36,24 @@ public static unsafe (Action<global::System.Collections.Specialized.NotifyCollec
...
-
-        public static global::System.Guid IID { get; } = new Guid(new global::System.ReadOnlySpan<byte>(new byte[] { 0xE1, 0x55, 0x01, 0x53, 0xA5, 0x28, 0x93, 0x56, 0x87, 0xCE, 0x30, 0x72, 0x4D, 0x95, 0xA0, 0x6D }));
-
-        public static IntPtr AbiToProjectionVftablePtr => INotifyCollectionChanged.Vftbl.AbiToProjectionVftablePtr;
+        }
+        public static global::System.Guid IID { get; } = new(GetIID());
+
+        private static ReadOnlySpan<byte> GetIID()
+            => FeatureSwitches.IsWuxMode
+                ? new(new byte[] { 0xd5, 0x67, 0xb1, 0x28, 0x31, 0x1a, 0x5b, 0x46, 0x9b, 0x25, 0xd5, 0xc3, 0xae, 0x68, 0x6c, 0x40 })
```
Sergio0694, line 44: Need to leave the `ROS<byte>` cast in all of these, otherwise we allocate an array instead of using RVA fields. Leaving this comment here but applies to all such IID initialization paths introduced in this PR, for all WUX/MUX types.

src/WinRT.Runtime/Projections/INotifyCollectionChanged.net5.cs
```diff
@@ -35,21 +37,22 @@ public static unsafe (Action<global::System.Collections.Specialized.NotifyCollec
...
             });
             return eventSource.EventActions;
-        }
-
-        public static global::System.Guid IID { get; } = new Guid(new global::System.ReadOnlySpan<byte>(new byte[] { 0xE1, 0x55, 0x01, 0x53, 0xA5, 0x28, 0x93, 0x56, 0x87, 0xCE, 0x30, 0x72, 0x4D, 0x95, 0xA0, 0x6D }));
-
-        public static IntPtr AbiToProjectionVftablePtr => INotifyCollectionChanged.Vftbl.AbiToProjectionVftablePtr;
+        }
+
+        public static global::System.Guid IID { get; } = GuidGenerator.GetWuxMuxIID(typeof(INotifyCollectionChanged).GetCustomAttribute<WuxMuxProjectedTypeAttribute>());
```
Sergio0694, line 42: Can we not hardcode this by switching on the flag here too, to avoid the attribute lookup?

src/WinRT.Runtime/Projections/INotifyCollectionChanged.net5.cs
```diff
@@ -35,21 +37,22 @@ public static unsafe (Action<global::System.Collections.Specialized.NotifyCollec
...
+        public static IntPtr AbiToProjectionVftablePtr => INotifyCollectionChanged.Vftbl.AbiToProjectionVftablePtr;
     }
 
     [DynamicInterfaceCastableImplementation]
     [EditorBrowsable(EditorBrowsableState.Never)]
-    [Guid("530155E1-28A5-5693-87CE-30724D95A06D")]
+    [Guid("530155E1-28A5-5693-87CE-30724D95A06D")]
+    [WuxMuxProjectedType(wuxIID: "CF75D69C-F2F4-486B-B302-BB4C09BAEBFA", muxIID: "530155E1-28A5-5693-87CE-30724D95A06D")]
```
Sergio0694, line 49: Do we actually need this, if we hardcode this above? Same for the vftbl type below.

src/WinRT.Runtime/Projections/INotifyPropertyChanged.net5.cs
```diff
@@ -2,32 +2,34 @@
...
 #if EMBED
     internal
 #else
     public
 #endif
     static class INotifyPropertyChangedMethods
-    {
-        public static global::System.Guid IID { get; } = new Guid(new global::System.ReadOnlySpan<byte>(new byte[] { 0x01, 0x76, 0xB1, 0x90, 0x65, 0xB0, 0x6E, 0x58, 0x83, 0xD9, 0x9A, 0xDC, 0x3A, 0x69, 0x52, 0x84 }));
+    {
+        public static global::System.Guid IID { get; } = GuidGenerator.GetWuxMuxIID(typeof(INotifyPropertyChanged).GetCustomAttribute<WuxMuxProjectedTypeAttribute>());
```
Sergio0694, line 19: Not sure why the comment here was resolved, I'm still seeing the attribute lookup? Or do you just plan on addressing all of these later when you get around to rebasing this PR? I won't leave other comments about this for now 😄

src/WinRT.Runtime/Projections/INotifyPropertyChanged.net5.cs
```diff
@@ -92,6 +97,7 @@ private static unsafe int Do_Abi_add_PropertyChanged_0(IntPtr thisPtr, IntPtr ha
                 *token = default;
                 try
                 {
+                    global::System.Diagnostics.Debugger.Launch();
```
Sergio0694, line 100: This also was resolved but the leftover `Debugger.Launch()` call is still here? 😅

src/WinRT.Runtime/Projections/PropertyChangedEventArgs.cs
```diff
@@ -86,7 +91,78 @@ public unsafe ObjectReferenceValue CreateInstance(string name)
                 MarshalInspectable<object>.DisposeAbi(__innerInterface);
             }
         }
+    }
+
+
+
+    [global::WinRT.ObjectReferenceWrapper(nameof(_obj))]
```
Sergio0694, line 98: This attribute should only be used downlevel. On .NET, shouldn't we also implement `IWinRTObject` here?

src/WinRT.Runtime/Projections/NotifyCollectionChangedEventHandler.cs
```diff
@@ -41,28 +42,32 @@ static unsafe NotifyCollectionChangedEventHandler()
             var nativeVftbl = ComWrappersSupport.AllocateVtableMemory(typeof(NotifyCollectionChangedEventHandler), sizeof(global::WinRT.Interop.IDelegateVftbl));
             Marshal.StructureToPtr(AbiToProjectionVftable, nativeVftbl, false);
             AbiToProjectionVftablePtr = nativeVftbl;
+
+            IID = Projections.UiXamlModeSetting == Projections.UiXamlMode.WindowsUiXaml
```
Sergio0694, line 46: Nit: this should also just be a switch with two RVA fields, rather than parsing at runtime

src/WinRT.Runtime/Projections/INotifyPropertyChanged.net5.cs
jkoritzinsky: I think I had some lost commits between rebases and different machines, sorry.

dongle-the-gadget:
Windows.UI.Xaml (System XAML) + Microsoft.UI.Xaml (WinUI 3)

asklar:
related: #723

Sergio0694:
0 difference in size in our minimal sample 😄

image

dongle-the-gadget:
Does it affect size when you switch to WUX mode?

Sergio0694:
Possibly, but that's not really a concern. I can test that too though.

driver1998:
Confirmed that data binding and INotifyPropertyChanged works on WUX.

But without XAML codegen it'll still be PITA to use...

image

dongle-the-gadget:
The XAML codegen is the work of another team.
________________________________

driver1998:
> The XAML codegen is the work of another team.

That's the issue I am concerned about right now, that other team might see System XAML as deprecated technology and refuse to update. Just like what happened to ARM64 .NET Framework.

Sergio0694:
But .NET Framework does support arm64 😅

dongle-the-gadget:
It does - 4.8.1

src/WinRT.Runtime/Interop/IID.tt
```diff
@@ -39,18 +39,25 @@ var entries = new (string Name, string IID, bool IsPublic)[]
...
+    ("MUX_INotifyPropertyChanged", "90B17601-B065-586E-83D9-9ADC3A695284", false),
+    ("WUX_INotifyPropertyChanged", "CF75D69C-F2F4-486B-B302-BB4C09BAEBFA", false),
+    ("MUX_INotifyCollectionChanged", "530155E1-28A5-5693-87CE-30724D95A06D", false),
+    ("WUX_INotifyCollectionChanged", "28B167D5-1A31-465B-9B25-D5C3AE686C40", false),
+    ("MUX_INotifyCollectionChangedEventArgsFactory", "5108EBA4-4892-5A20-8374-A96815E0FD27", false),
+    ("WUX_INotifyCollectionChangedEventArgsFactory", "B30C3E3A-DF8D-44A5-9A38-7AC0D08CE63D", false),
+    ("MUX_INotifyCollectionChangedEventArgs", "DA049FF2-D2E0-5FE8-8C7B-F87F26060B6F", false),
+    ("WUX_INotifyCollectionChangedEventArgs", "4CF68D33-E3F2-4964-B85E-945B4F7E2F21", false),
+    ("MUX_NotifyCollectionChangedEventHandler", "CA10B37C-F382-4591-8557-5E24965279B0", false),
+    ("WUX_NotifyCollectionChangedEventHandler", "8B0909DC-2005-5D93-BF8A-725F017BAA8D", false),
```
manodasanW, line 51: Is this the WUX one for the MUX IID?  Just noticed it matched up with the previous iid for this.

jkoritzinsky: Yep these need to be swapped. I checked the rest and they should be fine.

Sergio0694: Thank you for the quick fix Jeremy! 😄

src/Projections/Windows.UI.Xaml/Windows.UI.Xaml.csproj
```diff
@@ -0,0 +1,28 @@
...
+    <ProjectReference Include="..\..\Authoring\WinRT.SourceGenerator\WinRT.SourceGenerator.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" SetPlatform="Platform=x64"/>
+    <ProjectReference Include="..\Windows\Windows.csproj" />
+  </ItemGroup>
+
+  <PropertyGroup>
+    <CsWinRTFilters>
+-exclude Windows
+-include Windows.UI.Xaml
+-exclude Windows.UI.Xaml.Media.Animation.ConditionallyIndependentlyAnimatableAttribute
+-addition_exclude Windows.UI.Xaml.Media.Animation
```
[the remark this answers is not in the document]
jkoritzinsky: These additions are included any time you include the namespace. The Windows SDK projection as-is includes some types in this namespace, so it ends up providing the types in the additions publicly today.

This change allows the WUX projection to not doubly-define types that are already exposed publicly.

src/WinRT.Runtime/Projections/NotifyCollectionChangedEventArgs.cs
```diff
@@ -152,52 +159,56 @@ public static ObjectReferenceValue CreateMarshaler2(global::System.Collections.S
...
+        public static (int length, IntPtr data) FromManagedArray(global::System.Collections.Specialized.NotifyCollectionChangedEventArgs[] array) => MarshalInterfaceHelper<global::System.Collections.Specialized.NotifyCollectionChangedEventArgs>.FromManagedArray(array, FromManaged);
+        public static void DisposeMarshalerArray(MarshalInterfaceHelper<global::System.Collections.Specialized.NotifyCollectionChangedEventArgs>.MarshalerArray array) => MarshalInterfaceHelper<global::System.Collections.Specialized.NotifyCollectionChangedEventArgs>.DisposeMarshalerArray(array);
+        public static unsafe void DisposeAbiArray(object box) => MarshalInspectable<object>.DisposeAbiArray(box);
+
+        public static string GetGuidSignature()
+        {
+            if (FeatureSwitches.UseWindowsUIXamlProjections)
+            {
+                return "rc(Windows.UI.Xaml.Interop.NotifyCollectionChangedEventArgs;{4cf68d33-e3f2-4964-b85e-945b4f7e2f21})";
+            }
```
manodasanW, line 210: We can fix this separately, but I think the **MUX** IID is wrong here as it is using the WUX one.  It should probably be [Guid(3657736178u, 53984, 24552, 140, 123, 248, 127, 38, 6, 11, 111)]

src/WinRT.Runtime/Projections/PropertyChangedEventArgs.cs
```diff
@@ -1,175 +1,178 @@
...
+{
+    internal sealed unsafe class PropertyChangedEventArgsRuntimeClassFactory
+    {
+        private readonly ObjectReference<IUnknownVftbl> _obj;
+
+        public PropertyChangedEventArgsRuntimeClassFactory()
+        {
+            if (FeatureSwitches.UseWindowsUIXamlProjections)
+            {
+                _obj = ActivationFactory.Get("Windows.UI.Xaml.Data.PropertyChangedEventArgs").As<IUnknownVftbl>(IID.IID_WUX_PropertyChangedEventArgsRuntimeClassFactory);
```
manodasanW, line 20: nit: this can be addressed after, but we probably want to avoid the additional `As` call and object construction by doing what was done here: https://github.com/microsoft/CsWinRT/blob/c3c4aea1aced3d25e6f142096e599e3000f0740a/src/WinRT.Runtime/Projections/NotifyCollectionChangedEventArgs.cs#L27-L33

src/WinRT.Runtime/Projections/PropertyChangedEventArgs.cs
```diff
@@ -1,175 +1,178 @@
...
+        public static unsafe void DisposeAbiArray(object box) => MarshalInspectable<object>.DisposeAbiArray(box);
+
+        public static string GetGuidSignature()
+        {
+            if (FeatureSwitches.UseWindowsUIXamlProjections)
+            {
+                return "rc(Windows.UI.Xaml.Data.NotifyPropertyChangedEventArgs;{4f33a9a0-5cf4-47a4-b16f-d7faaf17457e})";
+            }
+
+            return "rc(Microsoft.UI.Xaml.Data.NotifyPropertyChangedEventArgs;{4f33a9a0-5cf4-47a4-b16f-d7faaf17457e})";
```
manodasanW, line 175: Same here, but here also I think Notify needs to be removed from the runtime class name.  Probably can be addressed separately as it is an issue before this PR.

src/Authoring/WinRT.SourceGenerator/WinRT.SourceGenerator.csproj
```diff
@@ -15,6 +15,7 @@
     <Description>C#/WinRT Authoring Source Generator Preview $(VersionString)</Description>
     <AssemblyTitle>C#/WinRT Authoring Source Generator Preview v$(VersionString)</AssemblyTitle>
     <Copyright>Copyright (c) Microsoft Corporation. All rights reserved.</Copyright>
+	  <IsRoslynComponent>true</IsRoslynComponent>
```
manodasanW, line 18: What is the implication of this and nit: identing.

manodasanW (approved):

src/Authoring/WinRT.SourceGenerator/WinRT.SourceGenerator.csproj
Sergio0694: I think that's just to enable the improved debugging experience, [see here](https://github.com/dotnet/roslyn-sdk/issues/850).

src/WinRT.Runtime/Projections/PropertyChangedEventArgs.cs
Sergio0694: Good catch, fixed in 000ca9bef419e0fdb3e90df1696e8b0ca341043a 👍

src/Projections/Windows.UI.Xaml/Windows.UI.Xaml.csproj
```diff
@@ -0,0 +1,28 @@
...
+    <ProjectReference Include="..\..\cswinrt\cswinrt.vcxproj" />
+    <InternalsVisibleTo Include="UnitTest" />
+    <ProjectReference Include="..\..\Authoring\WinRT.SourceGenerator\WinRT.SourceGenerator.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" SetPlatform="Platform=x64"/>
+    <ProjectReference Include="..\Windows\Windows.csproj" />
+  </ItemGroup>
+
+  <PropertyGroup>
+    <CsWinRTFilters>
+-exclude Windows
+-include Windows.UI.Xaml
```
dongle-the-gadget, line 22: Don't we need to also project `Windows.UI.Colors` and exclude the `Windows.UI.Color` addition? The Windows SDK projection doesn't include the `Colors` class as it is part of WUX.

Sergio0694: What was the reason again `Windows.UI.Color` is not projected in the Windows SDK too?

dongle-the-gadget: The Color struct is projected, but not the Colors class as it is WUX, that's why WinUI 3 has its own implementation of Colors.

Sergio0694: No I know that but I mean why does WinUI 3 have its own implementation, doesn't the `Color` type and the `Colors` type not have any dependencies on UWP anyway? I mean why couldn't WinUI 3 just reuse `Colors` too in the same way it's also reusing `Color` anyway? Am I missing something?

dongle-the-gadget: The `Windows.UI.Colors` class is implemented in WUX DLL, and so to use it you would have to load WUX, which I assume is unideal.

Screenshot of Windows.UI.Colors in OleViewDotNet

manodasanW: That was probably one of the reasons.  The other is probably that they want the capability to easily add to the Colors static class in an undocked way if they wanted without touching the system one.  Whereas the Color struct is most likely not going to change.

Sergio0694: Synced with Mano offline, we can revisit this and fix this in a follow up PR, resolving for now.

PR #1426, opened 2024-01-05
Fix diagnostics for namespaces with internal partial types

### LINKED ISSUE #1425: Error message for 'A public type has a namespace that shares no common prefix' doesn't specify the offending type
When using `<CsWinRTComponent>true</>` and tripping this error:

```
1>FooBar.cs(5,30,5,37): error CsWinRT1001: A public type has a namespace ('Foo.Bar.Baz') that shares no common prefix with other namespaces ('MyLibrary'). All types within a Windows Metadata file must exist in a sub namespace of the namespace that is implied by the file name. (https://github.com/microsoft/CsWinRT/tree/master/src/Authoring/WinRT.SourceGenerator/AnalyzerReleases.Unshipped.md)
```

It says what the namespace conflict is but doesn't give the right file/line number or tell me which type is the problem.

manodasanW:
The diagnostics analyzer for namespaces found all the namespaces which had public types and then made sure they are a subset of the one which the authored component is.  As part of determining public types, the check also considered partial types to be public with the assumption that it would later to be checked to figure out whether it is really public or not.  But for the namespace diagnostics this wasn't done as it is done in other scenarios.  In addition, partial types don't matter for namespace diagnostics as the diagnostics analyzer will see at least one of the partial types with its declared accessibility.  Due to that, this PR addresses this issue by making the namespace check only look for public types and not partial ones.

Fixes #1425

### DIFF
--- a/src/Authoring/WinRT.SourceGenerator/Generator.cs
+++ b/src/Authoring/WinRT.SourceGenerator/Generator.cs
@@ -224,6 +224,10 @@ namespace Generator
             //   once from the namespace declaration and once from the member's declaration
             if (syntaxNode is NamespaceDeclarationSyntax @namespace)
             {
+                // We only include the namespace if it has a public type as otherwise it won't.
+                // be projected.  For partial types, there would be one instance that we encounter
+                // which declares the accessibility and we will use that to determine the accessibility
+                // of the type for the purpose of determining whether to include the namespace.
                 if (HasSomePublicTypes(syntaxNode))
                 {
                     Namespaces.Add(@namespace);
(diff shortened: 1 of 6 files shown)


src/Authoring/WinRT.SourceGenerator/Generator.cs
```diff
@@ -224,6 +224,10 @@ public void OnVisitSyntaxNode(SyntaxNode syntaxNode)
             //   once from the namespace declaration and once from the member's declaration
             if (syntaxNode is NamespaceDeclarationSyntax @namespace)
             {
+                // We only include the namespace if it has a public type as otherwise it won't.
```
vineeththomasalex, line 227: Nit 
```suggestion
                // We only include the namespace if it has a public type as otherwise it won't
```

vineeththomasalex (approved):

PR #1429, opened 2024-01-07
Remove LINQ expression dependencies on NativeAOT

Sergio0694:
This PR attempts to fix the IL3050 warnings in CsWinRT due to System.Linq.Expressions being pulled in. The goal is to just remove that dependency entirely when on NativeAOT, and allow the linker to correctly trim all of that code, which should also improve the binary size a bit more. There were a few issues preventing this from happening:
- Some checks for NativeAOT (via `RuntimeFeature`) were using unsupported patterns ([see here](https://github.com/dotnet/runtime/blob/main/docs/design/tools/illink/feature-checks.md))
- Some paths were unconditionally relying on LINQ expressions (eg. `Nullable<T>`)

This PR fixes both of those issues for all occurrences.


dongle-the-gadget:
Seems like we are missing some types in `GetValueDelegateForFunctionPointer`?
- Point
- Size
- Rect
- Array types

Sergio0694:
I just copied those from what was being registered in `Projections.cctor`. How come those weren't there? 🤔
As for array types... Why would those ever be in a `Nullable<T>` instance? Isn't `IReference<T>` just WinRT's "boxed" type?

dongle-the-gadget:
`IReference<T>` can also store arrays.

src/WinRT.Runtime/Projections/EventHandler.cs
```diff
@@ -124,7 +124,14 @@ static class EventHandler<T>
         static unsafe EventHandler()
         {
 #if NET
-            if (!RuntimeFeature.IsDynamicCodeCompiled || EventHandlerMethods<T>.AbiToProjectionVftablePtr != default)
+            if (!RuntimeFeature.IsDynamicCodeCompiled)
```
manodasanW, line 127: Good catch on the unsupported patterns, did realize it.

src/WinRT.Runtime/Projections/EventHandler.cs
```diff
@@ -200,11 +207,16 @@ public NativeDelegateWrapper(ObjectReference<global::WinRT.Interop.IDelegateVftb
...
+                if (!RuntimeFeature.IsDynamicCodeCompiled)
+                {
+                    EventHandlerMethods<T>._Invoke(_nativeDelegate, sender, args);
+
+                    return;
+                }
 #endif
-                if (!useDynamicInvoke)
+
+                if (EventHandlerMethods<T>._Invoke == null)
```
manodasanW, line 219:
```suggestion
                if (EventHandlerMethods<T>._Invoke != null)

```

Sergio0694: Oops, good catch! Weird though that no test caught this 😅

Sergio0694:
Can confirm that this fully removes the IL3050 warning from LINQ expressions when publishing! 🎉

image

Removing the assembly entirely is handled by #1435.

Sergio0694:
Worth noting, this also saves ~128 KB!

src/WinRT.Runtime/Projections/EventHandler.cs
```diff
@@ -88,130 +88,149 @@ public unsafe static bool InitRcwHelper(delegate*<IObjectReference, object, T, v
...
+
+                if (EventHandlerMethods<T>._Invoke != null)
                 {
                     EventHandlerMethods<T>._Invoke(_nativeDelegate, sender, args);
-                }
+                }
                 else
                 {
+                    // Same as in the static constructor, we initialize the ABI delegate type manually here if needed
+                    global::System.Threading.Interlocked.CompareExchange(ref _abi_invoke_type, Projections.GetAbiDelegateType(typeof(void*), typeof(IntPtr), Marshaler<T>.AbiType, typeof(int)), null);
```
manodasanW, line 230: Any opportunity here to avoid calling `Projections.GetAbiDelegateType` on every invoke when AOT or the new optimizer changes isn't being used?

Sergio0694: Ah, yeah good catch! Added a check in 99d4b2dd5d466baa547b8c018cc90bf54d754bb3 to avoid unnecessary calls 🙂

src/WinRT.Runtime/Projections/Nullable.cs
```diff
@@ -1,66 +1,171 @@
...
+            {
+                global::WinRT.ExceptionHelpers.SetErrorInfo(__exception__);
+                return global::WinRT.ExceptionHelpers.GetHRForException(__exception__);
+            }
+            return 0;
+        }
+
+        internal static unsafe Delegate GetValueDelegateForAbi()
+        {
+            if (typeof(T) == typeof(int) ||
```
manodasanW, line 111: So because of [this](https://github.com/microsoft/CsWinRT/blob/ae695510af515a373bc010cf46aa188736b2b8de/src/WinRT.Runtime/ComWrappersSupport.cs#L594-L764), I feel like most of the types listed here won't come to this code path.  Do you think having some specific nullable optimizations for each type is better or do you think have a generic version that handles all here is better?  The main difference I see is with the former, we can list them as `UnManagedCallersOnly` whereas here we would need to do `GetfunctionPointerForDelegate`.  Or should we keep both as is, and have this as a fallback in case that code path wasn't used?

src/WinRT.Runtime/Projections/Nullable.cs
```diff
@@ -1,66 +1,171 @@
...
+#endif
+            {
+                Nullable<T>.Vftbl.get_Value_0_Type ??= Projections.GetAbiDelegateType(typeof(void*), Marshaler<T>.AbiType.MakeByRefType(), typeof(int));
+
+                return Delegate.CreateDelegate(
+                    Nullable<T>.Vftbl.get_Value_0_Type,
+                    typeof(BoxedValueIReferenceImpl<T>).GetMethod(nameof(Do_Abi_get_Value_0), BindingFlags.NonPublic | BindingFlags.Static)!.MakeGenericMethod(Marshaler<T>.AbiType));
+            }
+
+            throw new NotSupportedException("Failed to get the marshalling delegate for BoxedValueIReferenceImpl`1.");
```
manodasanW, line 159: Yea this should be fine here.  For all the other types, we should have handled them [here](https://github.com/microsoft/CsWinRT/blob/ae695510af515a373bc010cf46aa188736b2b8de/src/WinRT.Runtime/ComWrappersSupport.cs#L204C46-L212) using `StructTypeDetails` assuming you are on new CsWinRT and its source generator.  Maybe add to the error message the type name for which we failed.

src/WinRT.Runtime/Projections/Nullable.cs
Sergio0694: Oh, that's interesting, I didn't notice that. To clarify, all changes in this PR were solely driven by trying to fix all AOT warnings and allow all of that LINQ expression code to be trimmed out, basically I did the minimum amount of changes to satisfy that. I'm still seeing generic instantiations for all of these types, for instance:

image

So I'm inclined to say keeping both code paths can't hurt, and it might still help with trimming since it would make those generic instantiations get the specialized paths (even though we're not 100% sure they're actually executed at runtime). I would say perhaps let's just keep this for now and then we can potentially look into removing code once we've double checked that it is never in fact used anywhere, and also that it doesn't reintroduce any trimming regressions? 🤔

What do you think?

src/WinRT.Runtime/Projections/Nullable.cs
```diff
@@ -69,164 +174,326 @@ static unsafe BoxedValueIReferenceImpl()
...
+            if (typeof(T) == typeof(short)) return Nullable_short.GetGuidSignature();
+            if (typeof(T) == typeof(ushort)) return Nullable_ushort.GetGuidSignature();
+            if (typeof(T) == typeof(char)) return Nullable_char.GetGuidSignature();
+            if (typeof(T) == typeof(uint)) return Nullable_uint.GetGuidSignature();
+            if (typeof(T) == typeof(long)) return Nullable_long.GetGuidSignature();
+            if (typeof(T) == typeof(ulong)) return Nullable_ulong.GetGuidSignature();
+            if (typeof(T) == typeof(float)) return Nullable_float.GetGuidSignature();
+            if (typeof(T) == typeof(double)) return Nullable_double.GetGuidSignature();
+            if (typeof(T) == typeof(Guid)) return Nullable_guid.GetGuidSignature();
+            if (typeof(T) == typeof(global::System.TimeSpan)) return Nullable_TimeSpan.GetGuidSignature();
```
manodasanW, line 285: There is also technically `Nullable_Type`.

src/WinRT.Runtime/Projections/Nullable.cs
```diff
@@ -69,164 +174,326 @@ static unsafe BoxedValueIReferenceImpl()
...
+                // This is not safe on AOT, so in that case we just can't get a delegate.
+                if (RuntimeFeature.IsDynamicCodeCompiled)
+#endif
+                {
+                    get_Value_0_Type ??= Projections.GetAbiDelegateType(typeof(void*), Marshaler<T>.AbiType.MakeByRefType(), typeof(int));
+
+                    return Marshal.GetDelegateForFunctionPointer(ptr, get_Value_0_Type);
+                }
+
+                throw new NotSupportedException("Failed to get the marshalling delegate for Nullable`1.");
```
manodasanW, line 365: I think we will need to improve this scenario a bit here, but we can defer to another PR (possibly as part of mines) as it would break now anyways at runtime.

src/WinRT.Runtime/Projections/Nullable.cs
manodasanW: Sounds good, we can keep both for now.

src/WinRT.Runtime/Projections/Nullable.cs
Sergio0694: Done in https://github.com/microsoft/CsWinRT/pull/1429/commits/71b246a3a7ce36e37abab4e841876c5e68826df1 🙂

src/WinRT.Runtime/Projections/Nullable.cs
manodasanW: Note that in your sceenshot, that is all BoxedIReference**Array**Impl which is different from `BoxedIReferenceImpl`.  I know we have worked to do with arrays possibly.

src/WinRT.Runtime/Projections/Nullable.cs
Sergio0694: Added to the signature switch for good measure. Didn't wire up the rest since we mentioned we're not even entirely sure that this code is actually ever being executed (though it is being compiled, that part we're sure of). We can revisit this in a follow up PR and add that complexity if needed I think 🙂

src/WinRT.Runtime/Projections/Nullable.cs
Sergio0694: The type with a single type parameter does seem to only have one instantiation 🤔

image

manodasanW: For the first set, those come from [here](https://github.com/microsoft/CsWinRT/blob/ae695510af515a373bc010cf46aa188736b2b8de/src/WinRT.Runtime/ComWrappersSupport.cs#L774-L828).  I wasn't sure if I wanted to create specializations for each one so I had created a BoxedValueIreferenceImpl that takes both the type and the ABI type as generics so it was AOT compatible and had implemented it that way.  With your changes, there might be opportunity in a future PR to make it again use the one generic one if that is better.

Sergio0694:
Updated diff, sizoscope doesn't report everything but the .dll-s are actually **310 KB smaller!!** 🎉

image

PR #1431, opened 2024-01-08
Disable runtime marshalling for the CsWinRT repo.

jlaanstra:
Disable runtime marshalling for the CsWinRT repo.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -34,6 +34,7 @@
     <TestsBuildTFMs Condition="'$(CIBuildReason)' == 'CI'">net6.0-windows10.0.19041.0</TestsBuildTFMs>
     <FunctionalTestsBuildTFMs>net6.0;net8.0</FunctionalTestsBuildTFMs>
     <WindowsAppSDKVerifyWinrtRuntimeVersion>false</WindowsAppSDKVerifyWinrtRuntimeVersion>
+    <DisableRuntimeMarshalling>True</DisableRuntimeMarshalling>
   </PropertyGroup>
 
   <PropertyGroup Condition="!$(IsTargetFrameworkNet5OrGreater)">
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -34,7 +34,6 @@
     <TestsBuildTFMs Condition="'$(CIBuildReason)' == 'CI'">net6.0-windows10.0.19041.0</TestsBuildTFMs>
     <FunctionalTestsBuildTFMs>net6.0;net8.0</FunctionalTestsBuildTFMs>
     <WindowsAppSDKVerifyWinrtRuntimeVersion>false</WindowsAppSDKVerifyWinrtRuntimeVersion>
-    <DisableRuntimeMarshalling>True</DisableRuntimeMarshalling>
   </PropertyGroup>
 
   <PropertyGroup Condition="!$(IsTargetFrameworkNet5OrGreater)">
--- a/src/Authoring/Directory.Build.props
+++ b/src/Authoring/Directory.Build.props
@@ -2,4 +2,8 @@
 
   <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.props', '$(MSBuildThisFileDirectory)../'))" />
 
+  <PropertyGroup>
+    <DisableRuntimeMarshalling>True</DisableRuntimeMarshalling>
+  </PropertyGroup>
+
 </Project>
--- a/src/Projections/Directory.Build.props
+++ b/src/Projections/Directory.Build.props
@@ -2,6 +2,9 @@
 
   <PropertyGroup>
     <SimulateCsWinRTNugetReference>true</SimulateCsWinRTNugetReference>
+
+    <!-- Projections should not require runtime marshalling. -->
+    <DisableRuntimeMarshalling>True</DisableRuntimeMarshalling>
   </PropertyGroup>
 
   <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.props', '$(MSBuildThisFileDirectory)../'))" />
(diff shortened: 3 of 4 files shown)


Sergio0694:
Just so I understand, is this to start catching all remaining cases and address them? Because I can't see things just working with no changes at all after enabling this flag, especially given I know there's still other PRs removing more marshalling stuff 😅

Either way, absolutely love this, and it was about time! 🚀

src/Directory.Build.props
```diff
@@ -31,6 +31,7 @@
     <TestsBuildTFMs>net6.0-windows10.0.19041.0;net7.0-windows10.0.19041.0</TestsBuildTFMs>
     <TestsBuildTFMs Condition="'$(CIBuildReason)' == 'CI'">net6.0-windows10.0.19041.0;net7.0-windows10.0.19041.0</TestsBuildTFMs>
     <WindowsAppSDKVerifyWinrtRuntimeVersion>false</WindowsAppSDKVerifyWinrtRuntimeVersion>
+    <DisableRuntimeMarshalling>True</DisableRuntimeMarshalling>
```
Sergio0694, line 34: I don't think this actually does anything. You need to define the attribute, no? You can use this:

```xml
<!-- Emit the [DisableRuntimeMarshalling] attribute for all .NET 8 projects -->
<ItemGroup Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net8.0'))">
  <AssemblyAttribute Include="System.Runtime.CompilerServices.DisableRuntimeMarshallingAttribute" />
</ItemGroup>
```

Sergio0694 (changes requested):

manodasanW:
> Just so I understand, is this to start catching all remaining cases and address them? Because I can't see things just working with no changes at all after enabling this flag, especially given I know there's still other PRs removing more marshalling stuff 😅
> 
> Either way, absolutely love this, and it was about time! 🚀

The goal is to see what is going to break.  But I am curious which marshaling thing you are referring to?

Sergio0694:
> "The goal is to see what is going to break."

Gotcha, yeah that makes perfect sense 👍

> "I am curious which marshaling thing you are referring to?"

I was thinking of those in #1416, plus I expect some more stuff will also break. For instance:
- Some `Marshal` calls here and there, don't think all of them will work with disabled marshalling
- Same as above, specifically in the `As<T>()` code checking for COM import interfaces. On a related note, we should likely also do a check for NAOT and just skip that entirely (or throw), because NAOT just doesn't support built-in COM import at all.

manodasanW:
> Unrelated, it seems like the CI isn't running? Is this because the source branch is from another fork? I remember Jeremy also had to close his PR and re-open it from a branch in this same repo (for #1421).

Right, we don't automatically run CI for forks.

src/Directory.Build.props
jlaanstra: https://github.com/dotnet/sdk/pull/34388 :)

manodasanW:
> > "The goal is to see what is going to break."
> 
> Gotcha, yeah that makes perfect sense 👍
> 
> > "I am curious which marshaling thing you are referring to?"
> 
> I was thinking of those in #1416, plus I expect some more stuff will also break. For instance:
> 
> * Some `Marshal` calls here and there, don't think all of them will work with disabled marshalling
> * Same as above, specifically in the `As<T>()` code checking for COM import interfaces. On a related note, we should likely also do a check for NAOT and just skip that entirely (or throw), because NAOT just doesn't support built-in COM import at all.

That PR was trying to port the changes already in the staging branch to master.  The staging branch should already have blittable signatures.  It would be interesting to see which Marshal functions break when the tests run.

src/Directory.Build.props
Sergio0694: Ooh really cool, I had completely missed this! Thank you for adding it! 😄
It would seem I have some PRs to make in my own repo then eheheh

Sergio0694:
The changes look good to me, but I have to say I feel like I'm missing something here, or maybe I just don't completely understand what does disabling runtime marshalling affect. We have several uses of stuff like `Marshal.StructureToPtr` in CsWinRT, as well as some delegates with managed signatures, which we're using to get native function pointers for (eg. we do this in the ABI `Nullable<T>` type. I'm not sure I get how can eg. all of that continue to work with the changes in this PR 🤔

jlaanstra:
> The changes look good to me, but I have to say I feel like I'm missing something here, or maybe I just don't completely understand what does disabling runtime marshalling affect. We have several uses of stuff like `Marshal.StructureToPtr` in CsWinRT, as well as some delegates with managed signatures, which we're using to get native function pointers for (eg. we do this in the ABI `Nullable<T>` type. I'm not sure I get how can eg. all of that continue to work with the changes in this PR 🤔

This change is here to find all remaining cases where we rely on runtime marshalling with the goal of eventually not having any.

Sergio0694:
No I know that. I think I got tripped up by the CI being green, that's the part that was not adding up.
Then I remembered that this PR isn't actually triggering the CI, so ok it all makes sense now 😄

Question: would it not be better _not_ to merge this PR though, to avoid completely breaking the CI in the branch until we fix everything? As in, we can just keep rebasing this one as it spots more issues, and then we can gradually fix them in other PRs, and only merge this one once the CI for this one is actually all green? And on that note, can't we close and reopen this from a branch in this same repo like I mentioned, to get the CI to actually run from this branch?

jlaanstra:
> No I know that. I think I got tripped up by the CI being green, that's the part that was not adding up. Then I remembered that this PR isn't actually triggering the CI, so ok it all makes sense now 😄
> 
> Question: would it not be better _not_ to merge this PR though, to avoid completely breaking the CI in the branch until we fix everything? As in, we can just keep rebasing this one as it spots more issues, and then we can gradually fix them in other PRs, and only merge this one once the CI for this one is actually all green? And on that note, can't we close and reopen this from a branch in this same repo like I mentioned, to get the CI to actually run from this branch?

A manually kicked off build is actually green right now.

Sergio0694:
No performance difference whatsoever.

manodasanW (approved):

PR #1432, opened 2024-01-09
Fix test issues discovered from enabling AOT for authoring scenarios

manodasanW:
Enabling AOT for authoring scenarios discovered some scenarios that weren't handled.  This addresses most of them with one test failure remaining that I am still working on.

- As part of CsWinRT's C# collection support, CsWinRT implements the ABI interfaces for these types such as `IVector` and `IVectorView`.  As part of those implementation, some of those functions return managed objects for which we need to generate a vtable.  Some of these are generic objects which this PR handles by making sure those types are on the lookup table if one of the collection types that returns it is used as we do not know which function is going to be called from the native side.
- Fixes the `IsWinRTType` check to also handle types from authoring scenarios
- Fix a bunch of custom mapped value types that were missing the `WinRTExposedType` attribute.
- Fix IID optimizer which I found was generating the wrong guid for generics of custom mapped types which the `WinRTExposedType` attribute was added for.
- Fix the `GuidSignature` in the WinRTRuntimeTypeAttribute not being used for authoring types.
- Fix issue when adding to an event handler from native side due to `GetMetadataToken` doesn't work on AOT by replacing with `GetHashCode` as we were only using it get a unique value for the type to use as part of a token.
- Add partial for types that are returned as part of the AsAsync functions so that the AOT source generator can run on them.
- Fix issue where the AOT source generator was not running for authoring scenarios and thereby any internal types in them were not getting the `WinRTExposedType` attribute generated for them or the lookup table.  Public types worked before as they were handled by the authoring source generator itself.
- Fix issue where the Do_ABI functions in authoring scenarios can have generic types passed as parameters to them and those generic instantiation classes are not initialized.  This addresses it by making sure the `EnsureInitialized` is called for them.


src/WinRT.Runtime/Projections/IList.net5.cs
```diff
@@ -1250,7 +1250,7 @@ private static uint GetManyHelper(global::System.Collections.Generic.IList<T> so
         public static readonly IntPtr AbiToProjectionVftablePtr;
         static IList()
         {
-            if (IListMethods<T>.AbiToProjectionVftablePtr == default)
+            if (RuntimeFeature.IsDynamicCodeCompiled && IListMethods<T>.AbiToProjectionVftablePtr == default)
             {
```
manodasanW, line 1254: This and a couple other changes similar to this probably needs to be updated to reflect the findings that this conditional isn't compatible with the linker.  But it also looks like the other PR does fix this, so I am thinking of leaving this for now in this PR and having the other PR fix them the correct way when that conflict is resolved.  Thoughts?

src/WinRT.Runtime/Projections/Nullable.cs
```diff
@@ -1869,8 +1968,8 @@ public ComWrappers.ComInterfaceEntry[] GetExposedInterfaces()
 
         public static ComWrappers.ComInterfaceEntry[] GetExposedInterfaces(ComWrappers.ComInterfaceEntry delegateInterface)
         {
-            return new ComWrappers.ComInterfaceEntry[]
-            {
+            return
+            [
```
Sergio0694, line 1972: This syntax is C# 12, should we not avoid using it until we drop downlevel support, because these files could also be used in the embedded scenario, and people might not be using C# 12 there?

src/cswinrt/strings/additions/Windows.Foundation/TaskToAsyncOperationWithProgressAdapter.cs
```diff
@@ -15,7 +15,7 @@ namespace System.Threading.Tasks
 #if NET
     [global::System.Runtime.Versioning.SupportedOSPlatform("windows10.0.10240.0")]
 #endif
-    internal sealed class TaskToAsyncOperationWithProgressAdapter<TResult, TProgress>
+    internal sealed partial class TaskToAsyncOperationWithProgressAdapter<TResult, TProgress>
```
Sergio0694, line 18: How did you spot all of these missing `partial`? If this is required to let the generator add the necessary code to make things AOT friendly, should we consider having an analyzer (or the generator itself) spot such cases (ie. types matching some criteria, like being async operations or something) and emit a diagnostic if they're missing `partial`? So that you'd be warned rather than just have the generator silently skip processing the type. What do you think?

src/WinRT.Runtime/Projections/Nullable.cs
```diff
@@ -1831,19 +1909,40 @@ internal static class Nullable_Delegates
...
+    }
 }
 
 #if NET
-namespace WinRT
+    namespace WinRT
 {
     public sealed class StructTypeDetails<T, TAbi> : IWinRTExposedTypeDetails where T: struct where TAbi : unmanaged
     {
         private static readonly Guid PIID = GuidGenerator.CreateIID(typeof(ABI.System.Nullable<T>));
```
Sergio0694, line 1940: We could hardcode all the well known IIDs here as well to avoid the call to `GuidGenerator`. We can skip this in this PR though since I'm doing the same in #1429, so might as well just fix both occurrences there.

Sergio0694 (approved):
Left a couple notes, love this! Thank you! 😄

src/WinRT.Runtime/Projections/IList.net5.cs
Sergio0694: Yeah I think let's just leave this as is and then once this is merged and #1395 is merged, I can just rebase #1429 and do another pass on all runtime feature switches and fix any new occurrences from this PR as well. Will also make things easier to review since this PR wouldn't also be moving things around to change those switches 🙂

src/WinRT.Runtime/Projections/Nullable.cs
manodasanW: That is true, will address.

src/cswinrt/strings/additions/Windows.Foundation/TaskToAsyncOperationWithProgressAdapter.cs
manodasanW: I do have plans on adding diagnostics to help catch these and warn folks.  Right now, getting things functionally working has been the priority, soon I will start adding stuff like diagnostics.

Sergio0694: Cool, just figured I'd ask! We can just ignore this for now then, sounds good! 🙂

src/WinRT.Runtime/Projections/Nullable.cs
manodasanW: Lets defer on this discussion to the other PR where I will be responding to the nulllable changes soon.

PR #1435, opened 2024-01-10
Add support for feature switches, add option to disable dynamic objects support

### LINKED ISSUE #1434: FindTypeByName always roots ExpandoObject (and is also undefined behavior)
**Describe the bug**

This code in `TypeNameSupport.FindTypeByName` unconditionally roots `ExpandoObject`:

https://github.com/microsoft/CsWinRT/blob/f01956b448eb15f471f6db908917fdb86313e922/src/WinRT.Runtime/TypeNameSupport.cs#L105-L112

This shows up even in a minimal WinRT component with NAOT, not using anonymous/dynamic types at all:

image

Additionally, matching on `"<>f__AnonymousType"` is undefined behavior. The C# spec (nor Roslyn) guarantee that name prefix.

Sergio0694:
### Overview

- Adds a new `FeatureSwitches` class and the necessary infrastructure for custom feature switches
- Adds a new `CsWinRTEnableDynamicObjectsSupport` feature switch to disable support for dynamic objects
- Wires up the feature switches in the projections setup

### DIFF
--- a/src/WinRT.Runtime/Configuration/FeatureSwitches.cs
+++ b/src/WinRT.Runtime/Configuration/FeatureSwitches.cs
@@ -1,0 +1,94 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Runtime.CompilerServices;
+
+namespace WinRT;
+
+/// <summary>
+/// A container for all shared <see cref="AppContext"/> configuration switches for CsWinRT.
+/// </summary>
+/// <remarks>
+/// <para>
+/// This type uses a very specific setup for configuration switches to ensure ILLink can work the best.
+/// This mirrors the architecture of feature switches in the runtime as well, and it's needed so that
+/// no static constructor is generated for the type.
+/// </para>
+/// <para>
+/// For more info, see <see href="https://github.com/dotnet/runtime/blob/main/docs/workflow/trimming/feature-switches.md#adding-new-feature-switch"/>.
+/// </para>
+/// </remarks>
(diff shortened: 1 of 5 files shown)


src/WinRT.Runtime/Configuration/FeatureSwitches.cs
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    /// <summary>
+    /// Gets the default configuration value for a given feature switch.
+    /// </summary>
+    /// <param name="propertyName">The property name to retrieve the value for.</param>
+    /// <returns>The default value for the target <paramref name="propertyName"/>.</returns>
+    private static bool GetDefaultConfigurationValue(string propertyName)
+    {
+        if (propertyName == IsDynamicObjectsSupportEnabledPropertyName)
```
ahmed605, line 87: I think it's better to use a switch-case statement here instead?

Sergio0694: It doesn't really matter, this will all get trimmed out anyway 🙂

ahmed605: yeah but using a switch-case would be more cleaner in terms of code style in case more switches were introduced in the future

Sergio0694: I've removed this entirely and simplified the logic 😄

Sergio0694:
Saves about 19 KB on .NET 8:

image

manodasanW (approved):

PR #1436, opened 2024-01-11
Fixing issues discovered during validation

manodasanW:
Previous fixes to IID optimizer discovered issues where some of the generated class had incorrect IL.  This was particularly in cases where the generic was known but is a custom type mapping like Nullable<Vector3>.  This handles them by detecting if it is a generic or not before figuring out how to do a field access of the iidData field.

The generic instantiation classes for delegates were assuming they were IInspectable based when it isn't.  This addresses that by fixing the vtable index.


Sergio0694 (approved):
LGTM, good find! :shipit:

manodasanW:
CI already passed: 2.0.5-prerelease.240110.22

PR #1437, opened 2024-01-11
Rewrite MarshalGeneric<T> to minimize binary size on NativeAOT

Sergio0694:
This PR completely rewrites `MarshalGeneric<T>`, to reduce the binary size. Currently there's over 200 KB just in delegate size coming from various generic instantiations for this class alone. Probably much more if we consider all the rest of the code. This PR changes the structure of the class to use a static constructor where we do trimmable type switches and also optimize for lots of well known cases for primitive types and other ABI types that we know will always get constructed. Additionally, I've also introduced a much more compact system of constructing delegates for the various generated methods on the helper types coming from the generated projections, which also skip all those generated closures.

Here's what sizoscope is showing on a minimal WinRT component, for reference:

image

### DIFF
--- a/src/WinRT.Runtime/Marshalers.cs
+++ b/src/WinRT.Runtime/Marshalers.cs
@@ -580,15 +580,36 @@ namespace WinRT
             else if (typeof(T) == typeof(bool))
             {
                 // Same as above, but we do have an ABI type
+                HelperType = typeof(global::ABI.System.Boolean);
                 AbiType = typeof(byte);
+                MarshalerType = typeof(bool);
+                MarshalByObjectReferenceValueSupported = false;
+                CreateMarshaler = (Func<T, object>)(object)((Func<bool, bool>)global::ABI.System.Boolean.CreateMarshaler).WithObjectTResult();
+                CreateMarshaler2 = CreateMarshaler;
+                GetAbi = ((Func<bool, byte>)global::ABI.System.Boolean.GetAbi).WithObjectParams();
+                FromAbi = (Func<object, T>)(object)((Func<byte, bool>)global::ABI.System.Boolean.FromAbi).WithObjectT();
+                CopyAbi = ((Action<bool, IntPtr>)global::ABI.System.Boolean.CopyAbi).WithObjectT1();
+                FromManaged = (Func<T, object>)(object)((Func<bool, byte>)global::ABI.System.Boolean.FromManaged).WithObjectTResult();
+                CopyManaged = (Action<T, IntPtr>)(object)(Action<bool, IntPtr>)global::ABI.System.Boolean.CopyManaged;
+                DisposeMarshaler = ((Action<bool>)global::ABI.System.Boolean.DisposeMarshaler).WithObjectParams();
+                DisposeAbi = ((Action<byte>)global::ABI.System.Boolean.DisposeAbi).WithObjectParams();
 
-                return;
             }
             else if (typeof(T) == typeof(char))
             {
+                HelperType = typeof(global::ABI.System.Char);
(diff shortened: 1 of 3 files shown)


Sergio0694:
Publishing this, as I think it's in a decent state now and covering almost all scenarios. There's just one unit test that's failing only on .NET Core 3.1 for some reason, @manodasanW if you can I'd love a second pair of eyes on that whenever you have time 😄

Sergio0694:
Checked with sizoscope, saves ~270 KB already! 🎉

image

Sergio0694:
Saved another 13 KB on .NET 8, maybe a bit more on .NET 9! 🎉

image

Sergio0694:
Saved another 18 KB! 🎉

image

src/WinRT.Runtime/Marshalers.cs
```diff
@@ -485,158 +485,345 @@ class MarshalGeneric<T>
...
+            {
+                // Fallback case for all other types (could be anything, really). These would be reference types,
+                // which means we can make some assumptions on the shape of the helper type methods. Specifically,
+                // we expect the returned marshallers to be IObjectReference-s, and the ABI to be an IntPtr value.
+                HelperType = typeof(T).GetHelperType();
+                AbiType = typeof(T).GetAbiType();
+                MarshalerType = typeof(T).GetMarshalerType();
+                MarshalByObjectReferenceValueSupported = typeof(T).GetMarshaler2Type() == typeof(ObjectReferenceValue);
+#if NET
+                CreateMarshaler = HelperType.GetMethod("CreateMarshaler", BindingFlags.Public | BindingFlags.Static)?.CreateDelegate<Func<T, IObjectReference>>();
```
manodasanW, line 676: As part of one of the past performance improvements, we made a bunch of these `lazy` because we saw doing a bunch of `GetMethod` calls adds to the cost especially when we weren't using all of them either.  Now with this change, we are going back to doing a bunch of `GetMethod` calls at the same time.  We shouldn't go back to `lazy` as it has its own concerns, but do you think we should consider doing it lazily without using lazy?  Or would doing that add much more to the size compared to the perf benefits?

Sergio0694: The problem is that all of those members are public fields, so we cannot really do any lazy initialization when accessing them. We'd have to assign them to some stub delegate with that additional logic (which is what we were doing before), but that would mean introducing a whole lot of additional generic delegates again, which would increase the binary size. I'd be inclined to say this is fine, considering two things:
- We now have a whole lot of hardcoded paths (eg. for all primitive types and for well known projected types) that skip this reflection logic entirely. All of those were being instantiated previously, so I'm thinking the fact that will no longer be the case should already help with performance on that front.
- `GetMethod()` and `CreateDelegate` should be much faster on NativeAOT compared to CoreCLR, so especially there it shouldn't really be necessary to pay the increased binary size to account for that.

What do you think about taking this change and then just potentially reevaluate the lazy initialization logic in the future in case we do get some new traces showing that being a problem (which might not actually be the case)? 🙂

manodasanW: So, we have indeed reduced the scenarios where we use the generic marshalers as we source generate more of the code and there we know the actual type being used.  Given this brings size improvements and we have reduced the number of scenarios using this, lets go with this change and revisit if we notice perf issues as a result.

manodasanW (approved):

src/WinRT.Runtime/Marshalers.cs
Sergio0694: Sounds good, thank you! 🙂

PR #1441, opened 2024-01-13
Reduce enum stubs codegen with NativeAOT

Sergio0694:
This PR includes some optimizations to reduce the codegen a bit more for `Marshaler<T>` with enum types.
It also is a little performance optimizations in that all shared stubs are now properly cached with `beforefieldinit` too.

### DIFF
--- a/src/WinRT.Runtime/Marshalers.cs
+++ b/src/WinRT.Runtime/Marshalers.cs
@@ -1709,10 +1709,14 @@ namespace WinRT
     internal static class Marshaler
     {
         internal static Func<object, object> ReturnParameterFunc = (object box) => box;
-        internal static unsafe Action<object, IntPtr> CopyIntEnumFunc = 
-            (object value, IntPtr dest) => *(int*)dest.ToPointer() = (int)Convert.ChangeType(value, typeof(int));
-        internal static unsafe Action<object, IntPtr> CopyUIntEnumFunc =
-            (object value, IntPtr dest) => *(uint*)dest.ToPointer() = (uint)Convert.ChangeType(value, typeof(uint));
+
+        internal static unsafe void CopyIntEnum(object value, IntPtr dest) => *(int*)dest.ToPointer() = (int)Convert.ChangeType(value, typeof(int));
+
+        internal static unsafe void CopyIntEnum<T>(T value, IntPtr dest) => *(int*)dest.ToPointer() = (int)(object)value;
+
+        internal static unsafe void CopyUIntEnum(object value, IntPtr dest) => *(uint*)dest.ToPointer() = (uint)Convert.ChangeType(value, typeof(uint));
+
+        internal static unsafe void CopyUIntEnum<T>(T value, IntPtr dest) => *(uint*)dest.ToPointer() = (uint)(object)value;
     }
 
 #if EMBED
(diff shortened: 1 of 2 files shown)


Sergio0694:
Changing this back to draft as it doesn't actually seem to be improving the size.
There's one thing though that I don't understand: both before and after, we're instantiating all lambdas for all types:

image

But in both cases, all of this code is guarded under some `typeof(T).IsEnum` check:

https://github.com/microsoft/CsWinRT/blob/ae695510af515a373bc010cf46aa188736b2b8de/src/WinRT.Runtime/Marshalers.cs#L1568-L1581

@MichalStrehovsky would you know why is the linker apparently not respecting this? We're wasting a lot of size just compiling generic versions of all of those helpers that we know could not possibly be used. Why is this happening despite the check? 🤔

MichalStrehovsky:
> @MichalStrehovsky would you know why is the linker apparently not respecting this? We're wasting a lot of size just compiling generic versions of all of those helpers that we know could not possibly be used. Why is this happening despite the check?

This is constructing delegates, so:

* Are you on .NET 9?
* Do you have `Delegate.Method` somewhere in the program?

Without those, the whole program analysis will figure out the targets of reflection before RyuJIT starts eliminating unreachable codepaths and that's a done deal unfortunately.

Sergio0694:
Ooh I see, thank you! I guess this is just also blocked by #1444 then and fixing that will also improve this. Thank you! 🙂

Sergio0694:
@michalstrehovsky I don't get how this is still regressing on .NET 9, and we're not using `.Method` anymore. I see this:

image

With roots like this one:

image

I don't understand where are all of those generic instantiations coming from and on what closure. This PR literally only removes an existing lambda expression and replaces it with a cached, reused delegate. The weird thing is that I also don't see any generic instantiation over `WithTypedT1<T>`, so it's not like that path is being compiled multiple times. But then where are all of those new generic instantiations for that `<>c` type coming from? I don't get why is this PR regressing size 😅

Sergio0694:
Chatted with Michal, publishing this as it's not actually regressing, just not improving much.

src/WinRT.Runtime/Marshalers.cs
```diff
@@ -1672,12 +1672,24 @@ public static T FromAbi<T>(IntPtr nativeDelegate)
 
     internal static class Marshaler
     {
-        internal static Action<object> EmptyFunc = (object box) => { };
-        internal static Func<object, object> ReturnParameterFunc = (object box) => box;
-        internal static unsafe Action<object, IntPtr> CopyIntEnumFunc = 
-            (object value, IntPtr dest) => *(int*)dest.ToPointer() = (int)Convert.ChangeType(value, typeof(int));
-        internal static unsafe Action<object, IntPtr> CopyUIntEnumFunc =
-            (object value, IntPtr dest) => *(uint*)dest.ToPointer() = (uint)Convert.ChangeType(value, typeof(uint));
+        internal static readonly Action<object> EmptyFunc = Empty;
```
manodasanW, line 1675: I think this changed in one of your other PRs, so may need to rebase with those changes merged.

src/cswinrt/strings/WinRT.cs
```diff
@@ -98,6 +103,11 @@ private static void InvokeWithObjectParams<T>(this Action<T> func, object arg)
         {
             func.Invoke((T)arg);
         }
+
+        private static void InvokeWithTypedT1<T>(this Action<object, IntPtr> action, T arg1, IntPtr arg2)
+        {
```
manodasanW, line 108: Also probably want to bring in your draft PR to move all these out of WinRT.cs

Sergio0694: Yup, will update that too once this is merged! It's also why I left that still as a draft 👍

manodasanW (approved):

PR #1442, opened 2024-01-13
Omit non public methods from preserved metadata

Sergio0694:
This PR stops preserving reflection metadata for non public methods of helper types, given that they're:
1. All public anyway (from the generated projections, or from internal helpers that have been updated accordingly)
2. All other code was also just using `PublicMethods` as data flow annotations (which also produced warnings)

Also including a couple of very minor tweaks to some delegate creations, while at it.


Sergio0694:
Saves ~25 KB in a minimal WinRT sample (4225 KB -> **4200 KB**):

image

manodasanW (approved):

PR #1443, opened 2024-01-13
Trim out more legacy projections fallback paths

Sergio0694:
This PR just allows completely trimming out some fallback paths for legacy projections. These should never be reached when using NativeAOT, and even if they were, old projections are not supported in AOT scenarios anyway, so no point in keeping this.

### DIFF
--- a/src/WinRT.Runtime/GuidGenerator.cs
+++ b/src/WinRT.Runtime/GuidGenerator.cs
@@ -5,7 +5,6 @@ using System;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Reflection;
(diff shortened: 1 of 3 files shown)


manodasanW (approved):

manodasanW:
Looks like there are new authoring test failures:

[  FAILED  ] AuthoringTest.Arrays
[  FAILED  ] AuthoringTest.CustomVectorImplementations
[  FAILED  ] AuthoringTest.ExplicitInterfaces

manodasanW (approved):

Sergio0694:
All green, and only 1 failing authoring test now, which was already failing before too:

image

PR #1447, opened 2024-01-16
Add option to disable localized exception messages

Sergio0694:
This PR builds on top of #1435 and adds an option to disable localized exception messages, matching what the runtime also does when setting the `UseSystemResourceKeys` feature switch. The new flag allows trimming all of CsWinRT's embedded resources.


Sergio0694:
Saves 142 KB on .NET 9!! 🎉

image

nuget/Microsoft.Windows.CsWinRT.targets
```diff
@@ -237,6 +237,7 @@ $(CsWinRTInternalProjection)
   <!-- Default values for all custom CsWinRT runtimem feature switches -->
   <PropertyGroup>
     <CsWinRTEnableDynamicObjectsSupport Condition="'$(CsWinRTEnableDynamicObjectsSupport)' == ''">true</CsWinRTEnableDynamicObjectsSupport>
+    <CsWinRTUseExceptionResourceKeys Condition="'$(CsWinRTUseExceptionResourceKeys)' == ''">false</CsWinRTUseExceptionResourceKeys>
```
manodasanW, line 240: The naming was initially a bit confusing, but it looks like .NET uses the same naming.  To make sure my understanding is correct, the default state of false means resources are kept and is localized.  If set to true, localized resources are trimmed and the resource keys are used instead.

Sergio0694: Yup, that's correct! It's the same that .NET does, where it also defaults to false. Using "resource keys" means that whenever you're trying to load a localized resource, say, "Some_Resource", rather than actually loading the localized resource with that name from the embedded resources, you literally just return "Some_Resource" as a string. Which means you can then also skip bundling all those .resx files, and trim all of `ResourceManager`. If you're not setting this property, then there's no functional changes at all compared to without this PR 🙂

PR #1448, opened 2024-01-16
Rewrite event registration token logic to improve performance, trimming

Sergio0694:
This PR rewrites the logic to compute event registration tokens, see: https://github.com/dotnet/runtime/issues/96947.
This is both faster, and allows NativeAOT to trim signficantly more stuff, as it makes all delegate targets non-reflectable.

### DIFF
--- a/src/cswinrt/strings/WinRT.cs
+++ b/src/cswinrt/strings/WinRT.cs
@@ -1051,6 +1051,9 @@ namespace WinRT
         // Note this dictionary is also used as the synchronization object for this table
         private readonly Dictionary<int, object> m_tokens = new Dictionary<int, object>();
 
+        // The current counter used for the low 32 bits of the registration tokens.
+        private int m_low32Bits = 0x42424242;
+
         public EventRegistrationToken AddEventHandler(T handler)
         {
             // Windows Runtime allows null handlers.  Assign those the default token (token value 0) for simplicity
@@ -1072,7 +1075,33 @@ namespace WinRT
             // Get a registration token, making sure that we haven't already used the value. This should be quite
             // rare, but in the case it does happen, just keep trying until we find one that's unused. Note that
             // this mutable part of the token is just 32 bit wide (the lower 32 bits). The upper 32 bits are fixed.
-            int tokenLow32Bits = GetPreferredTokenLow32Bits(handler);
+            //
+            // Note that:
+            //   - If there is a handler assigned to the generated initial token value, it is not necessarily
+            //     this handler.
+            //   - If there is no handler assigned to the generated initial token value, the handler may still
+            //     be registered under a different token.
+            //
+            // Effectively the only reasonable thing to do with this value is to use it as a good starting point
+            // for generating a token for handler.
+            //
+            // We want to generate a token value that has the following properties:
+            //   1. Is quickly obtained from the handler instance (in this case, it doesn't depend on it at all).
+            //   2. Uses bits in the upper 32 bits of the 64 bit value, in order to avoid bugs where code
+            //      may assume the value is really just 32 bits.
+            //   3. Uuses bits in the bottom 32 bits of the 64 bit value, in order to ensure that code doesn't
+            //      take a dependency on them always being 0.
+            //
+            // The simple algorithm chosen here is to simply assign the upper 32 bits the metadata token of the
+            // event handler type, and the lower 32 bits to an incremental counter starting from some arbitrary
+            // constant. Using the metadata token for the upper 32 bits gives us at least a small chance of being
+            // able to identify a totally corrupted token if we ever come across one in a minidump or other scenario.
+            //
+            // We should feel free to change this algorithm as other requirements / optimizations become available.
+            // This implementation is sufficiently random that code cannot simply guess the value to take a dependency
+            // upon it. (Simply applying the hash-value algorithm directly won't work in the case of collisions,
+            // where we'll use a different token value).
+            int tokenLow32Bits = m_low32Bits++;
 
 #if NET6_0_OR_GREATER
             // When on .NET 6+, just iterate on TryAdd, which allows skipping the extra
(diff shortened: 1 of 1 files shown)


src/cswinrt/strings/WinRT.cs
```diff
@@ -1063,110 +1072,96 @@ private EventRegistrationToken AddEventHandlerNoLock(T handler)
...
+            int tokenLow32Bits = m_low32Bits++;
 
 #if NET6_0_OR_GREATER
             // When on .NET 6+, just iterate on TryAdd, which allows skipping the extra
             // lookup on the last iteration (as the handler is added rigth away instead).
-            while (!m_tokens.TryAdd(token, handler))
+            while (!m_tokens.TryAdd(tokenLow32Bits, handler))
             {
-                token = new EventRegistrationToken { Value = token.Value + 1 };
+                tokenLow32Bits++;
```
Sergio0694, line 1111: One possible concern I see is: in case the token was somehow already used, incrementing it here would mean we could "race past" the value in `m_low32Bits`, meaning that all subsequent registrations would then also go through unnecessary extra lookups as the value of those lower 32 bits "catches up" to the one that ended up being used here. Now, I can't say I know exactly when/how such a scenario might happen though, and whether it's a real concern 🤔

@jkotas thoughts on this? And if this is an actual concern, should we perhaps simply increment `m_low32Bits` for each loop iteration, so we're sure that the value we leave it on should be the very next "free" value we expect?

src/cswinrt/strings/WinRT.cs
```diff
@@ -1040,10 +1040,19 @@ protected override Delegate GetEventInvoke()
...
     // An event registration token table stores mappings from delegates to event tokens, in order to support
     // sourcing WinRT style events from managed code.
-    internal sealed class EventRegistrationTokenTable<T> where T : class, global::System.Delegate
+    internal sealed class EventRegistrationTokenTable<T>
+        where T : class, global::System.Delegate
     {
+        /// <summary>
+        /// The hashcode of the delegate type, being set in the upper 32 bits of the registration tokens.
+        /// </summary>
+        private static readonly int TypeOfTHashCode = typeof(T).GetHashCode();
```
jkotas, line 1049:
```suggestion
        private readonly int TypeOfTHashCode = typeof(T).GetHashCode();
```
I assume that `EventRegistrationTokenTable<T>` is used as a lazily initialized singleton. Is that correct? You can store the hashcode in an instance variable. It will be faster to access than a readonly static (in shared generic code in particular).

src/cswinrt/strings/WinRT.cs
jkotas: Yes, I think it is better to increment `m_low32Bits` so that the next registration does not need to go through the values that are in use again. (It is a corner case, unlikely to happen in practice.)

src/cswinrt/strings/WinRT.cs
Sergio0694: They're not singletons, we instantiate a different instance per-event and per-object, in every CCW, like so:

image

Do you think it's still worth it to switch to an instance field even in this case?
I assume it would just introduce a tiny increase in memory use? 🤔

src/cswinrt/strings/WinRT.cs
Sergio0694: Fixed in https://github.com/microsoft/CsWinRT/pull/1448/commits/d17885086c5686da370ecc6f66b9b552d8d1fcd7 🙂

src/cswinrt/strings/WinRT.cs
```diff
@@ -1040,10 +1040,19 @@ protected override Delegate GetEventInvoke()
 
     // An event registration token table stores mappings from delegates to event tokens, in order to support
     // sourcing WinRT style events from managed code.
-    internal sealed class EventRegistrationTokenTable<T> where T : class, global::System.Delegate
+    internal sealed class EventRegistrationTokenTable<T>
+        where T : class, global::System.Delegate
```
Joe4evr, line 1044: Is `where T : class` not implied from `where T : System.Delegate`? 😕

src/cswinrt/strings/WinRT.cs
jkotas: Ok, it can stay the way you have it now.

It may be a good idea to initialize the ID counter with a random number instead of 0x42424242, so that each instance gets its own range.

Also, https://learn.microsoft.com/en-us/uwp/api/windows.foundation.eventregistrationtoken.value says that a valid token should not be zero. This corner case is not handled currently.

Sergio0694: > "It may be a good idea to initialize the ID counter with a random number instead of 0x42424242, so that each instance gets its own range."

Gotcha, yeah we can do that. Random (possibly stupid) question: would it make sense to do something like `new object().GetHashCode()` (or something else) to get that initial pseudo-random seed per instance, to avoid rooting all of `Random` just for this one use? Since the goal here was to minimize code size, and the changes in this PR are already faster than what the original implementation was anyway 🤔

> "This corner case is not handled currently."

Why? Even when this incrementing counter wraps around and reaches 0, the actual event registration token would still not be 0 because the upper 32 bits would be set to the hashcode of the `T` type argument, no? 😅

jkotas: > to avoid rooting all of Random just for this one use? 

Guess how `new object().GetHashCode()` generates the hashcode: https://github.com/dotnet/runtime/blob/7de8f093508d1d3ed035860e63a16d2403c6ef21/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeHelpers.NativeAot.cs#L95

Also, Random is small (when not rooted for reflection).

> the upper 32 bits would be set to the hashcode of the T type argument, no?

What guarantees that this hashcode is non-zero?

Sergio0694: > "Guess how new object().GetHashCode() generates the hashcode"

Ha, well nevermind then ahahah

> "Also, Random is small (when not rooted for reflection)."

Got it, alright I'll just do that then, one less thing to worry about. Thank you! 🙂

> "What guarantees that this hashcode is non-zero?"

Yeah no that was just me clearly making a wrong assumption there 😅

I'm going to move the initialization to some static method that also does the check and uses some magic constant just in case that value ever was 0, just so we can make this fully WinRT compliant while we're at it. Good catch, thank you!

Sergio0694: Addressed both things in:
- 003a3ff86ec2fdf005b1630c0ba9547816d11e09
- 9ec2cb0d6f65bf337f349dc5568449ed661eb682

src/cswinrt/strings/WinRT.cs
jkotas: I agree `class` looks unnecessary here.

jkotas (approved):
LGTM

src/cswinrt/strings/WinRT.cs
```diff
@@ -1063,110 +1098,105 @@ private EventRegistrationToken AddEventHandlerNoLock(T handler)
...
+            // for generating a token for handler.
             //
-            // The hash code of a unicast delegate is not tied to the method being invoked, so in the case
-            // of a unicast delegate, the hash code of the target method is used instead of the full delegate
-            // hash code.
+            // We want to generate a token value that has the following properties:
+            //   1. Is quickly obtained from the handler instance (in this case, it doesn't depend on it at all).
+            //   2. Uses bits in the upper 32 bits of the 64 bit value, in order to avoid bugs where code
+            //      may assume the value is really just 32 bits.
+            //   3. Uuses bits in the bottom 32 bits of the 64 bit value, in order to ensure that code doesn't
```
manodasanW, line 1118:
```suggestion
            //   3. Uses bits in the bottom 32 bits of the 64 bit value, in order to ensure that code doesn't
```

src/cswinrt/strings/WinRT.cs
```diff
@@ -1063,110 +1098,105 @@ private EventRegistrationToken AddEventHandlerNoLock(T handler)
...
+                return false;
+            }
+
             lock (m_tokens)
             {
 #if NET6_0_OR_GREATER
                 // On .NET 6 and above, we can use a single lookup to both check whether the token
                 // exists in the table, remove it, and also retrieve the removed handler to return.
-                if (m_tokens.Remove(token, out handler))
+                if (m_tokens.Remove((int)token.Value, out object obj))
```
manodasanW, line 1180: Consider a comment here to indicate the cast to `int` means it uses the lower 32 bits which is what is tracked in the dictionary.

manodasanW (approved):

src/cswinrt/strings/WinRT.cs
Joe4evr: @jkotas For reference, there was some discussion on Discord looking into this. It was found that
* For any other base class constraint (aside from the special snowflake `System.Enum`), the compiler reports an error on the redundancy of the `class` constraint
* If you define `void M1<T>() where T : class, Delegate` and `void M2<T>() where T : Delegate`, [`M1` can be called from `M2` no problem](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBLANgHwAEAmARgFgAoKkgAkNIHYqBvK2j+lWgWVIB4AKgD4AFAEp2naQHcAFjFi1BtEPWJpaAERi4YAcwCGGGFI5tK0gL5UzXXsSFjJl6Z3mKYy1dt0Hjpq7mdtJ8ThIA3HY2lFZAA=)
  * Additionally, IL for `M2` contains `.param constraint` but it's not in `M1`. Not sure what difference that makes, but it does look weird that the technically *less* constrained method contains some kind of `constraint` instruction but the *more* constrained method does not.

Sergio0694:
Saves ~350 KB on .NET 9!! 🎉
I expect the delta to possibly be less than this once #1429 and #1437 are merged, but still, this is awesome!

image

(sizoscope only accounts for ~260 KB, but the .dll size diff is 4441 KB -> 4093 KB)

PR #1450, opened 2024-01-18
Drop all delegates and closures in MarshalInterface<T> on NAOT

Sergio0694:
Further size/performance optimization on top of #1437.
This removes all lambdas and closures in `MarshalInterface<T>` on NativeAOT.


Sergio0694:
Saves a few KBs, a bit more as your number of marshalled types goes up in larger applications:

image

manodasanW (approved):

PR #1453, opened 2024-01-18
Remove [Serializable] support from exceptions and seal them

Sergio0694:
This PR removes all `[Serializable]` support from exception types, as it's obsolete and shouldn't be used.
It also seals all exception types while at it. No consumer should be extending these anyway.


src/WinRT.Runtime/ExceptionHelpers.cs
```diff
@@ -439,7 +412,7 @@ public IObjectReference RealErrorObject
 
                 // Keep the error object alive so that user could retrieve error information
                 // using Data["RestrictedErrorReference"]
-                dict["__RestrictedErrorObjectReference"] = restrictedErrorObject == null ? null : new __RestrictedErrorObject(restrictedErrorObject);
+                dict["__RestrictedErrorObjectReference"] = restrictedErrorObject;
```
manodasanW, line 415: Are there any pointers to where it says using `Serializable` is now obsolete.  I see that the .NET exceptions still have the attribute.  The scenario I am concerned about is crash dump debugging.  Today when a crash happens, the exception is available as part of the stowed exception and the sos tools allow you to see the exception.  I would believe this would continue to work fine without the serializable attribute.  But then there is a 2nd scenario where the exception happens in one process and then is propagated to another process.  With some of the previous error origination fixes, I think this propagation of stowed exception error works now sometimes (there is still one scenario where it breaks).  But I think the exception object we pass is a managed object and thereby wouldn't be able to be inspected in the other process.  I have plans to try to see if we can improve it to serialize the exception and pass that instead which is what was originally intended as part of `RoOriginateLanguageException` API in its exception object parameter.

Sergio0694: There's now also an official announcement in the .NET repo about `BinaryFormatter` being completely removed in .NET 9: https://github.com/dotnet/runtime/issues/98245.

manodasanW: Based on that issue, it looks like `SetRemoteStackTrace` can be used to propagate the stack for cross process exceptions.  So that is probably what we can use when we get to looking at getting cross process exception propagation to work better.

manodasanW (approved):

PR #1454, opened 2024-01-19
Seal IObjectReference.TryAs<T> method

Sergio0694:
This PR seals `IObjectReference.TryAs<T>`, which was bloating the binary size due to combinatorial generics:

image

Should also be faster, given that the GVM dispatch isn't exactly the best.

### DIFF
--- a/src/WinRT.Runtime/ObjectReference.cs
+++ b/src/WinRT.Runtime/ObjectReference.cs
@@ -193,63 +193,11 @@ namespace WinRT
             // which would just explode the binary size on NativeAOT due to combinatorial generics.
             if (this is IObjectReferenceWithContext)
             {
-                return TryAsForObjectReferenceWithContext(this, iid, out objRef);
+                return ObjectReferenceWithContext<T>.TryAs(this, iid, out objRef);
             }
 
             // Normal path for IObjectReference or any IObjectReference<T>
-            return TryAsForObjectReference(this, iid, out objRef);
-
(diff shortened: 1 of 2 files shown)


Sergio0694:
Saves 186 KB on .NET 9! 🎉

image

src/WinRT.Runtime/ObjectReference.cs
```diff
@@ -167,34 +167,78 @@ public unsafe TInterface AsInterface<TInterface>()
...
         public int TryAs<
 #if NET
-        [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.NonPublicConstructors | DynamicallyAccessedMemberTypes.PublicFields)]
+            [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.NonPublicConstructors | DynamicallyAccessedMemberTypes.PublicFields)]
 #endif
-        T>(out ObjectReference<T> objRef) => TryAs(GuidGenerator.GetIID(typeof(T)), out objRef);
+            T>(out ObjectReference<T> objRef) => TryAs(GuidGenerator.GetIID(typeof(T)), out objRef);
 
-        public virtual unsafe int TryAs<
+        public unsafe int TryAs<
```
manodasanW, line 174: If someone doesn't override the `TryAs` function but does extend this class, does this actually break anything or no?

manodasanW (approved):

Sergio0694:
Actually saves 253 KB now with the latest changes too! 🎉

image

Issue #1324, opened 2023-04-19
CsWinRT should enable AOT and single-file analyzers


Sergio0694:
# Proposal: CsWinRT should enable AOT and single-file analyzers

## Summary
While CsWinRT currently doesn't support AOT yet (I mean it kinda works in some scenarios if you try really hard, but it's not officially supported nor really usable in general), one issue it has is also that it doesn't come with AOT and single-file analyzers at all. This makes it even more difficult to try to make things work, as it's not really easy to understand which paths might potentially cause issues.

Both analyzers should be enabled, and the relative annotations should be added when needed, so that warnings will correctly propagated to callsites. Note: this doesn't mean CsWinRT has to be AOT-compatible from the get go: enabling analyzers is orthogonal in this sense and it will at least provide users with a better developer experience. It should then also make it easier to migrate things, as it'll clearly highlight code paths that can potentially break down and when.

In general, CsWinRT should enable these two MSBuild flags and add all necessary annotations:

```xml
<EnableAotAnalyzer>true</EnableAotAnalyzer>
<EnableSingleFileAnalyzer>true</EnableSingleFileAnalyzer>
```

These are on top of the (already enabled) trim analyzer.
The various attributes that will be needed are:
- [`[RequiresDynamicCode]`](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.requiresdynamiccodeattribute)
- [`[RequiresAssemblyFiles]`](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.requiresassemblyfilesattribute)

## Rationale

* Gradual step towards making CsWinRT AOT compatible
* It will improve the developer experience using CsWinRT in AOT/single-file scenarios

PR #1459, opened 2024-01-21
Make marshalling 'System.Type' AOT-safe

Sergio0694:
### Contributes to #1324

While enabling the AOT analyzers, I noticed that marshalling `System.Type` was inherently not AOT-safe. Because this is a well known type, we can drop using `MarshalNonBlittable<T>` for it (as this is just not AOT-safe at all), and hardcode the right marshalling stubs instead, since we do know the ABI type at compile time in this case.

Opening this PR first to split the work. Will finish adding all AOT annotations in a follow up PR.


Sergio0694:
Regresses size by 1 KB, but it makes `Type` marshalling AOT safe and faster, I'll take it 😄

image

src/WinRT.Runtime/FundamentalMarshalers.cs
```diff
@@ -67,5 +71,183 @@ internal struct Char
         public static void DisposeMarshaler(char m) { }
         public static void DisposeAbi(ushort abi) { }
     }
+
+    // This partial declaration contains additional marshalling methods for the ABI type for System.Type.
+    // These are hardcoded for the ABI.System.Type type, which makes them AOT-safe. Previously, these were
+    // relying on AOT-unfriendly code in MarshalNonBlittable<T>. Because this is a well known type, we can
+    // just hardcode explicit marshalling methods for it instead to make using this type fully AOT-safe.
+    partial struct Type
```
manodasanW, line 79: Any advantages to putting this here rather than in `Type.cs` where `CreateMarshaler` and so on are already there.

Sergio0694: No specific reason other than I wanted to separate the definition from the rest of the stuff on the ABI type, as these methods are just internal helpers and they're not public. Maybe we could just add a `Type.ArrayMarshalling.cs` partial file? Otherwise I can also just move these to the same file if you have a strong preference 🙂

manodasanW: `Type.ArrayMarshalling.cs` might be better then.  It just sometimes it gets hard to find where certain marshalers are defined and it can get confusing when you see some defined in one file and someone might think that is all of them.

Sergio0694: Yup, no that makes perfect sense! Done in 9f265c3c31e45bb981887cde877fd32c71fc1223 🙂

manodasanW (approved):

PR #1460, opened 2024-01-21
Stop using ThreadLocal<T>, use [ThreadStatic]

Sergio0694:
We're already using `[ThreadStatic]` right below this, might as well just keep it consistent and use it for the stack of visited types too. This will save a little bit of binary size, as there was nothing else rooting the type otherwise (plus, one less lambda).

### DIFF
--- a/src/WinRT.Runtime/ComWrappersSupport.cs
+++ b/src/WinRT.Runtime/ComWrappersSupport.cs
@@ -11,7 +11,6 @@ using System.Linq;
 using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
-using System.Threading;
 using WinRT.Interop;
 
 #if NET
--- a/src/WinRT.Runtime/ComWrappersSupport.net5.cs
+++ b/src/WinRT.Runtime/ComWrappersSupport.net5.cs
@@ -6,7 +6,6 @@ using System.Collections;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
-using System.Linq.Expressions;
 using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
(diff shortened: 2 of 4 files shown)


Sergio0694:
This is weird... Barely any diffs because despite dropping over 10 KB worth of stuff from `ThreadLocal<T>`, the new version has... Over 10 KB of seemingly random numeric methods, value typles and other stuff that just makes no sense looking at the changes:

image

@MichalStrehovsky do you have any ideas on what could be going on? 🤔

Eg. this is the trace for that value tuple... But how is any change in this PR doing any of that?

image

MichalStrehovsky:
> @MichalStrehovsky do you have any ideas on what could be going on? 🤔

I don't know why is it only showing up after these changes; could be some visualization artifact or IDK, but the inclusion looks reasonable to me.

* We ran dataflow analysis on the `Marshaler<T>` static constructor. It saw this assignment:
  https://github.com/microsoft/CsWinRT/blob/329747f632701c60baec487253d066eb4a2d2722/src/WinRT.Runtime/Marshalers.cs#L1740
* AbiType is annotated as: please make all public methods available for runtime reflection:
  https://github.com/microsoft/CsWinRT/blob/329747f632701c60baec487253d066eb4a2d2722/src/WinRT.Runtime/Marshalers.cs#L1967-L1970
* So we keep `byte.DivRem`. If you reflection-invoke that method, it will return a boxed ValueTuple, so we need to treat ValueTuple as boxed, etc.

Note that dataflow analysis (the thing that computes dynamic use of reflection using the annotations, intrinsically recognized APIs, etc.) is a shared codebase with IL Linker and as such operates on uninstantiated method bodies. So we don't re-run it for each `T` but consider any `T` is potentially reachable when analyzing reflection use.

The only theory I have is that maybe before your change we avoided running dataflow analysis because we ended up running all `Marshaler<T>` static constructors at compile time. That's a bug. I filed it as https://github.com/dotnet/runtime/issues/97286.

You could check whether you're running into this bug by adding `<ItemGroup><IlcArg Include="--verbose" /></ItemGroup>` to your project and publishing again with/without your change. The verbose log (printed by ILC to stdout) will have a list of types that were/weren't preinitialized. Try to diff that, maybe something couldn't preinitialize due to the `ThreadStatic` anymore.

Sergio0694:
Re-run sizoscope after #1461, saves 16 KB now! 🎉

image

manodasanW (approved):

PR #1461, opened 2024-01-22
Drop [DynamicallyAccessedMembers] from Marshaler<T>.AbiType

Sergio0694:
Related to https://github.com/microsoft/CsWinRT/pull/1460#issuecomment-1903537019. This PR drops the `[DynamicallyAccessedMembers]` use from `Marshaler<T>.AbiType`, which was not needed and was causing a lot of unnecessary stuff to be preserved unnecessarily (for explicit `T` types that were not helper types). The public members for the helper types are already preserved separately, and this attribute was only used to suppress warnings within the static constructor. I've changed the logic there to just flow the annitations through a local variable instead.


manodasanW (approved):

Sergio0694:
Saves 51 KB on .NET 9! 🎉

image

PR #1464, opened 2024-01-24
Optimize ContextCallback on modern .NET

Sergio0694:
This PR optimizes the context callback logic on modern .NET with the following changes:
- Skipped using `ObjectReference<T>` entirely, doing manual interop instead
- Pass a pointer to our callback directly, skipping the closure + delegate + marshalling
- Also removed an unnecessary interface while at it

I'm expecting a minor performance increase, and minor binary size improvements


jlaanstra (approved):

manodasanW (approved):

Sergio0694:
Saves ~30 KB. Less when considering the other two PRs, but still a few KBs either way:

image

PR #1468, opened 2024-01-25
Reduce binary size in agile reference context cache

Sergio0694:
Fixes the small size regression from #1466.
We don't actually need the instantiations here at all.
This is exactly the same trick we also used in #1448.


Sergio0694:
Saves ~35 KB on .NET 9. Along with #1454 we should be back to par after #1466.

image

PR #1474, opened 2024-01-26
Add new api's to IObjectReference to safely use the underlying pointer.

jlaanstra:
Add a pattern similar to SafeHandle to safely access the pointer of an ObjectReference.

Depends on #1458


src/cswinrt/strings/WinRT.cs
```diff
@@ -303,13 +314,13 @@ internal abstract class State : IDisposable
         public System.Delegate del;
         public System.Delegate eventInvoke;
         private bool disposedValue;
-        private readonly IntPtr obj;
+        private readonly IObjectReference obj;
```
Sergio0694, line 317: This is potentially a semantics change, as the event source state wasn't previously keeping the RCW instance alive (unless indirectly captured by the wrapped delegate), but only the pointer to the native object. Are we sure changing this is not going to introduce some unwanted side effects and/or potentially any leaks anywhere?

PR #1475, opened 2024-01-26
Use typeof(DllModule).Assembly for faster resolution

Sergio0694:
Now that `ActivationFactory` is in WinRT.Runtime, we can use `typeof(Foo).Assembly` for faster resolution.

### DIFF
--- a/src/WinRT.Runtime/ActivationFactory.cs
+++ b/src/WinRT.Runtime/ActivationFactory.cs
@@ -47,7 +47,7 @@ namespace WinRT
 #if NET
             if (moduleHandle == IntPtr.Zero)
             {
-                NativeLibrary.TryLoad(fileName, Assembly.GetExecutingAssembly(), null, out moduleHandle);
+                NativeLibrary.TryLoad(fileName, typeof(DllModule).Assembly, null, out moduleHandle);
             }
 #endif
             if (moduleHandle == IntPtr.Zero)


manodasanW (approved):

jlaanstra:
On a related note, this is possible a change in behavior? Does it matter what Assembly we pass here?

jlaanstra (approved):

jlaanstra:
> On a related note, this is a change in behavior from the ActivationFactory PR. Does it matter what Assembly we pass here?

Sergio0694:
> "this is a change in behavior from the ActivationFactory PR. Does it matter what Assembly we pass here?"

Yeah I think that PR did change the behavior there. I can't really imagine what difference that would make though, other than... Maybe like, you have an app that's dynamically loading some plugin .dll that's in some other folder, and that .dll tries to activate some WinRT type that's in a .dll that's next to _that_ .dll, and not to your app's .dll? Not sure if that matters? 🤔

manodasanW:
> > "this is a change in behavior from the ActivationFactory PR. Does it matter what Assembly we pass here?"
> 
> Yeah I think that PR did change the behavior there. I can't really imagine what difference that would make though, other than... Maybe like, you have an app that's dynamically loading some plugin .dll that's in some other folder, and that .dll tries to activate some WinRT type that's in a .dll that's next to _that_ .dll, and not to your app's .dll? Not sure if that matters? 🤔

Yea I can see maybe some plugin scenarios being impacted.  Will need to monitor for any failures or reports.  The change to `AppContext.BaseDirectory` from several releases ago was more likely to run into similar issues here but never got any reports there.

PR #1483, opened 2024-01-30
Add 'EnableDefaultCustomTypeMappings' switch

Sergio0694:
We have a fair amount of binary size being used up due to all the default ABI type mappings from `Projections`'s type initializer, which ends up rooting a whole lot of stuff even if not actually used. This PR adds a new `CsWinRTEnableDefaultCustomTypeMappings` property (defaults to `true` to keep the current behavior), which allows developers to disable this behavior and trim all of that additional infrastructure. We should evaluate the best way to allow devs to then manually register the individual types they need. That could be via some new `RegisterAbiTypeMapping`-like methods or something.

sizoscope diff incoming...


Sergio0694:
Saves 586 KB on .NET 9, as a baseline (!!!):

image

Sergio0694:
Updated diff with the default always-on mappings, and the extra guard code, 583 KB saved on .NET 9:

image

src/WinRT.Runtime/Projections.CustomTypeMappings.g.cs
```diff
@@ -0,0 +1,1246 @@
...
+
+            if (Interlocked.CompareExchange(ref _NotifyCollectionChangedAction, 1, 0) == 1)
+            {
+                return;
+            }
+
+            RegisterCustomAbiTypeMapping(
+                typeof(NotifyCollectionChangedAction),
+                typeof(ABI.System.Collections.Specialized.NotifyCollectionChangedAction),
+                "Microsoft.UI.Xaml.Interop.NotifyCollectionChangedEventArgs",
```
manodasanW, line 898: NotifyCollectionChangedAction

Sergio0694: Fixed in https://github.com/microsoft/CsWinRT/pull/1483/commits/2413d8ed1fd6034238be93057240d79caee4e6b6!

manodasanW (approved):

PR #1484, opened 2024-01-30
Fix TFM check in generated activation factories

Sergio0694:
Fixes the .NET 7 checks in the code writer, which was causing legacy code to always be generated.
Draft until I can actually double-check that this does in fact fix the issue.

### DIFF
--- a/nuget/Microsoft.Windows.CsWinRT.Authoring.targets
+++ b/nuget/Microsoft.Windows.CsWinRT.Authoring.targets
@@ -52,20 +52,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     
     <PropertyGroup>
       <CsWinRTAuthoringInputs>$(CsWinRTAuthoringInputs) @(CsWinRTAuthoringDistinctWinMDs->'"%(FullPath)"', ' ') </CsWinRTAuthoringInputs>
-    </PropertyGroup>
-
-    <!--
-      Make sure 'CsWinRTExeTFM' is set in authoring scenarios as well. The items in this
-      group should be kept in sync with the ones in 'Microsoft.Windows.CsWinRT.targets'.
-    -->
-    <PropertyGroup>
-      <CsWinRTExeTFM Condition="$([MSBuild]::GetTargetFrameworkVersion('$(TargetFramework)')) == 5">net5.0</CsWinRTExeTFM>
-      <CsWinRTExeTFM Condition="$([MSBuild]::GetTargetFrameworkVersion('$(TargetFramework)')) == 6">net6.0</CsWinRTExeTFM>
-      <CsWinRTExeTFM Condition="$([MSBuild]::GetTargetFrameworkVersion('$(TargetFramework)')) == 7">net7.0</CsWinRTExeTFM>
-      <CsWinRTExeTFM Condition="$([MSBuild]::GetTargetFrameworkVersion('$(TargetFramework)')) >= 8">net8.0</CsWinRTExeTFM>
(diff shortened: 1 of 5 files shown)


Sergio0694:
I don't really get it, I keep always ending up with the downlevel code being generated:

image

@manodasanW do you have any ideas what I'm doing wrong here? 😅

manodasanW:
> I don't really get it, I keep always ending up with the downlevel code being generated:
> 
> image
> 
> @manodasanW do you have any ideas what I'm doing wrong here? 😅

target needs to be passed here when calling cswinrt.exe.  Right now it is passing empty.

https://github.com/microsoft/CsWinRT/blob/5b0252590bbab69cf9b3f0cc8ea6a08abbbca5a0/src/Authoring/WinRT.SourceGenerator/Generator.cs#L62

Sergio0694:
Mmh this works but only if I manually set `CsWinRTExeTFM`. Not sure why the .targets isn't setting this by default? 🤔

nuget/Microsoft.Windows.CsWinRT.Authoring.targets
```diff
@@ -52,6 +53,18 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <PropertyGroup>
       <CsWinRTAuthoringInputs>$(CsWinRTAuthoringInputs) @(CsWinRTAuthoringDistinctWinMDs->'"%(FullPath)"', ' ') </CsWinRTAuthoringInputs>
     </PropertyGroup>
+
+    <!--
+      Make sure 'CsWinRTExeTFM' is set in authoring scenarios as well. The items in this
+      group should be kept in sync with the ones in 'Microsoft.Windows.CsWinRT.targets'.
+    -->
+    <PropertyGroup>
+      <CsWinRTExeTFM Condition="$([MSBuild]::GetTargetFrameworkVersion('$(TargetFramework)')) == 5">net5.0</CsWinRTExeTFM>
```
dongle-the-gadget, line 62: CsWinRT doesn't support `net5.0` and `net8.0` as targets ATM, wouldn't this lead to a failure?

Sergio0694: I don't disagree, but let's remove .NET 5 from everywhere in a separate follow up PR 👍

nuget/Microsoft.Windows.CsWinRT.targets
```diff
@@ -159,6 +159,11 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       <CsWinRTInternalProjection Condition="'$(CsWinRTPrivateProjection)' == 'true'">-internal</CsWinRTInternalProjection>
       <CsWinRTEmbeddedProjection Condition="'$(CsWinRTEmbedded)' == 'true'">-embedded</CsWinRTEmbeddedProjection>
       <CsWinRTEmbeddedEnums Condition="'$(CsWinRTEmbeddedPublicEnums)' == 'true'">-public_enums</CsWinRTEmbeddedEnums>
+
```
manodasanW, line 162: You can probably simplify this and do this in one place by moving setting `CsWinRTExeTFM` out of the `CsWinRTGenerateProjection` target and instead to the top in the common property group.

src/Authoring/WinRT.SourceGenerator/Generator.cs
```diff
@@ -59,12 +59,13 @@ private void GenerateSources()
             string winmds = context.GetCsWinRTDependentMetadata();
```
manodasanW, line 59: Any reason to make `csWinRTExeTFM` a parameter rather than a variable here like the others.

PR #1486, opened 2024-01-31
Remove almost all LINQ uses

Sergio0694:
This PR removes all LINQ uses (except one `ToArray()` call) from CsWinRT. It leaves the use in the NS2.0 projections and in a couple of custom mapping types that are now optional, meaning they're already pay-for play anyway, so that's fine.


Sergio0694:
Saves 15 KB:

image

manodasanW (approved):

PR #1487, opened 2024-01-31
Hard-code the numeric scalar types in CoerceValue and simplify enum marshalling to remove any usage of Convert.ChangeType.

jkoritzinsky:
Convert.ChangeType was rooting a significant amount of code that would never be used (such as string.IConvertible.ToDateTime).

This PR will allow us to trim a significant amount of code in many cases (IConvertible isn't a commonly used type nowadays).

### DIFF
--- a/src/WinRT.Runtime/Projections/IPropertyValue.net5.cs
+++ b/src/WinRT.Runtime/Projections/IPropertyValue.net5.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using WinRT;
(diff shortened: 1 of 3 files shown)


src/WinRT.Runtime/Projections/IPropertyValue.netstandard2.0.cs
```diff
@@ -327,12 +327,46 @@ private static T CoerceValue<T>(object value)
                 {
                     return (T)(object)guid.ToString("D", global::System.Globalization.CultureInfo.InvariantCulture);
                 }
+                else if (typeof(T) == typeof(byte))
```
Sergio0694, line 330: We can probably avoid this change, we don't really care about NS2.0 anyway (it doesn't support trimming).

src/WinRT.Runtime/Projections/IPropertyValue.net5.cs
```diff
@@ -325,12 +326,46 @@ private static T CoerceValue<T>(object value)
                 {
                     return (T)(object)guid.ToString("D", global::System.Globalization.CultureInfo.InvariantCulture);
                 }
+                else if (typeof(T) == typeof(byte))
```
Sergio0694, line 329: Just so I'm sure I follow, how come we didn't need the 9x9 combinations anymore and just 9 instead?

Sergio0694 (commented):
Looks awesome!! Will share size diffs when the CI finishes 🙂

src/WinRT.Runtime/Projections/IPropertyValue.net5.cs
jkoritzinsky: We're not assuming anything about the passed in `value` type since we need to support user-defined-IConvertible (and other BCL-provided `IConvertible` implementations) anyway. We're just using the `Convert.ToX` methods instead of generic math since we'd need to use the `Convert.ToX` methods anyway and practically that branch would never be trimmable, no matter how smart we make ILC and RyuJIT.

src/WinRT.Runtime/Projections/IPropertyValue.netstandard2.0.cs
jkoritzinsky: Honestly I like the change for the consistency, as the two implementations should be the same anyway.

src/WinRT.Runtime/Projections/IPropertyValue.net5.cs
Sergio0694: I guess I'm still not sure I get it. How is this change allowing all that extra stuff to be trimmed out then? 🤔

jkoritzinsky: We never call `Convert.ToDateTime`, and a number of the other `ToX` methods, so we can trim out the interface definition and all of the implementations of all of those methods. `IConvertible.ToDateTime` in particular is the expensive one.

Sergio0694: Ah, gotcha, makes sense. Thank you! 🙂

src/WinRT.Runtime/Projections/IPropertyValue.net5.cs
```diff
@@ -325,12 +326,46 @@ private static T CoerceValue<T>(object value)
...
+                }
+                else if (typeof(T) == typeof(float))
+                {
+                    return (T)(object)Convert.ToSingle(value, global::System.Globalization.CultureInfo.InvariantCulture);
+                }
+                else if (typeof(T) == typeof(double))
+                {
+                    return (T)(object)Convert.ToDouble(value, global::System.Globalization.CultureInfo.InvariantCulture);
+                }
                 else
```
manodasanW, line 365: We do also call this with `DateTimeOffset` and `TimeSpan`.  Should those be handled here too?

https://github.com/microsoft/CsWinRT/blob/7bdf00b08d6a4d5f4554148d904e9c5d2f0aa00c/src/WinRT.Runtime/Projections/IPropertyValue.net5.cs#L614C1-L645

jkoritzinsky: DateTimeOffset and TimeSpan have never been supported in this scenario (we never coerce to them, we only support when they're already correctly-typed). I'd rather not add new functionality here unless we get an ask to do so.

manodasanW: So in those 2 cases, it will just be handled by the initial if that does `value is T u` right?

manodasanW: Actually I noticed that the block here was limited to NumericScalarTypes so this is equivalent to before.

manodasanW (approved):

PR #1491, opened 2024-02-01
Remove ComponentActivationFactory types

Sergio0694:
These two types weren't really doing anything. We can just implement the interface directly instead. It's also marginally better because creating the type directly is faster in some cases than going through the generated `Activator.CreateInstance` call.

### DIFF
--- a/src/WinRT.Runtime/Interop/IActivationFactory.cs
+++ b/src/WinRT.Runtime/Interop/IActivationFactory.cs
@@ -27,9 +27,9 @@ namespace WinRT.Interop
         /// Activates an instance of the target WinRT type.
         /// </summary>
         /// <returns>The resulting instance.</returns>
-        /// <exception cref="NotSupportedException">
-        /// Thrown if the operation is not supported by the activation factory type. For instance,
-        /// that is the case is the associated type is static or does not have a default constructor.
+        /// <exception cref="NotImplementedException">
+        /// Thrown if the operation is not available on the activation factory type in use. For instance,
+        /// that is the case if the associated type is static or does not have a default constructor.
         /// </exception>
         IntPtr ActivateInstance();
     }
--- a/src/WinRT.Runtime/Interop/IActivationFactory.cs
+++ b/src/WinRT.Runtime/Interop/IActivationFactory.cs
@@ -10,6 +10,9 @@ using WinRT.Interop;
 
 namespace WinRT.Interop
 {
+    /// <summary>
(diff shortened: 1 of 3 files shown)


src/WinRT.Runtime/Interop/IActivationFactory.cs
```diff
@@ -18,6 +21,14 @@ namespace WinRT.Interop
 #endif
     interface IActivationFactory
     {
+        /// <summary>
+        /// Activates an instance of the target WinRT type.
+        /// </summary>
+        /// <returns>The resulting instance.</returns>
+        /// <exception cref="NotSupportedException">
+        /// Thrown if the operation is not supported by the activation factory type. For instance,
+        /// that is the case is the associated type is static or does not have a default constructor.
```
manodasanW, line 30:
```suggestion
        /// that is the case if the associated type is static or does not have a default constructor.
```

src/cswinrt/code_writers.h
```diff
@@ -9623,13 +9623,20 @@ bind_list<write_parameter_name_with_modifier>(", ", signature.params())
...
-            "ComponentActivationFactory" : write_type_name_temp(w, type, "ActivatableComponentActivationFactory<%>", typedef_name_type::Projected);
+        auto is_activatable = !is_static(type) && has_default_constructor(type);
         auto type_name = write_type_name_temp(w, type, "%", typedef_name_type::Projected);
 
+        // If the type is activatable, we implement IActivationFactory by creating an
+        // instance and marshalling it to IntPtr. Otherwise, we just throw an exception.
+        auto activate_instance_body = is_activatable
+            ? w.write_temp(R"(% comp = new %();
+
+    return MarshalInspectable<%>.FromManaged(comp);)", type_name, type_name, type_name)
```
manodasanW, line 9634: Wonder if we should keep it NotImplementedException to keep it as an exception that has an HResult mapping for native callers.

src/WinRT.Runtime/Interop/IActivationFactory.cs
Sergio0694: Fixed in 9ce3b6c601b6da7155ee5a1354b6f51841f6b99e.

src/cswinrt/code_writers.h
Sergio0694: Ah, yeah I'm fine with keeping the exception type! Fixed in 9ce3b6c601b6da7155ee5a1354b6f51841f6b99e 🙂

manodasanW (approved):

PR #1493, opened 2024-02-01
Remove/move internal vftbl types

Sergio0694:
This PR tweaks two vftbl types we had in the shared code:
- Removes `IActivationFactoryVftbl`, as it was not needed (should also save minimal size)
- Moves `IReferenceTrackerTargetVftbl` to WinRT.Runtime, as it's only needed internally now
- Move `IDelegateVftbl` to WinRT.Runtime, only make it public downlevel (otherwise, internal)


src/WinRT.Runtime/Interop/IDelegateVftbl.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+
+namespace WinRT.Interop
+{
+    // This is internal both in the embedded case, but also if we're on modern .NET,
+    // because projections only use this vftbl type downlevel, and running older
+    // projections against a .NET version of CsWinRT is not supported.
```
manodasanW, line 10:
```suggestion
    // projections against a newer version of CsWinRT is not supported.
```

manodasanW (approved):

PR #1495, opened 2024-02-02
Simplify and optimize exception helpers

Sergio0694:
This PR includes the following changes to the exception helpers:
- Remove all vftbl types
- Remove all unused code
- Guard .NET only types with `#if NET` and remove fallback paths


Sergio0694:
Saves 9 KB. Eh, I mean I'll take, plus it's also a little bit faster 😆

image

src/WinRT.Runtime/Interop/ExceptionErrorInfo.cs
```diff
@@ -2,49 +2,17 @@
...
-
-        string GetReference();
-    }
-
+{
 #if NET
     [WinRTExposedType(typeof(ManagedExceptionErrorInfoTypeDetails))]
 #endif
-    internal sealed class ManagedExceptionErrorInfo : IErrorInfo, ISupportErrorInfo
+    internal sealed class ManagedExceptionErrorInfo
```
manodasanW, line 15: Are we no longer projecting these interfaces on the vtable for net standard?

Sergio0694: No, I'm just always using `IUnknownVftbl` and explicitly passing the IID of the error interface to use. All those interfaces were internal anyway and looking at the code it didn't look like the NS2.0 paths were really relying on any of those vtable bits either. All tests are passing too on all targets, so I guess that assumption was correct? 😅

manodasanW: The `ManagedExceptionErrorInfoTypeDetails` where the IIDs are explicitly passed is not today used on .NET Standard.  The following is in theory where this would be used in but only happens on down-level windows versions: https://github.com/microsoft/CsWinRT/blob/2b77d8412adb7b62bf8f1265a6d8b8c3169dde58/src/WinRT.Runtime/ExceptionHelpers.cs#L354

Sergio0694: Mmh right. I'm not sure I understand how are all the tests still passing then (are we missing some coverage here?), but yeah I'll double check that path and revert some downlevel changes if needed then 👍

Sergio0694: Ok I've reverted all downlevel changes and added a separate file for .NET 6+ with that simplified API.

manodasanW (approved):

PR #1497, opened 2024-02-03
Move EventSource<TDelegate> and associated types to WinRT.Runtime

Sergio0694:
This PR completes moving everything into WinRT.Runtime, and dropping all shared code. This also enabled some additional optimizations for `EventSource<T>`, like dropping the proxy delegates to subscribe/unsubscribe. We can just call the methods.


src/WinRT.Runtime/Interop/EventHandlerEventSource.cs
```diff
@@ -0,0 +1,54 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.ComponentModel;
+
+namespace WinRT.Interop
```
manodasanW, line 7: These types maybe should be in the ABI namespace.  Thoughts?

Sergio0694: Oh, yeah no, absolutely, I can move them 👍
So just eg. `ABI.System` I assume?

manodasanW: ABI.System works.  ABI.WinRT.Interop is probably fine too.

src/WinRT.Runtime/Interop/EventHandlerEventSource.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// An <see cref="EventSource{TDelegate}"/> implementation for <see cref="EventHandler"/>.
+    /// </summary>
+    /// <remarks>This type is only meant to be used by generated projections.</remarks>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+#if EMBED
+    internal
+#else
+    public
+#endif  
+    sealed unsafe class EventHandlerEventSource : EventSource<EventHandler>
```
manodasanW, line 19: Does this need to be public?  It looks like only the others are used from codewriters?

src/WinRT.Runtime/Projections/INotifyCollectionChanged.net5.cs
```diff
@@ -15,26 +15,25 @@ namespace ABI.System.Collections.Specialized
...
         {
             global::System.Threading.Interlocked.CompareExchange(ref _CollectionChanged, new(), null);
             return _CollectionChanged;
         }
-        private static global::System.Runtime.CompilerServices.ConditionalWeakTable<object, NotifyCollectionChangedEventSource> CollectionChanged => _CollectionChanged ?? MakeCollectionChangedTable();
+        private static global::System.Runtime.CompilerServices.ConditionalWeakTable<object, NotifyCollectionChangedEventHandlerEventSource> CollectionChanged => _CollectionChanged ?? MakeCollectionChangedTable();
 
 
-        public static unsafe (Action<global::System.Collections.Specialized.NotifyCollectionChangedEventHandler>, Action<global::System.Collections.Specialized.NotifyCollectionChangedEventHandler>) Get_CollectionChanged(IObjectReference obj, object thisObj)
+        public static unsafe global::WinRT.Interop.EventSource<global::System.Collections.Specialized.NotifyCollectionChangedEventHandler> Get_CollectionChanged(IObjectReference obj, object thisObj)
```
manodasanW, line 27: This is a new class, so this change is fine.  But looking at this, is there a breaking change here where these type of functions are now projected differently?

src/WinRT.Runtime/Interop/EventHandlerEventSource.cs
Sergio0694: Yeah I noticed the non-generic one wasn't actually used, I just figured it seemed consistent to expose this one too since it's kinda the same anyway? I'm happy to keep this internal though if you prefer. Wasn't sure if it could be useful at some point later.

src/cswinrt/code_writers.h
```diff
@@ -5289,20 +5289,18 @@ else
         for (auto&& evt : iface.EventList())
         {
                     w.write(R"(%
-% static unsafe (Action<%>, Action<%>) Get_%(% %, object _thisObj)
+% static unsafe global::WinRT.Interop.EventSource<%> Get_%(% %, object _thisObj)
```
manodasanW, line 5292: Wouldn't dropping the old one be a breaking change?

Sergio0694: So, my thinking is that this is not a breaking change because as of the current CsWinRT (as in, pre-AOT, ie. 2.0.6), these methods are in the various generated `<TYPE_NAME>Methods` classes, which are generated as internal (looking at eg. Win2D with ILSpy). So we should be able to change their signatures without affecting public consumers, since those would just go through the public events on the projected types instead, not the interop methods. What do you think?

manodasanW: I agree with if this is within a class that is internal, then just we can drop the old definition and just have the new one.  But you will see there are a couple scenarios that don't fall into this case.  i.e. ones which are public interfaces and have events on them such as `ABI.Windows.UI.Core.ICoreAcceleratorKeysMethods`.  For those, I think we will need to have both the old method and the new one to avoid this being a breaking change.  For new generated projections, we will use the new one and old ones are not broken as the old one is still there.  Thoughts?

src/WinRT.Runtime/Interop/EventHandlerEventSource.cs
Sergio0694: Fixed in https://github.com/microsoft/CsWinRT/pull/1497/commits/bdf347023a7197212454b2756fead3b1323141bc.

src/WinRT.Runtime/Projections/INotifyCollectionChanged.net5.cs
Sergio0694: Resolving this since it's the same as the one below.

src/WinRT.Runtime/Interop/EventHandlerEventSource.cs
manodasanW: I saw this was actually a return type in ICommand, so probably needs to be public.

src/WinRT.Runtime/Interop/EventSourceState{TDelegate}.cs
```diff
@@ -0,0 +1,162 @@
...
+// Licensed under the MIT License.
+
+using System;
+using System.ComponentModel;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+#nullable enable
+
+namespace WinRT.Interop
```
manodasanW, line 11: ABI?

Sergio0694: I thought you just wanted ABI for the implementation types specific to different event handler types? As in, whereas the base infrastructure would just be in WinRT.Interop. The rationale was that the former are tied to specific public types, whereas the latter are just part of the fundamental building blocks provided by CsWinRT to build stuff. I can change that though if you have a strong preference 🙂

manodasanW: I am thinking anything that isn't a projected type that consumers use would go under the ABI namespace.  So the building blocks would be in the ABI namespace so users don't need to worry about seeing it or so.  The exception is if it is internal, then it doesn't really matter.

Sergio0694: Mmh wait, but we already have a bunch of stuff in WinRT.Interop that's just meant for advanced scenarios, like `IID`, I thought the namespace would just be for "advanced scenarios" in a way? And these types are also hidden. But I can move them if you really prefer to not have them here, just let me know what you think.

manodasanW: I agree WinRT.Interop is like for advanced scenarios too.  IID was fine as that can be something some users may want to use.  And same goes for ActivationFactory.  But this I don't see what someone would want to use it for given it is all event implementation logic.  Thoughts?

Sergio0694: Sounds good to me! Done in 51e3e5740b3ca8a000d2510db3e48d9ef8301af7.

src/cswinrt/code_writers.h
```diff
@@ -3827,14 +3827,14 @@ evt.Name());
     void write_event_source_table(writer& w, Event const& evt)
     {
         w.write(R"(
-private volatile static global::System.Runtime.CompilerServices.ConditionalWeakTable<object, EventSource<%>> _%_;
-private static global::System.Runtime.CompilerServices.ConditionalWeakTable<object, EventSource<%>> Make%Table()
+private volatile static global::System.Runtime.CompilerServices.ConditionalWeakTable<object, global::WinRT.Interop.EventSource<%>> _%_;
```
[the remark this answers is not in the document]
Sergio0694: oops, I forgot 😅

manodasanW:
Somehow the changes in the last day regressed a .NET core 3 test:

Failed UnitTest.TestCSharp.TestEventSourceCaching [28 ms]
  Error Message:
   Assert.False() Failure
Expected: False
Actual:   True
  Stack Trace:
     at UnitTest.TestCSharp.TestEventSourceCaching() in D:\a\_work\1\s\src\Tests\UnitTest\TestComponentCSharp_Tests.cs:line 2902

Sergio0694:
Just out of curiosity, how come we're still testing .NET Core 3, since it's been EOL for years now? 😅

manodasanW:
> Just out of curiosity, how come we're still testing .NET Core 3, since it's been EOL for years now? 😅

Thats how we were testing our .NET standard projection.  Might eventually update that test to .NET framework.  Tried to do that last time during pipeline migration but ran into some other issues that need to be looked into.

Sergio0694:
That makes sense. Pushed a change to just change the namespace for the event source implementation types, trying to break things down to better investigate this. Will wait for the CI to finish running and then look into doing the other changes separately.

manodasanW:
Failure looks to be here, seems like event maybe never got removed?  https://github.com/microsoft/CsWinRT/blob/458d48d6738c930e7eb18170cfb58fcbf38868d9/src/Tests/UnitTest/TestComponentCSharp_Tests.cs#L2843

manodasanW:
Target type is not a projected type: ABI.WinRT.Interop.IWeakReferenceSource

 	WinRT.Runtime.dll!WinRT.TypeExtensions.GetHelperType(System.Type type) Line 131	C#
 	WinRT.Runtime.dll!ABI.WinRT.Interop.EventSourceCache.Create(WinRT.IObjectReference obj, int index, System.WeakReference<object> state) Line 77	C#
 	WinRT.Runtime.dll!ABI.WinRT.Interop.EventSource<System.EventHandler<int>>.Subscribe(System.EventHandler<int> handler) Line 103	C#
 	Test.dll!TestComponentCSharp.Class.StaticIntPropertyChanged.add(System.EventHandler<int> value) Line 315	C#
 	UnitTest.dll!UnitTest.TestCSharp.TestEventSourceCaching() Line 2806	C#

So at this line, it probably needs to be the projected type.

https://github.com/microsoft/CsWinRT/blob/462f56bbc4fa6615840438263c8fcb7bd920b5d4/src/WinRT.Runtime/Interop/EventSourceCache.cs#L75

Sergio0694:
Ooh I accidentally picked that one over `WinRT.Interop.IWeakReferenceSource` when changing the namespace 😅
Good catch, thank you!! Will make a new commit changing the namespace again and making sure to keep that as is.

manodasanW (approved):

PR #1504, opened 2024-02-15
Remove managed vtable support on .NET 6+

Sergio0694:
This PR includes these main changes:
- Removes support for managed vtables on .NET 6+ (this was no longer used anyway)
- Removes all `NonPublicConstructor` trimming annotations around `ObjectReference<T>`
  - These are no longer needed now that managed vtables are not supported anymore


dongle-the-gadget:
> Removes support for managed vtables on .NET 6+ (this was no longer used anyway)

Would this break over existing projections?

Sergio0694:
Technically yes, but because you'll get the new Windows SDK projections along with the new CsWinRT runtime library, that will not be a problem. I talked about this with @manodasanW and he mentioned that "the only [scenario that could break] so far I can think of is if someone adds an explicit reference to CsWinRT to get the new WinRT.Runtime for some reason and are not generating a projection. If they are generating a projection, they already need to be consuming the new Windows SDK projection probably.", so it should be fine. And we really need this change to be able to drop all those unnecessary annotations 🙂

src/WinRT.Runtime/ObjectReference.cs
```diff
@@ -571,27 +559,29 @@ public static ObjectReference<T> FromAbi(IntPtr thisPtr, Guid iid)
...
-#if !NET
-            if (typeof(T).IsGenericType)
-#else
+            // so we instead fall back to typeof(T).IsGenericType, which is not a JIT-time constant.
+#if NET
             if (RuntimeHelpers.IsReferenceOrContainsReferences<T>())
-#endif
             {
-                vftblT = (T)typeof(T).GetConstructor(BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.CreateInstance, null, new[] { typeof(IntPtr) }, null).Invoke(new object[] { thisPtr });
+                throw new NotSupportedException("Generic vtable types are not supported.");
```
manodasanW, line 574:
```suggestion
                throw new NotSupportedException("Generic vtable types with references are not supported.");
```

manodasanW (approved):

PR #1505, opened 2024-02-15
Add [WinRTImplementationTypeRcwFactory] type

Sergio0694:
This PR adds a new attribute to avoid reflection to create RCW instances. The attributes are the factories instead.

### DIFF
--- a/src/WinRT.Runtime/ComWrappersSupport.net5.cs
+++ b/src/WinRT.Runtime/ComWrappersSupport.net5.cs
@@ -228,6 +228,14 @@ namespace WinRT
                 return attribute.CreateInstance;
             }
 
+            if (!RuntimeFeature.IsDynamicCodeCompiled)
+            {
+                throw new NotSupportedException(
+                    $"Cannot create an RCW factory for implementation type '{implementationType}', because it doesn't have " +
+                    "a [WinRTImplementationTypeRcwFactory] derived attribute on it. The fallback path for older projections " +
+                    "is not trim-safe, and isn't supported in AOT environments. Make sure to reference updated projections.");
+            }
+
             return CreateRcwFallback(implementationType);
 
             [SuppressMessage("Trimming", "IL2070", Justification = "This fallback path is not trim-safe by design (to avoid annotations).")]
--- a/src/WinRT.Runtime/ComWrappersSupport.net5.cs
+++ b/src/WinRT.Runtime/ComWrappersSupport.net5.cs
@@ -39,7 +39,6 @@ namespace WinRT
         internal static readonly ConditionalWeakTable<Type, InspectableInfo> InspectableInfoTable = new();
         
         [ThreadStatic]
-        [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.NonPublicConstructors)]
         internal static Type CreateRCWType;
 
         private static ComWrappers _comWrappers;
@@ -87,12 +86,12 @@ namespace WinRT
             return InspectableInfoTable.GetValue(_this.GetType(), o => PregenerateNativeTypeInformation(o).inspectableInfo);
         }
 
-        public static T CreateRcwForComObject<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.NonPublicConstructors)] T>(IntPtr ptr)
+        public static T CreateRcwForComObject<T>(IntPtr ptr)
         {
             return CreateRcwForComObject<T>(ptr, true);
         }
 
-        private static T CreateRcwForComObject<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.NonPublicConstructors)] T>(IntPtr ptr, bool tryUseCache)
+        private static T CreateRcwForComObject<T>(IntPtr ptr, bool tryUseCache)
         {
             if (ptr == IntPtr.Zero)
             {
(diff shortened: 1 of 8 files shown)


src/cswinrt/code_writers.h
```diff
@@ -8450,7 +8452,13 @@ global::System.Collections.Concurrent.ConcurrentDictionary<RuntimeTypeHandle, ob
                     }
                 }),
             bind<write_class_members>(type, false, false),
-            bind<write_custom_query_interface_impl>(type));
+            bind<write_custom_query_interface_impl>(type),
+            settings.netstandard_compat ? "" : w.write_temp(R"(
+internal sealed class %RcwFactoryAttribute : WinRTImplementationTypeRcwFactoryAttribute
```
manodasanW, line 8457: From what I see, this is going inside of the RCW class.  Can we make this private?

src/cswinrt/code_writers.h
```diff
@@ -8450,7 +8452,13 @@ global::System.Collections.Concurrent.ConcurrentDictionary<RuntimeTypeHandle, ob
                     }
                 }),
             bind<write_class_members>(type, false, false),
-            bind<write_custom_query_interface_impl>(type));
+            bind<write_custom_query_interface_impl>(type),
+            settings.netstandard_compat ? "" : w.write_temp(R"(
```
manodasanW, line 8456: We can drop the netstandard check, this is a non netstandard code path.

src/cswinrt/code_writers.h
```diff
@@ -8450,7 +8452,13 @@ global::System.Collections.Concurrent.ConcurrentDictionary<RuntimeTypeHandle, ob
                     }
                 }),
             bind<write_class_members>(type, false, false),
-            bind<write_custom_query_interface_impl>(type));
+            bind<write_custom_query_interface_impl>(type),
```
manodasanW, line 8455: I think you are missing putting the generated attribute on the class itself?

src/cswinrt/code_writers.h
Sergio0694: Ooh, cool, thank you! Done in https://github.com/microsoft/CsWinRT/pull/1505/commits/f712e4a4cc3331ca321ca9fcf196fadc832133e2.

src/cswinrt/code_writers.h
Sergio0694: Ooops! Got tricked by the fallback path 😅
Fixed in https://github.com/microsoft/CsWinRT/pull/1505/commits/f712e4a4cc3331ca321ca9fcf196fadc832133e2.

src/cswinrt/code_writers.h
Sergio0694: It's not nested in the RCW, it's at the same level, so it has to be internal. I was following the same pattern of the `[WinRTExposedTypeDetails]` attributes, which are also generated as internal attributes. Since it's all just in the generated projections, it should mostly be the same? What do you think?

manodasanW: The only difference is the `WinRTExposedTypeDetails` are generated in the ABI namespace.  If this is at the same level, can we also generate this class in the ABI namespace instead similar to the WinRTExposedTypeDetails?

Sergio0694: Sure thing! Done in https://github.com/microsoft/CsWinRT/pull/1505/commits/2d0654ee8f390b91ae8578c46afce6994488ae00 🙂

src/WinRT.Runtime/ComWrappersSupport.net5.cs
```diff
@@ -220,8 +219,36 @@ public static void InitializeComWrappers(ComWrappers wrappers = null)
                 return obj => obj;
             }
 
-            var constructor = implementationType.GetConstructor(BindingFlags.NonPublic | BindingFlags.CreateInstance | BindingFlags.Instance, null, new[] { typeof(IObjectReference) }, null);
-            return (IInspectable obj) => constructor.Invoke(new[] { obj.ObjRef });
+            var attribute = implementationType.GetCustomAttribute<WinRTImplementationTypeRcwFactoryAttribute>();
```
manodasanW, line 222: inherit should probably be false here to get the tests to pass.

src/cswinrt/code_writers.h
```diff
@@ -8453,6 +8458,32 @@ global::System.Collections.Concurrent.ConcurrentDictionary<RuntimeTypeHandle, ob
...
+            return;
+        }
+
+        w.write(R"([global::System.ComponentModel.EditorBrowsable(global::System.ComponentModel.EditorBrowsableState.Never)]
+internal sealed class %RcwFactoryAttribute : global::WinRT.WinRTImplementationTypeRcwFactoryAttribute
+{
+    public override object CreateInstance(global::WinRT.IInspectable inspectable)
+        => new %(inspectable.ObjRef);
+}
+)", type.TypeName(), write_type_name_temp(w, type));
```
manodasanW, line 8484:
```suggestion
)", type.TypeName(), 
    bind<write_type_name>(type, typedef_name_type::Projected, false));
```

Sergio0694: Just for me to better understand, what's the difference between the two?

src/WinRT.Runtime/ComWrappersSupport.net5.cs
Sergio0694: Fixed in https://github.com/microsoft/CsWinRT/pull/1505/commits/cfc12166fb462d15769d93167974d367c6a5e30d 🙂

manodasanW (approved):

PR #1507, opened 2024-02-17
Remove [ObjectReferenceWrapper] on .NET 6+

Sergio0694:
This attribute is only needed downlevel (and it's not trim-safe). This PR marks it as obsolete on .NET 6+, makes it conditional in WinRT.Runtime, and stops generating it when targeting .NET 6+ by the code writer. Keeping the type for binary compatibility.

### DIFF
--- a/src/Projections/Test/TestHost.ProbeByHost.cs
+++ b/src/Projections/Test/TestHost.ProbeByHost.cs
@@ -22,7 +22,9 @@ namespace Windows.Foundation
 
 namespace ABI.Windows.Foundation
 {
+#if !NET6_0
     [global::WinRT.ObjectReferenceWrapper(nameof(_obj))]
+#endif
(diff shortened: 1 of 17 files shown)


src/Projections/Test/TestHost.ProbeByHost.cs
```diff
@@ -21,8 +21,10 @@ internal interface IActivationFactory
 }
 
 namespace ABI.Windows.Foundation
-{
-    [global::WinRT.ObjectReferenceWrapper(nameof(_obj))]
+{
+#if !NET6_0
```
manodasanW, line 25: Why not !NET here for consistency with the others.

Sergio0694: I didn't realize `NET` was also defined here, fixed! Thank you 🙂

manodasanW (approved):

PR #1509, opened 2024-02-21
Suppress incorrect CA2257 warnings

Sergio0694:
This PR suppresses some incorrect CS2257 warnings that were polluting our build output. As per [the docs](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2257):
> "Since a type that implements `IDynamicInterfaceCastable` may not implement a dynamic interface in metadata, **calls to an instance interface member that is not an explicit implementation defined on this type are likely to fail at run time**. To avoid run-time errors, mark new interface members static."

This clearly doesn't apply to these cases, since these members were just nested types, and not something you can invoke.


manodasanW:
Related to https://github.com/dotnet/roslyn-analyzers/issues/7106

manodasanW (approved):

jkoritzinsky (approved):

PR #1512, opened 2024-02-21
Allow authoring scenarios to make use of abstract classes in their implementation

manodasanW:
Abstract classes are not a concept in WinRT but authors of WinRT components may want to implement their components where they make of use abstract classes as their base class and might be public.  Given we don't need to worry about projecting them, this PR allows for this scenario by ignoring the base type if it is abstract.

Also fixing an issue where `CsWinRTWindowsMetadata` was never set in authoring scenarios because it was set as part of a target and thereby the source generator was getting an empty value.


Sergio0694 (approved):
This is awesome, thank you!! 🙌

PR #1513, opened 2024-02-22
Fix, obsolete and hide AsType<T>() on .NET 6

Sergio0694:
This PR fixes the `AsType<T>()` method on .NET 6 with the following changes:
- Makes it trim-safe using the new RCW factory attributes from #1505
- Makes the method obsolete and hidden on .NET 6 (as it shouldn't ever be used anyway)

### DIFF
--- a/src/WinRT.Runtime/ObjectReference.cs
+++ b/src/WinRT.Runtime/ObjectReference.cs
@@ -248,6 +248,40 @@ namespace WinRT
             {
                 return (T)attribute.CreateInstance(new IInspectable(this));
             }
+
+            if (!RuntimeFeature.IsDynamicCodeCompiled)
+            {
+                throw new NotSupportedException(
+                    $"Cannot create an RCW instance for type '{typeof(T)}', because it doesn't have a " +
+                    "[WinRTImplementationTypeRcwFactory] derived attribute on it. The fallback path for older projections " +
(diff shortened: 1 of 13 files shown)


src/WinRT.Runtime/ObjectReference.cs
```diff
@@ -231,15 +232,30 @@ public virtual unsafe int TryAs(Guid iid, out IntPtr ppv)
...
+        [Obsolete(AttributeMessages.GenericDeprecatedMessage)]
+        [EditorBrowsable(EditorBrowsableState.Never)]
+#endif
         public T AsType<T>()
         {
             ThrowIfDisposed();
+
+#if NET
+            // Same logic as in 'ComWrappersSupport.CreateFactoryForImplementationType', see notes there
+            var attribute = typeof(T).GetCustomAttribute<WinRTImplementationTypeRcwFactoryAttribute>(inherit: false);
```
manodasanW, line 245: We may still want a non-trim safe fallback as we did in the other PR just in case someone is using it and hadn't moved to new versions of all projections.

Sergio0694: Done in 63272620e92b9af61fe4b15a5d6bdbeec006c056 👍

manodasanW (approved):

Sergio0694:
Merging this, all tests are passing except that `CustomTypes` that's a known issue on NAOT and unrelated.

PR #1514, opened 2024-02-22
Report native memory pressure to GC

manodasanW:
This PR reports the native memory pressure of any CsWinRT projected RCWs to the .NET GC.  The .NET GC keeps track of managed memory and is not aware of native memory usage of anything that the C# managed objects are holding onto.  So, in the case of CsWinRT projected RCWs, the .NET GC is aware of the managed memory usage of CsWinRT's RCW projection implementation, but the native WinRT object being held on can be using much more memory that the GC isn't aware of.  By reporting this native memory pressure to the .NET GC, .NET is able to run the GC more often as needed based on memory usage and thereby making sure any collected objects are indeed finalized releasing the native memory.  There is another similar change on the WinUI side this works together with which reports WinUI native memory pressure.

- Reports native memory pressure based on the `GCPressure` attribute on RCWs and how much it indicates to report
- Reports basic native memory pressure for all `IObjectReferences` as they are holding onto WinRT objects.  This PR uses a simplified approach to achieving this by reporting it for any `IObjectReference`.  In theory, multiple `IObjectReferences` can be pointing to the same WinRT object and we don't need to report native memory pressure each time, but we shouldn't have too much of an impact by doing so.  In order to do the non-simplified approach, it will require another object allocation which isn't finalized until all the `IObjectReferences` held onto by an RCW or was from a QI from it is disposed.  This gets more complicated to implement due to needing to wait for all related `IObjectRefrences` to finalize so deferring on that until there is a need for it.
- Make use of the ComWrapper APIs that were added after .NET 5

Related: https://github.com/dotnet/runtime/issues/36762
Supersedes: https://github.com/microsoft/CsWinRT/pull/837


src/WinRT.Runtime/ComWrappersSupport.cs
```diff
@@ -30,6 +30,8 @@ namespace WinRT
 #endif
     static partial class ComWrappersSupport
     {
+        internal readonly static int GC_PRESSURE_BASE = 1000;
```
Sergio0694, line 33: Make this a `const`?

src/cswinrt/code_writers.h
```diff
@@ -6589,6 +6590,22 @@ private struct InterfaceTag<I>{};
...
                 }),
+            [&](writer& w)
+            {
+                if (!gc_pressure_amount) return;
+                w.write("GC.AddMemoryPressure(%);\n", gc_pressure_amount);
+            },
+            [&](writer& w)
+            {
+                if (!gc_pressure_amount) return;
+                w.write(R"(~%()
```
Sergio0694, line 6601: Is this adding finalizers for almost all types that didn't have one before this PR? Finalizers have non trivial overhead (especially if this would be for so many objects), are we sure it's something worth adding just to track the memory pressure? IIRC the GC wasn't really doing much at all with this information anyway.

manodasanW: Its not all types, it is only a handful which have the special `GCPressure` attribute to indicate their native objects use significant memory.

https://learn.microsoft.com/en-us/uwp/api/windows.foundation.metadata.gcpressureattribute?view=winrt-22621

Sergio0694: Oh I see, so we're only opting in with a finalizer for those and the others just get that baseline from `IObjectReference` which already has a finalizer anyway, got it. Thank you! 🙂

src/WinRT.Runtime/ComWrappersSupport.cs
manodasanW: Addressed.

Sergio0694 (approved):

PR #1519, opened 2024-02-29
Fix nullable and IPropertySet scenarios

manodasanW:
- Refactor nullable to be better AOT friendly.  We now detect that we are getting a nullable runtimeclass and instead of trying to create an instance of the nullable type and its helper type, we figure out what it is a nullable of and use that to get the nullable value.  This way we avoid having to call `MakeGenericType` to create the nullable helper type.
- We also now use `WinRTExposedTypeDetails` which is placed as part of an attribute on structs and delegates to implement an interface which assists in handling resolving the nullable for that type.
- Changed `MarshalInterface` to reduce the dependency on the helper type.  We were mostly using it to get the IID and for some .NET standard scenarios.  We are able to get the IID using the projected type too so doing that instead as it leads to avoiding `MakeGenericType` calls.  With this change, we might be calculating the IID one additional time than we used to.
- `IPropertySet` is a non-generic interface that inherits other generic interfaces.  If we had a function return an instance of it and the class that implements it was trimmed, we were running into scenarios where the IDIC cast calls failed due to it was for a generic interface.  Addressing that issue by generating a fallback `impl class` for scenarios where we have interfaces implementing generic interfaces.  We were already doing this for generic interfaces, but we also now do it for this scenario too.  This allows to avoid IDIC casts and have the scenario work.
- As part of generating the `impl class`, it was found that depending on which interfaces you implement you can run into compiler errors due to `InterfaceTag` missing.  Given we are going away from `InterfaceTag` as it relied on IDIC casts, refactored most of the manually projected interfaces which didn't have the statics method class to have one and for it to instead use that.  There are still 2 left which will be addressed in another PR.

### DIFF
--- a/src/WinRT.Runtime/Marshalers.cs
+++ b/src/WinRT.Runtime/Marshalers.cs
@@ -1374,16 +1374,23 @@ namespace WinRT
     {
 #if NET
         [DynamicallyAccessedMembers(
-            DynamicallyAccessedMemberTypes.PublicFields | 
-            DynamicallyAccessedMemberTypes.NonPublicFields | 
+            DynamicallyAccessedMemberTypes.PublicFields |
             DynamicallyAccessedMemberTypes.PublicNestedTypes |
             DynamicallyAccessedMemberTypes.PublicMethods)]
 #endif
-        private static readonly Type HelperType = typeof(T).GetHelperType();
-        private static FieldInfo _ObjField;
+        private static Type _HelperType;
 
-        // Either a Func<T, IObjectReference> (JIT) or a boxed Guid (NativeAOT)
-        private static object _CreateMarshalerOrIid;
+#if NET
+        [DynamicallyAccessedMembers(
+            DynamicallyAccessedMemberTypes.PublicFields |
+            DynamicallyAccessedMemberTypes.PublicNestedTypes |
+            DynamicallyAccessedMemberTypes.PublicMethods)]
+#endif
+        private static Type HelperType => _HelperType ??= typeof(T).GetHelperType();
+
+        private static Func<T, IObjectReference> _CreateMarshaler;
+        private static object _Iid;
+        private static Guid IID => (Guid)(_Iid ??= GuidGenerator.CreateIID(typeof(T)));
 
         public static T FromAbi(IntPtr ptr)
         {
@@ -1402,6 +1409,7 @@ namespace WinRT
                 return null;
             }
 
+#if !NET
             if (TryGetObjFieldValue(value, out IObjectReference objectReference))
             {
                 IntPtr ptr = objectReference.GetRef();
@@ -1411,6 +1419,7 @@ namespace WinRT
                 // on the same thread (and as a result don't need to capture context).
                 return ObjectReference<IUnknownVftbl>.Attach(ref ptr);
             }
+#endif
 
             return CreateMarshalerCore(value);
         }
@@ -1422,12 +1431,14 @@ namespace WinRT
                 return new ObjectReferenceValue();
             }
 
+#if !NET
             if (TryGetObjFieldValue(value, out IObjectReference objectReference))
             {
                 return objectReference.AsValue();
             }
+#endif
 
-            return MarshalInspectable<T>.CreateMarshaler2(value, iid == default ? GuidGenerator.GetIID(HelperType) : iid, true);
+            return MarshalInspectable<T>.CreateMarshaler2(value, iid == default ? IID : iid, true);
         }
 
         public static IntPtr GetAbi(IObjectReference value) =>
@@ -1479,6 +1490,8 @@ namespace WinRT
 
         public static unsafe void DisposeAbiArray(object box) => MarshalInterfaceHelper<T>.DisposeAbiArray(box);
 
+#if !NET
+        private static FieldInfo _ObjField;
         private static bool TryGetObjFieldValue(T value, out IObjectReference objectReference)
         {
             // If the value passed in is the native implementation of the interface,
@@ -1496,6 +1509,7 @@ namespace WinRT
 
             return false;
         }
+#endif
 
         private static IObjectReference CreateMarshalerCore(T value)
         {
@@ -1503,17 +1517,15 @@ namespace WinRT
             // On NativeAOT, we can inline everything and skip creating any delegates
             if (!RuntimeFeature.IsDynamicCodeCompiled)
             {
-                _CreateMarshalerOrIid ??= GuidGenerator.GetIID(HelperType);
-
-                return MarshalInspectable<T>.CreateMarshaler<IUnknownVftbl>(value, (Guid)_CreateMarshalerOrIid, true);
+                return MarshalInspectable<T>.CreateMarshaler<IUnknownVftbl>(value, IID, true);
             }
 #endif
             // Otherwise, just use the fallback path
 #pragma warning disable IL3050 // https://github.com/dotnet/runtime/issues/97273
-            _CreateMarshalerOrIid ??= BindCreateMarshaler();
+            _CreateMarshaler ??= BindCreateMarshaler();
 #pragma warning restore IL3050
 
-            return ((Func<T, IObjectReference>)_CreateMarshalerOrIid)(value);
+            return _CreateMarshaler(value);
         }
 
 #if NET8_0_OR_GREATER
@@ -1521,7 +1533,6 @@ namespace WinRT
 #endif
         private static Func<T, IObjectReference> BindCreateMarshaler()
         {
-            Guid iid = GuidGenerator.GetIID(HelperType);
             var vftblType = HelperType.FindVftblType();
 
             if (vftblType is not null)
@@ -1529,10 +1540,10 @@ namespace WinRT
                 var methodInfo = typeof(MarshalInspectable<T>).GetMethod("CreateMarshaler", new Type[] { typeof(T), typeof(Guid), typeof(bool) }).
                     MakeGenericMethod(vftblType);
                 var createMarshaler = (Func<T, Guid, bool, IObjectReference>)methodInfo.CreateDelegate(typeof(Func<T, Guid, bool, IObjectReference>));
-                return obj => createMarshaler(obj, iid, true);
+                return obj => createMarshaler(obj, IID, true);
             }
 
-            return obj => MarshalInspectable<T>.CreateMarshaler<IUnknownVftbl>(obj, iid, true);
+            return obj => MarshalInspectable<T>.CreateMarshaler<IUnknownVftbl>(obj, IID, true);
         }
     }
 
--- a/src/WinRT.Runtime/Marshalers.cs
+++ b/src/WinRT.Runtime/Marshalers.cs
@@ -1388,7 +1388,7 @@ namespace WinRT
 #endif
         private static Type HelperType => _HelperType ??= typeof(T).GetHelperType();
 
-        private static Func<T, IObjectReference> _CreateMarshaler;
+        private static object _CreateMarshaler;
 
         // We are here using CreateIID on the projected type rather than GetIID on the helper type
         // This allows us to avoid needing to do MakeGenericType calls or 
@@ -1400,7 +1400,17 @@ namespace WinRT
         // one in the helper type, but we are also caching it here too so it should be only
         // one additional call.
         private static object _Iid;
-        private static Guid IID => (Guid)(_Iid ??= GuidGenerator.CreateIID(typeof(T)));
+        private static Guid IID => (Guid)(_Iid ??= GetIID());
+
+        private static Guid GetIID()
+        {
+            if (typeof(T).IsGenericType && typeof(T).GetGenericTypeDefinition() == typeof(Nullable<>))
+            {
+                return GuidGenerator.CreateIID(typeof(T));
+            }
+            
+            return GuidGenerator.GetIID(HelperType);
+        }
 
         public static T FromAbi(IntPtr ptr)
         {
@@ -1535,7 +1545,7 @@ namespace WinRT
             _CreateMarshaler ??= BindCreateMarshaler();
 #pragma warning restore IL3050
 
-            return _CreateMarshaler(value);
+            return ((Func<T, IObjectReference>)_CreateMarshaler)(value);
         }
 
 #if NET8_0_OR_GREATER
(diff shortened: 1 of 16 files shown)


manodasanW:
CI on rebased branch of these [changes](https://github.com/microsoft/CsWinRT/tree/2ce13b832113c3b6fb04cf0dad8dcdb580b0e34e) passed: 2.0.7-prerelease-ci.240228.1

src/WinRT.Runtime/Marshalers.cs
```diff
@@ -1374,16 +1374,33 @@ struct MarshalInterface<T>
...
-        private static object _CreateMarshalerOrIid;
+#if NET
+        [DynamicallyAccessedMembers(
+            DynamicallyAccessedMemberTypes.PublicFields |
+            DynamicallyAccessedMemberTypes.PublicNestedTypes |
+            DynamicallyAccessedMemberTypes.PublicMethods)]
+#endif
+        private static Type HelperType => _HelperType ??= typeof(T).GetHelperType();
+
+        private static Func<T, IObjectReference> _CreateMarshaler;
```
Sergio0694, line 1391: Why is this an explicit field? That was done on purpose (reusing the same field, I mean) to avoid one delegate generic instantiation on NAOT. Is this done now because we'll need that `IID` field on both paths?

src/WinRT.Runtime/Marshalers.cs
```diff
@@ -1374,16 +1374,33 @@ struct MarshalInterface<T>
...
+        // This allows us to avoid needing to do MakeGenericType calls or 
+        // registration of helper types for projected generic types like System.Nullable<>.
+        // By using CreateIID with the projected type, we are still able to get the IID
+        // but at the same time don't need the generic instance of the helper type to do so.
+        // In addition, other than for that, we don't need the helper type in MarshalInterface.
+        // This does mean we are doing the PIID calculation here instead of using the cached
+        // one in the helper type, but we are also caching it here too so it should be only
+        // one additional call.
+        private static object _Iid;
+        private static Guid IID => (Guid)(_Iid ??= GuidGenerator.CreateIID(typeof(T)));
```
Sergio0694, line 1403: Could we not check whether `T` is generic and a nullable type (which we're trying to make intrinsics in .NET 9), so that in those cases we wouldn't even have this additional call, and it'd just be pay-for-play for nullables?

src/WinRT.Runtime/Projections/Nullable.cs
```diff
@@ -2203,6 +2006,139 @@ internal static class Nullable_Delegates
...
+            finally
+            {
+                Marshal.Release(nullablePtr);
+            }
+        }
+    }
+
+    internal static class NullableType
+    {
+        internal static Guid GetIIDForBuiltInType(global::System.Type type)
```
Sergio0694, line 2032: All these type checks where in a generic context before so that everything could be correctly elided, due to generic specialization over value types. If we're just branching on a `Type` instance now, we're losing that, meaning we'll end up rooting every single path. This is not so much a big deal here but it is a bit below where each path also triggers a `NullableBlittable<T>` instantiation and delegate creation. Can we keep these methods generic?

src/WinRT.Runtime/Projections/Nullable.cs
```diff
@@ -2203,6 +2006,139 @@ internal static class Nullable_Delegates
...
+            if (type == typeof(global::System.Numerics.Quaternion)) return NullableBlittable<global::System.Numerics.Quaternion>.GetValue;
+            if (type == typeof(global::System.Numerics.Vector2)) return NullableBlittable<global::System.Numerics.Vector2>.GetValue;
+            if (type == typeof(global::System.Numerics.Vector3)) return NullableBlittable<global::System.Numerics.Vector3>.GetValue;
+            if (type == typeof(global::System.Numerics.Vector4)) return NullableBlittable<global::System.Numerics.Vector4>.GetValue;
+            if (type == typeof(global::System.EventHandler)) return Nullable_EventHandler.GetValue;
+            if (type.IsEnum && Enum.GetUnderlyingType(type) == typeof(int)) return (inspectable) => Nullable_IntEnum.GetValue(type, inspectable);
+            if (type.IsEnum && Enum.GetUnderlyingType(type) == typeof(uint)) return (inspectable) => Nullable_FlagsEnum.GetValue(type, inspectable);
+
+#if NET
+            var winrtExposedClassAttribute = type.GetCustomAttribute<WinRTExposedTypeAttribute>(false);
```
Sergio0694, line 2104: To reduce code size (after making this method generic), we could just outline this fallback code to a non-generic method that just takes a `Type` instance, that'd be fine, since the type isn't statically known at this point anyway.

Sergio0694 (commented):
Did a first review pass and left some notes 🙂

src/WinRT.Runtime/Marshalers.cs
manodasanW: right, IID is used as part of `BindCreateMarshaler` which is what was also stored here.  Do we want to instead make this `object` to avoid the generic instantiation in the delegate?  We can also in theory use the helper type to get the IID in that code path as it is not a NAOT code path.

src/WinRT.Runtime/Marshalers.cs
manodasanW: Trying to understand what you are proposing here.  Are you saying for nullable scenarios we should use CreateIID while for others we should do what we did before where we used the helper type to get the IID?

src/WinRT.Runtime/Projections/Nullable.cs
manodasanW: I can easily make `GetIIDForBuiltInType` generic.  But `GetValueFactory` is a bit harder given type is a parameter and it is called from `CreateTypedRcwFactory`.

src/WinRT.Runtime/Marshalers.cs
Sergio0694: So what I'm saying is, with this change, we're always using `CreateIID`, which is pretty expensive as it computes the whole signature and hashes it at runtime. And my understanding is that's only ever needed in those few nullable cases, which would otherwise not work on NAOT at all due to them not being able to resolve the helper type from here. So what I'm suggesting is: let's make this pay for play, and stick to resolving the helper type if `T` is not some `Nullable<>` type. Otherwise, we can keep doing this. Essentially I'm suggesting to make this something like this:
```csharp
private static readonly Guid IID = MakeIID();

private static Guid MakeIID()
{
    if (typeof(T).IsGenericType && typeof(T).GetGenericTypeDefinition() == typeof(Nullable<>))
    {
        return GuidGenerator.CreateIID(typeof(T));
    }

    return GuidGenerator.GetIID(HelperType);
}
```
We also don't need the `Guid` to be an `object` anymore and we can always pre-initialize it.
What do you think?

src/WinRT.Runtime/Marshalers.cs
Sergio0694: Yeah we could just make that an `object` to be fully pay for play here 🤔

src/WinRT.Runtime/Projections/Nullable.cs
Sergio0694: Ah, I see, that's all wired up from the other RCW factory path. Alright let me run some size diffs first, maybe it doesn't hurt too much and we can just say the small size regression there is worth it to make this work with the current approach. In theory that type is very small so maybe we're lucky and it's not a big deal. Be right back 😄

src/WinRT.Runtime/Marshalers.cs
manodasanW: I see what you mean.  I was debating if I wanted this just for `nullable` scenarios or for any `generic` scenario.  Right now, in theory I believe this fixes the other issue where the `generic collection helper type` hasn't been pre-initialized / added to the dictionary which I was planning on fixing but then realized this would probably address it.  But I agree for `non-generic interfaces`, we shouldn't call `CreateIID`.  And for now I can see us scoping this to nullable but I would say expanding it to any generic type can technically be fine too as we are going to be doing the same call within the helper type itself when calling its PIID property.

src/WinRT.Runtime/Projections/Nullable.cs
```diff
@@ -2203,6 +2006,139 @@ internal static class Nullable_Delegates
...
+
+            return default;
+        }
+
+        public static Func<IInspectable, object> GetValueFactory(global::System.Type type)
+        {
+            return ComWrappersSupport.CreateReferenceCachingFactory(GetValueFactoryInternal(type));
+        }
+
+        private static Func<IInspectable, object> GetValueFactoryInternal(global::System.Type type)
```
Sergio0694, line 2069: Hey @MichalStrehovsky, any ideas on how we could change this method to avoid to always unconditionally root all of those generic instantiations even if the types for those checks have been trimmed out? Right now this will otherwise always bring in like 35 KB worth of stuff that you might not even be using at all 🥲

src/WinRT.Runtime/Projections/Nullable.cs
```diff
@@ -390,35 +390,12 @@ internal static Delegate GetValueDelegateForFunctionPointer(IntPtr ptr)
...
-            if (typeof(T) == typeof(global::System.Numerics.Matrix4x4)) return IReferenceIIDs.IReferenceMatrix4x4_IID;
-            if (typeof(T) == typeof(global::System.Numerics.Plane)) return IReferenceIIDs.IReferencePlane_IID;
-            if (typeof(T) == typeof(global::System.Numerics.Quaternion)) return IReferenceIIDs.IReferenceQuaternion_IID;
-            if (typeof(T) == typeof(global::System.Numerics.Vector2)) return IReferenceIIDs.IReferenceVector2_IID;
-            if (typeof(T) == typeof(global::System.Numerics.Vector3)) return IReferenceIIDs.IReferenceVector3_IID;
-            if (typeof(T) == typeof(global::System.Numerics.Vector4)) return IReferenceIIDs.IReferenceVector4_IID;
-#endif
-
-            return GuidGenerator.CreateIID(typeof(Nullable<T>));
+            var iid = NullableType.GetIIDForBuiltInType(typeof(T));
```
Sergio0694, line 393: If we make that generic so it can be inlined, the linker might also be able to fully trim out this fallback path for types that are in fact primitive types. Could save some rooted metadata, since I think that `CreateIID` call will root some stuff.

src/WinRT.Runtime/Projections/Nullable.cs
MichalStrehovsky: If a `System.Type` is all you have, you can't avoid it.

We could optimize this out at the codegen phase because we know this once we have the whole program view. It's a JitInterface change and I've been avoiding those because they're painful. I might look at it because this is not the first time...

Optimizing it out during codegen is not panacea because whatever is under the `if` branch could still cause damage to whole program view that cannot be undone during codegen anymore (like seeing the delegate construction if the app also uses `Delegate.Method` with that delegate) but maybe it will be all that's needed here.

src/WinRT.Runtime/Projections/Nullable.cs
Sergio0694: Just updating here too for context, I guess we can just look at the two things separately. For the one creating RCW factories below, we might just have to leave that one as is at least for now. For this one, we can definitely make it generic though, which can help with inlining and also dropping some fallback code paths. In fact, what I'm thinking is we could change this method to do everything for arbitrary types, like so:

```csharp
public static Guid GetIID<T>()
{
    if (typeof(T) == typeof(int)) return Nullable_int.IID;
    if (typeof(T) == typeof(byte)) return Nullable_byte.IID;
    if (typeof(T) == typeof(bool)) return Nullable_bool.IID;
    if (typeof(T) == typeof(sbyte)) return Nullable_sbyte.IID;
    if (typeof(T) == typeof(short)) return Nullable_short.IID;
    if (typeof(T) == typeof(ushort)) return Nullable_ushort.IID;
    if (typeof(T) == typeof(char)) return Nullable_char.IID;
    if (typeof(T) == typeof(uint)) return Nullable_uint.IID;
    if (typeof(T) == typeof(long)) return Nullable_long.IID;
    if (typeof(T) == typeof(ulong)) return Nullable_ulong.IID;
    if (typeof(T) == typeof(float)) return Nullable_float.IID;
    if (typeof(T) == typeof(double)) return Nullable_double.IID;
    if (typeof(T) == typeof(Guid)) return Nullable_guid.IID;
    if (typeof(T) == typeof(global::System.Type)) return Nullable_Type.IID;
    if (typeof(T) == typeof(global::System.TimeSpan)) return Nullable_TimeSpan.IID;
    if (typeof(T) == typeof(global::System.DateTimeOffset)) return Nullable_DateTimeOffset.IID;
    if (typeof(T) == typeof(global::Windows.Foundation.Point)) return IReferenceIIDs.IReferenceOfPoint_IID;
    if (typeof(T) == typeof(global::Windows.Foundation.Size)) return IReferenceIIDs.IReferenceOfSize_IID;
    if (typeof(T) == typeof(global::Windows.Foundation.Rect)) return IReferenceIIDs.IReferenceOfRect_IID;
    if (typeof(T) == typeof(global::System.Numerics.Matrix3x2)) return IReferenceIIDs.IReferenceMatrix3x2_IID;
    if (typeof(T) == typeof(global::System.Numerics.Matrix4x4)) return IReferenceIIDs.IReferenceMatrix4x4_IID;
    if (typeof(T) == typeof(global::System.Numerics.Plane)) return IReferenceIIDs.IReferencePlane_IID;
    if (typeof(T) == typeof(global::System.Numerics.Quaternion)) return IReferenceIIDs.IReferenceQuaternion_IID;
    if (typeof(T) == typeof(global::System.Numerics.Vector2)) return IReferenceIIDs.IReferenceVector2_IID;
    if (typeof(T) == typeof(global::System.Numerics.Vector3)) return IReferenceIIDs.IReferenceVector3_IID;
    if (typeof(T) == typeof(global::System.Numerics.Vector4)) return IReferenceIIDs.IReferenceVector4_IID;

    return GuidGenerator.CreateIID(typeof(Nullable<T>));
}
```
That `PIID` field initializer above can then also just be `NullableType.GetIID<T>();`.
(the method can also be public, since it's in an internal type already).

manodasanW: Addressed.

src/WinRT.Runtime/Projections/Nullable.cs
manodasanW: Addressed.

src/WinRT.Runtime/Projections/Nullable.cs
manodasanW: Connected to the other feedback where this function can't be generic due to how it is used right now.

Sergio0694 (approved):
LGTM, let's get this in! 🚀

PR #1520, opened 2024-03-01
Obsolete Marshaler<T>.RefAbiType on .NET 6+

Sergio0694:
This property is never used on .NET and it's rooting some type builder stuff. Obsoleting it and disabling it.


manodasanW:
Checked back a couple versions, looks unused so should be fine.

manodasanW (approved):

PR #1521, opened 2024-03-01
Stop rooting nested types for helper types

Sergio0694:
While looking at sizoscope traces with @jkoritzinsky, he noticed that the linker was rooting a considerable amount of code due to virtual methods on nested types for helper types, ie. all of the vtable types. These vtable types are no longer used at all with new projections (and NAOT will not even attempt to find them at all, and will just fail if using old projections). This PR updates the trimming annotations to stop rooting all of these nested vtable types, making the fallback path using older projections not trim-safe. It will still work though, and as long as trimming is not using, consumers using old projections (even indirectly) will have see no behavior change. This makes the back-compat support pay-for-play and avoid the binary size increase for customers using new projections and NAOT especially, and it also makes the behavior more consistent across NAOT and CoreCLR with trimming.

See:


dongle-the-gadget:
This would break over existing trimmed users.

Sergio0694:
Yes, but we could detail that in our release notes, and it would only break users of other 3rd party WinRT components that don't get updated in sync (which combined with the users of C#, and using trimming, makes for a very small number of users). Those users can wait to make sure their dependencies are updated before bumping CsWinRT, or alternatively they can make sure to manually root those projection assemblies as a temporary workaround. But this way we can make this pay for play.

src/WinRT.Runtime/TypeExtensions.cs
```diff
@@ -155,23 +149,41 @@ public static Type GetGuidType(this Type type)
...
-            }
-            if (helperType.IsGenericType && vftblType is object)
+#if NET8_0_OR_GREATER
+            [RequiresDynamicCode(AttributeMessages.MarshallingOrGenericInstantiationsRequiresDynamicCode)]
+#endif
+#if NET
+            [DynamicDependency(DynamicallyAccessedMemberTypes.All, "ABI.Windows.Foundation.IAsyncActionWithProgress`1+Vftbl", "Microsoft.Windows.SDK.NET")]
+            [DynamicDependency(DynamicallyAccessedMemberTypes.All, "ABI.Windows.Foundation.IAsyncOperationWithProgress`2+Vftbl", "Microsoft.Windows.SDK.NET")]
+            [DynamicDependency(DynamicallyAccessedMemberTypes.All, "ABI.Windows.Foundation.IAsyncOperation`1+Vftbl", "Microsoft.Windows.SDK.NET")]
+            [DynamicDependency(DynamicallyAccessedMemberTypes.All, "ABI.Windows.Foundation.Collections.IMapChangedEventArgs`1+Vftbl", "WinRT.Runtime")]
```
manodasanW, line 159: These are also Microsoft.Windows.SDK.NET

src/WinRT.Runtime/TypeExtensions.cs
```diff
@@ -155,23 +149,41 @@ public static Type GetGuidType(this Type type)
...
-            if (vftblType is null)
-            {
-                return null;
-            }
-            if (helperType.IsGenericType && vftblType is object)
+#if NET8_0_OR_GREATER
+            [RequiresDynamicCode(AttributeMessages.MarshallingOrGenericInstantiationsRequiresDynamicCode)]
+#endif
+#if NET
+            [DynamicDependency(DynamicallyAccessedMemberTypes.All, "ABI.Windows.Foundation.IAsyncActionWithProgress`1+Vftbl", "Microsoft.Windows.SDK.NET")]
```
manodasanW, line 156: You are missing the generic collections from [here](https://github.com/microsoft/CsWinRT/blob/dc4610e557a85b11d4d5b22dc1957fab36c3e5bc/src/WinRT.Runtime/Projections.cs#L76-L84)

src/WinRT.Runtime/TypeExtensions.cs
Sergio0694: Oh whoops, fixed! 😅

src/WinRT.Runtime/TypeExtensions.cs
Sergio0694: Fixed in https://github.com/microsoft/CsWinRT/pull/1521/commits/9d234477b577cd3089b7d0ef7160cacd718ed4c2! 🙂

Sergio0694:
Saves 213 KB 🎉

image

src/WinRT.Runtime/TypeExtensions.cs
```diff
@@ -155,23 +149,49 @@ public static Type GetGuidType(this Type type)
...
+            [DynamicDependency(DynamicallyAccessedMemberTypes.All, "ABI.System.EventHandler`1+Vftbl", "WinRT.Runtime")]
+            [DynamicDependency(DynamicallyAccessedMemberTypes.All, "ABI.System.System.Collections.Generic.KeyValuePair`2+Vftbl", "WinRT.Runtime")]
+            [DynamicDependency(DynamicallyAccessedMemberTypes.All, "ABI.System.System.Collections.Generic.IEnumerable`1+Vftbl", "WinRT.Runtime")]
+            [DynamicDependency(DynamicallyAccessedMemberTypes.All, "ABI.System.System.Collections.Generic.IEnumerator`1+Vftbl", "WinRT.Runtime")]
+            [DynamicDependency(DynamicallyAccessedMemberTypes.All, "ABI.System.System.Collections.Generic.IList`1+Vftbl", "WinRT.Runtime")]
+            [DynamicDependency(DynamicallyAccessedMemberTypes.All, "ABI.System.System.Collections.Generic.IReadOnlyList`1+Vftbl", "WinRT.Runtime")]
+            [DynamicDependency(DynamicallyAccessedMemberTypes.All, "ABI.System.System.Collections.Generic.IDictionary`2+Vftbl", "WinRT.Runtime")]
+            [DynamicDependency(DynamicallyAccessedMemberTypes.All, "ABI.System.System.Collections.Generic.IReadOnlyDictionary`2+Vftbl", "WinRT.Runtime")]
```
jkoritzinsky, line 169: Is this list fixed or can it grow with additional custom projections?

For the types defined outside of WinRT.Runtime, is this temporary until an updated projection is produced?

What about embedded-CsWinRT scenarios? Those would need different assembly names here.

manodasanW: This list is fixed, it is only for generic interfaces and developers can't define their own generic interface.  This is just for compat with pre-existing projections being used in apps that have trimming enabled and have an updated Windows SDK projection / WinRT.Runtime.  All of the newly generated projections will not need the vtbl.  This also means for embedded scenarios, if they have this it means the updated version of CsWinRT is being used to generate the embedded projection and thereby it should be using the newly generated projection which doesn't need the vftbl type.

manodasanW (approved):

Issue #970, opened 2021-08-19
DataTransferManager and Share API


Marv51:
**Describe the bug**
In WinUI 2 I had a "Share" functionality to share an image with Windows 10 system APIs. `DataTransferManager` has an event called DataRequested that you need to subscribe to that makes this work.

In WinUI3 with CsWinRT this is pretty difficult, because you need to get your `DataTransferManager from IDataTransferManagerInterop, like this:

```csharp
using Windows.ApplicationModel.DataTransfer;
using Windows.Win32.UI.Shell;

var guid = Guid.Parse("a5caee9b-8708-49d1-8d36-67d25a8da00c");
var iop = DataTransferManager.As<IDataTransferManagerInterop>();
var manager = iop.GetForWindow(GetMainHwnd(), guid);
var dtm = manager as DataTransferManager; // cast is always null here, I believe manager should be cast to an interface, but no interface for DataTransferManager exists.
```

However, it is unclear to me how to use that returned `object`. I can't cast to a `DataTransferManager`, maybe because that would only work with an interface? But DataTransferManager does not have any interface in [the API docs](https://docs.microsoft.com/en-us/uwp/api/windows.applicationmodel.datatransfer.datatransfermanager).

I'm getting the IDataTransferManagerInterop from CsWin32,[ I asked on their repo](https://github.com/microsoft/CsWin32/issues/246), but no solution could be found.

Is this a projection issue with CsWinRT, or should I be asking somewhere else?

angelazhangmsft:
@Marv51 there are details in #966 that may help with this scenario. You will need to do something like the following, where `eventArg` is the `DataTransferManager` object instead.

```csharp
var args = SomeEventArgType.FromAbi(((IInspectable)eventArg).ThisPtr);
```

Marv51:
@angelazhangmsft Thank you so much! Unfortunately, when I cast to IInspectable I get an InvalidCastException. 

Could this be happening, because in CsWin32 the object is marshalled as IUnknown?
```csharp
[Guid("3A3DCD6C-3EAB-43DC-BCDE-45671CE800C8"),InterfaceType(ComInterfaceType.InterfaceIsIUnknown),ComImport()]
public interface IDataTransferManagerInterop 
{
  [return: MarshalAs(UnmanagedType.IUnknown)]
  unsafe object GetForWindow(win32.Foundation.HWND appWindow, global::System.Guid* riid);
  //...
```
Full exception:
[...]

PR #1528, opened 2024-03-07
Add DataTransferManger interop APIs

Gaoyifei1011:
Add DataTransferManger interop APIs

### DIFF
--- a/src/Tests/UnitTest/ComInteropTests.cs
+++ b/src/Tests/UnitTest/ComInteropTests.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Runtime.InteropServices;
+using Windows.ApplicationModel.DataTransfer;
 using Windows.ApplicationModel.DataTransfer.DragDrop.Core;
 using Windows.Graphics.Display;
 using Windows.Graphics.Printing;
@@ -160,6 +161,13 @@ namespace UnitTest
         {
             Assert.Throws<COMException>(() => DisplayInformationInterop.GetForWindow(new IntPtr(0)));
             Assert.Throws<COMException>(() => DisplayInformationInterop.GetForMonitor(new IntPtr(0)));
+        }
+
+        [Fact]
+        public void TestDataTransferManager()
+        {
+            Assert.Throws<COMException>(() => DataTransferManagerInterop.GetForWindow(new IntPtr(0)));
+            DataTransferManagerInterop.ShowShareUIForWindow(new IntPtr(0));
         }
     }
 }
--- a/src/Tests/UnitTest/ComInteropTests.cs
+++ b/src/Tests/UnitTest/ComInteropTests.cs
@@ -167,7 +167,7 @@ namespace UnitTest
         public void TestDataTransferManager()
         {
             Assert.Throws<COMException>(() => DataTransferManagerInterop.GetForWindow(new IntPtr(0)));
-            DataTransferManagerInterop.ShowShareUIForWindow(new IntPtr(0));
+            Assert.Throws<COMException>(() => DataTransferManagerInterop.ShowShareUIForWindow(new IntPtr(0)));
         }
     }
 }
--- a/src/cswinrt/strings/ComInteropHelpers.cs
+++ b/src/cswinrt/strings/ComInteropHelpers.cs
@@ -534,7 +534,9 @@ namespace Windows.ApplicationModel.DataTransfer
             try
             {
                 // IDataTransferManagerInterop inherits IUnknown (3 functions) and provides GetForWindow giving a total of 5 functions
-                global::WinRT.ExceptionHelpers.ThrowExceptionForHR((*(delegate* unmanaged[Stdcall]<global::System.IntPtr, global::System.IntPtr, global::System.Guid, out global::System.IntPtr, int>**)thisPtr)[3](thisPtr, appWindow, riid, out ptr));
+                fixed(Guid* _riid = &riid)
+                global::WinRT.ExceptionHelpers.ThrowExceptionForHR((*(delegate* unmanaged[Stdcall]<global::System.IntPtr, global::System.IntPtr, global::System.Guid*, global::System.IntPtr*, int>**)thisPtr)[3](thisPtr, appWindow, _riid, &ptr));
+
                 return global::WinRT.MarshalInspectable<global::Windows.ApplicationModel.DataTransfer.DataTransferManager>.FromAbi(ptr);
             }
             finally
--- a/src/cswinrt/strings/ComInteropHelpers.cs
+++ b/src/cswinrt/strings/ComInteropHelpers.cs
@@ -504,7 +504,7 @@ namespace Windows.ApplicationModel.DataTransfer
         private static readonly global::System.Guid riid = new global::System.Guid(0xA5CAEE9B, 0x8708, 0x49D1, 0x8D, 0x36, 0x67, 0xD2, 0x5A, 0x8D, 0xA0, 0x0C);
 
 #if NET
-        private static global::WinRT.IObjectReference objectReference => global::WinRT.ActivationFactory.Get<global::WinRT.IObjectReference>("Windows.ApplicationModel.DataTransfer.DataTransferManager", IDataTransferManagerInterop_IID);
+        private static global::WinRT.IObjectReference objectReference => global::WinRT.ActivationFactory.Get("Windows.ApplicationModel.DataTransfer.DataTransferManager", IDataTransferManagerInterop_IID);
 #else
         private static global::WinRT.ObjectReference<global::WinRT.Interop.IUnknownVftbl> objectReference => global::WinRT.ActivationFactory.Get<global::WinRT.Interop.IUnknownVftbl>("Windows.ApplicationModel.DataTransfer.DataTransferManager", IDataTransferManagerInterop_IID);
 #endif
--- a/src/cswinrt/strings/ComInteropHelpers.cs
+++ b/src/cswinrt/strings/ComInteropHelpers.cs
@@ -528,7 +528,8 @@ namespace Windows.ApplicationModel.DataTransfer
         internal static unsafe DataTransferManager GetForWindow(global::WinRT.IObjectReference _obj, global::System.IntPtr appWindow, in global::System.Guid riid)
         {
             global::System.IntPtr thisPtr = _obj.ThisPtr;
-            global::System.IntPtr ptr = new global::System.IntPtr();
+            global::System.IntPtr ptr = default;
+
 
             try
             {
--- a/src/cswinrt/strings/ComInteropHelpers.cs
+++ b/src/cswinrt/strings/ComInteropHelpers.cs
@@ -512,4 +512,5 @@ namespace Windows.ApplicationModel.DataTransfer
             dataTransferManagerInterop.ShowShareUIForWindow(appWindow);
         }
     }
+#endif
 }
--- a/src/cswinrt/strings/ComInteropHelpers.cs
+++ b/src/cswinrt/strings/ComInteropHelpers.cs
@@ -499,14 +499,12 @@ namespace Windows.ApplicationModel.DataTransfer
 #endif
     static class DataTransferManagerInterop
     {
-        private static readonly global::System.Guid IDataTransferManagerInterop_IID = new global::System.Guid(0x3A3DCD6C, 0x3EAB, 0x43DC, 0xBC, 0xDE, 0x45, 0x67, 0x1C, 0xE8, 0x00, 0xC8);
-
         private static readonly global::System.Guid riid = new global::System.Guid(0xA5CAEE9B, 0x8708, 0x49D1, 0x8D, 0x36, 0x67, 0xD2, 0x5A, 0x8D, 0xA0, 0x0C);
 
 #if NET
-        private static global::WinRT.IObjectReference objectReference => global::WinRT.ActivationFactory.Get("Windows.ApplicationModel.DataTransfer.DataTransferManager", IDataTransferManagerInterop_IID);
+        private static global::WinRT.IObjectReference objectReference = global::WinRT.ActivationFactory.Get("Windows.ApplicationModel.DataTransfer.DataTransferManager", new global::System.Guid(0x3A3DCD6C, 0x3EAB, 0x43DC, 0xBC, 0xDE, 0x45, 0x67, 0x1C, 0xE8, 0x00, 0xC8));
 #else
-        private static global::WinRT.ObjectReference<global::WinRT.Interop.IUnknownVftbl> objectReference => global::WinRT.ActivationFactory.Get<global::WinRT.Interop.IUnknownVftbl>("Windows.ApplicationModel.DataTransfer.DataTransferManager", IDataTransferManagerInterop_IID);
+        private static global::WinRT.ObjectReference<global::WinRT.Interop.IUnknownVftbl> objectReference = global::WinRT.ActivationFactory.Get<global::WinRT.Interop.IUnknownVftbl>("Windows.ApplicationModel.DataTransfer.DataTransferManager", new global::System.Guid(0x3A3DCD6C, 0x3EAB, 0x43DC, 0xBC, 0xDE, 0x45, 0x67, 0x1C, 0xE8, 0x00, 0xC8));
 #endif
 
         public static global::Windows.ApplicationModel.DataTransfer.DataTransferManager GetForWindow(global::System.IntPtr appWindow)
@@ -523,11 +521,7 @@ namespace Windows.ApplicationModel.DataTransfer
 
     internal static class IDataTransferManagerInteropMethods
     {
-#if NET
         internal static unsafe global::Windows.ApplicationModel.DataTransfer.DataTransferManager GetForWindow(global::WinRT.IObjectReference _obj, global::System.IntPtr appWindow, in global::System.Guid riid)
-#else
-        internal static unsafe global::Windows.ApplicationModel.DataTransfer.DataTransferManager GetForWindow(global::WinRT.ObjectReference<global::WinRT.Interop.IUnknownVftbl> _obj, global::System.IntPtr appWindow, in global::System.Guid riid)
-#endif
         {
             global::System.IntPtr thisPtr = _obj.ThisPtr;
             global::System.IntPtr ptr = default;
@@ -546,11 +540,7 @@ namespace Windows.ApplicationModel.DataTransfer
             }
         }
 
-# if NET
         internal static unsafe void ShowShowShareUIForWindow(global::WinRT.IObjectReference _obj, global::System.IntPtr appWindow)
-#else
-        internal static unsafe void ShowShowShareUIForWindow(global::WinRT.ObjectReference<global::WinRT.Interop.IUnknownVftbl> _obj, global::System.IntPtr appWindow)
-#endif
         {
             global::System.IntPtr thisPtr = _obj.ThisPtr;
 
--- a/src/cswinrt/strings/ComInteropHelpers.cs
+++ b/src/cswinrt/strings/ComInteropHelpers.cs
@@ -499,18 +499,55 @@ namespace Windows.ApplicationModel.DataTransfer
 #endif
     static class DataTransferManagerInterop
     {
-        private static IDataTransferManagerInterop dataTransferManagerInterop = DataTransferManager.As<IDataTransferManagerInterop>();
+        private static readonly global::System.Guid IDataTransferManagerInterop_IID = new global::System.Guid("3A3DCD6C-3EAB-43DC-BCDE-45671CE800C8");
 
-        public static DataTransferManager GetForWindow(IntPtr appWindow)
+        private static readonly global::System.Guid riid = new global::System.Guid("A5CAEE9B-8708-49D1-8D36-67D25A8DA00C");
+
+        private static global::WinRT.FactoryObjectReference<global::WinRT.IObjectReference> factoryObjectReference;
+
+        static DataTransferManagerInterop()
         {
-            Guid iid = GuidGenerator.CreateIID(typeof(IDataTransferManager));
-            return (DataTransferManager)dataTransferManagerInterop.GetForWindow(appWindow, iid);
+            // typeName = typeof(Windows.ApplicationModel.DataTransfer.DataTransferManager).FullName
+            factoryObjectReference = global::WinRT.ActivationFactory.Get<global::WinRT.IObjectReference>("Windows.ApplicationModel.DataTransfer.DataTransferManager", IDataTransferManagerInterop_IID);
         }
 
-        public static void ShowShareUIForWindow(IntPtr appWindow)
+        public static DataTransferManager GetForWindow(global::System.IntPtr appWindow)
         {
-            dataTransferManagerInterop.ShowShareUIForWindow(appWindow);
+            return IIDataTransferManagerInteropMethods.GetForWindow(factoryObjectReference, appWindow, riid);
+        }
+
+        public static void ShowShareUIForWindow(global::System.IntPtr appWindow)
+        {
+            IIDataTransferManagerInteropMethods.ShowShowShareUIForWindow(factoryObjectReference, appWindow);
         }
     }
 #endif
+
+    internal static class IIDataTransferManagerInteropMethods
+    {
+        internal static unsafe DataTransferManager GetForWindow(global::WinRT.IObjectReference _obj, global::System.IntPtr appWindow, in global::System.Guid riid)
+        {
+            global::System.IntPtr thisPtr = _obj.ThisPtr;
+            global::System.IntPtr ptr = new global::System.IntPtr();
+
+            try
+            {
+                // IDataTransferManagerInterop inherits IUnknown (3 functions) and provides GetForWindow giving a total of 5 functions
+                global::WinRT.ExceptionHelpers.ThrowExceptionForHR((*(delegate* unmanaged[Stdcall]<global::System.IntPtr, global::System.IntPtr, global::System.Guid, out global::System.IntPtr, int>**)thisPtr)[3](thisPtr, appWindow, riid, out ptr));
+                return global::WinRT.MarshalInspectable<global::Windows.ApplicationModel.DataTransfer.DataTransferManager>.FromAbi(ptr);
+            }
+            finally
+            {
+                global::WinRT.MarshalInspectable<DataTransferManager>.DisposeAbi(ptr);
+            }
+        }
+
+        internal static unsafe void ShowShowShareUIForWindow(global::WinRT.IObjectReference _obj, global::System.IntPtr appWindow)
+        {
+            global::System.IntPtr thisPtr = _obj.ThisPtr;
+
+            // IDataTransferManagerInterop inherits IUnknown (3 functions) and provides ShowShowShareUIForWindow giving a total of 5 functions
+            global::WinRT.ExceptionHelpers.ThrowExceptionForHR((*(delegate* unmanaged[Stdcall]<global::System.IntPtr, global::System.IntPtr, int>**)thisPtr)[4](thisPtr, appWindow));
+        }
+    }
 }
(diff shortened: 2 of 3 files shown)


src/cswinrt/strings/ComInteropHelpers.cs
```diff
@@ -485,3 +485,31 @@ public static DisplayInformation GetForMonitor(IntPtr monitor)
...
+    [global::System.Runtime.Versioning.SupportedOSPlatform("Windows10.0.10240.0")]
+#endif
+#if EMBED
+    internal
+#else
+    public
+#endif
+    static class DataTransferManagerInterop
+    {
+        private static IDataTransferManagerInterop dataTransferManagerInterop = DataTransferManager.As<IDataTransferManagerInterop>();
```
Sergio0694, line 502: This should be readonly, no?

Sergio0694, line 502: You shouldn't use `As<T>`, always pass an explicit IID to avoid rooting reflection metadata.

src/cswinrt/strings/ComInteropHelpers.cs
```diff
@@ -485,3 +485,31 @@ public static DisplayInformation GetForMonitor(IntPtr monitor)
...
+
+        public static DataTransferManager GetForWindow(IntPtr appWindow)
+        {
+            Guid iid = GuidGenerator.CreateIID(typeof(IDataTransferManager));
```
Sergio0694, line 506: Don't use `CreateIID` in cases where we already know the IID. We should just hardcode it.

src/cswinrt/strings/ComInteropHelpers.cs
```diff
@@ -485,3 +485,31 @@ public static DisplayInformation GetForMonitor(IntPtr monitor)
...
+            return (DataTransferManager)dataTransferManagerInterop.GetForWindow(appWindow, iid);
+        }
+
+        public static void ShowShareUIForWindow(IntPtr appWindow)
```
Sergio0694, line 510: Should this use `WindowId` instead? Also has any of this gone through API review?

src/cswinrt/strings/ComInteropHelpers.cs
```diff
@@ -485,3 +485,31 @@ public static DisplayInformation GetForMonitor(IntPtr monitor)
...
+#if UAC_VERSION_1
+#if NET
+    [global::System.Runtime.Versioning.SupportedOSPlatform("Windows10.0.10240.0")]
+#endif
+#if EMBED
+    internal
+#else
+    public
+#endif
+    static class DataTransferManagerInterop
```
Sergio0694, line 500: Might be nice to add XML docs.

Sergio0694 (changes requested):

src/cswinrt/strings/ComInteropHelpers.cs
dongle-the-gadget: WindowId requires 20348+ SDK or WASDK, so I think keeping it in IntPtr is okay.

src/cswinrt/strings/ComInteropHelpers.cs
Gaoyifei1011: image

Currently, other similar interop apis do not have the readonly keyword, and if you need to add the readonly keyword, you may need to create a new PR and change all the interop apis

src/cswinrt/strings/ComInteropHelpers.cs
Gaoyifei1011: > Don't use `CreateIID` in cases where we already know the IID. We should just hardcode it.

image

Currently, other similar interoperable apis are not found to have hard coding, and if you need to add hard coding, you may need to create a new PR and change all the hard coding

src/cswinrt/strings/ComInteropHelpers.cs
Gaoyifei1011: > You shouldn't use `As<T>`, always pass an explicit IID to avoid rooting reflection metadata.

image

At present, other similar interop apis do not find that they use other methods related to As<T>, do you have any good alternative ways to change all the interop apis?

manodasanW:
I think for now as long as this new class is similar for the most part to the other interop classes, it is good.  These look to be the [docs](https://learn.microsoft.com/en-us/windows/win32/api/shobjidl_core/nn-shobjidl_core-idatatransfermanagerinterop) for this interface.  I will be retargeting this PR to the staging branch given this is an API surface change and that is our next planned assembly version change.

manodasanW:
Ran CI build.

src/cswinrt/strings/ComInteropHelpers.cs
```diff
@@ -485,3 +485,31 @@ public static DisplayInformation GetForMonitor(IntPtr monitor)
...
+            Guid iid = GuidGenerator.CreateIID(typeof(IDataTransferManager));
+            return (DataTransferManager)dataTransferManagerInterop.GetForWindow(appWindow, iid);
+        }
+
+        public static void ShowShareUIForWindow(IntPtr appWindow)
+        {
+            dataTransferManagerInterop.ShowShareUIForWindow(appWindow);
+        }
+    }
+}
```
manodasanW, line 515: missing `#endif` at end for UAC_VERSION_1 is causing CI failure.

  ComInteropHelpers.cs(525,1): error CS1027: #endif directive expected

Gaoyifei1011: image
Has updated just now

src/cswinrt/strings/ComInteropHelpers.cs
dongle-the-gadget: As far as I'm aware, the reason why these classes are using `As<T>` is because CsWinRT previously didn't have a way to easily get the ABI for activation factories, so this was done instead. This has been fixed with the `ActivationFactory` class in the AOT branch though.

Marv51:
Hey this sounds amazing! I would be so happy for this to be properly supported. 
I looked at the class and it looks like this would fix my issues: 
#970, microsoft/microsoft-ui-xaml#4886, microsoft/CsWin32#246 and microsoft/win32metadata#867
(I really tried to make this happen in 2021...)

src/cswinrt/strings/ComInteropHelpers.cs
manodasanW: IntPtr is also what we use for other interop interfaces here.

src/Tests/UnitTest/ComInteropTests.cs
```diff
@@ -161,5 +162,12 @@ public void TestDisplayInformation()
             Assert.Throws<COMException>(() => DisplayInformationInterop.GetForWindow(new IntPtr(0)));
             Assert.Throws<COMException>(() => DisplayInformationInterop.GetForMonitor(new IntPtr(0)));
         }
+
+        [Fact]
+        public void TestDataTransferManager()
+        {
+            Assert.Throws<COMException>(() => DataTransferManagerInterop.GetForWindow(new IntPtr(0)));
```
manodasanW, line 169: This test is currently failing.  It could be this API is silently handling invalid hwnds whereas other similar APIs don't.  

  Failed UnitTest.ComInteropTests.TestDataTransferManager [8 ms]
  Error Message:
   Assert.Throws() Failure
Expected: typeof(System.Runtime.InteropServices.COMException)
Actual:   (No exception was thrown)
  Stack Trace:
     at UnitTest.ComInteropTests.TestDataTransferManager() in D:\a\_work\1\s\src\Tests\UnitTest\ComInteropTests.cs:line 169
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)

manodasanW: I think we try removing the assert and check if the return is not null.  If it isn't we go with that similar to `TestUserConsentVerifier`.  Otherwise, we end up asserting it is null.

Gaoyifei1011: > I think we try removing the assert and check if the return is not null. If it isn't we go with that similar to `TestUserConsentVerifier`. Otherwise, we end up asserting it is null.

However, when I use the ComImport active import IDataTransferManagerInterop and sets the window handle to IntPtr. Zero, take the initiative to call the test method, it triggers the COMException. I am also curious about what happened to it, did the code I submitted go wrong again?
然而，当我使用 ComImport 主动导入IDataTransferManagerInterop 并将窗口句柄设置为 IntPtr.Zero，主动调用测试的方法时，它触发了 COMException。我也很好奇它究竟发生了什么，是我提交的代码再一次发生了错误吗？

image

manodasanW: That is a good point, something must be going wrong.  Let me take another look.

manodasanW: So the issue probably is right now when CsWinRT projects that interface, it assumes it is implementing IInspectable.  But in reality, it implements IUnknown.  So, I think we are ending up calling the wrong function.

Gaoyifei1011: > So the issue probably is right now when CsWinRT projects that interface, it assumes it is implementing IInspectable. But in reality, it implements IUnknown. So, I think we are ending up calling the wrong function.

Well, that's really unexpected. So I would like to ask you, is there any good solution?

manodasanW: In theory we can try to make CsWinRT aware of this scenario by trying to add an attribute similar to `ProjectionInternal` but for this scenario and have CsWinRT project the right thing based on that.  But that can get complicated and we only need it here.  So, I think we follow the approach we did with `WindowNative`/`IWindowNativeMethods` and `InitializeWithWindow`/`IInitializeWithWindowMethods`.  Both of them are interfaces that basically only implement `IUnknown` and are similar interop interfaces defined in this file.  Essentially, we will handroll the projection for these 2 methods rather than adding the interface to the idl file for it to be autogenerated.  You can probably just copy the generated `IDisplayInformationStaticsInteropMethods` static class with the current code and adjust the function pointer index on it to reflect it only implements `IUnknown`.

manodasanW: And to handle the difference where we have a static class here which the other 2 scenarios don't, I think we got a couple options.  We either keep the idl declaration for the purposes of just doing the `As` call and then to get the ThisPtr for it.  Or we just do our own `ActivationFactory.Get("Windows.ApplicationModel.DataTransfer.DataTransferManager", IID)`  call and use that `IObjectReference`.  This is probably acceptable given we avoid the IDIC cast in the other scenario and in both cases we have our own cached version of the interface.

Gaoyifei1011: > In theory we can try to make CsWinRT aware of this scenario by trying to add an attribute similar to but for this scenario and have CsWinRT project the right thing based on that. But that can get complicated and we only need it here. So, I think we follow the approach we did with / and /. Both of them are interfaces that basically only implement and are similar interop interfaces defined in this file. Essentially, we will handroll the projection for these 2 methods rather than adding the interface to the idl file for it to be autogenerated. You can probably just copy the generated static class with the current code and adjust the function pointer index on it to reflect it only implements .`ProjectionInternal``WindowNative``IWindowNativeMethods``InitializeWithWindow``IInitializeWithWindowMethods``IUnknown``IDisplayInformationStaticsInteropMethods``IUnknown`

image
image
Thank you for your help. I found out that in the IWindowNative and IInitializeWithWindow interfaces the content is an object instance object, but in the IDataTransferMangerInterop it is passing in the DataTransferManager to the interface, and I don't know how to do this
感谢您的帮助。我发现在IWindowNative 和 IInitializeWithWindow 接口传入的内容是一个 object 实例对象，然而在 IDataTransferMangerInterop 中，它传入的是DataTransferManager 转换成接口，我不太清楚这样该怎么操作

manodasanW: Right, that is where my other comment comes into play.  I think we replace the current static variable for `dataTransferManagerInterop` with a static variable for `ActivationFactory.Get("Windows.ApplicationModel.DataTransfer.DataTransferManager", IID_IDataTransferManagerInterop)`.  This would give us an `IObjectReference` for the interop interface.  Then we can use the ThisPtr from the `IObjectReference` for the function call rather than the `ThisPtr` from marshaling the object in the `WindowHandle` scenario.

Gaoyifei1011: > Right, that is where my other comment comes into play. I think we replace the current static variable for `dataTransferManagerInterop` with a static variable for `ActivationFactory.Get("Windows.ApplicationModel.DataTransfer.DataTransferManager", IID_IDataTransferManagerInterop)`. This would give us an `IObjectReference` for the interop interface. Then we can use the ThisPtr from the `IObjectReference` for the function call rather than the `ThisPtr` from marshaling the object in the `WindowHandle` scenario.

Thank you very much, I have successfully called it out, and has successfully displayed the share window, the code needs to be sorted out and submitted later.
非常感谢，我已经成功调用出来了，并已成功显示出分享的窗口，代码需要整理一下，稍后提交。

Gaoyifei1011: > Right, that is where my other comment comes into play. I think we replace the current static variable for `dataTransferManagerInterop` with a static variable for `ActivationFactory.Get("Windows.ApplicationModel.DataTransfer.DataTransferManager", IID_IDataTransferManagerInterop)`. This would give us an `IObjectReference` for the interop interface. Then we can use the ThisPtr from the `IObjectReference` for the function call rather than the `ThisPtr` from marshaling the object in the `WindowHandle` scenario.

I have a question for you. IInitializeWithWindowMethods, IWindowPrivateMethods both static class is on WinRT.Interop namespace, So IDataTransferManagerInteropMethods is in ABI.WinRT.Interop namespace, or WinRT.Interop namespace?
有一个问题想请教您一下。IInitializeWithWindowMethods，IWindowPrivateMethods这两个静态类是放在WinRT.Interop命名空间下的，那么IDataTransferManagerInteropMethods是放到ABI.WinRT.Interop命名空间下，还是 WinRT.Interop 命名空间下？

Gaoyifei1011: > Right, that is where my other comment comes into play. I think we replace the current static variable for `dataTransferManagerInterop` with a static variable for `ActivationFactory.Get("Windows.ApplicationModel.DataTransfer.DataTransferManager", IID_IDataTransferManagerInterop)`. This would give us an `IObjectReference` for the interop interface. Then we can use the ThisPtr from the `IObjectReference` for the function call rather than the `ThisPtr` from marshaling the object in the `WindowHandle` scenario.

I have a question for you. IInitializeWithWindowMethods, IWindowPrivateMethods both static class is on WinRT.Interop namespace, So IDataTransferManagerInteropMethods is in ABI.WinRT.Interop namespace, or WinRT.Interop namespace?
有一个问题想请教您一下。IInitializeWithWindowMethods，IWindowPrivateMethods这两个静态类是放在WinRT.Interop命名空间下的，那么IDataTransferManagerInteropMethods是放到ABI.WinRT.Interop命名空间下，还是 WinRT.Interop 命名空间下？

manodasanW: It looks like none of the static methods class in this file is under the ABI namespace.  This is probably because they are all marked internal, so it is not on the public API surface.  I would just put this internal static methods class in the `Windows.ApplicationModel.DataTransfer` namespace.

src/cswinrt/strings/ComInteropHelpers.cs
```diff
@@ -485,3 +485,69 @@ public static DisplayInformation GetForMonitor(IntPtr monitor)
...
+#else
+    public
+#endif
+    static class DataTransferManagerInterop
+    {
+        private static readonly global::System.Guid IDataTransferManagerInterop_IID = new global::System.Guid("3A3DCD6C-3EAB-43DC-BCDE-45671CE800C8");
+
+        private static readonly global::System.Guid riid = new global::System.Guid("A5CAEE9B-8708-49D1-8D36-67D25A8DA00C");
+
+        private static global::WinRT.FactoryObjectReference<global::WinRT.IObjectReference> factoryObjectReference;
```
manodasanW, line 506: CI error: D:\a\_work\1\s\src\Projections\Windows\Generated Files\net8.0\CsWinRT\ComInteropHelpers.cs(515,42): error CS0234: The type or namespace name 'FactoryObjectReference<>' does not exist in the namespace 'WinRT'

I believe `FactoryObjectReference` is only in main, it is no longer used in the `staging/AOT` branch.  The return type for `global::WinRT.ActivationFactory.Get` is just an IObjectReference on `#if NET`.  You will probably also need to handle the else case for '#if NET' as the function signature is slightly different as it takes an generic which here would be `IUnknownVftbl`.

See [here](https://github.com/microsoft/CsWinRT/blob/9486d45237bbbfa2a3aadb2341dd0118164540dd/src/WinRT.Runtime/ActivationFactory.cs#L233-L237).

src/cswinrt/strings/ComInteropHelpers.cs
```diff
@@ -485,3 +485,69 @@ public static DisplayInformation GetForMonitor(IntPtr monitor)
...
+        }
+    }
+#endif
+
+    internal static class IIDataTransferManagerInteropMethods
+    {
+        internal static unsafe DataTransferManager GetForWindow(global::WinRT.IObjectReference _obj, global::System.IntPtr appWindow, in global::System.Guid riid)
+        {
+            global::System.IntPtr thisPtr = _obj.ThisPtr;
+            global::System.IntPtr ptr = new global::System.IntPtr();
```
manodasanW, line 531:
```suggestion
            global::System.IntPtr ptr = default;

```

src/cswinrt/strings/ComInteropHelpers.cs
```diff
@@ -485,3 +485,69 @@ public static DisplayInformation GetForMonitor(IntPtr monitor)
...
+
+            try
+            {
+                // IDataTransferManagerInterop inherits IUnknown (3 functions) and provides GetForWindow giving a total of 5 functions
+                global::WinRT.ExceptionHelpers.ThrowExceptionForHR((*(delegate* unmanaged[Stdcall]<global::System.IntPtr, global::System.IntPtr, global::System.Guid, out global::System.IntPtr, int>**)thisPtr)[3](thisPtr, appWindow, riid, out ptr));
```
manodasanW, line 536:
```suggestion
                global::WinRT.ExceptionHelpers.ThrowExceptionForHR((*(delegate* unmanaged[Stdcall]<global::System.IntPtr, global::System.IntPtr, global::System.Guid, global::System.IntPtr*, int>**)thisPtr)[3](thisPtr, appWindow, riid, &ptr));

```

manodasanW: This addresses 2 things.  The first is it removes `out` to make the function pointer delegate call blittable.  We fixed most of them but apparently we missed one above.  Will look at fixing at that in another PR, but we should make sure this is blittable.  This other is the GUID parameter is `REFIID` which is a pointer to a GUID as is in the other generated ones which pass the IID as a parameter.

src/cswinrt/strings/ComInteropHelpers.cs
```diff
@@ -485,3 +485,69 @@ public static DisplayInformation GetForMonitor(IntPtr monitor)
...
+#endif
+    static class DataTransferManagerInterop
+    {
+        private static readonly global::System.Guid IDataTransferManagerInterop_IID = new global::System.Guid("3A3DCD6C-3EAB-43DC-BCDE-45671CE800C8");
+
+        private static readonly global::System.Guid riid = new global::System.Guid("A5CAEE9B-8708-49D1-8D36-67D25A8DA00C");
+
+        private static global::WinRT.FactoryObjectReference<global::WinRT.IObjectReference> factoryObjectReference;
+
+        static DataTransferManagerInterop()
```
manodasanW, line 508: if possible, I would just do an inline variable declaration rather than in a static constructor.

src/cswinrt/strings/ComInteropHelpers.cs
```diff
@@ -485,3 +485,69 @@ public static DisplayInformation GetForMonitor(IntPtr monitor)
...
+    [global::System.Runtime.Versioning.SupportedOSPlatform("Windows10.0.10240.0")]
+#endif
+#if EMBED
+    internal
+#else
+    public
+#endif
+    static class DataTransferManagerInterop
+    {
+        private static readonly global::System.Guid IDataTransferManagerInterop_IID = new global::System.Guid("3A3DCD6C-3EAB-43DC-BCDE-45671CE800C8");
```
manodasanW, line 502: For both these guids, we probably want to avoid GUID string parsing.  I would replace it to the least with the one that take the hex values like [here](https://github.com/microsoft/CsWinRT/blob/9486d45237bbbfa2a3aadb2341dd0118164540dd/src/cswinrt/strings/ComInteropHelpers.cs#L91).

Gaoyifei1011: > For both these guids, we probably want to avoid GUID string parsing. I would replace it to the least with the one that take the hex values like [here](https://github.com/microsoft/CsWinRT/blob/9486d45237bbbfa2a3aadb2341dd0118164540dd/src/cswinrt/strings/ComInteropHelpers.cs#L91).

Thank you for your suggestions in the code. I have adjusted the code according to your comments

src/cswinrt/strings/ComInteropHelpers.cs
```diff
@@ -499,33 +499,35 @@ namespace Windows.ApplicationModel.DataTransfer
 #endif
     static class DataTransferManagerInterop
     {
-        private static readonly global::System.Guid IDataTransferManagerInterop_IID = new global::System.Guid("3A3DCD6C-3EAB-43DC-BCDE-45671CE800C8");
+        private static readonly global::System.Guid IDataTransferManagerInterop_IID = new global::System.Guid(0x3A3DCD6C, 0x3EAB, 0x43DC, 0xBC, 0xDE, 0x45, 0x67, 0x1C, 0xE8, 0x00, 0xC8);
```
dongle-the-gadget, line 502: This is wrong endianess.

manodasanW: Correct me if I am wrong, but this uses the format where the values goes as parameters and not in the byte array where we need to flip it for endianness, so I think this is correct right?

https://learn.microsoft.com/en-us/dotnet/api/system.guid.-ctor?view=net-8.0#system-guid-ctor(system-int32-system-int16-system-int16-system-byte-system-byte-system-byte-system-byte-system-byte-system-byte-system-byte-system-byte)

src/cswinrt/strings/ComInteropHelpers.cs
```diff
@@ -485,3 +485,77 @@ public static DisplayInformation GetForMonitor(IntPtr monitor)
...
+    public
+#endif
+    static class DataTransferManagerInterop
+    {
+        private static readonly global::System.Guid IDataTransferManagerInterop_IID = new global::System.Guid(0x3A3DCD6C, 0x3EAB, 0x43DC, 0xBC, 0xDE, 0x45, 0x67, 0x1C, 0xE8, 0x00, 0xC8);
+
+        private static readonly global::System.Guid riid = new global::System.Guid(0xA5CAEE9B, 0x8708, 0x49D1, 0x8D, 0x36, 0x67, 0xD2, 0x5A, 0x8D, 0xA0, 0x0C);
+
+#if NET
+        private static global::WinRT.IObjectReference objectReference => global::WinRT.ActivationFactory.Get<global::WinRT.IObjectReference>("Windows.ApplicationModel.DataTransfer.DataTransferManager", IDataTransferManagerInterop_IID);
```
manodasanW, line 507: CI error D:\a\_work\1\s\src\Projections\Windows\Generated Files\net6.0\CsWinRT\ComInteropHelpers.cs(516,110): error CS0308: The non-generic method 'ActivationFactory.Get(string)' cannot be used with type arguments

```suggestion
        private static global::WinRT.IObjectReference objectReference => global::WinRT.ActivationFactory.Get("Windows.ApplicationModel.DataTransfer.DataTransferManager", IDataTransferManagerInterop_IID);
```

manodasanW, line 507: Can we also just do `=` here to cache rather than making it a property call that calls the function each time?

src/cswinrt/strings/ComInteropHelpers.cs
```diff
@@ -485,3 +485,77 @@ public static DisplayInformation GetForMonitor(IntPtr monitor)
...
+        {
+            IDataTransferManagerInteropMethods.ShowShowShareUIForWindow(objectReference, appWindow);
+        }
+    }
+#endif
+
+    internal static class IDataTransferManagerInteropMethods
+    {
+#if NET
+        internal static unsafe global::Windows.ApplicationModel.DataTransfer.DataTransferManager GetForWindow(global::WinRT.IObjectReference _obj, global::System.IntPtr appWindow, in global::System.Guid riid)
```
manodasanW, line 527: `ObjectReference` implements `IObjectReference` so we don't actually need the else case here or below.

manodasanW:
I think once the last set of feedback is addressed, we should be good with this change.

src/cswinrt/strings/ComInteropHelpers.cs
Gaoyifei1011: image
If you do this, then I can add IDataTransferManagerInterop_IID directly to the parameter
如果这样做，那我就直接把 IDataTransferManagerInterop_IID 添加到参数中

src/Tests/UnitTest/ComInteropTests.cs
```diff
@@ -161,5 +162,12 @@ public void TestDisplayInformation()
             Assert.Throws<COMException>(() => DisplayInformationInterop.GetForWindow(new IntPtr(0)));
             Assert.Throws<COMException>(() => DisplayInformationInterop.GetForMonitor(new IntPtr(0)));
         }
+
+        [Fact]
+        public void TestDataTransferManager()
+        {
+            Assert.Throws<COMException>(() => DataTransferManagerInterop.GetForWindow(new IntPtr(0)));
+            DataTransferManagerInterop.ShowShareUIForWindow(new IntPtr(0));
```
manodasanW, line 170: We should Assert.Throws<COMException> here.  Latest CI run showed the below exception here which is expected and also shows the function is being called:

System.Runtime.InteropServices.COMException : Invalid window handle. (0x80070578)

Gaoyifei1011: > 我们应该在这里 Assert.Throws 类似于上一行。最新的 CI 运行在此处显示以下异常，这是预期的，并且还显示正在调用函数：
> 
> System.Runtime.InteropServices.COMException：无效的窗口句柄。（0x80070578）

image
I tried to write this test method with reference to IPlayToManagerInterop because the two interfaces have similar function call names and parameters. As you said just now, if in DataTransferManagerInterop.ShowShareUIForWindow add test statements, does that also should add a new PR, In PlyaToManagerInterop.ShowPlayToUIForWindow() to add similar testing statement?

Gaoyifei1011: > We should Assert.Throws here similar to the previous line. Latest CI run showed the below exception here which is expected and also shows the function is being called:
> 
> System.Runtime.InteropServices.COMException : Invalid window handle. (0x80070578)

image
I tried to write this test method with reference to IPlayToManagerInterop because the two interfaces have similar function call names and parameters. As you said just now, if in DataTransferManagerInterop.ShowShareUIForWindow add test statements, does that also should add a new PR, In PlyaToManagerInterop.ShowPlayToUIForWindow() to add similar testing statement?

manodasanW: That other test isn't failing, so I would leave it as is.  It is possible that API handles invalid handle without erroring whereas this one doesn't.

Gaoyifei1011: > That other test isn't failing, so I would leave it as is. It is possible that API handles invalid handle without erroring whereas this one doesn't.

Well. Thank you for your answer.

manodasanW (approved):

manodasanW:
Thanks for this change, will merge once CI is done.

manodasanW:
CI finished and all tests passed.

PR #1530, opened 2024-03-09
Fix issue where RuntimeConfiguration defaults are not used

manodasanW:
In certain scenarios, the runtime configuration file may not be there.  When this happens, the defaults we set in msbuild do not take effect and we can end up defaulting to false.  This means all switches are disabled and can cause issues like custom type mappings not being enabled.  This addresses it by making the defaults in code so we can default to the correct value.


Sergio0694 (approved):
LGTM! The linker will replace the whole method body so this is fine with that too!

PR #1532, opened 2024-03-09
Add ActivationHandler to match the C++/WinRT winrt_activation_handler.

### LINKED ISSUE #1136: CsWinRT should support activation hooks to match C++/WinRT.
## Summary
C++/WinRT provides ```winrt_activation_handler ``` to allow hooking activation. This is useful for testing.

## Rationale
* Useful for testing.
* Matches C++/WinRT functionality.

jlaanstra:
Fixes #1136

### DIFF
--- a/src/Tests/UnitTest/TestComponentCSharp_Tests.cs
+++ b/src/Tests/UnitTest/TestComponentCSharp_Tests.cs
@@ -3193,6 +3193,8 @@ namespace UnitTest
         }
 
         [Guid("59C7966B-AE52-5283-AD7F-A1B9E9678ADD")]
+        [global::WinRT.WindowsRuntimeType("Windows.Foundation.UniversalApiContract")]
+        [global::WinRT.WindowsRuntimeHelperType(typeof(ICustomGuidHelperStatics))]
         interface ICustomGuidHelperStatics
         {
             public static readonly IntPtr AbiToProjectionVftablePtr;
@@ -3215,7 +3217,7 @@ namespace UnitTest
 
                 try
                 {
-                    __result = global::WinRT.ComWrappersSupport.FindObject<global::Windows.Foundation.ICustomGuidHelperStatics>(thisPtr).CreateNewGuid();
+                    __result = global::WinRT.ComWrappersSupport.FindObject<ICustomGuidHelperStatics>(thisPtr).CreateNewGuid();
                     *result = __result;
 
                 }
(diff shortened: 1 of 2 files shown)


src/WinRT.Runtime/ActivationFactory.cs
```diff
@@ -197,8 +197,27 @@ public static unsafe (ObjectReference<I> obj, int hr) GetActivationFactory<I>(st
 #endif
     static class ActivationFactory
     {
+#if NET
+        /// <summary>
+        /// This provides a hook into activation to hook/mock activation of WinRT types.
+        /// </summary>
+        public static Func<string, Guid, IntPtr> ActivationHandler { get; set; }
```
manodasanW, line 204: Looks good, is there an easy test case we can add for this just to make sure we don't regress it and that it works.

src/WinRT.Runtime/ActivationFactory.cs
```diff
@@ -197,8 +197,27 @@ public static unsafe (ObjectReference<I> obj, int hr) GetActivationFactory<I>(st
 #endif
     static class ActivationFactory
     {
+#if NET
```
manodasanW, line 200: And to confirm, all of this is guarded for NET to not worry about the vftable types in the .NET standard scenario?

jlaanstra: Yes and because tests almost always target a specific framework anyway.

src/WinRT.Runtime/ActivationFactory.cs
jlaanstra: Will add a test.

src/Tests/UnitTest/TestComponentCSharp_Tests.cs
```diff
@@ -3191,6 +3191,42 @@ public void TestActivationFactoriesFromMultipleContexts()
 
             Assert.Null(exception);
         }
+
+        class CustomGuidHelper : IGuidHelperStatics
```
manodasanW, line 3195: How were we able to implement this interface, it looks to be projected as an internal interface?

src/cswinrt/code_writers.h
```diff
@@ -7755,25 +7761,18 @@ NativeMemory.Free((void*)abiToProjectionVftablePtr);
...
-                for_typedef(w, get_type_semantics(iface.Interface()), [&](auto interface_type)
-                {
-                    if (type == interface_type && is_overridable(iface))
-                    {
-                        hasOverridableAttribute = true;
-                    }
-                });
-
-                if (hasOverridableAttribute)
+                if ((attributed_type.activatable || attributed_type.statics || attributed_type.overridable) && attributed_type.type == type)
```
manodasanW, line 7768: Do we want to start generating the vftbl for these scenarios?  In most cases other than I guess for this test case, they are not used.  I wonder if there is another way to handle scenarios like the test case.

src/Tests/UnitTest/TestComponentCSharp_Tests.cs
jlaanstra: Hmm that's a good point actually. It compiled as I ran the test locally. Let me see why.

src/cswinrt/code_writers.h
jlaanstra: I think it makes sense if the interface is public.

manodasanW (approved):

PR #1534, opened 2024-03-10
Add AsInterface<T> support for generated COM interfaces.

### LINKED ISSUE #1533: Add support for generated COM interfaces
# Proposal: Add support for generated COM interfaces

## Summary
.NET 8 introduces source generated COM wrappers, which provides a friendly, AOT-safe way to use COM in .NET. Currently for C#/WinRT users, the only way to query into custom COM types is to use the built-in COM interop, which is not AOT-safe at all.

## Rationale
* Better AOT support for custom COM types.

## Important Notes
Sample usage code, using `As<T>`:
```cs
[GeneratedComInterface]
[Guid("3faca0d2-e7f1-4e9c-82a6-404fd6e0aab8")]
internal partial interface IFooNative
{
    void Method(int i);
}

Foo fooObject = new Foo(); // Foo is a WinRT class, in this case.
IFooNative nativeObject = fooObject.As<IFooNative>(); // Casting the Foo object to an "IFooNative" object.
nativeObject.Method(0); // Calling a Method exposed in IFooNative
```
## Open Questions
There are several strategies that I could think of to implement this:
- Reuse `StrategyBasedComWrappers` from the runtime.
- Remodel our existing COM Wrappers implementation.
- Create or remodel an existing `IDynamicInterfaceCastable` implementation, considering that we don't actually need to handle CCWs here 🤔?

dongle-the-gadget:
Closes #1533.

> [!NOTE]
> I would recommend reviewing this with whitespaces disabled.


src/WinRT.Runtime/IWinRTObject.net5.cs
```diff
@@ -1,188 +1,270 @@
...
-}
\ No newline at end of file
+#if EMBED
+    internal
+#else
+    public
+#endif
+    interface IWinRTObject : IDynamicInterfaceCastable
+#if NET8_0_OR_GREATER
+        , IUnmanagedVirtualMethodTableProvider
```
manodasanW, line 23: I will need to check if adding an interface here is considered an API compat breaking change or if it is fine if we also implement it.

jkoritzinsky: Theoretically, there could be a diamond problem if two interfaces were to both implement `IWinRTObject` and then a class were to implement both of those interfaces. Other than that case (which shouldn't happen given the expectation of where this interface is implemented), this should be fine.

Alternatively, `IWinRTObject`'s `IDynamicInterfaceCastable` implementation could also provide an implementation for this interface lazily.

src/WinRT.Runtime/IWinRTObject.net5.cs
```diff
@@ -1,188 +1,270 @@
...
+                    ManagedType = details.Implementation.TypeHandle
+                };
+
+                if (!AdditionalTypeData.TryAdd(interfaceType, result))
+                {
+                    bool found = AdditionalTypeData.TryGetValue(interfaceType, out object newInfo);
+                    System.Diagnostics.Debug.Assert(found);
+                    result = (IIUnknownCacheStrategy.TableInfo)newInfo;
+                    StrategyBasedComWrappers.DefaultIUnknownStrategy.Release(ppv);
+                }
```
jkoritzinsky, line 209: Instead of using the `DefaultIUnknownStrategy` here, you should go down the same path as the rest of the CsWinRT interface types to handle the QI if possible (so as to handle COM marshalling, agile objects, etc). The only default strategy you need to use here is the interface details strategy.

dongle-the-gadget: Should be fixed now, as `IObjectReference.TryAs` is now used.

src/WinRT.Runtime/IWinRTObject.net5.cs
```diff
@@ -1,188 +1,269 @@
...
+            bool vtableLookup = LookupGeneratedVTableInfo(interfaceType, out _, out int qiResult);
+            if (vtableLookup)
+            {
+                return true;
+            }
+            else if (qiResult < 0 && throwIfNotImplemented)
+            {
+                // A qiResult of less than zero means the call to QueryInterface has failed.
+                ExceptionHelpers.ThrowExceptionForHR(qiResult);
+            }
```
jkoritzinsky, line 48: Any reason we can't use the `QueryInterfaceCache` to remove some of this code?

jkoritzinsky (commented):
Can we add a unit test?

src/WinRT.Runtime/IWinRTObject.net5.cs
dongle-the-gadget: I'm not sure if `QueryInterfaceCache` would help us here, since firstly it's run rather early in the code, and secondly even if I were to add the QI result to `QueryInterfaceCache`, the lack of the interface in `QueryInterfaceCache` wouldn't necessarily mean the QI failed (for instance, if the interface isn't actually a generated COM interface).

manodasanW:
> > Can we add a unit test?
> 
> Does the CI run the unit tests in .NET 8 or only in .NET 6? The `Directory.Build.props` file specifies that the unit tests are only up to .NET 6 locally but .NET 8 on CI?

CI runs all 3 TFMs listed (netcoreapp3.1, net6.0, net8.0).  For local runs we drop net8 just to save time, but you can manually add it to `AppBuildTFMs` for your testing purposes.

dongle-the-gadget:
I've copied the COM Import unit test and modified it to use generated COM interfaces instead, though I'm not too certain about the signatures of both.

src/Tests/UnitTest/ComGenerationTests.cs
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    public class ComGenerationTests
+    {
+        private static readonly Guid IID_IComInterop = new Guid("15651B9F-6C6B-4CC0-944C-C7D7B0F36F81");
+
+        [Fact]
+        public void TestHWND()
+        {
+            var comInterop = Class.ComInterop.As<IComInteropGenerated>();
```
jkoritzinsky, line 38: With the implementation you've done, this should be castable with just a regular cast. Something like the following instead of using `.As<T>()`.

```suggestion
            var comInterop = (IComInteropGenerated)(object)Class.ComInterop;
```

jkoritzinsky (commented):
One update to the tests, the product code looks good to me (pending merge conflict resolution)

src/Tests/UnitTest/ComGenerationTests.cs
dongle-the-gadget: I’m using `As<T>` here in order to match the pattern used by the built-in COM interop, but I suppose that should also work too.

manodasanW:
Looks like the following 3 authoring tests regressed in non AOT scenarios.

[  FAILED  ] 3 tests, listed below:
[  FAILED  ] AuthoringTest.Statics
[  FAILED  ] AuthoringTest.Events
[  FAILED  ] AuthoringTest.ExplicitInterfaces

dongle-the-gadget:
Oh, I don't test against authoring as building the WinRT projection for the Windows namespace is very time consuming on my machine.

However though, I'm not sure how Authoring utilizes `IWinRTObject`, but if it turns out to rely on the signature of the interface I guess we could implement `IUnmanagedVirtualMethodTableProvider` lazily through `IDynamicInterfaceCastable`, like Jeremy suggested.

manodasanW:
> Oh, I don't test against authoring as building the WinRT projection for the Windows namespace is very time consuming on my machine.
> 
> However though, I'm not sure how Authoring utilizes `IWinRTObject`, but if it turns out to rely on the signature of the interface I guess we could implement `IUnmanagedVirtualMethodTableProvider` lazily through `IDynamicInterfaceCastable`, like Jeremy suggested.

I will see if I can get the exception stacks when I get a chance and get back to you.  That should tell us what is failing.

manodasanW:
The exception was `Method 'GetVirtualMethodTableInfoForKey' in type 'NativeDelegateWrapper' from assembly 'AuthoringTest, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null' does not have an implementation.`  But looking more into it, it looks like the AuthoringTest was built for .NET 8 with the .NET 8 WinRT.Runtime, but the native app that consumes it bundled a .NET 6 version of WinRT.Runtime.  I think this is a test authoring error rather than a real scenario.  So, I tried a change to fix the test and it seems to work.  I think you should add that change to your PR and that should address the test failures.

Change: https://github.com/microsoft/CsWinRT/commit/fe2813313a4bfb70b7559200d955e264d6e36e77

dongle-the-gadget:
I’ve cherry picked the change in, thanks.

src/WinRT.Runtime/IWinRTObject.net5.cs
```diff
@@ -1,217 +1,298 @@
...
+                {
+                    result = tableInfo;
+                    return true;
+                }
+                return false;
+            }
+
+            if (StrategyBasedComWrappers.DefaultIUnknownInterfaceDetailsStrategy.GetIUnknownDerivedDetails(interfaceType) is IIUnknownDerivedDetails details)
+            {
+                qiResult = NativeObject.TryAs(details.Iid, out ObjectReference<IUnknownVftbl> objRef);
```
manodasanW, line 220: I was taking one last look at this before signing off and I noticed there is nothing that keeps the `objref` alive.  Do we want to add it to `QueryInterfaceCache` so that it keeps it alive?  The TableInfo just seems to point to `ThisPtr` which is an `IntPtr`.

src/WinRT.Runtime/IWinRTObject.net5.cs
```diff
@@ -1,217 +1,298 @@
...
+
+#if NET8_0_OR_GREATER
+        unsafe VirtualMethodTableInfo IUnmanagedVirtualMethodTableProvider.GetVirtualMethodTableInfoForKey(Type type)
+        {
+            if (!LookupGeneratedVTableInfo(type.TypeHandle, out IIUnknownCacheStrategy.TableInfo? result, out int qiHResult))
+            {
+                Marshal.ThrowExceptionForHR(qiHResult);
+            }
+
+            return new((void*)NativeObject.ThisPtr, result.Value.Table);
```
manodasanW, line 268: Should this be the ptr from the table / cache or is the `Ptr` that may not be QIed for the interface already fine?

manodasanW (approved):

manodasanW:
Thanks for addressing all the feedback, will merge once CI passes.

Issue #824, opened 2021-04-26
 Question: Is there any planned support for Memory<> and Span<>?


k62o:
Hi all,

I'm creating an API with IDL 3 that has the following signature (or a similar one):
`void Process(UInt8[] value);`

With C++/WinRT (2.0.210403.2) I have a nice `array_view` API:
`void Process(array_view<uint8_t> value);`
This allows to support many data types with no overhead.

With C#/WinRT (1.2.5) I have an `Array` API:
`void Process(byte[] value);`
This version does not allow us to pass `Memory<byte>` or `Span<byte>` that we use in other parts of our application. We may need to create a somewhat useless temporary array.

Do you have plan for `Memory<T>` and `Span<T>` support instead of `T[]`?

Regards,
Guillaume

teo-tsirpanis:
At first there could be APIs to interoperate between .NET's spans and memories and WinRT's `IBuffer` like the following:

``` diff
namespace System.Runtime.InteropServices.WindowsRuntime;

public static class WindowsRuntimeBufferExtensions
{
+   public static Memory<byte> AsMemory(this IBuffer source);

+   public static Memory<byte> AsMemory(this IBuffer source, uint offset);

+   public static Memory<byte> AsMemory(this IBuffer source, uint offset, int length);

+   public static IBuffer AsWindowsRuntimeBuffer(this Memory<byte> source);

// Gets the underlying IBuffer if the given memory is backed by one.
+   public static bool TryGetWindowsRuntimeBuffer(ReadOnlyMemory<byte> memory, out IBuffer? buffer, out uint offset, out int length);

+   public static void CopyTo(this IBuffer source, Span<byte> destination);

+   public static void CopyTo(this IBuffer source, uint sourceIndex, Span<byte> destination);

+   public static void CopyTo(this ReadOnlySpan<byte> source, IBuffer destination);

+   public static void CopyTo(this ReadOnlySpan<byte> source, IBuffer destination, uint destinationIndex);
}
```

A next step would be adding `Memory<byte>` overloads to WinRT APIs that accept `IBuffer`s. If there is a way to mark the `IBuffer` parameters that their content will not be changed they could be `ReadOnlyMemory`ies instead.

Passing spans instead of `IBuffer`s is unfortunately impossible; the latter are objects in the heap and the former are ref structs and not allowed to be stored there.

PR #1535, opened 2024-03-12
Add support for IBuffer.AsMemory() and Memory.AsBuffer().

jlaanstra:
Contributes to #824

WindowsRuntimeBuffer is now internally represented by a ```Memory<byte>``` instead of a ```byte[]```. Also simplified a few codepaths to use the CopyTo functionality from Span<T>.

```AsMemory()``` follows the approach taken by ```ToArray()``` in that the capacity of the ```IBuffer``` is ignored and the Memory represents the length of the buffer.

### DIFF
--- a/src/cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBufferExtensions.cs
+++ b/src/cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBufferExtensions.cs
@@ -55,7 +55,7 @@ namespace System.Runtime.InteropServices.WindowsRuntime
             if (source.Length - offset < capacity) throw new ArgumentException(global::Windows.Storage.Streams.SR.Argument_InsufficientArrayElementsAfterOffset);
             if (capacity < length) throw new ArgumentException(global::Windows.Storage.Streams.SR.Argument_InsufficientBufferCapacity);
 
-            return new WindowsRuntimeBuffer(source.AsMemory().Slice(offset, capacity - offset), length);
+            return new WindowsRuntimeBuffer(source.AsMemory().Slice(offset, capacity), length);
         }
 
 #endregion (Byte []).AsBuffer extensions
@@ -115,13 +115,13 @@ namespace System.Runtime.InteropServices.WindowsRuntime
             Debug.Assert(destinationIndex <= int.MaxValue);
 
             // If destination is backed by a managed memory, use the memory instead of the pointer as it does not require pinning:
-            Span<byte> destSpan = destination.TryGetUnderlyingData(out Memory<byte> destDataArr) ? destDataArr.Span.Slice(destinationIndex) : destination.GetSpanForCapacity(destinationIndex);
+            Span<byte> destSpan = destination.TryGetUnderlyingData(out Memory<byte> destDataArr) ? destDataArr.Span.Slice((int)destinationIndex) : destination.GetSpanForCapacity(destinationIndex);
             source.CopyTo(destSpan);
 
             // Update Length last to make sure the data is valid
             if (destinationIndex + source.Length > destination.Length)
             {
-                destination.Length = destinationIndex + source.Length;
+                destination.Length = destinationIndex + (uint)source.Length;
             }
         }
 
@@ -250,7 +250,7 @@ namespace System.Runtime.InteropServices.WindowsRuntime
             if (source == null) throw new ArgumentNullException(nameof(source));
             if (destination == null) throw new ArgumentNullException(nameof(destination));
 
-            CopyTo(source, sourceIndex, destination.AsSpan(destinationIndex, count));
+            CopyTo(source, sourceIndex, destination.AsSpan(destinationIndex, count), count);
         }
 
 #endregion (IBuffer).CopyTo extensions for copying to a (Byte [])
@@ -282,8 +282,8 @@ namespace System.Runtime.InteropServices.WindowsRuntime
             Debug.Assert(destinationIndex <= int.MaxValue);
 
             // If source are destination are backed by managed arrays, use the arrays instead of the pointers as it does not require pinning:
-            Span<byte> srcSpan = source.TryGetUnderlyingData(out Memory<byte> srcDataArr) ? srcDataArr.Span.Slice((int)sourceIndex, (int)count) : source.GetSpanForCapacity((int)sourceIndex);
-            Span<byte> destSpan = source.TryGetUnderlyingData(out Memory<byte> destDataArr) ? destDataArr.Span.Slice((int)destinationIndex) : destination.GetSpanForCapacity((int)destinationIndex);
+            Span<byte> srcSpan = source.TryGetUnderlyingData(out Memory<byte> srcDataArr) ? srcDataArr.Span.Slice((int)sourceIndex, (int)count) : source.GetSpanForCapacity(sourceIndex);
+            Span<byte> destSpan = destination.TryGetUnderlyingData(out Memory<byte> destDataArr) ? destDataArr.Span.Slice((int)destinationIndex) : destination.GetSpanForCapacity(destinationIndex);
 
             srcSpan.CopyTo(destSpan);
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBufferExtensions.cs [hunk 1]
```diff
@@ -140,6 +156,25 @@ public static byte[] ToArray(this IBuffer source, uint sourceIndex, int count)
 #endregion (IBuffer).ToArray extensions for copying to a new (Byte [])
 
 
+#region (IBuffer).AsMemory extensions for copying to a new (Byte [])
+
+        public static Memory<byte> AsMemory(this IBuffer source)
```
Sergio0694, line 161: This is deceivingly unsafe. The GC could collect the underlying `IBuffer` while someon's using the span.

jlaanstra: I don't follow. If the IBuffer is backed by a Memory<byte> that should be safe. If not, the Memory instance should keep the IBufferMemoryManager alive which is keeping the IBuffer alive?

src/cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBufferExtensions.cs
```diff
@@ -96,17 +115,14 @@ public static void CopyTo(this byte[] source, int sourceIndex, IBuffer destinati
...
-            IntPtr destPtr = destination.GetPointerAtOffset(destinationIndex);
-            Marshal.Copy(source, sourceIndex, destPtr, count);
+            // If destination is backed by a managed memory, use the memory instead of the pointer as it does not require pinning:
+            Memory<byte> destDataArr;
+            if (!destination.TryGetUnderlyingData(out destDataArr))
+            {
+                destDataArr = destination.AsMemory();
+            }
+
+            source.AsMemory(sourceIndex, count).CopyTo(destination.AsMemory().Slice((int)destinationIndex));
```
manodasanW, line 125: destDataArr?

src/cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBufferExtensions.cs
```diff
@@ -468,6 +469,38 @@ internal unsafe WindowsRuntimeBufferUnmanagedMemoryStream(IBuffer sourceBuffer,
...
+
+            public override Span<byte> GetSpan()
+            {
+                IntPtr sourcePtr = _buffer.GetPointerAtOffset(0);
+                return new Span<byte>((byte*)sourcePtr.ToPointer(), (int)_buffer.Capacity);
+            }
+
+            public override MemoryHandle Pin(int elementIndex = 0)
+            {
+                // IBuffer is responsable for pinning so assume pinned
```
manodasanW, line 493:
```suggestion
                // IBuffer is responsible for pinning so assume pinned

```

src/cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBufferExtensions.cs
```diff
@@ -51,11 +53,28 @@ public static IBuffer AsBuffer(this byte[] source, int offset, int length, int c
...
+            return new WindowsRuntimeBuffer(source, source.Length);
+        }
+
+        public static IBuffer AsBuffer(this Memory<byte> source, int offset, int length)
+        {
+            if (offset < 0) throw new ArgumentOutOfRangeException(nameof(offset));
+            if (length < 0) throw new ArgumentOutOfRangeException(nameof(length));
+            if (source.Length - offset < length) throw new ArgumentException(global::Windows.Storage.Streams.SR.Argument_InsufficientArrayElementsAfterOffset);
+
+            return new WindowsRuntimeBuffer(source.Slice(offset), source.Length);
```
manodasanW, line 74: should this be length instead or no?

src/cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBufferExtensions.cs
Sergio0694: Not sure why you resolved this, I didn't get a chance to reply. This is not safe because of this:
```csharp
IBuffer buffer = GetSomeBuffer();
Span<byte> span = buffer.AsMemory().Span;
// <-- GC runs here
DoStuff(span); // Boom 💥
```

Sergio0694 (changes requested):
I don't think we should expose these APIs as is, because they are effectively unsafe, but hide that. Having a custom memory manager wrapping an `IBuffer` is equivalent to having a memory manager with a finalizer (because `IBuffer` will have a finalizer), which is explicitly not recommended specifically because it's unsafe.

We should definitely expose a way to convert between `IBuffer` and `Memory<T>` types, but I'm wondering whether it wouldn't be better to do so via an API that makes the fact they're unsafe explicit, and with proper documentation, so developers can know what to do. For instance, something like this:
```csharp
public static class WindowsRuntimeMarshal
{
    public static Memory<byte> AsMemoryUnsafe(IBuffer buffer);
}
```

Just for context, pasting it here again, this is the scenario I'm worried about:
```csharp
IBuffer buffer = GetSomeBuffer();
Span<byte> span = buffer.AsMemory().Span;
// <-- GC runs here
DoStuff(span); // Boom 💥
```

Also cc. @tannergooding, @GrabYourPitchforks, thoughts? 🙂

src/cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBufferExtensions.cs
jlaanstra: Doesn't that equally apply to string?

```csharp
string buffer = GetSomeString();
Span<char> span = buffer.AsMemory().Span;
// <-- GC runs here
DoStuff(span); // Boom 💥

**Comment by jlaanstra** [comment]
> I don't think we should expose these APIs as is, because they are effectively unsafe, but hide that. Having a custom memory manager wrapping an `IBuffer` is equivalent to having a memory manager with a finalizer (because `IBuffer` will have a finalizer), which is explicitly not recommended specifically because it's unsafe.
> 
> We should definitely expose a way to convert between `IBuffer` and `Memory<T>` types, but I'm wondering whether it wouldn't be better to do so via an API that makes the fact they're unsafe explicit, and with proper documentation, so developers can know what to do. For instance, something like this:
> 
> ```cs
> public static class WindowsRuntimeMarshal
> {
>     public static Memory<byte> AsMemoryUnsafe(IBuffer buffer);
> }
> ```
> 
> Just for context, pasting it here again, this is the scenario I'm worried about:
> 
> ```cs
> IBuffer buffer = GetSomeBuffer();
> Span<byte> span = buffer.AsMemory().Span;
> // <-- GC runs here
> DoStuff(span); // Boom 💥
> ```
> 
> Also cc. @tannergooding, @GrabYourPitchforks, thoughts? 🙂

Doesn't this apply equally to string?

```cs
string buffer = GetSomeString();
Span<char> span = buffer.AsMemory().Span;
// <-- GC runs here
DoStuff(span); // Boom 💥
```

src/cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBufferExtensions.cs
jlaanstra: I resolved as we discussed offline, but maybe there was some confusion around what I meant :). Either way we can continue the conversation outside this comment.

Sergio0694:
> "Doesn't that equally apply to string?"

No, because a `string` is a managed object and the span will keep it alive (as it's an interior pointer to it).

jlaanstra:
Superseded by #1556 1556

PR #1536, opened 2024-03-12
Fix bug in the AsStream extensions where the Length isn't properly updated on the IBuffer.

jlaanstra:
If you write to an IBuffer via the AsStream() method and you increase the length, the length of the underlying IBuffer doesn't update.


jlaanstra:
Probably also want to fix the managed path which has the same issue.

manodasanW (approved):

PR #1537, opened 2024-03-13
Optimize AsTask when the underlying IAsyncOperation is wrapping a task.

jlaanstra:
No need for all the wrapping. This is similar to various IBuffer optimizations.

### DIFF
--- a/src/cswinrt/strings/additions/Windows.Foundation/Windows.Foundation.cs
+++ b/src/cswinrt/strings/additions/Windows.Foundation/Windows.Foundation.cs
@@ -25,12 +25,14 @@ namespace System
                 throw new ArgumentNullException(nameof(source));
             }
 
+#if NET
             if (source is ITaskAwareAsyncInfo asyncInfo && asyncInfo.Task is Task task)
             {
                 return cancellationToken.CanBeCanceled ?
                     task.WaitAsync(cancellationToken) :
                     task;
             }
+#endif
 
             switch (source.Status)
             {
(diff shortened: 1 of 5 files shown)


src/cswinrt/strings/additions/Windows.Foundation/Windows.Foundation.cs
```diff
@@ -25,8 +25,10 @@ public static Task AsTask(this IAsyncAction source, CancellationToken cancellati
                 throw new ArgumentNullException(nameof(source));
             }
 
-            // TODO: Handle the scenario where the 'IAsyncAction' is actually a task (i.e. originated from native code
-            // but projected into an IAsyncAction)
+            if (source is ITaskAwareAsyncInfo asyncInfo && asyncInfo.Task is Task task)
```
manodasanW, line 28: As discussed earlier, we should figure out what this means for the cancellation token.

jlaanstra: Handled.

src/cswinrt/strings/additions/Windows.Foundation/Windows.Foundation.cs
```diff
@@ -25,8 +25,12 @@ public static Task AsTask(this IAsyncAction source, CancellationToken cancellati
                 throw new ArgumentNullException(nameof(source));
             }
 
-            // TODO: Handle the scenario where the 'IAsyncAction' is actually a task (i.e. originated from native code
-            // but projected into an IAsyncAction)
+            if (source is ITaskAwareAsyncInfo asyncInfo && asyncInfo.Task is Task task)
+            {
+                return cancellationToken.CanBeCanceled ?
```
manodasanW, line 30: Did a CI build to confirm, but WaitAsync is .NET 6+ it seems so probably need a if NET guard around this to not run it on our NET standard projection.

manodasanW (approved):

PR #1545, opened 2024-03-20
Fix remaining scenarios still calling FindVftblType

manodasanW:
RCWs implementing IAsyncOperation<> happened to be calling FindVftblType.  This removes the remaining scenario still calling it.

### DIFF
--- a/src/cswinrt/code_writers.h
+++ b/src/cswinrt/code_writers.h
@@ -2241,7 +2241,7 @@ private IObjectReference Make__%()
                         {
                             w.write(R"(global::System.Threading.Interlocked.CompareExchange(ref __%, GetDefaultInterfaceObjRef(%), null);)", objrefname, get_class_hierarchy_index(classType));
                         } 
-                        else if (distance(ifaceType.GenericParam()) == 0 || ifaceType.TypeNamespace() == "Windows.Foundation.Collections")
+                        else
                         {
                             if (distance(ifaceType.GenericParam()) != 0)
                             {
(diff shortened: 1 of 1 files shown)


src/cswinrt/code_writers.h
```diff
@@ -2269,12 +2269,6 @@ private IObjectReference Make__%()
                                 bind<write_type_name>(semantics, typedef_name_type::StaticAbiClass, true)
                             );
                         }
-                        else
-                        {
-                            w.write(R"(global::System.Threading.Interlocked.CompareExchange(ref __%, (IObjectReference)typeof(IObjectReference).GetMethod("As", Type.EmptyTypes).MakeGenericMethod(typeof(%).GetHelperType().FindVftblType()).Invoke(((IWinRTObject)this).NativeObject, null), null);)",
```
Sergio0694, line 2274: Mmh... How was this working at all on NAOT before? 🤔

manodasanW: It probably wasn't, there was only like 2 classes that went into this block before.  Looking at if there is an easy test I can add for this scenario.

Sergio0694 (approved):
LGTM! Approving it now since we didn't have a test for this anyway, feel free to either just push a commit adding one to this PR, or make a separate PR later just adding that test. Either way, the change itself is clearly a good one 😄

src/cswinrt/code_writers.h
manodasanW: Confirmed new code path we are now using for all generic types implemented by RCWs is already tested by the `DeviceInformation.FindAllAsync` tests which have an RCW type implementing a collection type.

PR #1547, opened 2024-03-21
Disable NAOT native exports for non-publish builds

### LINKED ISSUE #1546: `CsWinRTAotExportsEnabled` incorrectly assumes `PublishAot == true` implies a publish is actually occurring
Microsoft.Windows.CsWinRT.targets defines a property [`CsWinRTAotExportsEnabled`](https://github.com/microsoft/CsWinRT/blob/staging/AOT/nuget/Microsoft.Windows.CsWinRT.targets#L26):

```msbuild
   <!--
      If the AOT optimizer is enabled, and we're publishing with NativeAOT, automatically set CsWinRTAotExportsEnabled as well.
      Only do this if the property is not already set by the user, so we respect any existing preference.
    -->
    <CsWinRTAotExportsEnabled Condition="'$(CsWinRTAotExportsEnabled)' == '' and '$(CsWinRTAotOptimizerEnabled)' == 'true' and '$(PublishAot)' == 'true'">true</CsWinRTAotExportsEnabled>
    <CsWinRTAotExportsEnabled Condition="'$(CsWinRTAotExportsEnabled)' != 'true'">false</CsWinRTAotExportsEnabled>
```

However, `PublishAot` being true does not indicate that a build was actually an NAOT build, instead it indicates that a publish build will produce NAOT’d assemblies. This results in build failures when expected artefacts don’t exist. For example, [WinRT.Host.dll](https://github.com/microsoft/CsWinRT/blob/staging/AOT/nuget/Microsoft.Windows.CsWinRT.Authoring.targets#L62) is outputted conditional on `CsWinRTAotExportsEnabled` not being true.

I'm not sure how often this idiom is used across the targets files, so I can't recommend a specific fix, but it's probably something like setting `CsWinRTAotExportsEnabled` to true in a target that runs only before the publish target and ensuring that the only place `PublishAot` is checked is when that's set.

Sergio0694:
## Fixes #1546

This PR disables the NAOT native exports and restores the bundled native host for non-publish builds.

### DIFF
--- a/nuget/Microsoft.Windows.CsWinRT.targets
+++ b/nuget/Microsoft.Windows.CsWinRT.targets
@@ -21,9 +21,10 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
     <!--
       If the AOT optimizer is enabled, and we're publishing with NativeAOT, automatically set CsWinRTAotExportsEnabled as well.
-      Only do this if the property is not already set by the user, so we respect any existing preference.
+      Only do this if the property is not already set by the user, so we respect any existing preference. Note that this is only
+      set for publish builds, as 'PublishAot' might also be set when doing normal builds (which would still use CoreCLR).
     -->
-    <CsWinRTAotExportsEnabled Condition="'$(CsWinRTAotExportsEnabled)' == '' and '$(CsWinRTAotOptimizerEnabled)' == 'true' and '$(PublishAot)' == 'true'">true</CsWinRTAotExportsEnabled>
+    <CsWinRTAotExportsEnabled Condition="'$(CsWinRTAotExportsEnabled)' == '' and '$(CsWinRTAotOptimizerEnabled)' == 'true' and '$(PublishAot)' == 'true' and '$(_IsPublishing)' == 'true'">true</CsWinRTAotExportsEnabled>
     <CsWinRTAotExportsEnabled Condition="'$(CsWinRTAotExportsEnabled)' != 'true'">false</CsWinRTAotExportsEnabled>
 
     <!--


MichalStrehovsky:
This looks good to me in the sense that I don't think there's any other way to do what you need. `_IsPublishing` is obviously an internal implementation detail of the SDK, but the SDK doesn't provide any other way to do this so...

Cc @baronfel @nagilson

nagilson:
Yes, keep in mind this will not work for ms build t publish, only when running dotnet publish on CLI (and eventually VS publish, but not at the moment.)

Sergio0694:
Mmh when it "doesn't work", does that mean that `_IsPublishing` would just not be set to anything? Because I'm wondering if that's the case, perhaps would it be better for us to check for `'$(_IsPublishing)' != 'false'"` rather than `'$(_IsPublishing)' == 'true'"`? The rationale being that with the former, `dotnet publish` would work the same, but `msbuild -t:publish` would then  also work (as the property would just not be set to anything). Basically it'd be the same as today, except we'd get the small improvement that `dotnet build` would correctly no longer generate those exports (as `_IsPublish` would presumably be `false` then?). Thinking that might be better than just outright breaking consumers using `msbuild` to publish, since that's relatively common. Does that make sense? 🤔

MichalStrehovsky:
I don't think you'll ever see `_IsPublishing == false`.

`dotnet publish` sets the property to true. Publish in VS currently doesn't. They're working on making it set it. `dotnet build /t:Publish` would not have it either and will still do a publish (with some extra hurdles, but it will do a publish too).

nagilson:
> I don't think you'll ever see `_IsPublishing == false`.
> 
> `dotnet publish` sets the property to true. Publish in VS currently doesn't. They're working on making it set it. `dotnet build /t:Publish` would not have it either and will still do a publish (with some extra hurdles, but it will do a publish too).

It's a good thought, but @MichalStrehovsky  is right, currently thats pretty much never false.

Sergio0694:
I see. Thank you both for the additional info! I'll revert this to draft. Seems like we just can't do it for now 😅

nagilson:
Thank you for checking in with us! I think its a good idea. If youre ok with it not being the same in msbuild t publish, then I dont see why we wouldnt merge this once vs has the behavior, which id expect sooner rather than later.

@agocke is tracking that.

nagilson (approved):

Sergio0694:
Closing this one since it's still not clear whether this approach would fully address the issue, and there's still no properly officially supported way of handling this. Current consumers (including in Windows) are not blocked by this currently anyway). We can revisit in 3.0 at this point, ideally when the .NET SDK also provides a proper way to handle this.

PR #1549, opened 2024-03-22
Addressed a couple missed out on unmanaged function pointers

manodasanW:


src/cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBuffer.cs
```diff
@@ -72,14 +72,15 @@ public static IBuffer Create(byte[] data, int offset, int length, int capacity)
...
         {
             if (t_winRtMarshalProxy != null)
                 return;
 
             try
             {
-                int hr = RoGetBufferMarshaler(out IntPtr proxyPtr);
+                IntPtr proxyPtr = default;
+                int hr = RoGetBufferMarshaler(&proxyPtr);
                 IMarshal proxy = new ABI.Com.IMarshal(ObjectReference<ABI.Com.IMarshal.Vftbl>.Attach(ref proxyPtr));
```
Sergio0694, line 84: Unrelated, but how come we're using a vtable type here? Should we fix this separately?

Sergio0694 (approved):
LGTM, good find!

jlaanstra (approved):

src/cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBuffer.cs
manodasanW: I noticed that too, I plan on addressing this separately.  But it also not a managed type, so it still AOT compatible.

PR #1550, opened 2024-03-22
Fix trimming regression for creating RCW types using ABI pointer

manodasanW:
#1505 enabled a better way that doesn't use reflection other than attributes to construct RCW types via their internal constructor.  But the removal of the previous `DynamicallyAccessed constructor attribute` regressed scenarios where someone may have an updated Windows SDK projection but are still using older projections for other components and have trimming enabled.  This brings back the attribute so we don't regress that scenario but at the same time still uses the new approach for any newly generated projection.  Given this is scoped to non-public constructors and we are already referencing these same constructors in the new way from attributes, it probably shouldn't make much of a negative size impact.

### DIFF
--- a/src/WinRT.Runtime/Marshalers.cs
+++ b/src/WinRT.Runtime/Marshalers.cs
@@ -1570,7 +1570,11 @@ namespace WinRT
 #else
     public
 #endif
-    static class MarshalInspectable<T>
+    static class MarshalInspectable<
+#if NET
+        [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.NonPublicConstructors)]
+#endif
+        T>
     {
         public static IObjectReference CreateMarshaler<V>(
             T o,
--- a/src/WinRT.Runtime/ComWrappersSupport.cs
+++ b/src/WinRT.Runtime/ComWrappersSupport.cs
@@ -502,7 +502,12 @@ namespace WinRT
             };
         }
 
-        internal static Func<IInspectable, object> CreateTypedRcwFactory(Type implementationType, string runtimeClassName = null)
+        internal static Func<IInspectable, object> CreateTypedRcwFactory(
+#if NET
+            [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.NonPublicConstructors)]
+#endif
+            Type implementationType,
+            string runtimeClassName = null)
         {
             // If runtime class name is empty or "Object", then just use IInspectable.
             if (implementationType == null || implementationType == typeof(object))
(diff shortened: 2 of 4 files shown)


src/WinRT.Runtime/Marshalers.cs
```diff
@@ -1570,7 +1570,11 @@ private static IObjectReference CreateMarshalerCore(T value)
 #else
     public
 #endif
-    static class MarshalInspectable<T>
+    static class MarshalInspectable<
```
Sergio0694, line 1573: I really, _really_ would like to avoid changing this back, and same for `ComWrappersSupport`. It's not so much for the size regression here (though I do want to run sizoscope later as well), but also because this attribute is viral and it will have to be propagated everywhere. This means that every single code path ultimately using any of these APIs will also need to have the attribute applied to it, which is extremely clunky and annoying, and something that all library authors and app developers using these CsWinRT APIs in some way would have to do. It also means that you'd be unnecessarily rooting non public constructors for all managed types in authoring scenarios where you want to get a CCW for them (via `FromManaged`), or you'd have to manually suppress those yourself, which again is extremely clunky and not intuitive at all. For instance, I hit this last scenario in multiple places in ComputeSharp, in my Win2D interop library.

Can we find some other way to work around this that doesn't require us to revert these changes? If this is a niche enough scenario, can we perhaps just use documentation to tell people that hit this (if any) to manually root things via the .rd.xml file, or something else? Or, could we maybe come up with some other fallback path with `[DynamicDependency]` or something, that we could do only when not on NAOT?

Basically anything but bring these annotations back 😭

jlaanstra:
> can we perhaps just use documentation to tell people that hit this (if any) to manually root things via the .rd.xml file, or something else?

rd.xml is totally unsupported I believe.

src/WinRT.Runtime/Marshalers.cs
manodasanW: Let me see if maybe it is possible to reduce the number of attributes we need maybe?  DynamicDependency probably won't work because this can be referencing any projection assembly.  This is used during a core scenario (getting an existing object from native code), so would be hard to ask folks to workaround it.

Sergio0694: Another idea could be to have the generator emit the `[DynamicDependency]` attributes on some method (say, a module initializer, or a partial declaration of something or whatever) after identifying what types are needed? At least it would be pay for play. We could also maybe introduce some MSBuild property for devs to indicate which outdated projections they're using, and we could emit those annotations only for types from there? Basically I really want to avoid adding the annotations for everyone. If the issue is that one backcompat scenario, let's try to find some workaround specifically for that without making the dev experience worse for everyone 🥲

PR #1554, opened 2024-03-25
Adding diagnostics and code fix for adding partial keyword

manodasanW:
Adding diagnostics and code fix for adding partial keyword for classes implementing WinRT interfaces if not already to make them AOT and trimming compatible.

### DIFF
--- a/src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
+++ b/src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
@@ -20,15 +20,11 @@ namespace Generator
         {
             var properties = context.AnalyzerConfigOptionsProvider.Select(static (provider, _) => (provider.IsCsWinRTAotOptimizerEnabled(), provider.IsCsWinRTComponent()));
 
-            var vtableAttributesOrDiagnostics = context.SyntaxProvider.CreateSyntaxProvider(
+            var vtableAttributesToAdd = context.SyntaxProvider.CreateSyntaxProvider(
                     static (n, _) => NeedVtableAttribute(n),
                     static (n, _) => GetVtableAttributeToAdd(n)
                 ).Where(vtableAttribute => vtableAttribute != null);
 
-            var diagnostics = vtableAttributesOrDiagnostics.Where(vtableAttribute => vtableAttribute.Diagnostic != null);
-            var vtableAttributesToAdd = vtableAttributesOrDiagnostics.Where(vtableAttribute => vtableAttribute.Diagnostic == null);
-
-            context.RegisterImplementationSourceOutput(diagnostics.Collect().Combine(properties), GenerateDiagnostics);
             context.RegisterImplementationSourceOutput(vtableAttributesToAdd.Collect().Combine(properties), GenerateVtableAttributes);
 
             var vtablesToAddOnLookupTable = context.SyntaxProvider.CreateSyntaxProvider(
@@ -57,10 +53,8 @@ namespace Generator
 
         private static VtableAttribute GetVtableAttributeToAdd(GeneratorSyntaxContext context)
         {
-            var classDeclaration = context.Node as ClassDeclarationSyntax;
-            var symbol = context.SemanticModel.GetDeclaredSymbol(classDeclaration);
-            bool isPartial = classDeclaration.Modifiers.Any(m => m.IsKind(SyntaxKind.PartialKeyword));
-            return GetVtableAttributeToAdd(symbol, GeneratorHelper.IsWinRTType, context.SemanticModel.Compilation.Assembly, false, "", !isPartial);
+            var symbol = context.SemanticModel.GetDeclaredSymbol(context.Node as ClassDeclarationSyntax);
+            return GetVtableAttributeToAdd(symbol, GeneratorHelper.IsWinRTType, context.SemanticModel.Compilation.Assembly, false);
         }
 
         private static string ToFullyQualifiedString(ISymbol symbol)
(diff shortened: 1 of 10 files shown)


src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
```diff
@@ -20,11 +20,15 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
...
-            var vtableAttributesToAdd = context.SyntaxProvider.CreateSyntaxProvider(
+            var vtableAttributesOrDiagnostics = context.SyntaxProvider.CreateSyntaxProvider(
                     static (n, _) => NeedVtableAttribute(n),
                     static (n, _) => GetVtableAttributeToAdd(n)
                 ).Where(vtableAttribute => vtableAttribute != null);
 
+            var diagnostics = vtableAttributesOrDiagnostics.Where(vtableAttribute => vtableAttribute.Diagnostic != null);
+            var vtableAttributesToAdd = vtableAttributesOrDiagnostics.Where(vtableAttribute => vtableAttribute.Diagnostic == null);
+
+            context.RegisterImplementationSourceOutput(diagnostics.Collect().Combine(properties), GenerateDiagnostics);
```
Sergio0694, line 31: Can we move this to a diagnostic analyzer? Emitting diagnostics from generators is highly frowned upon.

manodasanW: Yea I was debating on this, I did it here initially because I wasn't sure if it was worth doing 2 different analysis for the same syntax for the same thing for one diagnostic.  But yea I can move it to a different analyzer.

Sergio0694: I'm hoping that in the analyzer we'd be able to do a very minimal amount of processing, as we only need to discover whether these types are exposed in some way, but not do all the actual work to actually compute the vtables and whatnot. We can also probably optimize the way we compare eg. types and whatnot, but resolving symbols from the compilation start and then comparing those, rather than formatting types and comparing their names. But also in general analyzers run out of process and don't impact the core IDE inner loop, so the Roslyn teams recommends using them for diagnostics whenever possible.

manodasanW: Addressed.

src/Authoring/WinRT.SourceGenerator/WinRTAotCodeFixer.cs
```diff
@@ -0,0 +1,110 @@
...
+                    namedType.TypeKind == TypeKind.Class &&
+                    !namedType.IsAbstract &&
+                    !namedType.IsStatic)
+                {
+                    // Make sure this is a class that we would generate the WinRTExposedType attribute on
+                    // and that it isn't already partial.
+                    if (!GeneratorHelper.IsPartial(namedType) &&
+                        !GeneratorHelper.IsWinRTType(namedType) &&
+                        !GeneratorHelper.HasNonInstantiatedWinRTGeneric(namedType) &&
+                        !GeneratorHelper.HasWinRTExposedTypeAttribute(namedType))
```
Sergio0694, line 49: Do the same to check for the `WinRTExposedTypeAttribute` as well.

Sergio0694 (commented):
Left a few perf suggestions for the analyzer 🙂

src/Authoring/WinRT.SourceGenerator/WinRTAotCodeFixer.cs
```diff
@@ -0,0 +1,110 @@
...
+    [DiagnosticAnalyzer(LanguageNames.CSharp), Shared]
+    public sealed class WinRTAotDiagnosticAnalyzer : DiagnosticAnalyzer
+    {
+        private static ImmutableArray<DiagnosticDescriptor> _supportedDiagnostics = ImmutableArray.Create(WinRTRules.ClassNotAotCompatible);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => _supportedDiagnostics;
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
```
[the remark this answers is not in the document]
manodasanW: I did think about this, right now this is info diagnostics and not warnings.  So I scoped it to files that are written / not generated.  But I might expand it to generated code too.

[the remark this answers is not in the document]
Sergio0694: Oooh I see, right that makes sense. Then yeah, just `None` makes sense!

src/Authoring/WinRT.SourceGenerator/Helper.cs
```diff
@@ -225,6 +226,40 @@ public static bool IsWinRTType(ISymbol type, Func<ISymbol, bool> isAuthoringWinR
...
+                type.DeclaredAccessibility == Accessibility.Public)
+            {
+                // Authoring diagnostics will make sure all public types are valid WinRT types.
+                return true;
+            }
+
+            bool isProjectedType = HasAttributeWithType(type, winrtRuntimeTypeAttribute);
+            if (!isProjectedType & type.ContainingNamespace != null)
+            {
+                isProjectedType = MappedCSharpTypes.ContainsKey(string.Join(".", type.ContainingNamespace.ToDisplayString(), type.MetadataName));
```
Sergio0694, line 248: Not sure this will work. Doesn't `ContainingNamespace` only return the closest enclosing namespace? Ie. if we have some types with nested namespaces (eg. "System.ComponentModel.DataErrorsChangedEventArgs"), wouldn't that just return the closest namespace name (eg. "ComponentModel")? 🤔

Sergio0694 (approved):
LGTM! Might want to add a couple unit tests, but otherwise seems fine!

src/Authoring/WinRT.SourceGenerator/Helper.cs
manodasanW: It does get the closest namespace, but `ToDisplayString` makes it print out the entire namespace including parent namespaces.  Did a quick test and that seems to confirm that.

manodasanW:
> LGTM! Might want to add a couple unit tests, but otherwise seems fine!

Tests will be upcoming in another PR along with some other tests.

PR #1555, opened 2024-03-25
Add RcwReflectionFallbackGenerator type

Sergio0694:
### Follow up to #1505, supersedes #1550

This PR adds a generator to emit fallback `[DynamicDependency]` annotations to not break consumers of old projections.


src/Authoring/WinRT.SourceGenerator/RcwReflectionFallbackGenerator.cs
```diff
@@ -0,0 +1,332 @@
...
+        return Visit(assemblySymbol.GlobalNamespace);
+    }
+
+    /// <summary>
+    /// Checks whether a given type symbol has an attribute with a specified type.
+    /// </summary>
+    /// <param name="typeSymbol">The type to check.</param>
+    /// <param name="attributeTypeSymbol">The attribute to look for.</param>
+    /// <returns>Whether <paramref name="typeSymbol"/> has an attribute with type <paramref name="attributeTypeSymbol"/>.</returns>
+    private static bool ContainsAttributeWithType(ITypeSymbol typeSymbol, ITypeSymbol attributeTypeSymbol)
```
manodasanW, line 250: Same as Generator.GeneratorHelper.HasAttributeWithType

src/Authoring/WinRT.SourceGenerator/RcwReflectionFallbackGenerator.cs
```diff
@@ -0,0 +1,332 @@
...
+        /// </summary>
+        public PortableExecutableReference Reference { get; } = executableReference;
+
+        /// <summary>
+        /// Gets the <see cref="Microsoft.CodeAnalysis.Compilation"/> object for <see cref="Reference"/>.
+        /// </summary>
+        /// <returns>The <see cref="Microsoft.CodeAnalysis.Compilation"/> object for <see cref="Reference"/>.</returns>
+        /// <exception cref="InvalidOperationException">Thrown if the <see cref="Microsoft.CodeAnalysis.Compilation"/> object has been collected.</exception>
+        /// <remarks>
+        /// This method should only be used from incremental steps immediastely following a change in the metadata reference
```
manodasanW, line 308:
```suggestion
        /// This method should only be used from incremental steps immediately following a change in the metadata reference
```

src/Authoring/WinRT.SourceGenerator/RcwReflectionFallbackGenerator.cs
```diff
@@ -0,0 +1,332 @@
...
+        {
+            return false;
+        }
+
+        // Scan all dependent assemblies to look for CsWinRT with version < 2.0.8
+        foreach (AssemblyIdentity assemblyIdentity in dependentAssemblies)
+        {
+            if (assemblyIdentity.Name == "WinRT.Runtime")
+            {
+                return assemblyIdentity.Version < new Version(2, 0, 8);
```
manodasanW, line 206: Given this is an assembly version comparison, we should probably increment the version for this branch to like 2.1.0 until the actual version is determined.

https://github.com/microsoft/CsWinRT/blob/3b983c31ac1889bc78bd45b422bb6b49f0c58ae6/build/AzurePipelineTemplates/CsWinRT-Variables.yml#L5C35-L5C42

manodasanW: Checked in a change to address this so that I can also do a Windows SDK projection build to try them together.

src/Authoring/WinRT.SourceGenerator/RcwReflectionFallbackGenerator.cs
Sergio0694: Mmh not seeing any methods with that name 🤔

manodasanW: https://github.com/microsoft/CsWinRT/blob/3b983c31ac1889bc78bd45b422bb6b49f0c58ae6/src/Authoring/WinRT.SourceGenerator/Helper.cs#L320

Sergio0694: Ooh I see, it was added in a recent PR and this branch was outdated ahah, thanks! 😄

manodasanW (approved):

manodasanW:
One thought, we should also have a disable similar to the force opt-in.  But we can add that later separate from this PR.

Sergio0694:
Sure thing, I can make another PR to add a "ForceOptOut" property for this too! 🙂

PR #1556, opened 2024-03-26
Add WindowsRuntimeMarshal.TryGetDataUnsafe and use it in various places.

jlaanstra:
Allows unsafe access to the underlying buffer data.


src/cswinrt/strings/additions/Windows.Storage.Streams/IBufferByteAccess.cs
```diff
@@ -18,6 +18,20 @@ namespace ABI.Windows.Storage.Streams
...
     using global::System.ComponentModel;
 
+    internal static class IBufferByteAccessMethods
+    {
+        public static unsafe IntPtr get_Buffer(IObjectReference _obj)
+        {
+            var ThisPtr = _obj.ThisPtr;
+
+            IntPtr __retval = default;
+            global::WinRT.ExceptionHelpers.ThrowExceptionForHR((*(delegate* unmanaged[Stdcall]<IntPtr, IntPtr*, int>**)ThisPtr)[6](ThisPtr, &__retval));
```
manodasanW, line 28: Is the index here right?  The interface looks to implement IUnknown.

https://learn.microsoft.com/en-us/windows/win32/api/robuffer/ns-robuffer-ibufferbyteaccess

src/cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBufferExtensions.cs
```diff
@@ -248,6 +249,27 @@ public static void CopyTo(this IBuffer source, uint sourceIndex, IBuffer destina
 
 #endregion (IBuffer).CopyTo extensions for copying to an (IBuffer)
 
+        public static bool TryGetDataUnsafe(this IBuffer buffer, out IntPtr dataPtr)
+        {
+            if (buffer is IWinRTObject winrtObj && winrtObj.HasUnwrappableNativeObject &&
+                winrtObj.NativeObject.TryAs<IUnknownVftbl>(global::ABI.Windows.Storage.Streams.IBufferByteAccessMethods.IID, out var objRef) >= 0)
```
Sergio0694, line 255: We can make this more efficient by just using the `TryAs` overload returning a raw pointer and doing the marshalling manually, no? That way we can avoid allocating a new `ObjectReference`, since it's not really needed anyway here. Basically we can just do manual RAII on that returned COM object.

Sergio0694 (changes requested):

src/cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBufferExtensions.cs
```diff
@@ -248,6 +249,32 @@ public static void CopyTo(this IBuffer source, uint sourceIndex, IBuffer destina
...
 
+        public static bool TryGetDataUnsafe(this IBuffer buffer, out IntPtr dataPtr)
+        {
+            if (buffer == null)
+            {
+                dataPtr = IntPtr.Zero;
+                return false;
+            }
+
+            if (ComWrappersSupport.TryUnwrapObject(buffer, out var objRef))
```
manodasanW, line 260: We lost the QI to IBufferByteAccess in the conversion?

src/cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBufferExtensions.cs
jlaanstra: Not anymore as I'm using ComWrapperSupport now.

src/cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBufferExtensions.cs
jlaanstra: Yup will add back.

src/cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBufferExtensions.cs
Sergio0694: But that still has the same problem. Shouldn't we keep the faster path on modern .NET?

jlaanstra: That would certainly be ideal. @manodasanW any suggestion for how to do that?

manodasanW: So `TryUnwrapObject` should be optimized already for .NET 5+.  But you can avoid the object construction on the QI afterwards using the other `TryAs` method that gives you the ptr as an out and then update the methods static class to take a `IntPtr` instead.  You would need to call Marshal.Release afterwards.  There is also the `AsValue` method which takes an `IID` that achieves this, but it throws when QI fails so you probably don't want that here.

src/cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeMarshal.cs
```diff
@@ -0,0 +1,62 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
manodasanW, line 1: This is a new function right that you wrote, I would just use the standard MIT licensing we use in other files.

manodasanW: I would still have something like this: https://github.com/microsoft/CsWinRT/blob/9854573b9880650efd87c5be41a553d125877c92/src/WinRT.Runtime/CastExtensions.cs#L1-L2

src/cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeMarshal.cs
```diff
@@ -0,0 +1,57 @@
...
+                return false;
+            }
+
+            if (ComWrappersSupport.TryUnwrapObject(buffer, out var unwrapped) &&
+                unwrapped.TryAs(global::ABI.Windows.Storage.Streams.IBufferByteAccessMethods.IID, out IntPtr ThisPtr) >= 0)
+            {
+                try
+                {
+                    IntPtr __retval = default;
+                    global::WinRT.ExceptionHelpers.ThrowExceptionForHR((*(delegate* unmanaged[Stdcall]<IntPtr, IntPtr*, int>**)ThisPtr)[3](ThisPtr, &__retval));
```
Sergio0694, line 35: Why the roundtrip to a managed exception? We could just check the HRESULT instead.

src/cswinrt/strings/additions/Windows.Storage.Streams/IBufferByteAccess.cs
```diff
@@ -18,6 +18,24 @@ namespace ABI.Windows.Storage.Streams
     using global::System.Runtime.InteropServices;
     using global::System.ComponentModel;
 
+    internal static class IBufferByteAccessMethods
```
Sergio0694, line 21: Where are these new memberes being used?

src/cswinrt/strings/additions/Windows.Storage.Streams/IBufferByteAccess.cs
```diff
@@ -18,6 +18,24 @@ namespace ABI.Windows.Storage.Streams
...
+    {
+        public static unsafe IntPtr get_Buffer(IObjectReference _obj)
```
manodasanW, line 23: is this unused?

src/cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeBufferExtensions.cs
```diff
@@ -639,17 +641,21 @@ public override void WriteByte(byte value)
...
         }  // class WindowsRuntimeBufferUnmanagedMemoryStream
-
+
         private static Span<byte> GetSpanForCapacityUnsafe(this IBuffer buffer, uint offset)
         {
             Debug.Assert(0 <= offset);
             Debug.Assert(offset < buffer.Capacity);
 
-            unsafe
+            if (!WindowsRuntimeMarshal.TryGetDataUnsafe(buffer, out IntPtr buffPtr))
```
Sergio0694, line 650: Technically a GC hole (the JIT might reorder that property access), add a `GC.KeepAlive`.

src/cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeMarshal.cs
```diff
@@ -0,0 +1,57 @@
...
+    /// Contains extension methods that expose operations on WinRT <code>Windows.Foundation.IBuffer</code>.
+    /// </summary>
+#if EMBED
+    internal
+#else
+    public
+#endif
+    static partial class WindowsRuntimeMarshal
+    {
+        public static unsafe bool TryGetDataUnsafe(IBuffer buffer, out IntPtr dataPtr)
```
Sergio0694, line 21: Add some XML docs with also some remarks explaining why this is unsafe and how to use it safely?

src/cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeMarshal.cs
jlaanstra: We aren't roundtripping. We still want to throw if the call fails.

src/cswinrt/strings/additions/Windows.Storage.Streams/IBufferByteAccess.cs
jlaanstra: It is now :).

src/cswinrt/strings/additions/Windows.Storage.Streams/IBufferByteAccess.cs
jlaanstra: They are no longer used, removed. Just need to keep the IID.

manodasanW:
D:\a\_work\1\s\src\Projections\Windows\Generated Files\net8.0\CsWinRT\Windows.Storage.Streams.cs(8700,37): error CS0234: The type or namespace name 'IBufferByteAccessMethods' does not exist in the namespace 'ABI.Windows.Storage.Streams' (are you missing an assembly reference?) [D:\a\_work\1\s\src\Projections\Windows\Windows.csproj::TargetFramework=net8.0]

src/cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeMarshal.cs
```diff
@@ -0,0 +1,71 @@
...
+        {
+            if (buffer == null)
+            {
+                dataPtr = IntPtr.Zero;
+                return false;
+            }
+
+            if (ComWrappersSupport.TryUnwrapObject(buffer, out var unwrapped) &&
+                unwrapped.TryAs(global::ABI.Windows.Storage.Streams.IBufferByteAccessMethods.IID, out IntPtr ThisPtr) >= 0)
```
manodasanW, line 44:
```suggestion
                unwrapped.TryAs(IID, out IntPtr ThisPtr) >= 0)

```

src/cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeMarshal.cs
Sergio0694: Oops, sorry I misread, nevermind 😅

src/cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeMarshal.cs
```diff
@@ -0,0 +1,72 @@
...
+    /// </summary>
+#if EMBED || !NET // for netstandard this type conflicts with the type in the BCL so make it internal
+    internal
+#else
+    public
+#endif
+    static partial class WindowsRuntimeMarshal
+    {
+#if NET
+        private static global::System.Guid IID { get; } = new Guid(new global::System.ReadOnlySpan<byte>(new byte[] { 0xEF, 0x0F, 0x5A, 0x90, 0x53, 0xBC, 0xDF, 0x11, 0x8C, 0x49, 0x00, 0x1E, 0x4F, 0xC6, 0x86, 0xDA }));
```
Sergio0694, line 26: Remove this and just use `global::WinRT.Interop.IID.IID_IBufferByteAccess`.

Sergio0694 (changes requested):

src/cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeMarshal.cs
```diff
@@ -0,0 +1,72 @@
...
+#else
+        private static global::System.Guid IID { get; } = new Guid(new byte[] { 0xEF, 0x0F, 0x5A, 0x90, 0x53, 0xBC, 0xDF, 0x11, 0x8C, 0x49, 0x00, 0x1E, 0x4F, 0xC6, 0x86, 0xDA });
+#endif
+        /// <summary>
+        /// Returns a pointer to the underlying data representation of the <see cref="IBuffer"/>.
+        /// Callers are responsible for ensuring that the buffer is kept alive while the pointer is in use.
+        /// </summary>
+        /// <param name="buffer">The buffer to get the data pointer for.</param>
+        /// <param name="dataPtr">The pointer to the underlying data representation of the buffer.</param>
+        public static unsafe bool TryGetDataUnsafe(IBuffer buffer, out IntPtr dataPtr)
```
Sergio0694, line 36: While we're at it, should we enable nullability annotations in this file? On `NET`, we could also make that `buffer` parameter be `[NotNullWhen(true)] IBuffer? buffer`, which could be kinda nice.

src/cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeMarshal.cs
```diff
@@ -0,0 +1,73 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace System.Runtime.InteropServices.WindowsRuntime
+{
+    using System.Diagnostics;
+    using System.Diagnostics.CodeAnalysis;
+    using System.IO;
+    using System.Threading;
```
manodasanW, line 9: I feel like we can simplify this list such as remove threading?

manodasanW (approved):

src/cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeMarshal.cs
```diff
@@ -0,0 +1,73 @@
...
+        /// Returns a pointer to the underlying data representation of the <see cref="IBuffer"/>.
+        /// Callers are responsible for ensuring that the buffer is kept alive while the pointer is in use.
+        /// </summary>
+        /// <param name="buffer">The buffer to get the data pointer for.</param>
+        /// <param name="dataPtr">The pointer to the underlying data representation of the buffer.</param>
+        public static unsafe bool TryGetDataUnsafe(
+#if NET
+            [NotNullWhen(true)]
+#endif
+            IBuffer? buffer, out IntPtr dataPtr)
```
Sergio0694, line 36: Remind me again, why not use `byte*` for the return type?

src/cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeMarshal.cs
```diff
@@ -0,0 +1,71 @@
...
+            if (buffer == null)
+            {
+                dataPtr = IntPtr.Zero;
+                return false;
+            }
+
+            if (ComWrappersSupport.TryUnwrapObject(buffer, out var unwrapped) &&
+                unwrapped.TryAs(global::WinRT.Interop.IID.IID_IBufferByteAccess, out IntPtr ThisPtr) >= 0)
+            {
+                try
```
Sergio0694, line 45: We could optimize this to not need this `try/finally`:
```csharp
IntPtr __retval;
int hresult = (*(delegate* unmanaged[Stdcall]<IntPtr, IntPtr*, int>**)ThisPtr)[3](ThisPtr, &__retval);
dataPtr = __retval;

Marshal.Release(ThisPtr);

global::WinRT.ExceptionHelpers.ThrowExceptionForHR(hresult);

return true;
```

But also, why are we throwing in the first place here, given the method is named "Try"?

Sergio0694 (changes requested):

src/cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeMarshal.cs
jlaanstra: Try doesn't mean we never throw. Try here means we are not throwing if the type doesn't implement IBufferByteAccess. If it does implement it but the implementation is bad we still throw.

Sergio0694 (approved):

src/cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeMarshal.cs
hez2010: I don't think it's an optimization though.
`try/finally` doesn't have performance impact for fast path (no exception being thrown) and it can make sure the finally block being executed once there's an exception. You can check the codegen, unless you don't have `catch` blocks in a `try`, it shouldn't make any meaningful impact on the performance.

PR #1557, opened 2024-03-28
Fix more warnings in WinRT.Runtime

Sergio0694:
This PR includes more fixes for warnings in WinRT.Runtime:
- Unused fields
- Missing `[DynamicallyAccessedMembers]` annotations
- Obsolete `AsInterface<I>` and `As<I>` methods

### DIFF
--- a/src/WinRT.Runtime/ComWrappersSupport.cs
+++ b/src/WinRT.Runtime/ComWrappersSupport.cs
@@ -485,7 +485,11 @@ namespace WinRT
             };
         }
 
-        private static Func<IInspectable, object> CreateCustomTypeMappingFactory(Type customTypeHelperType)
+        private static Func<IInspectable, object> CreateCustomTypeMappingFactory(
+#if NET
+            [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods)]
+#endif
+            Type customTypeHelperType)
         {
             var fromAbiMethod = customTypeHelperType.GetMethod("FromAbi", BindingFlags.Public | BindingFlags.Static);
             if (fromAbiMethod is null)
--- a/src/WinRT.Runtime/IInspectable.cs
+++ b/src/WinRT.Runtime/IInspectable.cs
@@ -2,6 +2,8 @@
 // Licensed under the MIT License.
 
 using System;
+using System.ComponentModel;
+using System.Diagnostics.CodeAnalysis;
 using System.Runtime.InteropServices;
 using WinRT.Interop;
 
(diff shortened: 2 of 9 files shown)


manodasanW (approved):

Sergio0694:
Hey @michalstrehovsky, would you be able to help investigate some small size regressions that are caused by this PR?

<details>
    <summary><b>Sizoscope diff (click to expand):</b></summary>
<br/>

image
</details>

I'm trying to figure out what's causing all that `ICustomProperty` goo, plus those System.Reflection.* bits and more to get rooted. I'm also seeing a bunch of stuff like, reflection over editor browsable attributes and stuff showing up in the root paths..??

Here's the full diff with MSTAT info:
- [fix-trim-warnings-before.zip](https://github.com/microsoft/CsWinRT/files/14921852/fix-trim-warnings-before.zip)
- [fix-trim-warnings-after.zip](https://github.com/microsoft/CsWinRT/files/14921854/fix-trim-warnings-after.zip)

I'll also look into it, but if you have any ideas, please let me know! 😅
Thank you!

Sergio0694:
It's weird because I can see `ICustomProperty` being rooted by `Do_Abi_GetCustomProperty_0` and `Do_Abi_GetIndexedProperty_1` in `ManagedCustomPropertyProviderVftbl`, which is then rooted by `ComWrappersSupport.GetInterfaceTableEntries`, but the part that I don't understand is, it seems like that method hasn't been modified in a while, and most importantly it hasn't been modified in this PR. So why is the diff showing up here?? 🤔

MichalStrehovsky:
> It's weird because I can see `ICustomProperty` being rooted by `Do_Abi_GetCustomProperty_0` and `Do_Abi_GetIndexedProperty_1` in `ManagedCustomPropertyProviderVftbl`, which is then rooted by `ComWrappersSupport.GetInterfaceTableEntries`, but the part that I don't understand is, it seems like that method hasn't been modified in a while, and most importantly it hasn't been modified in this PR. So why is the diff showing up here?? 🤔

I cannot make good sense of it because the staging/AOT branch is not github searchable and it very difficult to navigate this repo (classes are not in same-named files, etc.)

The difference is that when we process custom attributes on `MarshalInterface<ICustomProperty>`, one compilation saw a `WindowsRuntimeHelperType(typeof(ICustomProperty))` that said to make methods and fields reflection visible, the other didn't see an attribute with such instruction. Either attribute missing, different annotation on the attribute, previously no matching methods/fields on ICustomProperty, etc. As I wrote, very difficult to navigate this PR. Hopefully you can make sense of it from this.

Sergio0694:
Mmh that's weird. I mean that's over `ICustomProperty`, which was referenced by a bunch of places ultimately from `ManagedCustomPropertyProviderVftbl`. I'm still not entirely sure I understand how could that not also get rooted in the previous branch. Either way, should all be fixed by #1565 now, which allowed trimming all of that entirely 😄

PR #1559, opened 2024-03-31
Reference CsWinRTEnableLogging in msbuild targets

hez2010:
CsWinRTEnableLogging is not referenced anywhere else thus the MSBuild extension won't recognize it as a known property.
Reference it to make it be included in the completion list.

image

BTW I would suggest to add references to all new available msbuild properties introduced in the AOT branch to the target file.


nuget/Microsoft.Windows.CsWinRT.targets
```diff
@@ -16,6 +16,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <AllowUnsafeBlocks Condition="$(CsWinRTEnabled)">true</AllowUnsafeBlocks>
     <CoreCompileDependsOn>CsWinRTIncludeProjection;CsWinRTRemoveWinMDReferences;$(CoreCompileDependsOn)</CoreCompileDependsOn>
     <TrackFileAccess Condition="'$(CsWinRTComponent)' != 'true'">false</TrackFileAccess> 
+    <CsWinRTEnableLogging>$(CsWinRTEnableLogging)</CsWinRTEnableLogging>
```
manodasanW, line 19: Given this property is only used during authoring scenarios today, can we move this to the authoring target?

https://github.com/microsoft/CsWinRT/blob/master/nuget/Microsoft.Windows.CsWinRT.Authoring.targets

manodasanW (approved):

PR #1560, opened 2024-04-02
Fixing various crashes discovered with latest builds and removing vftbl

manodasanW:
- `NotifyCollectionChangedEventArgs` still had `vftbls`.  Updating the implementation for it to remove that and also rewriting parts to simplify the number of functions we actually need and avoiding `.GUID` calls.
- Fixing issue where `MarshalInterface<>.FromManaged` calls still rely on the helper type to get the PIID, but the helper type isn't registered for the generic interfaces.  To address that, making sure the helper type is registered in the RCW scenario where it isn't today.
- Fixing issue where XAML can give a generic interface like `Microsoft.UI.XAML.IObservableVector<GraidentStop>` and ask for the `IXamlType` for it but this is failing on AOT because we don't want to rely on `MakeGenericType` to construct that type instance.  Instead, I am doing what we do in the non-generic code path where we rely on `assembly.GetType` to get the type if it does exist and isn't trimmed.  If it is trimmed or native code for it wasn't generated, then it won't be found and that is fine as it means there were no C# references / bindings to that type and thereby the XAML compiler won't generate the code which asks for it.  Even if it does ask for it, we will go down the fallback path when null is returned where we return a fake type that XAML doesn't actually use due to no C# references.


src/WinRT.Runtime/TypeNameSupport.cs
```diff
@@ -187,6 +187,33 @@ private static Type FindTypeByNameCore(string runtimeClassName, Type[] genericTy
...
+                        nameBuilder.Append('[');
+                        nameBuilder.Append(genericTypes[i].AssemblyQualifiedName);
+                        nameBuilder.Append(']');
+                    }
+                    nameBuilder.Append(']');
+                    var resolvedGenericType = resolvedType.Assembly.GetType(nameBuilder.ToString());
+                    if (resolvedGenericType is not null)
+                    {
+                        return resolvedGenericType;
+                    }
```
Sergio0694, line 215: This works, but I'm wondering if there isn't a better way to do this... Essentially we're roundtripping to a `string` just to try to load a type with that fully qualified name, which seems a bit wasteful (and is the fully instantiated generic type name even documented to have exactly that format?). It kinda seems like here we'd want some `TryMakeGenericType` API that simply tries to _look up_ whether a given generic type exists, but just returns `null` if it doesn't, and it doesn't root any additional reflection or type builder stuff.

@agocke @MichalStrehovsky @jkoritzinsky do you know if there's a better way to do that, or if an API equivalent to what I described already exists? If not, do you think it would be worth it to create a proposal for that? It seems like it shouldn't be that costly to implement..? 🤔

MichalStrehovsky: This is basically `resolvedType.MakeGenericType(genericTypes)`, but done in an obscure way so that there's only trimming warnings and no AOT warnings, no?

This will throw if the type cannot be constructed on AOT; it's not a lookup. We don't have a lookup API - try/catch would be needed for that.

Sergio0694: Yeah that's what I was wondering. So essentially I'm asking:
- If we just used `MakeGenericType` with a try/catch and a suppression for the AOT warning, would that not cause a bunch of type loader cruft to be rooted (for `MakeGenericType` as a whole) even if we don't need it? I don't love that 🥲
- Thoughts about an API proposal for that lookup API? It seems like there's no "good" replacement for it currently, and it seems quite useful in scenarios like this on AOT? It could also be implemented in a way that makes sure it wouldn't root any unnecessary reflection and type loader thing, since it should only ever do a lookup?

MichalStrehovsky: If you have Type.GetType/Assembly.GetType anywhere, you're already rooting the type loader. CsWinRT would need to fully stop using reflection to even entertain the idea of getting rid of the type loader. The type loader gets rooted from many code paths (do you read custom attributes anywhere? that needs type loader. do you have a generic virtual method anywhere? that's also type loader. etc.)

I don't think it's possible to take it out in the current architecture and a new API won't help.

Sergio0694: Oh... Well that's a bummer. I thought the type system stuff could at least get partially trimmed. I suppose we could do sizoscope just to double check what is going on here, but essentially you're saying that just using `MakeGenericType` would have no size regressions at all if we were to use it here? If so, would you recommend to just do that with a warning suppression, or to stick with the current approach building the type name and trying to get that? 🤔

manodasanW: I am thinking if we confirm that using `MakeGenericType` has no size changes, we just use that and suppress the warning if any given it sounds like both ways are equivalent.

MichalStrehovsky: `Assembly.GetType` will literally call `MakeGenericType` as part of the operations so this is just a buggier and slower version of it.

You can definitely suppress the warning around the `Nullable_Delegate` handling because if the generic parameter is a reference type (and a delegate is a reference type), `MakeGeneric` will work.

I don't immediately see how/why it could be possible to suppress the warning for the other cases though.

Sergio0694: Can we not just use `[UnconditionalSuppress]` and a `try/catch`, so that if we happen to have a value type type argument that has been trimmed out, we just catch that and assume the type is missing? I mean wouldn't that attribute make ILC not complain about that on NAOT too when publishing, ie. not show any warnings?

MichalStrehovsky: > Can we not just use [UnconditionalSuppress] and a try/catch, so that if we happen to have a value type type argument that has been trimmed out, we just catch that and assume the type is missing?

This will introduce a difference between F5 debug time and publish time.

The code as it is there right now doesn't have this problem because MakeGeneric is guarded by IsDynamicCodeSupported. This property turns into false as soon as someone puts PublishAot=true in the project (even if we do have a JIT). That way there is a consistent user experience between developing the app and running the result of publish (i.e. if you were to run into the exception, you'd see it at a point where it's easy to debug).

If the suppression introduces a behavioral difference between F5 debug time and publish time, it's not a suppression, it's a bug.

Sergio0694: I see. I guess I'm not sure I understand how to properly fix this at all then 😅

MichalStrehovsky: > I see. I guess I'm not sure I understand how to properly fix this at all then 😅

1. Does this involve value type generic arguments or reference type generic arguments? Reference type can be checked/asserted and then the suppression skipped.
2. If there are value types, is there a bound? Like only Guid, int, double, etc. We could just special case those instead of MakeGeneric?
3. Is there a functional subset we could carve out and call "AOT safe" (have this throw for IsDynamicCodeSupported for anything else - you cannot use _all of_ CsWinRT, but some basic predetermined set works)?

If there are no constraints we can impose and in the limit this could be `KeyValuePair<KeyValuePair<KeyValuePair<int, double>, double>, KeyValuePair<int>...>` then yeah, this won't work with AOT.

src/WinRT.Runtime/Interop/IID.g.cs
```diff
@@ -15,6 +15,7 @@ namespace WinRT.Interop
 #endif
     static class IID
     {
+
```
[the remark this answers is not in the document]
manodasanW: Looks like VS causes issues when LF is used for new line for tt files based on VS bug.  Switched tt file to CRLF and things seem to be better now.

src/WinRT.Runtime/TypeNameSupport.cs
manodasanW: Looking into the WinUI scenarios that are using this, it looks like the generic arguments are probably reference types especially given I believe the generic argument would probably be derived from `Microsoft.UI.Xaml.DepedencyObject`.  Due to that, per Michal's comments, I will scope the generic args to reference arguments and not throw in those scenarios.  Generic args which are value types will continue to throw.

src/WinRT.Runtime/TypeNameSupport.cs
```diff
@@ -187,18 +187,27 @@ private static Type FindTypeByNameCore(string runtimeClassName, Type[] genericTy
             {
                 if (genericTypes != null)
                 {
+                    if (resolvedType == typeof(global::System.Nullable<>) && genericTypes[0].IsDelegate())
+                    {
+#pragma warning disable IL3050 // Generic type is reference type.
```
Sergio0694, line 192: Suppressing warnings like this will only affect the analyzer, but ILC will still produce warnings at compile time. That's also why the CI build failed (also, this way we could double check that those warnings do fail the build, nice 😄). We should probably move this code in the `if (genericTypes != null)` block into a small helper function, and add `[UnconditionalSuppressMessage]` over it for IL3050. That way ILC will also not warn for it.

Sergio0694 (changes requested):

src/WinRT.Runtime/TypeNameSupport.cs
agocke: Since attribute suppressions have whole-method granularity, splitting out helper methods is a good practice. You can use local functions as well, if you don't want to add a method to the parent class.

Sergio0694: Yup, using local functions is the approach we've been using through CsWinRT already in similar situations (basically for all code paths only meant for non-AOT scenarios). It's quite convenient indeed and avoids cluttering types! 😄

manodasanW: Addressed with local function.

src/WinRT.Runtime/TypeNameSupport.cs
```diff
@@ -187,26 +187,39 @@ private static Type FindTypeByNameCore(string runtimeClassName, Type[] genericTy
...
+                    return typeof(ABI.System.Nullable_Delegate<>).MakeGenericType(genericTypes);
+                }
+
+#if NET
+                foreach (var type in genericTypes)
+                {
+                    if (type.IsValueType)
                     {
-                        throw new NotSupportedException($"Cannot provide generic type from '{runtimeClassName}'.");
+                        if (!RuntimeFeature.IsDynamicCodeCompiled)
```
Sergio0694, line 213: Nit: can hoist the check here? Ie. just check for `IsDynamicCodeCompiled` once and skip the whole loop if we're not on NAOT. I mean technically the code is the same since it's a constant, but seems easier to read and could help the JIT on CoreCLR just a little bit more to remove the entire loop?

manodasanW: good point, addressed.

Sergio0694 (approved):
LGTM! :shipit:

PR #1561, opened 2024-04-05
Use MethodInvoker in MarshalGenericFallback<T>

Sergio0694:
This PR updates `MarshalGenericFallback<T>` on .NET 8 to use `MethodInvoker` instead of `MethodInfo.Invoke`, to invoke all the generated marshalling methods on the resolved helper types. This is faster and also avoids allocating arrays for all arguments.

### DIFF
--- a/src/WinRT.Runtime/Marshalers.cs
+++ b/src/WinRT.Runtime/Marshalers.cs
@@ -33,6 +33,28 @@ namespace WinRT
             }
         }
 #endif
+
+#if NET8_0_OR_GREATER
+#nullable enable
+        /// <summary>
+        /// Tries to create a <see cref="MethodInvoker"/> for a given method on a helper type.
+        /// </summary>
+        /// <param name="helperType">The input helper type.</param>
+        /// <param name="methodName">The name of the method to get.</param>
+        /// <returns>The resulting <see cref="MethodInvoker"/> instance, if the method was present.</returns>
+        public static MethodInvoker? TryGetMethodInvoker([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods)] Type helperType, string methodName)
+        {
+            MethodInfo? method = helperType.GetMethod(methodName, BindingFlags.Public | BindingFlags.Static);
+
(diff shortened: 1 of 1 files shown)


src/WinRT.Runtime/Marshalers.cs
```diff
@@ -740,33 +779,52 @@ internal sealed class MarshalGenericFallback<T>
...
-            _createMarshaler2 = helperType.GetMethod("CreateMarshaler2", BindingFlags.Public | BindingFlags.Static);
-            _disposeAbi = helperType.GetMethod("DisposeAbi", BindingFlags.Public | BindingFlags.Static);
-            _createMarshalerArray = helperType.GetMethod("CreateMarshalerArray", BindingFlags.Public | BindingFlags.Static);
-            _getAbiArray = helperType.GetMethod("GetAbiArray", BindingFlags.Public | BindingFlags.Static);
-            _fromAbiArray = helperType.GetMethod("FromAbiArray", BindingFlags.Public | BindingFlags.Static);
-            _fromManagedArray = helperType.GetMethod("FromManagedArray", BindingFlags.Public | BindingFlags.Static);
-            _disposeMarshalerArray = helperType.GetMethod("DisposeMarshalerArray", BindingFlags.Public | BindingFlags.Static);
-            _disposeAbiArray = helperType.GetMethod("DisposeAbiArray", BindingFlags.Public | BindingFlags.Static);
+#if NET8_0_OR_GREATER
+            _createMarshaler = MarshalExtensions.TryGetMethodInvoker(helperType, "CreateMarshaler");
```
Sergio0694, line 791: @MichalStrehovsky additional question: is creating a `MethodInvoker` instance fast enough that in this scenario (ie. we're already doing some reflection lookups) it doesn't matter and we continue eagerly initializing these instances, or is it recommended to load them on the fly with some kind of lazy loading strategy? Ie. is this current code ok here?

MichalStrehovsky: No need for lazy loading.

manodasanW (approved):

PR #1565, opened 2024-04-09
Add 'CsWinRTEnableICustomPropertyProviderSupport' feature switch

Sergio0694:
While checking #1557, I noticed some very annoying traces in sizoscope:

image

None of this is needed if you're just building a minimal component, but it's rooted by default.
This PR adds a new 'CsWinRTEnableICustomPropertyProviderSupport feature switch to allow opting out.
The default is the current behavior (ie. support enabled), so no semantics changes are introduced in this PR.


manodasanW (approved):

Sergio0694:
Double checked, it does correctly trim out all the `ICustomProperty` stuff, as far as I can tell:

image

PR #1566, opened 2024-04-09
Fix trim warnings in 'AdaptiveFromAbiHelper'

Sergio0694:
This PR tries to fix the trim warnings left in `AdaptiveFromAbiHelper`. Not a huge fan, but... It should work? 😅

### DIFF
--- a/src/WinRT.Runtime/Projections/Bindable.net5.cs
+++ b/src/WinRT.Runtime/Projections/Bindable.net5.cs
@@ -373,6 +373,8 @@ namespace ABI.System.Collections
 {
     using global::Microsoft.UI.Xaml.Interop;
     using global::System;
+    using global::System.Diagnostics.CodeAnalysis;
+    using global::System.Reflection;
     using global::System.Runtime.CompilerServices;
 
 #if EMBED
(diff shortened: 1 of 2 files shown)


src/WinRT.Runtime/Projections/Bindable.net5.cs
```diff
@@ -396,22 +398,84 @@ internal unsafe interface IEnumerable : global::System.Collections.IEnumerable,
...
-
+        {
+#if NET8_0_OR_GREATER
+            private readonly MethodInvoker _enumerator;
+#else
+            private readonly MethodInfo _enumerator;
+#endif
+
+#if NET
+            [DynamicDependency(DynamicallyAccessedMemberTypes.PublicMethods, typeof(IEnumerable<>))]
```
Sergio0694, line 409: @MichalStrehovsky I could use a second pair of eyes to double check this is actually safe. I'm not 100% sure 😅

src/WinRT.Runtime/Projections/Bindable.net5.cs
```diff
@@ -396,22 +398,84 @@ internal unsafe interface IEnumerable : global::System.Collections.IEnumerable,
...
+            [SuppressMessage("Trimming", "IL2070", Justification = "We're using '[DynamicDependency]' to preserve public methods of 'IEnumerable<T>'.")]
+            [SuppressMessage("Trimming", "IL2075", Justification = "We're using '[DynamicDependency]' to preserve public methods of 'IEnumerable<T>'.")]
```
MichalStrehovsky, line 411: This suppression/dynamicdependency combo is avoidable. Replace the `var methodInfo` line with

```csharp
var methodInfo = enumGenericType.GetMemberWithSameMetadataDefinitionAs(typeof(IEnumerable<>).GetMethod("GetEnumerator"));
```

Sergio0694: TIL, that's neat! 👀

Out of curiosity, what happens if the type implements two `IEnumerable<T>`-s? Will this just throw?

MichalStrehovsky: I don't understand the question. You call the API on `IEnumerable<SomeType>` with `IEnumerable<>.Method` and it returns `IEnumerable<SomeType>.Method`.

Sergio0694: Oh my bad, yeah that made no sense, I misread your comment. I can see that the call is actually on the constructed interface type, not on the concrete type that implements the interface. Ignore that 😅

src/WinRT.Runtime/Projections/Bindable.net5.cs
```diff
@@ -396,22 +398,82 @@ internal unsafe interface IEnumerable : global::System.Collections.IEnumerable,
...
+
+                    enumGenericType = GetEnumerableOfTInterface(runtimeType);
+                }
+
+#if NET8_0_OR_GREATER
+                var methodInfo = (MethodBase)enumGenericType?.GetMemberWithSameMetadataDefinitionAs(EnumerableOfTGetEnumerator);
+
+                _enumerator = methodInfo is null ? null : MethodInvoker.Create(methodInfo);
+#else
+                _enumerator = enumGenericType?.GetMethod("GetEnumerator");
```
manodasanW, line 459: Any reason you didn't use `GetMemberWithSameMetadataDefinitionAs` on the else case here on NET 6?  I assume we can cast to MethodBase and be able to call invoke on it.

manodasanW (approved):

PR #1573, opened 2024-04-17
Add WinRTRuntimeClassName attribute for use in scenarios where we previously used interface name

manodasanW:
In scenarios where we have a type that isn't a WinRT type, but implements WinRT interfaces, we previously went through the interfaces and determined the most derived one to return as the runtime class name.  This is not trim friendly.  With these changes, we will do this in the AOT source generator rather than at runtime and instead at runtime lookup the attribute on the type.  This also allows the opportunity for someone in this scenario to specify their own class name if desired as today if there are multiple interfaces that don't inherit from each other, we will need to choose one and that might not be what is desired by the author.

In addition to the scenarios where we put the attribute on the type, there are a couple scenarios where the type isn't authored in the component but rather a system type.  To handle these scenarios, we use the lookup table similar to what we do in the CCW vtable scenario.

### DIFF
--- a/src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
+++ b/src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
@@ -56,7 +56,7 @@ namespace Generator
         private static VtableAttribute GetVtableAttributeToAdd(GeneratorSyntaxContext context)
         {
             var symbol = context.SemanticModel.GetDeclaredSymbol(context.Node as ClassDeclarationSyntax);
-            return GetVtableAttributeToAdd(symbol, GeneratorHelper.IsWinRTType, context.SemanticModel.Compilation.Assembly, context.SemanticModel.Compilation.HasImplicitConversion, false);
+            return GetVtableAttributeToAdd(symbol, GeneratorHelper.IsWinRTType, context.SemanticModel.Compilation, false);
         }
 
         private static string ToFullyQualifiedString(ISymbol symbol)
@@ -155,8 +155,7 @@ namespace Generator
         internal static VtableAttribute GetVtableAttributeToAdd(
             ITypeSymbol symbol, 
             Func<ISymbol, bool> isWinRTType, 
-            IAssemblySymbol assemblySymbol, 
-            Func<ITypeSymbol, ITypeSymbol, bool> isImplicitConversion,
+            Compilation compilation,
             bool isAuthoring, 
             string authoringDefaultInterface = "")
         {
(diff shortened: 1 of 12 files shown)


src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
```diff
@@ -843,6 +947,33 @@ private static ComWrappers.ComInterfaceEntry[] LookupVtableEntries(Type type)
...
+                    source.AppendLine($$"""
+                                private static string LookupRuntimeClassName(Type type)
+                                {
+                                    string typeName = type.ToString();
+                                """);
+
+                    foreach (var vtableAttribute in value.vtableAttributes.ToImmutableHashSet().Where(v => !string.IsNullOrEmpty(v.RuntimeClassName)))
+                    {
+                        source.AppendLine($$"""
+                                if (typeName == "{{vtableAttribute.VtableLookupClassName}}")
```
Sergio0694, line 963: Just so I understand, are we switching on the type name here and not just on the type itself to avoid rooting the types? Otherwise we do we round-trip to `string` rather than comparing `Type` instances directly?

src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
```diff
@@ -843,6 +947,33 @@ private static ComWrappers.ComInterfaceEntry[] LookupVtableEntries(Type type)
                 source.AppendLine($$"""
                                                 return default;
                                             }
+                                    """);
+
+                if (hasRuntimeClasNameEntries)
+                {
+                    source.AppendLine($$"""
+                                private static string LookupRuntimeClassName(Type type)
```
Sergio0694, line 955: Nit: is the generated file with nullability annotations? If so, the return should be `string?`.

manodasanW: Right now, it isn't.

src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
manodasanW: Rooting might be one scenario, but if I recall the main scenario I did this before was to avoid triggering the static constructors on those types unnecessarily during the lookup.

SingleAccretion: > I did this before was to avoid triggering the static constructors on those types unnecessarily during the lookup.

It is unexpected that `typeof(X)` would trigger a static constructor by itself. Where was this observed?

manodasanW: You are right, it wasn't the static constructor.  Looking back at the history, it was the module initializer when the types came from different modules.  We were registering the lookup table as part of the module initializer and when we were going through the lookup table, we triggered other module initializers that triggered registration of those lookup tables and thereby the lock we had.

https://github.com/microsoft/CsWinRT/pull/1377#issuecomment-1808497979

https://github.com/microsoft/CsWinRT/pull/1377/commits/c9b190c1381d8f1bf9159a589b4ed0cde7fc9028

FWIW, the other scenario this allows us to handle is nested private classes (i.e. `typeof(global::System.Collections.Generic.List<global::TestClass.NestedClass>)`) as that is the first issue I hit when I tried an experiment to go back to not using strings to confirm the findings from before.

Sergio0694: Ooh that makes sense, the private nested type is actually one I can see would be pretty common. I also have a few of those types in my ComputeSharp WinRT component, for some types implementing interop COM types from Win2D interfaces needed for custom effects. Perhaps at least leave a detailed comment over that code (in the generator code, not in the actual generated code) to explain all this for future reference?

Sergio0694 (approved):

PR #1578, opened 2024-04-19
Support for exception handling from unmanaged caller

### LINKED ISSUE #1571: Feature request: Exception callbacks
# Proposal: Exception callbacks

## Summary
In out-of-proc scenario, we can only see `COMException` without any message from the client if any unhandled exception happens on the server. But we cannot intercept it without wrapping every API with a try-catch block manually.
This makes it difficult to do unified unhandled exception handling and logging. 
I propose to add the support for exception callbacks which allows propagate the exception to a user specified handler to handle unhandled exceptions:

```cs
public class WinRT
{
    public event WinRTExceptionDelegate OnUnhandledException;
}

public delegate void WinRTExceptionDelegate(object? sender, WinRTExceptionEventArgs e);

public class WinRTExceptionEventArgs
{
    public WinRTExceptionEventArgs(Exception ex) => this.Exception = ex;

    public Exception Exception { get; }
    public bool Handled { get; set; }
}
```

## WinRT codegen
```cs
        [UnmanagedCallersOnly(CallConvs = new[] { typeof(CallConvStdcall) })]
        private static unsafe int Do_Abi_Foo(IntPtr thisPtr, ...)
        {
            FooClass? @this = null;
            try
            {
                @this = global::WinRT.ComWrappersSupport.FindObject<global::FooClass>(thisPtr);
                @this.Foo(...);
            }
            catch (Exception __exception__)
            {
                var args = new WinRTExceptionEventArgs(__exception__);
                WinRT.OnUnhandledException?.Invoke(@this, args);
                if (args.Handled) return 0;
                global::WinRT.ExceptionHelpers.SetErrorInfo(__exception__);
                return global::WinRT.ExceptionHelpers.GetHRForException(__exception__);
            }
            return 0;
        }
```

hez2010:
Closes #1571

Codegen example:

```cs
[UnmanagedCallersOnly(CallConvs = new[] { typeof(CallConvStdcall) })]
private static unsafe int Do_Abi_GetPoint_2(IntPtr thisPtr, global::Windows.Foundation.Point* __return_value__)
{
    global::AuthoringTest.BasicClass __this__ = default;
    global::Windows.Foundation.Point ____return_value__ = default;
    *__return_value__ = default;
    try
    {
        ____return_value__ = (__this__ = global::WinRT.ComWrappersSupport.FindObject<global::AuthoringTest.BasicClass>(thisPtr)).GetPoint();
        *__return_value__ = ____return_value__;

    }
    catch (Exception __exception__)
    {
        if (global::WinRT.ExceptionHelpers.TryHandleWinRTServerException(__this__, __exception__))
        {
            return 0;
        }
        global::WinRT.ExceptionHelpers.SetErrorInfo(__exception__);
        return global::WinRT.ExceptionHelpers.GetHRForException(__exception__);
    }
    return 0;
}
```

### DIFF
--- a/src/WinRT.Runtime/ExceptionHelpers.cs
+++ b/src/WinRT.Runtime/ExceptionHelpers.cs
@@ -116,9 +116,14 @@ namespace WinRT
 
         public static bool TryHandleWinRTServerException(object sender, Exception ex)
         {
-            UnhandledWinRTServerExceptionEventArgs args = new UnhandledWinRTServerExceptionEventArgs(ex);
-            UnhandledWinRTServerException?.Invoke(sender, args);
-            return args.Handled;
+            if (UnhandledWinRTServerException != null)
+            {
+                UnhandledWinRTServerExceptionEventArgs args = new UnhandledWinRTServerExceptionEventArgs(ex);
+                UnhandledWinRTServerException.Invoke(sender, args);
+                return args.Handled;
+            }
+
+            return false;
         }
 
         public static void ThrowExceptionForHR(int hr)
--- a/src/WinRT.Runtime/ExceptionHelpers.cs
+++ b/src/WinRT.Runtime/ExceptionHelpers.cs
@@ -98,6 +98,29 @@ namespace WinRT
             return true;
         }
 
+        /// <summary>
+        /// Unhandled WinRT server exception event.
+        /// </summary>
+        public static event EventHandler<UnhandledWinRTServerExceptionEventArgs> UnhandledWinRTServerException;
+
+        public class UnhandledWinRTServerExceptionEventArgs : EventArgs
+        {
+            public Exception Exception { get; }
+            public bool Handled { get; set; }
+
+            public UnhandledWinRTServerExceptionEventArgs(Exception exception)
+            {
+                Exception = exception;
+            }
+        }
+
+        public static bool TryHandleWinRTServerException(object sender, Exception ex)
+        {
+            UnhandledWinRTServerExceptionEventArgs args = new UnhandledWinRTServerExceptionEventArgs(ex);
+            UnhandledWinRTServerException?.Invoke(sender, args);
+            return args.Handled;
+        }
+
         public static void ThrowExceptionForHR(int hr)
         {
             if (hr < 0)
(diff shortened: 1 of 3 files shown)


src/WinRT.Runtime/ExceptionHelpers.cs
```diff
@@ -98,6 +98,29 @@ private static bool Initialize()
...
+
+            public UnhandledWinRTServerExceptionEventArgs(Exception exception)
+            {
+                Exception = exception;
+            }
+        }
+
+        public static bool TryHandleWinRTServerException(object sender, Exception ex)
+        {
+            UnhandledWinRTServerExceptionEventArgs args = new UnhandledWinRTServerExceptionEventArgs(ex);
```
Sergio0694, line 119: Definitely don't do this, this will always allocate the args even if there's no handler.

Sergio0694 (changes requested):

src/WinRT.Runtime/ExceptionHelpers.cs
hez2010: Resolved.

src/WinRT.Runtime/ExceptionHelpers.cs
```diff
@@ -98,6 +98,34 @@ private static bool Initialize()
...
+
+            public UnhandledWinRTServerExceptionEventArgs(Exception exception)
+            {
+                Exception = exception;
+            }
+        }
+
+        public static bool TryHandleWinRTServerException(object sender, Exception ex)
+        {
+            if (UnhandledWinRTServerException != null)
```
Sergio0694, line 119: Hoist this to a local to avoid the second static field fetch and also to guard against concurrent mutations.

Sergio0694 (changes requested):

src/WinRT.Runtime/ExceptionHelpers.cs
hez2010: Done.

jkotas:
```
        if (global::WinRT.ExceptionHelpers.TryHandleWinRTServerException(__this__, __exception__))
        {
            return 0;
        }
        global::WinRT.ExceptionHelpers.SetErrorInfo(__exception__);
        return global::WinRT.ExceptionHelpers.GetHRForException(__exception__);
```
I assume that this block is repeated a lot. You may want to turn into an API or a per-assembly helper to make the code smaller.

Sergio0694:
Closing this one, we don't plan on adding this feature to 2.x. Thank you! 🙂

PR #1583, opened 2024-04-22
Add 'CsWinRTEnableIDynamicInterfaceCastableSupport' feature switch

Sergio0694:
This PR adds the 'CsWinRTEnableIDynamicInterfaceCastableSupport' feature switch to allow trimming of all IDIC-related code. This would also include fixing the size regression introduced by #1566 (which can't be fixed without making the code non trim-safe).

> [!NOTE]
> Opening as draft until I get a sizoscope diff.


dongle-the-gadget:
This would break `As<T>` right?

Sergio0694:
Only if it gets to the fallback IDIC cast from `IInspectable` to the interface type, otherwise it'll still work.

manodasanW:
Note that this is just a feature switch for size improvements if you do not need the IDIC functionality.  If you do need it, you would have the feature switch on which is also the default.

dongle-the-gadget:
> Only if it gets to the fallback IDIC cast from IInspectable to the interface type, otherwise it'll still work

The only case this would be true is if the interface is using built-in COM Interop, which already doesn't support NativeAOT anyway.

Sergio0694:
Saves **52 KB** in the minimal sample, plus more incremental savings as you have more enumerable types 🚀

image

src/WinRT.Runtime/Projections/INotifyDataErrorInfo.net5.cs
```diff
@@ -33,6 +33,13 @@ public static unsafe bool get_HasErrors(IObjectReference obj)
 
         public static unsafe global::System.Collections.IEnumerable GetErrors(IObjectReference obj, string propertyName)
         {
+            if (!FeatureSwitches.IsDynamicObjectsSupportEnabled)
```
manodasanW, line 36: EnableIDynamicInterfaceCastableSupport?

Sergio0694: Oops, fixed in d3ffb5c9620ca9727ca6ed47987e96dc1c233288, thank you! Also renamed the one for dynamic objects to make it consistent with the other names, that's why I tripped up and had accidentally used it here 😅

manodasanW (approved):

PR #1585, opened 2024-04-23
Fix CCW for boxing scenarios of generic types

manodasanW:
- This addresses a couple boxing scenarios that are seen in WinUI apps.  Previously we had logic where if we detected boxing when calling a CsWinRT projected function, we ensured that the type being boxed if generic was on the lookup table.  But this didn't handle scenarios where another function was first called which is when the boxing happened and then that called a CsWinRT projected function.  This is often seen in XAML compiler generated code.  This addresses that by expanding the check for CsWinRT projected function to also include functions within the same assembly.  In theory, we should also handle cross library calls to other libraries but deferring on addressing that to later based on need.  Ideally, we also detect whether there is a CsWinRT projected function actually being called by the other method, but doing a flow analysis from initialization to various functions being called with it as a parameter to then the CsWinRT projected function being called seems not trivial to get working in a source generator.
- Refactoring to share common code
- Make RcwFallbackFactory not run when assembly version is null / 0 (i.e. CsWinRT local builds)
- Fixing blittable check to not include static fields


Sergio0694 (commented):
Looks good! Just so I understand, there's no new changes to the generated code, it's just that with these tweaks we'll end up emitting the vtable initializers in more cases than before? But eg. if you had a scenario where the previous checks were already sufficient to discover those types, then the generated code would be identical?

manodasanW:
> Looks good! Just so I understand, there's no new changes to the generated code, it's just that with these tweaks we'll end up emitting the vtable initializers in more cases than before? But eg. if you had a scenario where the previous checks were already sufficient to discover those types, then the generated code would be identical?

Right the generated code didn't change.  There are basically more scenarios we are detecting to generate the code.

Sergio0694 (approved):

PR #1589, opened 2024-04-24
Remove even more closure allocations in ObjectReferenceWithContext<T>

Sergio0694:
Builds on top of #1587 and further improves things. This drops all closures for `Context` callback invocations.


Sergio0694:
Just for context (ha!), saves 3 KB from target. Target uses 4 KB more than its own target, set is just +1 KB for everything.

image

Honestly would've still been worth it regardless given the perf/memory improvement, but still 😄

manodasanW (approved):

PR #1591, opened 2024-04-26
Seal and tweak 'IWinRTObject' members

Sergio0694:
This PR tweaks some members of `IWinRTObject` following advice from @michalstrehovsky:
- Seal all the "Fallback" methods and other helpers we only need from CsWinRT. This will avoid an interface dispatch and will turn them into normal instance methods. Also, literally nobody should have implemented these helper methods anyway.
- Also make them all `internal`, since nobody should ever use them outside of CsWinRT
- Delete the leftover and not working `GetOrCreateTypeHelperData`, once again nobody should have ever used this.


src/WinRT.Runtime/IWinRTObject.net5.cs
```diff
@@ -35,7 +35,7 @@ bool IDynamicInterfaceCastable.IsInterfaceImplemented(RuntimeTypeHandle interfac
             return IsInterfaceImplementedFallback(interfaceType, throwIfNotImplemented);
         }
 
-        bool IsInterfaceImplementedFallback(RuntimeTypeHandle interfaceType, bool throwIfNotImplemented)
+        internal sealed bool IsInterfaceImplementedFallback(RuntimeTypeHandle interfaceType, bool throwIfNotImplemented)
```
jkoritzinsky, line 38: I think these fallbacks can be called by RCW types. Can we make them `protected` or `protected internal` instead?

```suggestion
        protected sealed bool IsInterfaceImplementedFallback(RuntimeTypeHandle interfaceType, bool throwIfNotImplemented)

```

Sergio0694: > "can be called by RCW types"

Just so I understand, do you have some scenario in mind? The generated RCWs will never call those methods (and there's no references to them at all in `code_writers`, which also how the CI still passed. Can you think of a case where a user would want/need to actually manually implement an RCW type and then call those fallback methods? I can't personally think of one. And would that even be a scenario we want/need to support in the first place? 🤔

manodasanW (approved):

PR #1592, opened 2024-04-26
Set 'TreatWarningsAsErrors' almost everywhere

Sergio0694:
This PR enables 'TreatWarningsAsErrors' in all published projects, plus in the functional and authoring tests. I've also removed all AOT warnings being marked explicitly as errors, since the new property replaces this (and also treats all other warnings the same).


src/cswinrt/strings/additions/Windows.Storage.Streams/WindowsRuntimeStreamExtensions.cs
```diff
@@ -42,10 +45,13 @@ static class WindowsRuntimeStreamExtensions
...
-                                                                                                where TValue : class
+        private static void AssertMapContains<TKey, [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)] TValue>(
+            ConditionalWeakTable<TKey, TValue> map,
+            TKey key,
+            TValue value,
+            bool valueMayBeWrappedInBufferedStream)
+            where TKey : class
+            where TValue : class
         {
             TValue valueInMap;
```
manodasanW, line 56: Was this the line that needed that attribute?

manodasanW (approved):

PR #1595, opened 2024-04-29
Obsolete reflection-based 'ObjectReference' APIs

Sergio0694:
This makes them consistent with `FromAbi` not taking an IID, which is also obsolete and not safe.

### DIFF
--- a/src/WinRT.Runtime/ObjectReference.cs
+++ b/src/WinRT.Runtime/ObjectReference.cs
@@ -139,11 +139,12 @@ namespace WinRT
             Dispose();
         }
 
-        public ObjectReference<T> As<
 #if NET
-        [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicFields)]
+        [RequiresUnreferencedCode(AttributeMessages.GenericRequiresUnreferencedCodeMessage)]
+        [Obsolete(AttributeMessages.GenericDeprecatedMessage)]
+        [EditorBrowsable(EditorBrowsableState.Never)]
 #endif
-        T>() => As<T>(GuidGenerator.GetIID(typeof(T)));
+        public ObjectReference<T> As<T>() => As<T>(GuidGenerator.GetIID(typeof(T)));
 
         public unsafe ObjectReference<T> As<T>(Guid iid)
         {
--- a/src/WinRT.Runtime/ObjectReference.cs
+++ b/src/WinRT.Runtime/ObjectReference.cs
@@ -183,11 +183,12 @@ namespace WinRT
 #endif
         }
 
-        public int TryAs<
 #if NET
-            [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicFields)]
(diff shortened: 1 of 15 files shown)


src/WinRT.Runtime/ObjectReference.cs
```diff
@@ -139,11 +139,12 @@ protected IObjectReference(IntPtr thisPtr)
...
 
-        public ObjectReference<T> As<
 #if NET
-        [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicFields)]
+        [RequiresUnreferencedCode(AttributeMessages.GenericRequiresUnreferencedCodeMessage)]
+        [Obsolete(AttributeMessages.GenericDeprecatedMessage)]
+        [EditorBrowsable(EditorBrowsableState.Never)]
 #endif
-        T>() => As<T>(GuidGenerator.GetIID(typeof(T)));
+        public ObjectReference<T> As<T>() => As<T>(GuidGenerator.GetIID(typeof(T)));
```
manodasanW, line 147: For this and below, I am wondering if there is any backcompat risk by dropping the dynamicallyaccessedmembers attribute,

manodasanW: I assume the vftbl dynamically accessed attributed we put as a fallback before takes care of all scenarios here?

src/cswinrt/code_writers.h
```diff
@@ -9969,17 +9969,26 @@ public static IntPtr Make()
...
 #endif
-    I>()
+public static ObjectReference<I> ActivateInstance<I>()
 {
     IntPtr instance = _factory.ActivateInstance();
 
     return ObjectReference<IInspectable.Vftbl>.Attach(ref instance, global::WinRT.Interop.IID.IID_IInspectable).As<I>();
 }
 
+public static ObjectReference<I> ActivateInstance<I>(Guid iid)
```
manodasanW, line 9985: Is anyone calling this?

Sergio0694: No, it'd be a new API. Just figured it might make sense to add so people using the current one would have something else to migrate to that's safe and not obsolete? Though if you don't think anyone would be using either one we can also just remove both of them directly. Honestly I was wondering if we could do that in the first place.

src/WinRT.Runtime/ObjectReference.cs
Sergio0694: Which one are you thinking about? My thinking here was also that this method would not be producing a trim warning anyway, so people upgrading would still be warned. Effectively speaking you'd definitely want to use new projections in general when using trimming to get all the new improvements and bug fixes.

manodasanW: My comment was in reference to this: https://github.com/microsoft/CsWinRT/blob/0b4bc36f92961979545cbc8b3869e4ddc202d12b/src/WinRT.Runtime/TypeExtensions.cs#L163-L176

Sergio0694:
Saves 8 KB, and in particular gets rid of the `__Canon` instantiation of `ObjectReference<T>` 🎉

image

PR #1597, opened 2024-04-30
Reduce 'ObjectReference<T>' generic instantiations in WinRT.Runtime

Sergio0694:
This PR removes some generic instantiations over vtable types with `ObjectReference<T>` for types in WinRT.Runtime.


src/cswinrt/strings/additions/Windows.Storage.Streams/IMarshal.cs
```diff
@@ -210,43 +219,81 @@ private static int Do_Abi_DisconnectObject_5(IntPtr thisPtr, uint dwReserved)
...
+#else
             Marshal.ThrowExceptionForHR(_obj.Vftbl.UnmarshalInterface_3(ThisPtr, pStm, riid, ppv));
+#endif
         }
 
         public unsafe void ReleaseMarshalData(IntPtr pStm)
         {
+#if NET
+            IntPtr thisPtr = ThisPtr;
+            Marshal.ThrowExceptionForHR(((delegate* unmanaged[Stdcall]<IntPtr, IntPtr, int>)(*(void***)thisPtr)[6])(thisPtr, pStm));
```
manodasanW, line 279: indexing seems wrong starting from here.  There is 2 6.

src/WinRT.Runtime/Interop/IMarshal.net5.cs
```diff
@@ -0,0 +1,271 @@
...
+
+        public unsafe void UnmarshalInterface(IntPtr pStm, Guid* riid, IntPtr* ppv)
+        {
+            IntPtr thisPtr = ThisPtr;
+            Marshal.ThrowExceptionForHR(((delegate* unmanaged[Stdcall]<IntPtr, IntPtr, Guid*, IntPtr*, int>)(*(void***)thisPtr)[6])(thisPtr, pStm, riid, ppv));
+        }
+
+        public unsafe void ReleaseMarshalData(IntPtr pStm)
+        {
+            IntPtr thisPtr = ThisPtr;
```
manodasanW, line 261: same here and below.

src/cswinrt/strings/additions/Windows.Storage.Streams/IMarshal.cs
Sergio0694: Good catch! I assume this also means we are missing some tests?

manodasanW: I believe we have tests for the overall scenario, but maybe not this specific function.  I am not sure which specific scenario that this IMarshal function gets called in and if it even does for how we use it.  Will need to take a look.

Sergio0694: Fixed in https://github.com/microsoft/CsWinRT/pull/1597/commits/4ffd01739cf83b5c5020826a4e74f7e15905b759.

src/WinRT.Runtime/Interop/IMarshal.net5.cs
Sergio0694: Fixed in https://github.com/microsoft/CsWinRT/pull/1597/commits/942670d67698e6b3a29d2fc39eea9172413d310e.

Sergio0694:
Saves 16 KB:

image

manodasanW (approved):

PR #1598, opened 2024-05-01
Strip generated authoring metadata types

Sergio0694:
This PR strips all generated code from the `ABI.Impl` types in authoring scenarios. The types are only used for metadata lookups, and they're never actually instantiated, so they don't actually need to contain any logic to execute at runtime.

### DIFF
--- a/src/cswinrt/code_writers.h
+++ b/src/cswinrt/code_writers.h
@@ -8267,8 +8267,66 @@ return global::System.Runtime.InteropServices.CustomQueryInterfaceResult.NotHand
         auto wrapped_type_name = write_type_name_temp(w, type, "%", typedef_name_type::Projected);
         auto default_interface_name = get_default_interface_name(w, type, false, true);
 
-        // This type can be empty, as it is only used for metadata lookup, but not as implementation
-        w.write(R"(%%[global::WinRT.ProjectedRuntimeClass(typeof(%))]
+        if (settings.netstandard_compat)
+        {
+            auto base_semantics = get_type_semantics(type.Extends());
+            auto from_abi_new = !std::holds_alternative<object_type>(base_semantics) ? "new " : "";
+
+            // Fallback path for .NET Standard, with all interfaces of the user defined type implemented
+            // on the authoring metadata type as well. This is because on this target, CsWinRT will go
+            // through the list of implemented interfaces to construct the CCW and do other things. In
+            // theory, the type could be abstract and without actually providing an implementation of the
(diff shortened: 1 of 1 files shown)


src/cswinrt/code_writers.h
```diff
@@ -8266,31 +8266,16 @@ return global::System.Runtime.InteropServices.CustomQueryInterfaceResult.NotHand
...
         auto default_interface_name = get_default_interface_name(w, type, false, true);
-        auto base_semantics = get_type_semantics(type.Extends());
-        auto from_abi_new = !std::holds_alternative<object_type>(base_semantics) ? "new " : "";
 
+        // This type can be empty, as it is only used for metadata lookup, but not as implementation
         w.write(R"(%%[global::WinRT.ProjectedRuntimeClass(typeof(%))]
-%internal % partial class %%
-{
-public %(% comp)
+%internal % partial class %
```
manodasanW, line 8272: It is looking like we are going to need to a fallback where we have the interfaces for .NET Standard (given we don't generate the winrt exposed type attribute with the vtables for it today).  We happen to have someone relying on that support it seems.  On .NET Standard, we can choose to either generate this as an abstract class with the interfaces implemented with no content or we can just do if else where on .NET Standard we have the old logic and else we have your new generation code.

Sergio0694: Kept the original code to reduce risk and make things simpler, done! 🙂

src/cswinrt/code_writers.h
```diff
@@ -8266,10 +8266,19 @@ return global::System.Runtime.InteropServices.CustomQueryInterfaceResult.NotHand
...
-        w.write(R"(%%[global::WinRT.ProjectedRuntimeClass(typeof(%))]
+        if (settings.netstandard_compat)
+        {
+            auto base_semantics = get_type_semantics(type.Extends());
+            auto from_abi_new = !std::holds_alternative<object_type>(base_semantics) ? "new " : "";
+
+            // Fallback path for .NET Standard, with all interfaces of the user defined type implemented
+            // on the authoring metadata type as well. This is because on this target, CsWinRT will go
+            // through the list of implemented interfaces to construct the CCW and do other things. In
+            // theory, the type could be abstract and without actually providing an implementatio of the
```
manodasanW, line 8278:
```suggestion
            // theory, the type could be abstract and without actually providing an implementation of the
```

manodasanW (approved):

PR #1575, opened 2024-04-17
Add 'CsWinRTEnableIReferenceSupport' feature switch

Sergio0694:
This PR adds a new 'CsWinRTEnableIReferenceSupport' feature switch to trim out all `IReference*<T>` support.
Opening as draft until I get sizoscope diffs.

PR #1603, opened 2024-05-07
Add missing check for 'IReference<T>' support

Sergio0694:
Leftover from #1575 that was causing a bunch more stuff to be rooted.


Sergio0694:
Saves 112 KB, now down to 1845 KB! 🎉

image

manodasanW (approved):

PR #1604, opened 2024-05-08
Fixing couple reported issues

### LINKED ISSUE #1584: System.InvalidCastException: 'Specified cast is not valid.' on the staging/AOT branch
**Describe the bug**

```
WinRT.Runtime.dll!WinRT.ExceptionHelpers.ThrowExceptionForHR.__Throw|38_0(int hr = -2147467262) line 115
	: D:\source\repos\microsoft\CsWinRT\src\WinRT.Runtime\ExceptionHelpers.cs(115)
WinRT.Runtime.dll!WinRT.ExceptionHelpers.ThrowExceptionForHR(int hr = -2147467262) line 118
	: D:\source\repos\microsoft\CsWinRT\src\WinRT.Runtime\ExceptionHelpers.cs(118)
WinRTServerTest.dll!ABI.WinRTServer.ITestClassClassMethods.HelloAsync(WinRT.IObjectReference _obj = {WinRT.ObjectReference<WinRT.Interop.IUnknownVftbl>}, WinRTServer.BinaryDelegate func = {Method = {System.Reflection.RuntimeMethodInfo}}, int x = 24, int y = 35) line 556
	: D:\source\repos\WinRTServer\WinRTServerTest\obj\x64\Debug\net8.0-windows10.0.22621.0\win-x64\Generated Files\CsWinRT\WinRTServer.cs(556)
WinRTServerTest.dll!WinRTServer.TestClass.HelloAsync(WinRTServer.BinaryDelegate func = {Method = {System.Reflection.RuntimeMethodInfo}}, int x = 24, int y = 35) line 254
	: D:\source\repos\WinRTServer\WinRTServerTest\obj\x64\Debug\net8.0-windows10.0.22621.0\win-x64\Generated Files\CsWinRT\WinRTServer.cs(254)
WinRTServerTest.dll!Program.<Main>$(string[] args = {string[0]}) line 15
	: D:\source\repos\WinRTServer\WinRTServerTest\Program.cs(15)
WinRTServerTest.dll!Program.<Main>(string[] args = {string[0]})

Inner exception:

System.Private.CoreLib.dll!System.Runtime.InteropServices.Marshal.ThrowExceptionForHR(int errorCode) line 1259
	: System.Runtime.InteropServices\Marshal.cs(1259)
WinRT.Runtime.dll!WinRT.ComWrappersSupport.CreateCCWForObjectForABI(object obj = {System.Threading.Tasks.TaskToAsyncOperationAdapter<WinRTServer.HelloStruct>}, System.Guid iid = {System.Guid}) line 172
	: D:\source\repos\microsoft\CsWinRT\src\WinRT.Runtime\ComWrappersSupport.net5.cs(172)
WinRT.Runtime.dll!WinRT.ComWrappersSupport.CreateCCWForObjectForMarshaling(object obj = {System.Threading.Tasks.TaskToAsyncOperationAdapter<WinRTServer.HelloStruct>}, System.Guid iid = {System.Guid}) line 1165
	: D:\source\repos\microsoft\CsWinRT\src\WinRT.Runtime\ComWrappersSupport.cs(1165)
WinRT.Runtime.dll!WinRT.MarshalInspectable<Windows.Foundation.IAsyncOperation<WinRTServer.HelloStruct>>.CreateMarshaler2(Windows.Foundation.IAsyncOperation<WinRTServer.HelloStruct> o = {System.Threading.Tasks.TaskToAsyncOperationAdapter<WinRTServer.HelloStruct>}, System.Guid iid = {System.Guid}, bool unwrapObject = true) line 1789
	: D:\source\repos\microsoft\CsWinRT\src\WinRT.Runtime\Marshalers.cs(1789)
WinRT.Runtime.dll!WinRT.MarshalInterface<Windows.Foundation.IAsyncOperation<WinRTServer.HelloStruct>>.CreateMarshaler2(Windows.Foundation.IAsyncOperation<WinRTServer.HelloStruct> value = {System.Threading.Tasks.TaskToAsyncOperationAdapter<WinRTServer.HelloStruct>}, System.Guid iid = {System.Guid}) line 1619
	: D:\source\repos\microsoft\CsWinRT\src\WinRT.Runtime\Marshalers.cs(1619)
WinRT.Runtime.dll!WinRT.MarshalInterface<Windows.Foundation.IAsyncOperation<WinRTServer.HelloStruct>>.FromManaged(Windows.Foundation.IAsyncOperation<WinRTServer.HelloStruct> value = {System.Threading.Tasks.TaskToAsyncOperationAdapter<WinRTServer.HelloStruct>}) line 1647
	: D:\source\repos\microsoft\CsWinRT\src\WinRT.Runtime\Marshalers.cs(1647)
```

**To Reproduce**

1. git clone https://github.com/hez2010/WinRTServer
2. switch to your build of CsWinRT and projections

A convenient way to switch to your customized projection is:

```xml
<FrameworkReference Remove="Microsoft.Windows.SDK.NET.Ref" />
<Reference Include="Microsoft.Windows.SDK.NET">
  <HintPath>path\to\Microsoft.Windows.SDK.NET.dll</HintPath>
</Reference>
```

**Expected behavior**
No exception should be thrown.

**Version Info**
CsWinRT on staging/AOT branch: 91fca8cd846d5ed351c45c5d842abef0ad964cc7

**Additional context**

manodasanW:
- Incrementing the assembly version to reflect what it should actually be
- Fixing AsAsyncOperation and similar async scenarios where we weren't previously generating the vtable for generic classes implementing generic WinRT interfaces.  We still don't do that in general, but at least the async scenarios are now handled by using the global vtable lookup table when we detect such a function being used.
- Adding property to be able to control whether the global vtable lookup table generation happens.  This is by default disabled if we detect that you are trying to generate a projection as we don't need global vtable lookup table for those scenarios.  Otherwise, it is by default enabled.
- Fixing issue where when we were asked to generate the vtable for a boxed struct, we were failing to because we were restricting checking the CCW type to classes.
- Fixing issue where when using foreach, it was possible for an element to be get added to the list and thereby that would fail due to change in collection while doing for each.  Given we never remove from the list and only add, replacing it with a normal loop where we check count before looping.  Also doing it for the other scenarios where we were doing similar checks given those were running into a writer lock being taken while a reader lock was held.
- Fixing issue where both the component generator and the AOT generator were generating the same generic interface instantiation static classes.  We now detect that the component generator would have generated that and don't do it in the AOT generator.  This issue started showing up after fixing the issue where our collection adapter classes weren't generated for the non authoring scenarios even though we probably should.

Fixes #1584


nuget/Microsoft.Windows.CsWinRT.targets
```diff
@@ -56,6 +57,21 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <Error Text="Support for .NET 5 ended with C#/WinRT 2.0. For .NET 5 support, use C#/WinRT version 1.6.5. See https://github.com/microsoft/CsWinRT/discussions/1232" />
   </Target>
 
+  <!-- Note this runs before the msbuild editor config file is generated because that is what is used to pass properties to the source generator. -->
+  <Target Name="CsWinRTSetGeneratorProperties" BeforeTargets="GenerateMSBuildEditorConfigFile;GenerateMSBuildEditorConfigFileCore">
+    <PropertyGroup>
+      <CsWinRTCcwLookupTableGeneratorEnabled Condition="'$(CsWinRTCcwLookupTableGeneratorEnabled)' == '' and
+                                                        '$(CsWinRTComponent)' == 'true'">true</CsWinRTCcwLookupTableGeneratorEnabled>		
+      <!-- If the lookup generator is enabled, the AOT source generator generates vtable entries for generic types and boxing scenarios that it detects. -->
+	  <!-- This is intended to be ran on projections, so we try to detect that and disable it if a projection is being generated. -->
```
Sergio0694, line 66: Did you mean to _not_ be run on projections?
```suggestion
	  <!-- This is not intended to be ran on projections, so we try to detect that and disable it if a projection is being generated. -->
```

src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
```diff
@@ -53,10 +92,72 @@ private static bool NeedVtableAttribute(SyntaxNode node)
...
+            { "System.Runtime.InteropServices.WindowsRuntime.AsyncInfo.CompletedActionWithProgress<TProgress>()", "System.Threading.Tasks.TaskToAsyncActionWithProgressAdapter`1" },
+            { "System.Runtime.InteropServices.WindowsRuntime.AsyncInfo.FromExceptionWithProgress<TProgress>(System.Exception)", "System.Threading.Tasks.TaskToAsyncActionWithProgressAdapter`1" },
+            { "System.Runtime.InteropServices.WindowsRuntime.AsyncInfo.CanceledActionWithProgress<TProgress>()", "System.Threading.Tasks.TaskToAsyncActionWithProgressAdapter`1" }
+        };
+
+        // Detect if AsAsyncOperation or similar function is being called and if so,
+        // make sure the generic adapter type we use for it is on the lookup table.
+        private static VtableAttribute GetVtableAttributesForTaskAdapters(GeneratorSyntaxContext context)
+        {
+            if (context.SemanticModel.GetSymbolInfo(context.Node as InvocationExpressionSyntax).Symbol is IMethodSymbol symbol)
```
Sergio0694, line 143: This is quite expensive. We're asking for symbol info for _all_ method invocations. Fred suggested to filter this in the predicate of the syntax provider by checking the name of the method of the invocation expression. Since methods can't be aliased, we should be able to use that to exclude all methods that _for sure_ can't possibly be matches. That would cut down virtually all methods that wouldn't have matched.

src/WinRT.Runtime/ComWrappersSupport.cs
```diff
@@ -181,7 +181,8 @@ internal static List<ComInterfaceEntry> GetInterfaceTableEntries(Type type)
             // which means the attribute lives on the authoring metadata type.
             if (winrtExposedClassAttribute == null)
             {
-                var authoringMetadaType = type.GetRuntimeClassCCWType();
+                // Using GetCCWType rather than GetRuntimeClassCCWType given we want to handle boxed value types.
```
Sergio0694, line 184: Should we think about this will play with cases where `IReference<T>` support is disabled? Would thast just fail elsewhere when we actually try to add the `IReference<T>` support, or should we add a check here for that feature flag and throw right away if the type is a value type, or something else?

Sergio0694 (changes requested):

src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
manodasanW: Yea I was concerned about the same and was planning on doing some form of contains check over a list once I have the NullPointerException figured out.

src/WinRT.Runtime/ComWrappersSupport.cs
manodasanW: I think when we check the attribute, that is already going to return an empty list for structs and delegates.  So, I don't think we need a check here.  And we want to recognize here that there was indeed an attribute and not go on the fallback code path.

src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
manodasanW: Addressed.

src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
```diff
@@ -53,10 +98,96 @@ private static bool NeedVtableAttribute(SyntaxNode node)
...
+        // This is used to try to determine types which a component source generator will process.
+        private static bool IsComponentType(SyntaxNode node)
+        {
+            return node is ClassDeclarationSyntax declaration &&
+                !declaration.Modifiers.Any(m => m.IsKind(SyntaxKind.StaticKeyword) || m.IsKind(SyntaxKind.AbstractKeyword)) &&
+                declaration.Modifiers.Any(m => m.IsKind(SyntaxKind.PublicKeyword)) &&
+                !GeneratorHelper.IsWinRTType(declaration); // Making sure it isn't an RCW we are projecting.
+        }
+
+        private static (VtableAttribute, List<VtableAttribute>) GetVtableAttributeToAdd(GeneratorSyntaxContext context)
```
Sergio0694, line 111: This is used in a transform, but `List<T>` isn't equatable. Should we return an `EquatableArray<T>` instead? I assume `VtableAttribute` is already equatable?

src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
```diff
@@ -53,10 +98,96 @@ private static bool NeedVtableAttribute(SyntaxNode node)
...
+        };
+
+        // There are several async operation related methods that can be called.
+        // But they are all under the AsyncInfo static class or is the AsAsyncOperation
+        // extension method.
+        static bool IsAsyncOperationMethodCall(SyntaxNode node)        
+        {
+            if (node is InvocationExpressionSyntax method)
+            {
+                var methodStr = method.ToString();
```
Sergio0694, line 152: This is nice, but I'm wondering whether there's a more efficient way to avoid calling `ToString()` on every single invocation just to check the name. I suppose we could manually check for `IdentifierNameSyntax` or `SimpleMemberAccessSyntax`, but it feels a bit brittle. @333fred do you know if there's a better, efficient way to get the name of the method for a given `InvocationExpressionSyntax`?

src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
manodasanW: Yea I started getting inconsistent with this, will address.

manodasanW: Addressed.

src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
333fred: Walk through any member access expressions to the tokens you are about. Looking at this code, you're not handling `System.Runtime.InteropServices.WindowsRuntime.AsyncInfo` formats or `global::System.Runtime.InteropServices.WindowsRuntime.AsyncInfo`. I would always recommend using a syntax visualizer here to understand the format.

manodasanW: Thanks, I was able check the member accesses and get it working.

Sergio0694 (approved):

PR #1609, opened 2024-05-10
Remove generated 'As<I>()' methods for static types

Sergio0694:
This PR removes the generated `As<I>()` methods on static RCW types. This method is:
- Unnecessary, now that `ActivationFactory` is public
- Only works with either `[ComImport]` types, or via IDIC
- It doesn't exist on non-static RCW types, which is not consistent (eg. Win2D uses that)

Devs should just use `ActivationFactory` directly, which also exposes new high perf methods.

### DIFF
--- a/src/cswinrt/code_writers.h
+++ b/src/cswinrt/code_writers.h
@@ -2501,8 +2501,6 @@ Marshal.Release(inner);
 
     void write_attributed_types(writer& w, TypeDef const& type)
     {
-        bool factory_written{};
-
         for (auto&& [interface_name, factory] : get_attributed_types(w, type))
         {
             if (factory.activatable)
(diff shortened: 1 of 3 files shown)


hez2010:
Isn't it an ABI breaking change and breaks all existing apps?

src/cswinrt/code_writers.h
```diff
@@ -2513,43 +2511,6 @@ Marshal.Release(inner);
             {
                 write_composable_constructors(w, factory.type, type, factory.visible ? "public"sv : "protected"sv);
             }
-            else if (factory.statics)
```
manodasanW, line 2516: We probably can't stop generating these functions due to back compat.  When we generate the Windows SDK projection for instance, the backcompat check there will break.

Sergio0694:
Closing this too, moving to 3.0.

PR #1610, opened 2024-05-12
Fix potential type incompatibility for in-proc WinRT components

### LINKED ISSUE #1510: Unable to cast object of type 'TaskToAsyncOperationAdapter' to type 'IAsyncOperation'
**Describe the bug**
I'm authoring an in-process WinRT server, and it has the following method:

```cs
public struct Data
{
    public int X;
}

class Foo
{
    public IAsyncOperation<Data> GetDataAsync()
    {
        return AsyncInfo.Run(async (_) => 
        {
            await Task.Delay(1000);
            return new Data { X = 42 };
        });
    }
}
```

**To Reproduce**
Create a client that call into the server method:

```cs
var foo = new Foo();
await foo.GetDataAsync();
```

You will get:

```
Unhandled exception. System.InvalidCastException: Unable to cast object of type 'System.Threading.Tasks.TaskToAsyncOperationAdapter`1[WinRTServer.Data]' to type 'Windows.Foundation.IAsyncOperation`1[WinRTServer.Data]'.
   at WinRT.MarshalInspectable`1.FromAbi(IntPtr ptr)
   at WinRT.MarshalInterface`1.FromAbi(IntPtr ptr)
   at ABI.WinRTServer.IFooClassMethods.GetDataAsync(IObjectReference _obj)
   at WinRTServer.Foo.GetDataAsync()
```

Note that the `WinRTServer.Data` in `TaskToAsyncOperationAdapter<WinRTServer.Data>` comes from the server, while the one in `IAsyncOperation<WinRTServer.Data>` comes from the client.

This issue doesn't reproduce if you are using an out-of-process WinRT server. It only reproduces for in-process WinRT server.

A simple repro can be found here (checkout the repro branch): https://github.com/hez2010/WinRTServer/tree/repro

**Expected behavior**
No exception being thrown. 

**Version Info**
CsWinRT 2.0.7

hez2010:
We need to create a new RCW for this case.

Fixes #1510.


hez2010:
Can confirm the issue has been fixed:
image

src/cswinrt/strings/additions/Windows.Foundation/Windows.Foundation.cs
```diff
@@ -302,12 +302,12 @@ internal void CompleteFromAsyncActionWithProgress(IAsyncActionWithProgress<TProg
 
         internal void CompleteFromAsyncOperation(IAsyncOperation<TResult> asyncInfo, AsyncStatus asyncStatus)
         {
-            Complete(asyncInfo, ai => ((IAsyncOperation<TResult>)ai).GetResults(), asyncStatus);
+            Complete(asyncInfo, ai => ai.As<IAsyncOperation<TResult>>().GetResults(), asyncStatus);
```
Sergio0694, line 305: This `As<T>` overload is deprecated. You need to pass an IID.

src/WinRT.Runtime/Marshalers.cs
```diff
@@ -1304,7 +1304,20 @@ public static T FromAbi(IntPtr ptr)
...
                 if (IUnknownVftbl.IsReferenceToManagedObject(iunknownPtr))
                 {
-                    return (T)ComWrappersSupport.FindObject<object>(iunknownPtr);
+                    // We use a global instance of ComWrappers, but it's possible to use different projections of the same type
+                    // when both server and client are managed and in the same process.
+                    // In this case, we need to check if the object is of the same type as the one we're trying to cast to.
+                    // If it's not, we need to create an RCW for the object.
+                    // We cannot use T directly here as it may lead to invalid cast due to the above reason.
+                    var obj = ComWrappersSupport.FindObject<object>(iunknownPtr);
+                    if (typeof(T).IsAssignableFrom(obj.GetType()))
```
Sergio0694, line 1313: Why not `is obj is T`?

Sergio0694 (changes requested):

src/cswinrt/strings/additions/Windows.Foundation/Windows.Foundation.cs
manodasanW: For IID, it would be like IAsyncOperationMethods<TResult>.IID.

hez2010: Here it's calling into https://github.com/microsoft/CsWinRT/blob/9854573b9880650efd87c5be41a553d125877c92/src/WinRT.Runtime/CastExtensions.cs#L29. I don't see where it is deprecated.

hez2010: Update: the change here is unnecessary. it's a leftover when I was debugging the issue.
Reverting the commit.

hez2010:
PTAL.

manodasanW:
I can see this being a feasible solution to this problem with different projections.  Going to run the CI pipelines to make sure everything continues to pass.

hez2010:
Do you think we should handle this in the projection as well?
i.e. turning all `global::WinRT.ComWrappersSupport.FindObject<T>(thisPtr)` into `global::WinRT.ComWrappersSupport.FindObject<T>(thisPtr) is T __obj__ ? __obj__ : global::WinRT.ComWrappersSupport.CreateRcwForComObject<T>(thisPtr)`

manodasanW:
> Do you think we should handle this in the projection as well? i.e. turning all `global::WinRT.ComWrappersSupport.FindObject<T>(thisPtr)` into `global::WinRT.ComWrappersSupport.FindObject<object>(thisPtr) is T __obj__ ? __obj__ : global::WinRT.ComWrappersSupport.CreateRcwForComObject<T>(thisPtr)`

All of the `FindObject` calls in the generated projections are part of the `Do_ABI_` implementations of the various interfaces.  I feel like if we are ending up there, we have already decided that is the right interface for the call.  If it was for another projection, I feel like it would end up on the helper type for that interface from that projection.  Until we find a scenario that impacts those, I would hold off on updating those.

manodasanW (approved):

manodasanW:
CI passed, thanks for this change.

PR #1611, opened 2024-05-12
Fixing issues from GetObjectReferenceFromInterface change

manodasanW:
- It seems that the `IUnknown` interface and the default interface are the same in our test component scenarios.  This caused for an issue introduced by the `GetObjectReferenceFromInterface` change to go unnoticed where we were triggering a QI to `IUnknown` and then using it as if it were the default interface.  Added a test to catch this scenario with an explicit `IUnknown` and also updated the projections to call a new version of `GetObjectReferenceForInterface` which can be passed the IID and whether a QI is needed or is already done.
- Went through all the `GetObjectReferenceForInterface` calls and made sure the correct IID is being passed and that we are avoiding QIs where we can.  In some cases, such as the `ObjRefFromABI` scenarios, the update to the IID doesn't probably make any impact but updated it for correctness.
- Had some previous tests that weren't checked in, adding those here.
- Added a new `ComWrappersHelper.Init` to allow the IID to be passed so we can have the correct IID to refer to and also avoid an additional QI.  Avoid the QI is important here to make sure we hold onto the right inner from aggregation.  The IID is passed for correctness but doesn't impact much as we typically do a QI with the result anyways.  In the future in case we do further optimizations here, making sure the correct IID is being passed.
- Reducing the scenarios where we call `GuidGenerator.GetIID` function to get to a point where we can remove our dependency on the IID optimizer tool that runs post build.


src/WinRT.Runtime/ComWrappersSupport.cs
```diff
@@ -113,7 +113,10 @@ internal unsafe static bool IsFreeThreaded(IObjectReference objRef)
 
         public static IObjectReference GetObjectReferenceForInterface(IntPtr externalComObject)
         {
-            return GetObjectReferenceForInterface<IUnknownVftbl>(externalComObject, IID.IID_IUnknown);
+            // Here the ptr itself might not point to IUnknown, but we are using IUnknown for the purposes of getting
+            // an agile reference if needed.  Due to that and to keep back compat, making sure to not trigger a QI
+            // In addition, the ptr is already pointing to the correct interface which the IObjectReference is expected for.
+            return GetObjectReferenceForInterface<IUnknownVftbl>(externalComObject, IID.IID_IUnknown, false);
```
Sergio0694, line 119: Nit: I'd always explicitly use `requireQI: true` (or `false`) at these callsites just for extra clarity.

Sergio0694, line 119: Question: in theory, like in this case, could we not _always_ skip the QI for `IUnknown`? I mean, since any COM interface will always extend `IUnknown` regardless. As in, unless we specifically want to QI just because we want to test for object identity (in which case `IUnknown` is the only way to do that), isn't that always not needed? Just trying to understand this better.

src/WinRT.Runtime/ComWrappersSupport.cs
```diff
@@ -134,6 +137,11 @@ public static ObjectReference<T> GetObjectReferenceForInterface<T>(IntPtr extern
             return GetObjectReferenceForInterface<T>(externalComObject, iid, true);
         }
 
+        public static IObjectReference GetObjectReferenceForInterface(IntPtr externalComObject, Guid iid, bool requireQI)
```
Sergio0694, line 140: Nit: it'd be nice to add proper XML docs to these APIs, and in particular for this one we might want to call out in the remarks that callers have the responsibility of passing valid `iid` values for the input objects, otherwise the behavior will be weird. That is, because if you pass `requireQI: false`, the call will still succeed, but then you would get a failure potentially much later on when the implementation is trying to use that `iid` to get an agile reference from somewhere else entirely.

src/WinRT.Runtime/ComWrappersSupport.net5.cs
```diff
@@ -364,9 +364,25 @@ class ComWrappersHelper
             object thisInstance,
```
Sergio0694, line 364: TIL the `ComWrappersHelper` type exists. Weird one.
- Why is this type not static?
- Why does this exist at all? These methods feel like they should just be factory methods on `IObjectReference`:
```csharp
static IObjectReference CreateUnsafe(
    bool isAggregation,
    object thisInstance,
    IntPtr newInstance,
    IntPtr inner,
    Guid iid);
```

src/WinRT.Runtime/Interop/IActivationFactory.cs
```diff
@@ -70,13 +71,45 @@ public static unsafe IObjectReference ActivateInstanceUnsafe(IObjectReference ob
...
+            IntPtr thisPtr = objectReference.ThisPtr;
+            IntPtr instancePtr;
+
+            ExceptionHelpers.ThrowExceptionForHR((*(delegate* unmanaged[Stdcall]<IntPtr, IntPtr*, int>**)thisPtr)[6](thisPtr, &instancePtr));
+
+            GC.KeepAlive(objectReference);
+
+            try
+            {
+                return ComWrappersSupport.GetObjectReferenceForInterface<IUnknownVftbl>(instancePtr, iid, true);
```
Sergio0694, line 105: Just so I understand: how come we're always requiring QI on .NET, but not downlevel? Can we never skip it here?

src/WinRT.Runtime/Projections/EventHandler.cs
```diff
@@ -323,13 +325,14 @@ static unsafe EventHandler()
...
             AbiToProjectionVftablePtr = nativeVftbl;
+            ComWrappersSupport.RegisterDelegateFactory(typeof(global::System.EventHandler), CreateRcw);
         }
 
 #if !NET
         public static global::System.Delegate AbiInvokeDelegate { get; }
 #endif
 
-        private static readonly Guid IID = new(0xc50898f6, 0xc536, 0x5f47, 0x85, 0x83, 0x8b, 0x2c, 0x24, 0x38, 0xa1, 0x3b);
+        public static Guid IID { get; } = global::WinRT.Interop.IID.IID_EventHandler;
```
Sergio0694, line 335: We don't need the duplicate storage here, we can probably optimize it:
```suggestion
        public static Guid IID => global::WinRT.Interop.IID.IID_EventHandler;

```

src/WinRT.Runtime/Projections/NotifyCollectionChangedEventHandler.cs
```diff
@@ -40,12 +40,13 @@ static unsafe NotifyCollectionChangedEventHandler()
...
             *(global::WinRT.Interop.IDelegateVftbl*)nativeVftbl = AbiToProjectionVftable;
-            AbiToProjectionVftablePtr = nativeVftbl;
+            AbiToProjectionVftablePtr = nativeVftbl;
+            ComWrappersSupport.RegisterDelegateFactory(typeof(global::System.Collections.Specialized.NotifyCollectionChangedEventHandler), CreateRcw);
         }
 
         public static global::System.Delegate AbiInvokeDelegate { get; }
 
-        private static readonly Guid IID = new(0x8B0909DC, 0x2005, 0x5D93, 0xBF, 0x8A, 0x72, 0x5F, 0x01, 0x7B, 0xAA, 0x8D);
+        public static Guid IID { get; } = global::WinRT.Interop.IID.IID_NotifyCollectionChangedEventHandler;
```
Sergio0694, line 49: Same here. If we return the RVA field directly the JIT can inline this across calls too:
```suggestion
        public static Guid IID => global::WinRT.Interop.IID.IID_NotifyCollectionChangedEventHandler;
```

src/WinRT.Runtime/Projections/PropertyChangedEventHandler.cs
```diff
@@ -39,12 +39,13 @@ static unsafe PropertyChangedEventHandler()
...
             *(IDelegateVftbl*)nativeVftbl = AbiToProjectionVftable;
-            AbiToProjectionVftablePtr = nativeVftbl;
+            AbiToProjectionVftablePtr = nativeVftbl;
+            ComWrappersSupport.RegisterDelegateFactory(typeof(global::System.ComponentModel.PropertyChangedEventHandler), CreateRcw);
         }
 
         public static global::System.Delegate AbiInvokeDelegate { get; }
 
-        private static readonly Guid IID = new(0xE3DE52F6, 0x1E32, 0x5DA6, 0xBB, 0x2D, 0xB5, 0xB6, 0x09, 0x6C, 0x96, 0x2D);
+        public static Guid IID { get; } = global::WinRT.Interop.IID.IID_PropertyChangedEventHandler;
```
Sergio0694, line 48: Same here:
```suggestion
        public static Guid IID => global::WinRT.Interop.IID.IID_PropertyChangedEventHandler;
```

src/cswinrt/code_writers.h
```diff
@@ -2178,17 +2180,46 @@ ComWrappersSupport.RegisterObjectForInterface(this, ThisPtr);
...
-                settings.netstandard_compat ? "" : "ComWrappersHelper.Init(_inner, false);");
+                    class_type.TypeName(),
+                    has_base_type ? ":base(global::WinRT.DerivedComposed.Instance)" : "",
+                    objrefname,
+                    bind<write_type_name>(default_type_semantics, typedef_name_type::StaticAbiClass, true),
+                    bind([&](writer& w)
+                    {
+                        if (is_manually_gen_default_interface)
+                        {
+                            w.write("_defaultLazy = new Lazy<%>(() => (%)new SingleInterfaceOptimizedObject(typeof(%), _inner));", default_interface_name, default_interface_name, default_interface_name);
```
Sergio0694, line 2219: Not blocking on this PR, but I think we really should optimize these paths later. We're generating a bunch of lambdas, captures, and some overhead for `Lazy<T>`, etc. We can probably make this simpler and leaner.

src/cswinrt/code_writers.h
```diff
@@ -9037,11 +9074,33 @@ private static unsafe int Do_Abi_Invoke%
...
+                    {
+                        w.write(R"(
+public static global::System.Guid IID{ get; } = new Guid(new byte[]{ % });
+)",
+                            bind<write_guid_bytes>(type));
+                    }
+                    else
+                    {
+                        w.write(R"(
+public static global::System.Guid IID { get; } = new Guid(new global::System.ReadOnlySpan<byte>(new byte[] { % }));
```
Sergio0694, line 9099: Why not generate the same code as in out T4 file for GUIDs, to directly return RVA fields as IIDs?

src/cswinrt/code_writers.h
```diff
@@ -9037,11 +9074,33 @@ private static unsafe int Do_Abi_Invoke%
...
-                if (type_params.empty()) return;
-                w.write(R"(
-public static Guid PIID = GuidGenerator.CreateIID(typeof(%));)",
-                    type_name
-                );
+                if (!type_params.empty())
+                {
+                    // Generating both PIID and IID for backcompat consistency and to have a common property to rely on
+                    // similar to what we do in the manual projections.
+                    w.write(R"(
```
Sergio0694, line 9081: Not sure I get it, why do we need consistency if this is only for new projections? And why do we need two in the first place rather than just IID even for the generic case? 🤔

src/WinRT.Runtime/Interop/IActivationFactory.cs
manodasanW: In both scenarios, there is a QI given we get `IInspectable`.  We are just optimizing here by doing it earlier.  On !NET, we will have a QI done right after this when it goes into the default interface constructor.  I just had an opportunity to reduce an object allocation on NET by doing the QI here and thereby simplifying the code in the consumer.

src/cswinrt/code_writers.h
manodasanW: There are only 2 interfaces using this scenario now, I will probably be looking at removing them if possible later when I get to it.  I already had a change I was working on for this, will need to get back to it at some point.

src/cswinrt/code_writers.h
manodasanW: PIID was pre-existing, so if there was a previously generated projection that had a dependency on the old version of this, removing it can break it.  At the same time, I have been generalizing across all the interfaces having an IID property that they can call to get the GUID rather than checking if it is a generic or not first and switching between PIID and IID.  Due to that, I added both here.

src/WinRT.Runtime/ComWrappersSupport.net5.cs
manodasanW: This type probably should have been static, I think we just missed doing that when we added it.  It is an interesting proposal whether it should have been on IObjectReference, but we had kept it separate as it was meant to a set of helper methods to help deal with WinUI reference tracking setup.  That was why it was put under a helper class.  For now, leaving it as is.

src/WinRT.Runtime/ComWrappersSupport.cs
Sergio0694: Oh, nono, it was just me double checking that I was getting things right. As in, for the future, can we skip the QI in all cases where don't specifically need to check for object identity, or are there some more gotchas to keep in mind? Basically I'm trying to understand when exactly I'd also have to be careful if I were adding some new code that needed this API, to set that `requireQI` to `true` 🙂

src/cswinrt/code_writers.h
Sergio0694: Oh that's awesome, sounds good!

src/WinRT.Runtime/ComWrappersSupport.cs
manodasanW: Addressed.

src/WinRT.Runtime/ComWrappersSupport.cs
manodasanW: Added some docs here.

src/cswinrt/code_writers.h
manodasanW: Yea I think that makes sense, updated both places where we generate similar IID to be in the same format.

PR #1619, opened 2024-05-15
Trim 'ABI.System.EventHandler' when not needed

Sergio0694:
This PR enables fully trimming `ABI.System.EventHandler` when not needed, making it fully pay-for-play.


Sergio0694:
Can confirm it fully trims out `EventHandler` and related stuff (saves ~3 KB):

image

manodasanW (approved):

PR #1625, opened 2024-05-22
Fix 'CsWinRTIncludes' check in .targets file

Sergio0694:
Small follow up to #1604. This PR fixes the .targets file to make sure that `CsWinRTCcwLookupTableGeneratorEnabled` is correctly set to `false` in cases where `CsWinRTInclude` is set to some `string` value, including multilines. This also fixes Win2D.


dongle-the-gadget:
There's also `CsWinRTParam`, so make sure to check that.

Sergio0694:
That property wasn't being checked before this PR either, so I just kept that behavior as is.

manodasanW (approved):

PR #1626, opened 2024-05-27
Minor changes to IIDOptimizer and AOTOptimizer

manodasanW:
- One of our recent changes to convert the IID properties in the methods class to the new format where it is a ref caused the AOT optimizer to detect it as a boxing scenario (going from byte[] to readonlyspan) when it isn't.  Adding logic to detect that.
- Removing misleading WinRT.Runtime message from IIDOptimizer which caused confusions to folks when they see it
- Fixing IIDOptimizer to only write DLL if it patched something as I have seen in updated projections that it doesn't patch anything anymore in most cases as expected given we use the IID instead, but it is still writing its internal implementation type when not needed.


Sergio0694 (approved):
Nice!

PR #1628, opened 2024-05-31
Fix marshaling in AOT source generator

manodasanW:
- When we were marshaling bool, we were previously trying to cast to a byte from bool or vice versa.  This results in a compiler error.  Addressing it by replacing it with a check for != 0 or check for true / false.
- Array marshaling was also broken for bool, fixing it by making it use MarshalNonBlittable rather than trying to find the ABI type for bool.


Scottj1s (approved):

src/Authoring/WinRT.SourceGenerator/Helper.cs
```diff
@@ -603,7 +603,14 @@ public static string GetMarshalerClass(string type, string abiType, TypeKind kin
                 }
                 else
                 {
-                    return "global::ABI." + type;
+                    if (isArray)
+                    {
+                        return $$"""MarshalNonBlittable<{{type}}>""";
```
Sergio0694, line 608: Just so I understand, this doesn't seem to be specific to `bool` values, so won't this change cause _all_ arrays of value types that are not the same as their ABI types to use `MarshalNonBlittable`? ...Is this intended? 🤔

manodasanW: Yea I was making it match what code_writers did: https://github.com/microsoft/CsWinRT/blob/6257088ab16ce74bf04726c5062c606d77c0ca8f/src/cswinrt/code_writers.h#L4370

Sergio0694 (approved):

src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
```diff
@@ -836,7 +837,18 @@ private static bool NeedVtableOnLookupTable(SyntaxNode node)
                     node is ReturnStatementSyntax;
         }
 
-        private static EquatableArray<VtableAttribute> GetVtableAttributesToAddOnLookupTable(GeneratorSyntaxContext context)
+        private static (EquatableArray<VtableAttribute>, EquatableArray<VtableAttribute>) GetVtableAttributesToAddOnLookupTable(GeneratorSyntaxContext context)
+        {
+            // Get the lookup table assuming we are running in an authoring scenario and not in an authoring scenario
```
Sergio0694, line 842: > "assuming we are running in an authoring scenario and not in an authoring scenario"

I assume this is a typo? 😄

manodasanW: Not really, we get the lookup table as if this was potentially an authoring component and also as if we weren't.  At this point, we don't know which one it is as we can't consult properties.  So, we return for both scenarios and then choose the right one based on properties when we can actually check it.  Updated the comment a bit to reflect that.

PR #1637, opened 2024-06-11
Add 'CsWinRTMergeReferencedActivationFactories' option

Sergio0694:
- Add the new `[WinRTAssemblyExportsType]` attribute
- Generate a type with a unique name referenced by it, in each authored component
- Add the `CsWinRTMergeReferencedActivationFactories` option to merge all activation factories


Sergio0694:
Seems to work fine when testing locally 🙂

<details>
    <summary><b>Disabled (click to expand):</b></summary>
<br/>

image

image

image
</details>

<details>
    <summary><b>Enabled (click to expand):</b></summary>
<br/>

image
</details>

src/Authoring/WinRT.SourceGenerator/Generator.cs
```diff
@@ -269,6 +276,40 @@ public static int DllCanUnloadNow()
...
+        /// Generates the native exports for a WinRT component.
+        /// </summary>
+        /// <param name="context">The <see cref="GeneratorExecutionContext"/> value to use to produce source files.</param>
+        public static void GenerateWinRTExportsType(GeneratorExecutionContext context)
+        {
+            if (context.Compilation.AssemblyName is not { Length: > 0 } assemblyName)
+            {
+                return;
+            }
+
```
nagya, line 290: I recommend referring to the exact set of allowed characters in assembly vs namespace names and transforming assemblyName accordingly here. For example, '-' is valid in assembly names, but not namespace names, which has resulted in similar bugs before, e.g. https://github.com/OmniSharp/generator-aspnet/issues/256

Sergio0694: Yup, just asked @333Fred about this as well (he mentioned II.6.2 and II.5.2/3 of ECMA 335), I'll push a fix for this 👍

src/Authoring/WinRT.SourceGenerator/MergeReferencedActivationFactoriesGenerator.cs
```diff
@@ -0,0 +1,199 @@
...
+                            /// <param name="fullyQualifiedTypeName">The marshalled fully qualified type name of the activation factory to retrieve.</param>
+                            /// <returns>The pointer to the activation factory that corresponds with the class specified by <paramref name="fullyQualifiedTypeName"/>.</returns>
+                            internal static unsafe IntPtr TryGetDependentActivationFactory(ReadOnlySpan<char> fullyQualifiedTypeName)
+                            {                  
+                    """);
+
+                foreach (string assemblyExportsTypeName in assemblyExportsTypeNames)
+                {
+                    builder.AppendLine($$"""
+                                    IntPtr obj = global::{{assemblyExportsTypeName}}.GetActivationFactory(fullyQualifiedTypeName);
```
manodasanW, line 125: Is repeating this going to hit multiple variable declaration of the same name?

src/Authoring/WinRT.SourceGenerator/MergeReferencedActivationFactoriesGenerator.cs
```diff
@@ -0,0 +1,199 @@
...
+
+                        /// <inheritdoc cref="Module"/>
+                        partial class Module
+                        {
+                            /// <summary>
+                            /// Tries to retrieve the activation factory from all dependent WinRT components.
+                            /// </summary>
+                            /// <param name="fullyQualifiedTypeName">The marshalled fully qualified type name of the activation factory to retrieve.</param>
+                            /// <returns>The pointer to the activation factory that corresponds with the class specified by <paramref name="fullyQualifiedTypeName"/>.</returns>
+                            internal static IntPtr TryGetDependentActivationFactory(ReadOnlySpan<char> fullyQualifiedTypeName)
```
manodasanW, line 90: I feel like the code in the else would of just handled this scenario just fine as the loop just wouldn't run.  Or did I miss any other differences between both?

src/Authoring/WinRT.SourceGenerator/MergeReferencedActivationFactoriesGenerator.cs
Sergio0694: Ah, good catch! Fixed 🙂

src/Authoring/WinRT.SourceGenerator/MergeReferencedActivationFactoriesGenerator.cs
Sergio0694: I think I added this as an optimization in the default case, so we don't need to create the intermediate builder. But then again in the default case, this output node will only ever run once, so it's likely not needed. Removed!

manodasanW (approved):

PR #1640, opened 2024-06-15
Add WinRT.StubExe project and .targets to support WinRT activation for apps

Sergio0694:
#### Depends on: https://github.com/dotnet/runtime/pull/103504

This PR includes two changes:
- Adds a new WinRT.StubExe project to produce a native .exe that will invoke a custom `Main` from applications, which will leverage the new `CustomNativeMain` functionality from .NET 9 (see related PR there). This enables apps to both be executed normally, as well as act as implementation .dll for WinRT components (eg. for OOP background tasks).
- Some .targets changes to wire things up when publishing (configuring NativeAOT props, bundling the stub .exe, etc.)

> [!NOTE]
> Opening as a draft as things are still very much up in the air at this point.


src/Authoring/WinRT.StubExe/WinRT.StubExe.cpp
```diff
@@ -0,0 +1,52 @@
...
+    FARPROC pEntryPoint = GetProcAddress(hModule, "__managed__Main");
+
+    if (!pEntryPoint)
+    {
+        return GetLastError();
+    }
+
+    // We need to parse the arguments to get back 'argc', 'argv' for the managed entry point
+    int argc;
+    LPWSTR* argv = CommandLineToArgvW(lpCmdLine, &argc);
```
jkotas, line 48: This is broken legacy API with number of corner-case bugs: https://github.com/dotnet/runtime/issues/77644

Is there no way to convince C++ compiler to build WinRT app with regular C main?

src/Authoring/WinRT.StubExe/WinRT.StubExe.vcxproj
```diff
@@ -0,0 +1,104 @@
...
+  <Import Project="$(VCTargetsPath)\Microsoft.Cpp.props" />
+  <ImportGroup Label="ExtensionSettings">
+  </ImportGroup>
+  <ImportGroup Label="Shared">
+  </ImportGroup>
+  <ImportGroup Label="PropertySheets">
+    <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
+  </ImportGroup>
+  <PropertyGroup Label="UserMacros" />
+  <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
```
jkotas, line 54: This looks redundant with `$(Configuration)' == 'Debug'` below

src/Authoring/WinRT.StubExe/WinRT.StubExe.vcxproj
```diff
@@ -0,0 +1,104 @@
...
+      <IntrinsicFunctions>true</IntrinsicFunctions>
+      <SDLCheck>true</SDLCheck>
+      <PreprocessorDefinitions>NDEBUG;_WINDOWS;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <ConformanceMode>true</ConformanceMode>
+    </ClCompile>
+    <Link>
+      <SubSystem>Windows</SubSystem>
+      <EnableCOMDATFolding>true</EnableCOMDATFolding>
+      <OptimizeReferences>true</OptimizeReferences>
+      <GenerateDebugInformation>true</GenerateDebugInformation>
```
jkotas, line 91: I think you will need to enable a more options here to pass https://github.com/microsoft/binskim

src/Authoring/WinRT.StubExe/WinRT.StubExe.cpp
dongle-the-gadget: Sergio is using the GUI subsystem here, which uses WinMain rather than C main.

Sergio0694: Oh, TIL. Technically speaking this stub .exe doesn't have to be a WinRT app whatsoever (it literally just needs to be some native executable jumping into that custom main, that's it, it doesn't do anything WinRT-related on its own, nor does it have to know about it even). I just created a C++ project as my understanding was that compiling flat C required a bit of trickery compared to vanilla C++, but if there's a simple way to do that, absolutely, we could do that! I was mostly just prototyping things here so far 🙂

dongle-the-gadget: Apparently you can change the `/ENTRY` linker flag to `mainCRTStartup` to use C main.

https://learn.microsoft.com/en-us/cpp/build/reference/entry-entry-point-symbol?view=msvc-170

Sergio0694: Ooh good find! Done in 6d4caaeb4ad781d9ba122794e8adf2222ab7e515 😄

src/Authoring/WinRT.StubExe/WinRT.StubExe.vcxproj
Sergio0694: Ah, yeah, leftover, forgot to remove this one. Thank you! 🙂

Sergio0694:
Spoke at length with @michalstrehovsky, @jkoritzinsky and @AaronRobinsonMSFT, and I think the conclusion is that the simplest/best way to do this is to port over some code from [DNNE](https://github.com/AaronRobinsonMSFT/DNNE/) and have a custom MSBuild task just invoke `cl.exe` to build/link a C program just forwarding to the custom main. This will keep things simple enough (and all the code we need is already there in DNNE), will let us have a stub .exe directly linking to the custom main, without the need for `LoadLibrary` at runtime or anything else, and it will also allow us to customize the .exe when needed in a single step, by just passing additional arguments to the linker (eg. `/APPCONTAINER`). Going to close this one 🙂

PR #1646, opened 2024-06-20
Fix boxing of types

manodasanW:
We were able to box types from C# to native, but getting box types from native in C# didn't get properly unboxed.  In addition, if we had a collection of types, we failed to properly generate the IID for it on AOT due to not finding the helper type.  Addressing both of those issues and adding some tests.


src/WinRT.Runtime/Projections.cs
```diff
@@ -41,9 +41,7 @@ static Projections()
             // They're also very cheap anyway and commonly used, so this keeps things simpler.
             RegisterCustomAbiTypeMappingNoLock(typeof(bool), typeof(ABI.System.Boolean), "Boolean");
             RegisterCustomAbiTypeMappingNoLock(typeof(char), typeof(ABI.System.Char), "Char");
-
-            // Also always register Type, since it's "free" (no associated ABI type to root)
-            CustomTypeToAbiTypeNameMappings.Add(typeof(Type), "Windows.UI.Xaml.Interop.TypeName");
+            RegisterCustomAbiTypeMappingNoLock(typeof(Type), typeof(ABI.System.Type), "Windows.UI.Xaml.Interop.TypeName");
```
Sergio0694, line 44: Ah, this makes sense, I've always wondered why `Type` wasn't registering any ABI type at all here. However, if this is not longer free but is now rooting an ABI type preserving a bunch of stuff, we should likely move this to the group of custom ABI types. There's a T4 template for it, so we don't really need to write code manually. I would recommend not doing this until the WUX PR is merged, since that's adding a bunch of changes there and we'd likely end up with a bunch of hard to resolve conflicts (because of all the generated code).

If we want to merge this one before that, that's also fine though, I can move this there after the WUX PR is in.

Sergio0694 (approved):

src/WinRT.Runtime/Projections.cs
manodasanW: Yea I was debating whether there is more being rooted, I feel like everything this is rooting technically is also rooted by Marshaler< T > given it has a static constructor for `type`.  But I haven't check yet whether that was indeed rooting things or if that was good enough to detect Marshaler< T > isn't used with type as T and doesn't need to be kept.

Sergio0694: I can run sizoscope before and after these changes too for good measure. Will report back 👍

Sergio0694: It roots ~10 KB, and pulls in a bunch of types and instantiations that weren't here before:

image

I think there would be no harm in making this optional in the same way all other custom type mappings work? It would also be consistent, given that with these changes, `Type` also effectively becomes a special type with custom mapping.

manodasanW: That is interesting.  Yes we can do the custom mappings similar to how others work.  I do also have another thing which I can do is go back to doing what we had before and then just add it to the other map too manually and add a special case check in GetSignature for type to return its signature which is the main reason why I added the function call to `RegisterCustomAbiTypeMappingNoLock`.  The other scenario which is what is making it kept for reflection isn't really needed given `Type` is special cased already in Marshaler< T >

manodasanW: Latest changes should make it mostly similar to before.

Sergio0694: Looks good! Just so I understand though, how come for `Type` specifically we're doing this weird "halfway" combination of manually registering the custom type mappings in both directions but without the marshaller type, and hardcoding the signature in `GetIID()`, rather than just doing the same thing we're doing for all other custom mapping types?

manodasanW: Right, due to the hard coding in `Marshaler<T>` for Type, the helper type is unused and Type is already special cased in many different places, so this is just adding it to the one place it wasn't and we get the savings on not needing to mark the ABI type for type as dynamically reflectable.

PR #1648, opened 2024-06-25
Fixing various AOT bugs

### LINKED ISSUE #1631: CsWinRT doesn't generate code for IMapChangedEventArgs
The below code make CsWinRT produce invalid code:

```cs
public partial class ObservableDictionaryChangedEventArgs : IMapChangedEventArgs<string>
{
    public ObservableDictionaryChangedEventArgs(CollectionChange change, string key)
    {
        this.CollectionChange = change;
        this.Key = key;
    }

    public CollectionChange CollectionChange { get; private set; }
    public string Key { get; private set; }
}
```

where CsWinRT generates code for `ObservableDictionaryChangedEventArgs `:

```cs
using static WinRT.TypeExtensions;

namespace App1
{
[global::WinRT.WinRTRuntimeClassName("Windows.Foundation.Collections.IMapChangedEventArgs`1<String>")]
[global::WinRT.WinRTExposedType(typeof(ObservableDictionaryChangedEventArgsWinRTTypeDetails))]
partial class ObservableDictionaryChangedEventArgs
{
}

internal sealed class ObservableDictionaryChangedEventArgsWinRTTypeDetails : global::WinRT.IWinRTExposedTypeDetails
{
    public global::System.Runtime.InteropServices.ComWrappers.ComInterfaceEntry[] GetExposedInterfaces()
    {
        _ = global::WinRT.GenericHelpers.IMapChangedEventArgs_string.Initialized;

        return new global::System.Runtime.InteropServices.ComWrappers.ComInterfaceEntry[]
        {
            new global::System.Runtime.InteropServices.ComWrappers.ComInterfaceEntry
            {
                IID = global::ABI.Windows.Foundation.Collections.IMapChangedEventArgsMethods<string>.IID,
                Vtable = global::ABI.Windows.Foundation.Collections.IMapChangedEventArgsMethods<string>.AbiToProjectionVftablePtr
            },
};
        }
    }
}
```

but there's no `IMapChangedEventArgs_string` being generated.

Version used: CsWinRT 2.1.0-prerelease.240602.1

### LINKED ISSUE #1638: AOT fails with ABI.System.Threading in 1.6 Experimental
**Describe the bug**

AOT fails with ABI.System.Threading in 1.6 Experimental

**To Reproduce**

When Windows App SDK 1.6 Experimental and CsWinRT 2.1.0 were installed at the same time, a source generator error occurred when compiling the program

2>D:\Github\GetStoreApp\GetStoreApp\obj\x64\Debug\net9.0-windows10.0.22621.0\win-x64\WinRT.SourceGenerator\Generator.WinRTAotSourceGenerator\WinRTGenericInstantiation.g.cs(3728,33,3728,61): error CS0234: 命名空间“ABI.System”中不存在类型或命名空间名“Threading”(是否缺少程序集引用?)

**Expected behavior**

Successfully build with CsWinRT 2.1.0 preview and Windows App SDK 1.6 Experimental

**Version Info**

CsWinRT 2.1.0-preview
.NET SDK 9.0.0-preview4
Windows SDK 10.0.22621.35-preview

**Additional context**

image
image

manodasanW:
- Fixed issue where nested types were not correctly represented in the lookup table with a +.
- Fixed issue where multi-level generic types were not correctly represented in the lookup table where the generics need to come at the end after all the nested generic types
- Fix issue where ObservableCollection's use of SingleItemReadOnlyList causes interface not found issues due to the latter type is an internal type within the .NET runtime and thereby the source generator not being able to run on it.  Special casing to handle that if ObservableCollection is found to be used.
- Fix typo with IMapChangedEventArgs
- Fix issue where for KeyValuePair, we didn't do a proper WinRT type check before generating the vtable causing issues with withing ABI functions not being there due to they are not expected to be there.

Fixes #1638 
Fixes #1631 
Fixes https://github.com/microsoft/microsoft-ui-xaml/issues/9699
Fixes https://github.com/microsoft/microsoft-ui-xaml/issues/9710


src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
```diff
@@ -1082,6 +1116,26 @@ internal static void AddVtableAdapterTypeForKnownInterface(ITypeSymbol classType
...
+                        false,
+                        "SingleItemReadOnlyList",
+                        ImmutableArray<TypeInfo>.Empty,
+                        "System.Collections.Specialized.SingleItemReadOnlyList",
+                        genericInterfaces.ToImmutableArray(),
+                        ImmutableArray<GenericInterface>.Empty,
+                        false,
+                        false,
+                        false,
+                        "Microsoft.UI.Xaml.Interop.IBindableVector"));
```
Sergio0694, line 1136: Leaving a note here, we need to use `typeMapper` to pick the right namespace here for WUX/MUX

src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
```diff
@@ -1082,6 +1116,26 @@ internal static void AddVtableAdapterTypeForKnownInterface(ITypeSymbol classType
...
+                // ObservableCollection make use of an internal built-in type as part of its
+                // implementation for INotifyPropertyChanged.  Handling that manually here.
+                var genericInterfaces = new List<string>() { "System.Collections.IList", "System.Collections.IEnumerable" };
+                vtableAttributes.Add(
+                    new VtableAttribute(
+                        "System.Collections.Specialized",
+                        false,
+                        "SingleItemReadOnlyList",
+                        ImmutableArray<TypeInfo>.Empty,
+                        "System.Collections.Specialized.SingleItemReadOnlyList",
```
Sergio0694, line 1130: This doesn't seem right. We should not rely on internal implementation details of the BCL. There isn't even a guarantee that all versions of the BCL will have this same internal layout, nor that it won't change at any point in the future. Can you clarify what the issue is here exactly and why the current proposed solution needs to do this in the first place?

Sergio0694 (changes requested):

src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
manodasanW: I agree that is a risk and this was scoped enough that I did a tactical fix.  There isn't anything we are directly doing with the type here other than creating a `vtable` for it and if the signature of the type changes, you will get an `InvalidCastException` which is what folks are seeing today.

The scenario is you have an `ObservableCollection` binded in `XAML` and then you make a modification to it.  This triggers the `INotifyCollectionChanged` handler in `ObservableCollection` which are projected as it is a custom type mapping and the implementation of it passes a `SingleItemReadOnlyList` as the oldItem or so.  Since that isn't a type that the CsWinRT AOT source generator can see or run on, it doesn't get our vtable attribute.  Thereby when it goes across the ABI, `IList` isn't projected and thereby you get a `InvalidCastException`.

The other fix I am looking at possibly is to have a fallback where when the vtable attribute is missing, we only marshal the statically declared interface on the calling function.  But that is going to take longer to get working and do in a way that won't add more to the size in scenarios where it isn't needed.

Sergio0694: I see, yeah that's reasonable 🤔

> "when the vtable attribute is missing, we only marshal the statically declared interface on the calling function"

Yeah that's what I had in mind too, we'd just marshal as the interface type. If that's planned, that's fine 👍

src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
manodasanW: Fixed.

PR #1654, opened 2024-07-01
Fix invalid cast exception with stream APIs and make code fixer emit diagnostics as warnings

### LINKED ISSUE #1634: Loading images using .net file streams does not work when AOT is enabled
**Describe the bug**

Calling BitmapImage.SetSourceAsync() with a .net file stream using stream.AsRandomAccessStream() leads to WinRT originate error - 0x80004002 : 'Specified cast is not valid.'

**To Reproduce**

Code similar to the one below:

```csharp
using (FileStream stream = File.OpenRead(path))
{
  bitmapImage = new BitmapImage();
  await bitmapImage.SetSourceAsync(stream.AsRandomAccessStream());
}
```

works correctly when AOT is disabled but not when it is enabled

**Expected behavior**

.net file streams should work when AOT is enabled

**Version Info**

appsdk 1.6 experimental 1, CsWinRT 2.1.0-prerelease.240602.1

**Additional context**

### LINKED ISSUE #1649: AOT source generator should output warnings instead of messages if partial keyword is missing
# Proposal: The AOT source generator should output a warning instead of a message if the partial keyword is missing

## Summary

To make an application AOT compatible, the partial keyword needs to be added to certain class definitions. The AOT source generator outputs a very useful message if they are missing. However, this message gets lost very easily as there are often a huge number of other messages (spell check, ...). As the app will in most cases not work correctly without this keyword, I think at least a warning should be output in the specific window and the build output window.

## Rationale

* Currently, these messages are easily overlooked

manodasanW:
- Added a new `CsWinRTAotWarningLevel` property to control whether the warnings from the code fixer are emitted as warnings or info.  By default, for scenarios involving non built-in types, they will be emitted as warnings and built-in types will be emitted as info.  The property can be used to control that behavior.
- Fix InvalidCastException when using the stream extension APIs.  The stream extension APIs make use of generic types and `AsyncInfo.Run`, but given we disable the lookup table generation on the Windows SDK projection, they are not automatically made AOT safe.  Due to that, adding a manual implementation of what would get generated and only initializing it from the stream scenarios so that other scenarios can just trim it if the stream extension APIs are not used.

Fixes #1634
Fixes #1649


docs/aot.md
```diff
@@ -6,7 +6,7 @@ As of CsWinRT **2.1.0-preview** and Windows SDK projection 10.0.<sdk_ver>.**35-p
...
+If your app or library has non-WinRT classes that implement C#/WinRT projected interfaces or built-in .NET interfaces that are mapped to WinRT types and are passed across the ABI to other WinRT functions, the class needs to be marked `partial` in order to be AOT compatible.  Marking it `partial` allows the source generator distributed with C#/WinRT to add an attribute to the class which has the necessary logic to produce the WinRT vtable for it in a way that is both trimming and AOT compatible.  To help with this, there is a `code fixer` that will warn you when such types are not marked `partial`.  The deafult `warning level` (`CsWinRTAotWarningLevel`) for this code fixer is `1` which is to `not warn for scenarios involving only built-in types`.  But if you pass types implementing only built-in interfaces that are mapped to WinRT across the WinRT ABI, it recommended to set the warning level to `2` and mark such types `partial` or suppress the warning if those types are not passed across the WinRT ABI.
```
dongle-the-gadget, line 9:
```suggestion
If your app or library has non-WinRT classes that implement C#/WinRT projected interfaces or built-in .NET interfaces that are mapped to WinRT types and are passed across the ABI to other WinRT functions, the class needs to be marked `partial` in order to be AOT compatible.  Marking it `partial` allows the source generator distributed with C#/WinRT to add an attribute to the class which has the necessary logic to produce the WinRT vtable for it in a way that is both trimming and AOT compatible.  To help with this, there is a `code fixer` that will warn you when such types are not marked `partial`.  The default "warning level" (`CsWinRTAotWarningLevel`) for this code fixer is `1` which is to "not warn for scenarios involving only built-in types".  But if you pass types implementing only built-in interfaces that are mapped to WinRT across the WinRT ABI, it recommended to set the warning level to `2` and mark such types `partial` or suppress the warning if those types are not passed across the WinRT ABI.
```

nuget/readme.md
```diff
@@ -48,6 +48,7 @@ C#/WinRT behavior can be customized with these project properties:
 | CsWinRTRcwFactoryFallbackGeneratorForceOptIn | true \| *false | Forces the RCW factory fallback generator to be enabled (it only runs on .exe projects by default)  |
 | CsWinRTRcwFactoryFallbackGeneratorForceOptOut | true \| *false | Forces the RCW factory fallback generator to be disabled (overrides "ForceOptIn" as well)  |
 | CsWinRTMergeReferencedActivationFactories | true \| *false | Makes the native `DllGetActivationFactory` exported function for AOT scenarios also forward the activation call to all referenced WinRT components, allowing them to all be merged into a single executable or shared library |
+| CsWinRTAotWarningLevel | 0 \| *1 \| 2 | Specifies the warning level used by the code fixer for trimming and AOT compat (0 - no warnings, 1 - warnings for scenarios involving non built-in types, warnings for all scenarios) |
```
dongle-the-gadget, line 51:
```suggestion
| CsWinRTAotWarningLevel | 0 \| *1 \| 2 | Specifies the warning level used by the code fixer for trimming and AOT compat (0 - no warnings, 1 - warnings for scenarios involving non built-in types, 2 - warnings for all scenarios) |
```

lhak:
Thank you for trying to fix https://github.com/microsoft/CsWinRT/issues/1634. Unfortunately, I still see the same issue with this patch. Interestingly, if I copy  StreamTaskAdaptersImplementation.cs to my application and make sure its Init() method has been called, it works.

manodasanW:
> Thank you for trying to fix #1634. Unfortunately, I still see the same issue with this patch. Interestingly, if I copy StreamTaskAdaptersImplementation.cs to my application and make sure its Init() method has been called, it works.

Did you just use the updated `WinRT.Runtime` or did you also build the Windows SDK projection?  The stream changes are in the Windows SDK projection (Microsoft.Windows.SDK.NET.dll) so they will need that to be regenerated.

lhak:
Ok, I thought the updated code was in WinRT.Runtime.dll and I only used an updated cswinrt nuget package. So I guess everything will be fine once an updated Windows SDK nuget package is available.

PR #1655, opened 2024-07-02
Add 'WindowsUIXamlProjectionsAnalyzer' to error on invalid configs

Sergio0694:
Implements an analyzer to emit errors when the WUX/MUX configuration is invalid.

### DIFF
--- a/src/Authoring/WinRT.SourceGenerator/WindowsUIXamlProjectionsAnalyzer.cs
+++ b/src/Authoring/WinRT.SourceGenerator/WindowsUIXamlProjectionsAnalyzer.cs
@@ -1,0 +1,116 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Immutable;
+using System.Composition;
+using System.Threading;
+using Generator;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+#nullable enable
+
+namespace WinRT.SourceGenerator;
+
+/// <summary>
+/// A diagnostic analyzer that emits errors when 'UseUwp' and 'UseWinUI' are used incorrectly.
+/// </summary>
+[DiagnosticAnalyzer(LanguageNames.CSharp), Shared]
+public sealed class WindowsUIXamlProjectionsAnalyzer : DiagnosticAnalyzer
+{
+    /// <inheritdoc/>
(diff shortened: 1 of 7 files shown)


src/Authoring/WinRT.SourceGenerator/WindowsUIXamlProjectionsAnalyzer.cs
```diff
@@ -0,0 +1,116 @@
...
+    /// <param name="compilation">The input <see cref="Compilation"/> to check.</param>
+    /// <param name="assemblyName">The name of the assembly to check.</param>
+    /// <param name="token">The cancellation token for the operation.</param>
+    /// <returns>Whether <paramref name="compilation"/> references an assembly with a given name.</returns>
+    private static bool IsAssemblyReferenced(
+        Compilation compilation,
+        string assemblyName,
+        CancellationToken token)
+    {
+        foreach (MetadataReference metadataReference in compilation.References)
```
Sergio0694, line 91: @333fred just double checking, we can rely on this always returning all references including from transitive dependencies, right? Basically a scenario we want to check here is, you're referencing library A, and library A is referencing the System XAML projections, or WinAppSDK. That assembly would show up as a reference here even if your project isn't directly referencing them (eg. it's not adding an explicit NuGet package dependency to the package containing that assembly), right?

hez2010: The assemblies being passed to roslyn are resolved by msbuild targets, roslyn doesn't have any knowledge of "transitive dependencies".

Sergio0694: Right, I know Roslyn doesn't really process dependencies in terms of NuGet packages, but I meant like, practically speaking does this mean I'd always get all assemblies in all those scenarios being passed as references there?

Sergio0694:
Closing this, we're going to use .targets instead.

src/Authoring/WinRT.SourceGenerator/WindowsUIXamlProjectionsAnalyzer.cs
333fred: Yes, you would get all references passed to the compiler. That _does not_ mean that all references were resolved correctly by msbuild, or that there isn't some other build authoring error that would cause an assembly to be missing.

PR #1662, opened 2024-07-04
Skip checking COM objects on AOT

Sergio0694:
There will never be COM objects on AOT anyway. Also minor optimization to copying vtables.

### DIFF
--- a/src/WinRT.Runtime/ComWrappersSupport.net5.cs
+++ b/src/WinRT.Runtime/ComWrappersSupport.net5.cs
@@ -692,10 +692,9 @@ namespace WinRT
             public VtableEntries(List<ComInterfaceEntry> entries, Type type)
             {
                 Data = (ComInterfaceEntry*)RuntimeHelpers.AllocateTypeAssociatedMemory(type, sizeof(ComInterfaceEntry) * entries.Count);
-                for (int i = 0; i < entries.Count; i++)
-                {
-                    Data[i] = entries[i];
-                }
+
+                CollectionsMarshal.AsSpan(entries).CopyTo(new Span<ComInterfaceEntry>(Data, entries.Count));
+
                 Count = entries.Count;
             }
         }
--- a/src/WinRT.Runtime/ComWrappersSupport.net5.cs
+++ b/src/WinRT.Runtime/ComWrappersSupport.net5.cs
@@ -572,6 +572,13 @@ namespace WinRT
 
         private static unsafe bool IsRuntimeImplementedRCW(Type objType)
         {
+            // Built-in COM interop isn't supported in AOT environments,
+            // so this method can only ever return false. Just inline it.
+            if (!RuntimeFeature.IsDynamicCodeCompiled)
+            {
+                return false;
+            }
+
             bool isRcw = objType.IsCOMObject;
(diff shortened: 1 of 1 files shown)


manodasanW (approved):

src/WinRT.Runtime/ComWrappersSupport.net5.cs
```diff
@@ -572,6 +572,13 @@ static unsafe DefaultComWrappers()
 
         private static unsafe bool IsRuntimeImplementedRCW(Type objType)
         {
+            // Built-in COM interop isn't supported in AOT environments,
+            // so this method can only ever return false. Just inline it.
+            if (!RuntimeFeature.IsDynamicCodeCompiled)
+            {
+                return false;
+            }
+
```
MichalStrehovsky, line 581: How much does this really save? `!IsDynamicCodeCompiled ==> built-in COM is not supported` is a pretty major leap that might burn you in the future. What you'd want is `Marshal.IsBuiltInComInteropSupported` but such API doesn't exist.

This is basically assuming that `IsDynamicCodeCompiled` approximates `IsNativeAot` and that native AOT will never have built-in COM interop.

This approximation might be broken _today_ if you put `<PublishAot>true</PublishAot>` and `<BuiltInComInteropSupport>true</BuiltInComInteropSupport>` to the project file and F5 launch the app (not Publish). In this configuration, `RuntimeFeature.IsDynamicCodeCompiled` is false, but built-in COM works. It is a terrible configuration to set, but it's possible.

Cc @jkoritzinsky @AaronRobinsonMSFT

AaronRobinsonMSFT: I would agree. I think adding some public check is possible. I also tend to agree with @MichalStrehovsky here, what is this actually saving in practice?

Sergio0694: This wasn't so much to save binary size per se, but rather to make the AOT-safe version of CsWinRT "stricter" and simpler. We have several parts of the code that we kept for backwards-compatibility (including built-in COM support), like being able to use old generated projections that are not fully trim-safe, etc.

Starting from .NET 8, we're saying that all that only applies to non-AOT scenarios. If you enable AOT, you also opt-in into the stricter mode, where we require you to only use updated projections everywhere (old projections wouldn't work with AOT anyway), and disable some more leftover features. Dropping built-in COM support entirely is part of that. We've added support for interop with generated COM interfaces starting with the .NET 8 TFM instead.

dongle-the-gadget: The .NET 8 WinRT.Runtime (which implements generated COM interfaces) hasn't shipped though.

Sergio0694: Remember that AOT support hasn't shipped at all either, it's all experimental. We'll get there 🙂

MichalStrehovsky: As long as we're in agreement that `!RuntimeFeature.IsDynamicCodeCompiled` doesn't imply built-in COM is disabled (neither today, nor in the future), consider this resolved.

PR #1677, opened 2024-07-22
Make ICustomPropertyProvider AOT safe

manodasanW:
- Added `BindableCustomPropertyAttribute` to allow to specify that public properties from the class are bindable.  Either all public properties can be bindable by passing no arguments to the constructor.  Or the names of the properties and the types of the indexer types can be passed to specify which ones to make bindable.
- By adding this attribute to a class and marking it partial, an interface `Microsoft.UI.Xaml.Data.IBindableCustomPropertyImplementation` gets implemented on that class via the source generator which provides the implementation to support `ICustomPropertyProvider`.  This is done as an interface instead of an attribute because this allows to support generic types while the attribute model won't work with it.  In addition, we only need to worry about supporting classes and not other types.
- Added a bunch of tests to validate
- This model doesn't support built-in .NET types but those can be worked around by wrapping it around a class or so.  Support for that can be added via other means if needed in the future.

```
partial class Language : global::Microsoft.UI.Xaml.Data.IBindableCustomPropertyImplementation
{
    global::Microsoft.UI.Xaml.Data.BindableCustomProperty global::Microsoft.UI.Xaml.Data.IBindableCustomPropertyImplementation.GetProperty(string name)
    {
        if (name == "Name")
        {
            return new global::Microsoft.UI.Xaml.Data.BindableCustomProperty(
                true,
                false,
                "Name",
                typeof(string),
                static (instance) => ((Language)instance).Name,
                null,
                null,
                null);
        }
        if (name == "Value")
        {
            return new global::Microsoft.UI.Xaml.Data.BindableCustomProperty(
                true,
                true,
                "Value",
                typeof(int),
                static (instance) => ((Language)instance).Value,
                static (instance, value) => ((Language)instance).Value = (int)value,
                null,
                null);
        }
        return default;
    }

    global::Microsoft.UI.Xaml.Data.BindableCustomProperty global::Microsoft.UI.Xaml.Data.IBindableCustomPropertyImplementation.GetProperty(global::System.Type indexParameterType)
    {
        if (indexParameterType == typeof(int))
        {
            return new global::Microsoft.UI.Xaml.Data.BindableCustomProperty(
                true,
                true,
                "Item",
                typeof(string),
                null,
                null,
                static (instance, index) => ((Language)instance)[(int)index],
                static (instance, value, index) => ((Language)instance)[(int)index] = (string)value);
        }
        return default;
    }
}
```


src/WinRT.Runtime/Projections/ICustomPropertyProvider.net5.cs
```diff
@@ -289,25 +385,34 @@ static unsafe ManagedCustomPropertyProviderVftbl()
...
+                object target = global::WinRT.ComWrappersSupport.FindObject<object>(thisPtr);
+
+                if (target is global::Microsoft.UI.Xaml.Data.IBindableCustomPropertyImplementation bindableCustomPropertyImplementation)
+                {
+                    __result = bindableCustomPropertyImplementation.GetProperty(_name);
+                    *result = MarshalInterface<global::Microsoft.UI.Xaml.Data.ICustomProperty>.FromManaged(__result);
+                    return 0;
+                }
+
+                if (RuntimeFeature.IsDynamicCodeCompiled)
```
Sergio0694, line 403: Should we not immediately throw `NotSupportedException` here if on AOT and the target doesn't implement the interface, and include an error message to explain the attribute is missing on that type for this to work?

src/WinRT.Runtime/Projections/ICustomPropertyProvider.net5.cs
```diff
@@ -289,25 +385,34 @@ static unsafe ManagedCustomPropertyProviderVftbl()
...
+                {
+                    PropertyInfo propertyInfo = target.GetType().GetProperty(
```
Sergio0694, line 405: We should move this into a helper method and add the suppressions on it, right? Otherwise this will still warn on publish.

src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
```diff
@@ -101,6 +101,15 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
                 GenerateCCWForGenericInstantiation);
 
             context.RegisterImplementationSourceOutput(vtablesToAddOnLookupTable.Collect().Combine(properties), GenerateVtableLookupTable);
+
+            var bindableCustomPropertyAttributes = context.SyntaxProvider.CreateSyntaxProvider(
```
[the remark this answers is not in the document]
manodasanW: I haven't updated the rosyln version yet to include the version with that support.  As far as I can tell, it is only included starting from a certain VS 2022 version and I am not sure what the behavior will be if someone used an older VS 2022 version.

src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
```diff
@@ -224,6 +241,117 @@ private static (VtableAttribute, VtableAttribute) GetVtableAttributesForTaskAdap
...
+            }
+
+            List<BindableCustomProperty> bindableCustomProperties = new();
+
+            // Make all public properties in the class bindable including ones in base type.
+            if (attributeData.ConstructorArguments.Length == 0)
+            {
+                for (var curSymbol = symbol; curSymbol != null; curSymbol = curSymbol.BaseType)
+                {
+                    foreach (var propertySymbol in curSymbol.GetMembers().
```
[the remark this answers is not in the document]
manodasanW: We are using symbol data to get this data, so I believe it should.  But we can confirm after.

src/WinRT.Runtime/Attributes.cs
```diff
@@ -230,6 +230,39 @@ public WinRTAssemblyExportsTypeAttribute(Type type)
         public Type Type { get; }
     }
 
+    /// <summary>
+    /// An attribute used to indicate the properties which are bindable via the <see cref="Microsoft.UI.Xaml.Data.ICustomProperty"/> implementation provided for use in WinUI scenarios.
```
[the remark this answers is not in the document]
manodasanW: Leaving this like this.

src/WinRT.Runtime/Projections/ICustomPropertyProvider.net5.cs
manodasanW: I thought that was going to happen, but not seeing the warning strangely.  I even just fixed WarningsNotAsErrors to not include it.

manodasanW: I do see is sizoscope, none of this code seems to be included.

src/WinRT.Runtime/Projections/ICustomPropertyProvider.net5.cs
manodasanW: That is an interesting proposal, will try it out and test it with WinUI gallery to see if this doesn't cause any unexpected issues.

src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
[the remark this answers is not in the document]
Sergio0694: I was more worried about the fact you might see two different property symbols for the same member (one being the definition, one being the implementation part), and that the fact they'd be "duplicate" might cause problems.

src/WinRT.Runtime/Projections/ICustomPropertyProvider.net5.cs
Sergio0694: The reason why I'm suggesting that is that otherwise you'd have no way of removing the trim warnings and also make the behavior testable and reliable. Essentially, without it you'd be back in undefined behavior land, as you'd either still have those trim warnings during publish, or you'd have the code potentially crash at runtime due to no metadata.

src/WinRT.Runtime/Projections/ICustomPropertyProvider.net5.cs
Sergio0694: > "but not seeing the warning strangely"

Mmh that is weird, doesn't seem right. Have you tried with .NET 9? There's a bunch of missing warnings on .NET 8.

Sergio0694: I do see the warning and error on .NET 8 as well:

image

I really think we should do the same pattern we used elsewhere: move to a helper with the suppression, always throw on AOT, otherwise call the helper for backwards-compat.

manodasanW: Strange, I don't see those.  But addressed.

src/WinRT.Runtime/Projections/ICustomPropertyProvider.net5.cs
manodasanW: Addressed, will see how it works out with WinUI Gallery.

src/WinRT.Runtime/Projections/ICustomPropertyProvider.net5.cs
```diff
@@ -23,6 +25,105 @@ interface ICustomProperty
...
+
+        bool ICustomProperty.CanRead => _canRead;
+
+        bool ICustomProperty.CanWrite => _canWrite;
+
+        string ICustomProperty.Name => _name;
+
+        Type ICustomProperty.Type => _type;
+
+        object ICustomProperty.GetIndexedValue(object target, object index) => _getIndexedValue != null ? _getIndexedValue(target, index) : throw new NotImplementedException();
```
Sergio0694, line 104: Is `NotImplementedException` by design? Ie. should it not be `NotSupportedException` or some other exception?

src/WinRT.Runtime/Projections/ICustomPropertyProvider.net5.cs
```diff
@@ -23,6 +25,105 @@ interface ICustomProperty
...
+            _canWrite = canWrite;
+            _name = name;
+            _type = type;
+            _getValue = getValue;
+            _setValue = setValue;
+            _getIndexedValue = getIndexedValue;
+            _setIndexedValue = setIndexedValue;
+        }
+
+        bool ICustomProperty.CanRead => _canRead;
```
Sergio0694, line 96: How come all members are explicitly implemented?

src/WinRT.Runtime/Attributes.cs
```diff
@@ -230,6 +230,39 @@ public WinRTAssemblyExportsTypeAttribute(Type type)
...
+        /// </summary>
+        /// <param name="propertyNames">The name of the non-indexer public properties to mark as bindable.</param>
+        /// <param name="indexerPropertyTypes">The parameter type of the indexer public properties to mark as bindable.</param>
+        public BindableCustomPropertyAttribute(string[] propertyNames, Type[] indexerPropertyTypes)
+        {
+            PropertyNames = propertyNames;
+            IndexerPropertyTypes = indexerPropertyTypes;
+        }
+
+        internal string[] PropertyNames { get; }
```
Sergio0694, line 263: Nit: should these two be internal? All other attribute properties are public. Not a big deal, but I think it's common practice for these kind of attribute properties just storing constructor parameters to be public? We should also add XML docs if so.

src/WinRT.Runtime/Projections/ICustomPropertyProvider.net5.cs
manodasanW: ICustomProperty is internal and is just an internal implementation detail.  So explicitly implemented them.

src/WinRT.Runtime/Attributes.cs
manodasanW: I kept them internal to make it easier to change stuff if needed and the value of them didn't really make sense in the default constructor case where it really means all properties.

src/WinRT.Runtime/Projections/ICustomPropertyProvider.net5.cs
Sergio0694: Oh, duh, right ahahah

src/WinRT.Runtime/Projections/ICustomPropertyProvider.net5.cs
manodasanW: There are scenarios where these `NotImplementedException` can be seen as `NotSupportException`.  i.e. someone using the `indexer` function when it is a regular property.  But then there are scenarios where if the set accessor or get is not implemented by the user, maybe it should be `NotImplementedException` rather than `NotSupportedException`.  So, for consistency throughout and to avoid trying to figure out the actual intention, I think `NotImplementedException` probably makes sense to keep unless someone is looking for the exact hresult and there is a contract to maintain.

src/WinRT.Runtime/Projections/ICustomPropertyProvider.net5.cs
manodasanW: Things seem to be good with WinUI gallery testing, no unexpected issues.  It was actually useful in guiding to find the types to address.

src/Tests/FunctionalTests/CCW/Program.cs
```diff
@@ -527,4 +635,93 @@ public void Execute(object parameter)
...
+    public string Name { get; } = "Language";
+    public int Value { get; set; }
+    public string this[int i]
+    {
+        get => _values[i];
+        set => _values[i] = value;
+    }
+}
+
+[global::WinRT.BindableCustomProperty([nameof(Name), nameof(Derived)], [typeof(int)])]
```
Scottj1s, line 654: just testing various attribute qualifiers here, and in line 660 and 666?

src/Tests/TestComponentCSharp/Class.cpp
```diff
@@ -1553,6 +1558,67 @@ namespace winrt::TestComponentCSharp::implementation
...
+        auto customProperty = !isIndexer ? customPropertyProvider.GetCustomProperty(property) : customPropertyProvider.GetIndexedProperty(property, indexerType);
+        if (customProperty == nullptr)
+		{
+            return false;
+		}
+
+        if (validateOnlyExists)
+		{
+			return true;
+		}
```
Scottj1s, line 1584: indentation

src/Tests/FunctionalTests/CCW/Program.cs
manodasanW: Yea, including namespaces vs not.

src/WinRT.Runtime/Projections/ICustomPropertyProvider.net5.cs
```diff
@@ -289,25 +390,47 @@ static unsafe ManagedCustomPropertyProviderVftbl()
...
+                {
+                    __result = bindableCustomPropertyImplementation.GetProperty(_name);
+                    *result = MarshalInterface<global::Microsoft.UI.Xaml.Data.ICustomProperty>.FromManaged(__result);
+                    return 0;
+                }
+
+                if (!RuntimeFeature.IsDynamicCodeCompiled)
+                {
+                    throw new NotSupportedException(
+                        $"ICustomProperty support used by XAML binding for '{target.GetType()}' requires the type to marked with 'WinRT.BindableCustomPropertyAttribute'. " +
```
Scottj1s, line 411: any plans to document this in greater detail, like with an md on the repo?  Xaml Compiler could then point errors/warnings to that page.

manodasanW: I can get our aot.md updated to mention this in a follow up PR.

src/WinRT.Runtime/Projections/ICustomPropertyProvider.net5.cs
```diff
@@ -23,6 +25,105 @@ interface ICustomProperty
         bool CanWrite { get; }
         string Name { get; }
         global::System.Type Type { get; }
+    }
+
+    /// <summary>
+    /// An interface complementing <see cref="BindableCustomPropertyAttribute"/> providing the implementation to expose the specified properties.
+    /// </summary>
+#if EMBED
+    internal
```
Sergio0694, line 34: Nit: given this is all new code, perhaps we can enable nullability annotations around it? Both the interface and the property type. Ideally, the attribute as well (since eg. the properties can be `null` when the default constructor is used).

src/Tests/TestComponentCSharp/Class.cpp
```diff
@@ -1553,6 +1558,67 @@ namespace winrt::TestComponentCSharp::implementation
         return winrt::make<bindable_observable_vector>(vector);
     }
 
+    bool Class::ValidateBindableProperty(
+        WF::IInspectable const& bindableObject,
+        hstring property,
+        Windows::UI::Xaml::Interop::TypeName const& indexerType,
+        bool validateOnlyExists,
```
Scottj1s, line 1565: enums (or name:value syntax) would be easier to read at call sites

src/Tests/TestComponentCSharp/Class.cpp
```diff
@@ -1553,6 +1558,67 @@ namespace winrt::TestComponentCSharp::implementation
...
+		{
+            return false;
+		}
+
+        if (validateOnlyExists)
+		{
+			return true;
+		}
+
+        if (customProperty.Name() != property ||
```
Scottj1s, line 1586: what's the purpose of passing these bools in?  given they have to match customProperty, why not just use the latter? (they seem redundant)

Scottj1s (approved):

src/Tests/TestComponentCSharp/Class.cpp
manodasanW: This is validating expected vs actual.  I will have a follow up PR to address this along with your enum comment for easier readability.

src/WinRT.Runtime/Projections/ICustomPropertyProvider.net5.cs
manodasanW: I took an attempt at properly supporting nullable here.  If we were to annotate it, I would need to annotate the Funcs too which would mean things like the value possibly being null and the index possibly being null.  And then using that, I enabled nullable in the generated code and tried to properly annotate that, but we run into scenarios where we will need to then also detect if the user code types are nullable or not and based on that generate the appropriate code.  Given this interface and class are both to support just the generated code, is just an implementation detail, and not meant to be called directly by consumers, I am going to skip annotating these.

PR #1680, opened 2024-07-24
Fix merged component support in non-WinRT component projects

Sergio0694:
This PR adds support for generating WinRT exports to chain referenced components, from non-component projects.


jevansaks:
One issue is that we still need the merged winmd written to the output location. That's done in the authoring targets via `CsWinRTPlaceWinMDInOutputFolder`. We need an equivalent thing for projects that set CsWinRTMergeReferencedActivationFactories=true but not CsWinRTComponent=true.

Sergio0694:
Is that not a separate feature? I mean one can also just copy all the individual .winmd-s to the output and just indicate the merged .dll as the implementation .dll for all of those .winmd-s, right? Eg. UWP currently does that, there's no single merged .winmd.

jevansaks:
I'm ok with getting this fix and just documenting that it does only thing and you might have to do some other targets work yourself.

manodasanW (approved):

PR #1682, opened 2024-07-25
Fix source generator for exception types and fix is operator

manodasanW:
- The source generator was generating code when collections of exceptions were used that didn't compile.  This was because exception is a class which gets mapped in WinRT to a value type and they don't have all the same functions.  This addresses that and also the issue where the exception marshaler functions were also mismatching which ones to use.
- Also fixing the is operator check when used with collection types whose helper type haven't been registered and is not implemented by the object.  Right now, it throws an exception in that scenario when it should just say this object doesn't implement it.
- Also adding DisposeAbiArray which was missing for Uris


src/WinRT.Runtime/Marshalers.cs
```diff
@@ -668,7 +668,8 @@ static MarshalGeneric()
                 DisposeMarshaler = ABI.System.NonBlittableMarshallingStubs.NoOpFunc;
                 DisposeAbi = ABI.System.NonBlittableMarshallingStubs.NoOpFunc;
             }
-            else if (typeof(T).IsValueType)
+            else if (typeof(T).IsValueType ||
```
Sergio0694, line 671: I'm worried this and the path below might regress size, because now this entire branch must be preserved for all reference types as well (the `Exception` check can't be folded), whereas previously with just `IsValueType` that could be handled by the importer directly, as it's an intrinsic, and by the linker too 🤔

hez2010: > the `Exception` check can't be folded

While it can be folded by the JIT so I don't think it would result in size regression. https://godbolt.org/z/5nKG7MWon

Sergio0694: The point is this whole path could be removed entirely for all reference types, but now it must be kept in all cases.

src/WinRT.Runtime/Marshalers.cs
```diff
@@ -668,7 +668,8 @@ static MarshalGeneric()
                 DisposeMarshaler = ABI.System.NonBlittableMarshallingStubs.NoOpFunc;
                 DisposeAbi = ABI.System.NonBlittableMarshallingStubs.NoOpFunc;
             }
-            else if (typeof(T).IsValueType)
+            else if (typeof(T).IsValueType ||
+                     typeof(T) == typeof(Exception))
```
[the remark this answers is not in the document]
manodasanW: Technically generated code shouldn't use anything other than Exception but adding it just in case anyone manually calls it.

src/WinRT.Runtime/Marshalers.cs
manodasanW: Latest changes look to address it.  We no longer have a separate generic instantiation for each class, but just for `canon` and `exception`.  Before these changes we did.  With the latest, we just have the below:

image

src/WinRT.Runtime/IWinRTObject.net5.cs
```diff
@@ -167,7 +167,21 @@ internal sealed bool IsInterfaceImplementedFallback(RuntimeTypeHandle interfaceT
...
 
-            Type helperType = type.FindHelperType();
+            Type helperType;
+            try
+            {
+                helperType = type.FindHelperType();
+            }
+            catch (Exception)
+            {
+                if (throwIfNotImplemented)
```
Sergio0694, line 177: Use an exception filter instead to avoid unwinding and re-throwing?

src/WinRT.Runtime/Marshalers.cs
```diff
@@ -2026,6 +2028,26 @@ static Marshaler()
                 DisposeMarshalerArray = new Action<object>(ABI.System.Type.DisposeMarshalerArray);
                 DisposeAbiArray = new Action<object>(ABI.System.Type.DisposeAbiArray);
             }
+            else if (typeof(Exception).IsAssignableFrom(typeof(T)))
```
Sergio0694, line 2031: I wonder if putting this branch after the following (that's checking `IsValueType` could help). Have we seen any regressions at all with these current changes? I can also help validate with sizoscope.

src/WinRT.Runtime/IWinRTObject.net5.cs
```diff
@@ -167,7 +167,21 @@ internal sealed bool IsInterfaceImplementedFallback(RuntimeTypeHandle interfaceT
                 return false;
             }
 
-            Type helperType = type.FindHelperType();
+            Type helperType;
+            try
+            {
+                helperType = type.FindHelperType();
+            }
+            catch (Exception)
```
Sergio0694, line 175: Wait, when would this throw? We're never catching exceptions from `FindHelperType` elsewhere 🤔

manodasanW: This throws if we try to instantiate a custom mapped generic type like a collection type that we haven't registered already in our lookup table.  In most scenarios during Init, we would have already added it.  But this is an is operator scenario, so it isn't added.

src/WinRT.Runtime/Marshalers.cs
manodasanW: As far as I can tell with sizoscope, with the new changes, there doesn't seem to be any unintended additions as there was  with my initial changes.  See https://github.com/microsoft/CsWinRT/pull/1682#discussion_r1694421721

PR #1686, opened 2024-07-30
Fix property initializers and derived ObservableCollection

### LINKED ISSUE #1663: Conversion to generic interface breaks CCWs for collection types
**Describe the bug**

**To Reproduce**

Triggered by [this `ItemsSource` binding](https://github.com/Sergio0694/ComputeSharp/blob/b05a118d24f3b26457c9220c551e5bef5849d1c7/samples/ComputeSharp.SwapChain.WinUI/Views/MainWindow.xaml#L80). The viewmodel property is [this one](https://github.com/Sergio0694/ComputeSharp/blob/b05a118d24f3b26457c9220c551e5bef5849d1c7/samples/ComputeSharp.SwapChain.WinUI/ViewModels/MainViewModel.cs#L87).

```csharp
public ShaderRunnerViewModel[] ComputeShaderOptions { get; } = ...
```

Works.

```csharp
public IReadOnlyList<ShaderRunnerViewModel> ComputeShaderOptions { get; } = ...
```

Crashes.

> [!NOTE]
> I've tried using an array expression instead of a collection expression, I still get the same exact crash.

**Stack trace**

image

**Expected behavior**

Should work fine

**Version Info**

Latest from staging/AOT (016e9ecc0b130d066a18860098e5fa92fa0e9177)

**Additional context**

Possibly related to #1661.

### LINKED ISSUE #1685: Reordering GridView items causes crashes with AOT compilation
**Describe the bug**

Using a GridView with the ItemsSource set to a class derived from `ObservableCollection<Item>` causes a crash when items are reordered due to an InvalidCastException. A workaround it to add code according to the documentation (https://github.com/manodasanW/WinUI-Gallery/blob/16ed717700b929dcb6591d32a4f10cd8b102aa07/WinUIGallery/VtableInitialization.cs#L24-L53) with `typeName` set to `System.Collections.Specialized.SingleItemReadOnlyList`. 

**Expected behavior**

Should work without additional code

**Version Info**

CsWinRT 2.1.0-prerelease.240722.1
WindowsSDK 10.0.22621.37-preview

**Additional context**

manodasanW:
The source generator was not detecting property initializers during lookup table generation.  It was also not detecting scenarios where classes extended ObservableCollection.  Both of those are now addressed.

Fixes #1685 
Fixes #1663

### DIFF
--- a/src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
+++ b/src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
@@ -1091,13 +1091,13 @@ namespace Generator
             else if (context.Node is VariableDeclarationSyntax variableDeclaration)
             {
                 var leftSymbol = context.SemanticModel.GetSymbolInfo(variableDeclaration.Type).Symbol;
-                foreach (var variable in variableDeclaration.Variables)
+                if (leftSymbol is INamedTypeSymbol namedType)
                 {
-                    if (variable.Initializer != null)
+                    foreach (var variable in variableDeclaration.Variables)
                     {
-                        var instantiatedType = context.SemanticModel.GetTypeInfo(variable.Initializer.Value);
-                        if (leftSymbol is INamedTypeSymbol namedType)
+                        if (variable.Initializer != null)
                         {
+                            var instantiatedType = context.SemanticModel.GetTypeInfo(variable.Initializer.Value);
                             AddVtableAttributesForType(instantiatedType, namedType);
                         }
                     }
@@ -1106,12 +1106,12 @@ namespace Generator
             // Detect scenarios where the property declaration has an initializer and is to a boxed or cast type during initialization.
             else if (context.Node is PropertyDeclarationSyntax propertyDeclaration)
             {
-                var leftSymbol = context.SemanticModel.GetSymbolInfo(propertyDeclaration.Type).Symbol;
                 if (propertyDeclaration.Initializer != null)
                 {
-                    var instantiatedType = context.SemanticModel.GetTypeInfo(propertyDeclaration.Initializer.Value);
+                    var leftSymbol = context.SemanticModel.GetSymbolInfo(propertyDeclaration.Type).Symbol;
                     if (leftSymbol is INamedTypeSymbol namedType)
                     {
+                        var instantiatedType = context.SemanticModel.GetTypeInfo(propertyDeclaration.Initializer.Value);
                         AddVtableAttributesForType(instantiatedType, namedType);
                     }
                 }
--- a/src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
+++ b/src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
@@ -1279,7 +1279,7 @@ namespace Generator
                 }
             }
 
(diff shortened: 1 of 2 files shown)


src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
```diff
@@ -1102,6 +1103,19 @@ private static bool NeedVtableOnLookupTable(SyntaxNode node)
                     }
                 }
             }
+            // Detect scenarios where the property declaration has an initializer and is to a boxed or cast type during initialization.
+            else if (context.Node is PropertyDeclarationSyntax propertyDeclaration)
+            {
+                var leftSymbol = context.SemanticModel.GetSymbolInfo(propertyDeclaration.Type).Symbol;
```
Sergio0694, line 1109: Nit: move this inside the branch so we don't unnecessary resolve the symbol if there's no initializer?

src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
```diff
@@ -1102,6 +1103,19 @@ private static bool NeedVtableOnLookupTable(SyntaxNode node)
                     }
                 }
             }
+            // Detect scenarios where the property declaration has an initializer and is to a boxed or cast type during initialization.
+            else if (context.Node is PropertyDeclarationSyntax propertyDeclaration)
+            {
+                var leftSymbol = context.SemanticModel.GetSymbolInfo(propertyDeclaration.Type).Symbol;
+                if (propertyDeclaration.Initializer != null)
+                {
+                    var instantiatedType = context.SemanticModel.GetTypeInfo(propertyDeclaration.Initializer.Value);
```
Sergio0694, line 1112: Nit: same here, we can move resolving `instantiatedType` inside the inner branch?

src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
```diff
@@ -1102,6 +1103,19 @@ private static bool NeedVtableOnLookupTable(SyntaxNode node)
...
+            // Detect scenarios where the property declaration has an initializer and is to a boxed or cast type during initialization.
+            else if (context.Node is PropertyDeclarationSyntax propertyDeclaration)
+            {
+                var leftSymbol = context.SemanticModel.GetSymbolInfo(propertyDeclaration.Type).Symbol;
+                if (propertyDeclaration.Initializer != null)
+                {
+                    var instantiatedType = context.SemanticModel.GetTypeInfo(propertyDeclaration.Initializer.Value);
+                    if (leftSymbol is INamedTypeSymbol namedType)
+                    {
+                        AddVtableAttributesForType(instantiatedType, namedType);
```
Sergio0694, line 1115: Do we need any validation on `instantiatedType`? Or will `AddVtableAttributesForType` take care of that?

Sergio0694 (approved):
Nice! 🎉

src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
manodasanW: `AddVtableAttributesForType` will validate it.

PR #1692, opened 2024-08-01
Fix GetGuidSignature for event args

manodasanW:
Fixing IID in GetGuidSignature to reflect MUX versions.  Addressing comments from https://github.com/microsoft/CsWinRT/pull/1421#discussion_r1651683184


Sergio0694 (approved):
Oops 😅

PR #1701, opened 2024-08-03
Fix get type for nullable enums and fix WUX projections

### LINKED ISSUE #1697: EasingMode parsing failed with NAOT
**Describe the bug**

Similar to the bug #1696, but now with _EasingMode_ property.
Error: `Failed to create a Windows.Foundation.IReference1<Microsoft.UI.Xaml.Media.Animation.EasingMode> from the text 'EaseOut"`
image

**To Reproduce**

1. Install CommunityToolkit.WinUI.Animations package
2. Enable PublishAot in csproj
3. Use lines from the screenshot above

**Expected behavior**

Application does not crash. animations works correctly

**Version Info**

**CsWinRT**: 2.1.0-**preview3**
**WinUI**: 1.6-exp2
**WinSDK**: 10.0.22621.37-preview
**NET**: net8.0-windows10.0.22621
**Windows**: build 26100

**Additional context**

manodasanW:
- In the previous PR for nullable, we fixed the scenario where WinUI asks for the type of the nullable version of built-in types.  But it looks common for WinUI to do the same for nullable enums too.  This expands the approach we take with `WinRTExposedType` to enums and addresses this issue via an interface that it internally implements to get the nullable version of the enum.
- Also removing the throw we had in `ResolveGenericType` as we always catch that exception later and return null.  It was there primarily as a hint, but given we now properly support nullable in this scenario and we have fallbacks for the other scenarios such as during RCW creation, removing it.
- Noticed that the custom mapped WUX projections were missing the AOT attributes for `WinRTExposedType`, so adding them there.

Fixes #1697

### DIFF
--- a/src/WinRT.Runtime/Projections/Nullable.cs
+++ b/src/WinRT.Runtime/Projections/Nullable.cs
@@ -2285,7 +2285,32 @@ namespace ABI.System
             if (type == typeof(global::System.Numerics.Vector2)) return typeof(global::System.Nullable<global::System.Numerics.Vector2>);
             if (type == typeof(global::System.Numerics.Vector3)) return typeof(global::System.Nullable<global::System.Numerics.Vector3>);
             if (type == typeof(global::System.Numerics.Vector4)) return typeof(global::System.Nullable<global::System.Numerics.Vector4>);
- 
+
+#if NET
+            var winrtExposedClassAttribute = type.GetCustomAttribute<WinRTExposedTypeAttribute>(false);
+            if (winrtExposedClassAttribute == null)
+            {
+                var authoringMetadaType = type.GetAuthoringMetadataType();
+                if (authoringMetadaType != null)
+                {
+                    winrtExposedClassAttribute = authoringMetadaType.GetCustomAttribute<WinRTExposedTypeAttribute>(false);
+                }
+            }
+
+            if (winrtExposedClassAttribute != null && winrtExposedClassAttribute.WinRTExposedTypeDetails != null)
+            {
+                if (Activator.CreateInstance(winrtExposedClassAttribute.WinRTExposedTypeDetails) is IWinRTNullableTypeDetails nullableTypeDetails)
+                {
+                    return nullableTypeDetails.GetNullableType();
+                }
+            }
+
+            if (!RuntimeFeature.IsDynamicCodeCompiled)
+            {
+                throw new NotSupportedException($"Failed to construct nullable with type '{type}'.");
+            }
+#endif
+
             return null;
         }
     }
@@ -2311,6 +2336,7 @@ namespace ABI.System
     internal interface IWinRTNullableTypeDetails
     {
         object GetNullableValue(IInspectable inspectable);
+        Type GetNullableType();
     }
 
     public sealed class StructTypeDetails<T, TAbi> : IWinRTExposedTypeDetails, IWinRTNullableTypeDetails where T: struct where TAbi : unmanaged
@@ -2364,6 +2390,8 @@ namespace ABI.System
                 Marshal.Release(nullablePtr);
             }
         }
+
+        Type IWinRTNullableTypeDetails.GetNullableType() => typeof(global::System.Nullable<T>);
     }
 
     public abstract class DelegateTypeDetails<T> : IWinRTExposedTypeDetails, IWinRTNullableTypeDetails where T : global::System.Delegate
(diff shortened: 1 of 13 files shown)


manodasanW:
Working on doing a test to see if there is much impact to the size.

src/WinRT.Runtime/Projections/Nullable.cs
```diff
@@ -2364,6 +2390,8 @@ unsafe object IWinRTNullableTypeDetails.GetNullableValue(IInspectable inspectabl
                 Marshal.Release(nullablePtr);
             }
         }
+
+        Type IWinRTNullableTypeDetails.GetNullableType() => typeof(global::System.Nullable<T>);
```
Sergio0694, line 2394: Nullable support is disabled when the `IReference<T>` feature switch is disabled, right? I'm wondering if we should leverage that and make this method just throw (or return `null`?) when that feature switch is disabled, to avoid the generic `Nullable<T>` instantiation for all these types.

src/WinRT.Runtime/Projections/Nullable.cs
```diff
@@ -2418,6 +2446,53 @@ unsafe object IWinRTNullableTypeDetails.GetNullableValue(IInspectable inspectabl
                 Marshal.Release(nullablePtr);
             }
         }
+
+        // Delegates are handled separately.
+        Type IWinRTNullableTypeDetails.GetNullableType() => throw new NotImplementedException();
```
Sergio0694, line 2451: Nit: should this be `NotSupportedException`?

src/WinRT.Runtime/Projections/Nullable.cs
```diff
@@ -2418,6 +2446,53 @@ unsafe object IWinRTNullableTypeDetails.GetNullableValue(IInspectable inspectabl
                 Marshal.Release(nullablePtr);
             }
         }
+
+        // Delegates are handled separately.
+        Type IWinRTNullableTypeDetails.GetNullableType() => throw new NotImplementedException();
+    }
+
+    public sealed class EnumTypeDetails<T> : IWinRTExposedTypeDetails, IWinRTNullableTypeDetails where T : struct, Enum
```
Sergio0694, line 2454: Nit: should this be `unmanaged`, instead of `struct`?

src/WinRT.Runtime/Projections/Nullable.cs
```diff
@@ -2418,6 +2446,53 @@ unsafe object IWinRTNullableTypeDetails.GetNullableValue(IInspectable inspectabl
...
             }
         }
+
+        // Delegates are handled separately.
+        Type IWinRTNullableTypeDetails.GetNullableType() => throw new NotImplementedException();
+    }
+
+    public sealed class EnumTypeDetails<T> : IWinRTExposedTypeDetails, IWinRTNullableTypeDetails where T : struct, Enum
+    {
+        [SkipLocalsInit]
```
Sergio0694, line 2456: Didn't we enable this for the entire module of WinRT.Runtime? Should we?

src/WinRT.Runtime/Projections/Nullable.cs
```diff
@@ -2418,6 +2446,53 @@ unsafe object IWinRTNullableTypeDetails.GetNullableValue(IInspectable inspectabl
...
+                        IID = ABI.System.Nullable_IntEnum.GetIID(type),
+                        Vtable = ABI.System.Nullable_IntEnum.AbiToProjectionVftablePtr
+                    };
+                }
+            }
+
+            return entries.Slice(0, count).ToArray();
+        }
+
+        Type IWinRTNullableTypeDetails.GetNullableType() => typeof(global::System.Nullable<T>);
```
Sergio0694, line 2492: Same question here for guarding the generic type instantiation with a feature switch check.

src/WinRT.Runtime/Projections/Nullable.cs
```diff
@@ -2418,6 +2446,53 @@ unsafe object IWinRTNullableTypeDetails.GetNullableValue(IInspectable inspectabl
...
+
+            if (FeatureSwitches.EnableIReferenceSupport)
+            {
+                entries[count++] = new ComWrappers.ComInterfaceEntry
+                {
+                    IID = global::WinRT.Interop.IID.IID_IPropertyValue,
+                    Vtable = ABI.Windows.Foundation.ManagedIPropertyValueImpl.AbiToProjectionVftablePtr
+                };
+
+                var type = typeof(T);
```
Sergio0694, line 2470: Nit: don't store the type in a local, just use `typeof(T)` directly. It often leads to better codegen too, because some optimizations can break down due to the JIT not being able to properly follow the type across locals.

src/WinRT.Runtime/Projections/Nullable.cs
manodasanW: I don't believe so, it looks to be just applied to a couple places.  We can look into whether we should enable this widely after.

src/WinRT.Runtime/Projections/Nullable.cs
manodasanW: We do support getting a `Nullable` like type for delegates just in case the WinUI native side asks for it given in WinRT delegates still support `IReference`.  We use our own type to represent it rather than actually `Nullable` which doesn't support it.  It is handled elsewhere rather than here which is why this is `NotImplementedException` rather than saying it is not supported.

src/WinRT.Runtime/Projections/Nullable.cs
manodasanW: This is implemented as an internal interface and all uses of the internal interface are behind the feature check, so that should allow all of this to get trimmed away.  To confirm, I did some sizoscope comparisons, the interface implementations are all getting trimmed away, so there are no nullable instantiations.  As a result, I don't think we will need another feature check here.

Sergio0694: Ah, cool, that works too! Thank you for double checking 🙂

Sergio0694 (approved):

PR #1705, opened 2024-08-05
Add 'Windows.UI.Xaml.*' exception types

Sergio0694:
This PR adds the 4 `Windows.UI.Xaml.*` exception types we were missing, and uses a feature switch to throw either those or the `Microsoft.UI.Xaml.*` types. Either group will be fully trimmed depending on the feature switch, so no size impact expected.


src/WinRT.Runtime/ExceptionHelpers.Windows.cs
```diff
@@ -0,0 +1,118 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+
+namespace Windows.UI.Xaml
```
manodasanW, line 6: Probably want NET guard for WUX version to avoid any netstandard conflicts.

Sergio0694: Fixed 👍

src/WinRT.Runtime/ExceptionHelpers.cs
```diff
@@ -7,7 +7,6 @@
 using System.Runtime.CompilerServices;
 using System.Runtime.ExceptionServices;
 using System.Runtime.InteropServices;
-using System.Text;
 using WinRT.Interop;
```
manodasanW, line 10: namespace is used as part of encoding reference below.

src/WinRT.Runtime/ExceptionHelpers.Windows.net5.cs
```diff
@@ -0,0 +1,118 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
```
manodasanW, line 4: Can we also add #if NET for this entire file.

manodasanW (approved):

manodasanW:
CI got far enough to show that tests are now passing.  And previously everything else built fine.

PR #1717, opened 2024-08-16
Add 'CoCreateInstance' helper

Sergio0694:
This PR adds the following APIs:

```csharp
namespace WinRT;

public enum ClassContext : uint
{
    InProcServer,
    LocalServer
}

public class IInspectable
{
    public static T CoCreateInstance<T>(in Guid classId, in Guid interfaceId, ClassContext classContext)
        where T : class;
}
```

These can be used to easily create WinRT objects from out-of-proc COM servers.

For instance, for a Windows Package Manager type:

```csharp
PackageManager packageManager = IInspectable.CoCreateInstance<PackageManager>(
    CLSID_PackageManager,
    IID_IPackageManager,
    ClassContext.LocalServer);
```

### DIFF
--- a/src/WinRT.Runtime/IInspectable.cs
+++ b/src/WinRT.Runtime/IInspectable.cs
@@ -21,6 +21,22 @@ namespace WinRT
         FullTrust = PartialTrust + 1
     }
 
+    /// <summary>
+    /// Values that are used in activation calls to indicate the execution contexts in which an object is to be run.
+    /// </summary>
+    public enum ClassContext : uint
+    {
+        /// <summary>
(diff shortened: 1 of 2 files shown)


shmuelie:
As the IID is in the metadata for types, could we reference it that way instead of needing to declare it explicitly? (Sadly not true for CLSID)

Sergio0694:
Closing this, we're not really planning on adding new functionality to 2.x at this point. We can re-evaluate in the future whether we'd want a public API like this in 3.0, though the API surface and implementation would have to be different for sure.

PR #1730, opened 2024-09-02
Treat init-only properties as cannot write

hez2010:
We cannot write to init-only properties after initialized, treat them as cannot write to avoid compilation error. 

Alternatives: generate an `UnsafeAccessor` to mutate the property.


manodasanW:
Good catch, I think this is the right fix for this given this isn't a write property if it is specified as init only.

manodasanW (approved):

PR #1732, opened 2024-09-04
Fix source generator issues

### LINKED ISSUE #1721: SourceGenerator generates code using unsafe even if EnableUnsafeBlocks aren't enabled
Source generator generates code using unsafe even if EnableUnsafeBlocks isn't enabled.  We probably should check if it is not enabled and given a better diagnostic asking EnableUnsafeBlocks to be enabled.  If the CsWinRT package is referenced, this isn't an issue.  But when the source generator is ran standalone as part of the projection package, it is.

WinRT.SourceGenerator\Generator.WinRTAotSourceGenerator\WinRTGenericInstantiation.g.cs(680,31,680,54): error CS0227: Unsafe code may only appear if compiling with /unsafe

### LINKED ISSUE #1724: AOT generator ignores generic classes inheriting from generic types
**Describe the bug**
AOT generator does not generate glue code for generic classes inheriting from generic types, causing the process to crash when trying to use these classes.

**To Reproduce**
```csharp
// notice that there's no suggestion to add "partial" if it's missing
public partial class TestCollection<T> : ObservableCollection<T>
{
}
...
myListView.ItemsSource = new TestCollection<string>();
```

**Expected behavior**
COM glue is properly generated and process won't crash

**Version Info**

2.1.1

manodasanW:
This addresses a couple issues that came up with respect to our source generator:

- There are scenarios where we generate code making use of unsafe.  This can lead to compiler errors that might not be immediately obvious if unsafe isn't enabled and wasn't enabled by a CsWinRT package reference.  So now we detect if unsafe isn't enabled and instead produce a diagnostic with a code fix to allow unsafe blocks.  This is similar to what the other .NET sources generators do that also generate code using unsafe.  Fixes #1721
- We were generating code for generic interface instantiations in scenarios we can say for sure aren't WinRT scenarios.  Specifically, for .NET custom WinRT type mapping classes, we treated function calls to them as if they were WinRT classes and generated code to make them AOT safe when needed.  But given they are not actually WinRT classes like `string.Join`, we do not need to generate during those scenarios.  Addressed by differentiating the check to be for whether it is a WinRT class rather than just a WinRT type.
- Instantiations of generic types defined in the same assembly were not getting code generated for on the vtable lookup table.  This was due to the check we put where we said if the type is defined in the current assembly, we will probably be generating an attribute on it so we don't need to put it on the lookup table.  But that isn't true if it is a generic class implementing generic interfaces.  Addressing by making that check more accurate.  Fixes #1724


manodasanW:
Validated with a couple repro scenarios that they behave as expected with the fix.

PR #1795, opened 2024-09-20
[Authoring] Allow custom overloads

### LINKED ISSUE #667: Add support for authoring interfaces with GUID attribute specified

### LINKED ISSUE #1569: Windows.Foundation.Metadata.OverrideAttribute generated even when it is already manually specified.
**Describe the bug**
When authoring WinRT Components, CsWinRT will automatically generate the Override attribute for you. If I add my own Override attribute, I expect CsWinRT to not generate one anymore but it still does.

**To Reproduce**
1. Author a WinRT Component with override methods.
2. Give one of the overrides the Override attribute with a custom name.
3. See that the method will have two Override attributes.

**Expected behavior**
That the method only has the Override attribute I added.

**Version Info**

CsWinRT: 2.0.7
.NET SDK: 8.0.204
Windows SDK: 10.0.22621.0

dongle-the-gadget:
Closes #1569.
Closes #667.


Sergio0694 (commented):
Is there no unit test we can add for this?

dongle-the-gadget:
I think we would need a projection that doesn't have overloads, which from my understanding is none.

Sergio0694:
Mh. I mean, changes seem fine to me, let's just see if @manodasanW thinks we need tests or not 🙂

dongle-the-gadget:
Oh, one more thing. This change doesn't seem to work with CsWinMD (not the source generator itself), probably due to the Windows.winmd file not being imported properly, though CsWinMD is probably low priority at the moment.

manodasanW:
> I think we would need a projection that doesn't have overloads, which from my understanding is none.
> 
> Or possibly an IL inspector? I'm not sure if we have any ATM.



> Mh. I mean, changes seem fine to me, let's just see if @manodasanW thinks we need tests or not 🙂

Can't we just add another pair of functions similar to [this](https://github.com/microsoft/CsWinRT/blob/037f6aedf22178e278a87aef12dacaf0b74e57db/src/Tests/AuthoringTest/Program.cs#L215-L218) with a different method name and then use the Overload attribute on them to define the name that should be used for the overload?

src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
```diff
@@ -1561,6 +1561,10 @@ private void AddOverloadAttributeForInterfaceMethods(TypeDeclaration interfaceTy
                         continue;
                     }
 
+                    bool alreadyHasOverload = method.TryGetAttributeWithType(Model.Compilation.GetTypeByMetadataName("Windows.Foundation.Metadata.OverloadAttribute"), out AttributeData _);
+                    if (alreadyHasOverload)
```
manodasanW, line 1565: I wonder if we should be skipping when there is an attribute or if we should be checking the overloaded method name in the attribute if one is set and then using that below rather than the one we generate.  Once you add the test case, we should be able to see whether the overload attribute is ending up in the winmd or not.

dongle-the-gadget:
> Can't we just add another pair of functions similar to [this](https://github.com/microsoft/CsWinRT/blob/037f6aedf22178e278a87aef12dacaf0b74e57db/src/Tests/AuthoringTest/Program.cs#L215-L218) with a different method name and then use the Overload attribute on them to define the name that should be used for the overload?

I’m not sure how that would work for the test. Don’t we have to check the WinMD?

Sergio0694:
Superseded by #2275 for 2.x and #2454 for 3.0. Thank you!

PR #1802, opened 2024-09-28
Fix IsPartial check and escaping of assembly name

### LINKED ISSUE #1799: Source generator IsPartial check doesn't check parent classes
In scenarios where a nested class is `partial` but the parent isn't and the nested class meets the requirements for us to generate the sources, we result in a compiler error rather than a diagnostic.

### LINKED ISSUE #1800: Source generator not escaping dashes in assembly name
Our source generator has logic to escape the assembly name as it is used as part of the namespace for generated sources.  The escaping logic doesn't handle dashes which it should.

        public static string EscapeTypeNameForIdentifier(string typeName)
        {
            return Regex.Replace(typeName, """[(\ |:<>,\.)]""", "_");
        }

manodasanW:
- We previously didn't escape `-` and `@` in the assembly name when using it for an identifier or namespace.  This addresses that.
- Our `IsPartial` check wasn't complete where we didn't also check whether the parent classes were partial too to allow for us to generate source.  Due to that, we would generate source in such cases leading to a compiler error rather than raising a diagnostic as we did when the type itself wasn't partial.
- Improve the code fixer to handle making parent types partial.
- Add partial diagnostics for the `GeneratedCustomBindableProperty` attribute too

Fixes #1799 
Fixes #1800


Sergio0694 (approved):

manodasanW:
Failures are post build infra issues, tests passed.

PR #1804, opened 2024-09-30
Improve performance of source generator

manodasanW:
- Changed the ordering of the properties during the combine operations to allow us to know earlier whether authoring support is enabled.  This allows us to reduce how often we call `GetVtableAttributesToAddOnLookupTable` and thereby reducing calls to `GetSymbolInfo`.  This should reduce the number of calls by half and thereby improve performance.

### DIFF
--- a/src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
+++ b/src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
@@ -29,16 +29,18 @@ namespace Generator
 
             var propertiesAndAssemblyName = properties.Combine(assemblyName);
 
-            var typeMapper = context.AnalyzerConfigOptionsProvider.Select((options, ct) => options.GetCsWinRTUseWindowsUIXamlProjections()).Select((mode, ct) => new TypeMapper(mode));
+            var typeMapperAndProperties = context.AnalyzerConfigOptionsProvider
+                .Select(static (options, ct) => options.GetCsWinRTUseWindowsUIXamlProjections())
+                .Select(static (mode, ct) => new TypeMapper(mode))
+                .Combine(properties);
 
             var vtablesToAddFromClassTypes = context.SyntaxProvider.CreateSyntaxProvider(
                     static (n, _) => NeedVtableAttribute(n),
                     static (n, _) => n)
-                .Combine(typeMapper)
-                .Combine(properties)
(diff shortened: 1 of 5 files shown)


src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
```diff
@@ -19,8 +19,28 @@ public class WinRTAotSourceGenerator : IIncrementalGenerator
...
+                    // check for such as component support where the generator introduces the reference and embedded support.
+                    // We will allow for an override for this by detecting whether CsWinRTAotWarningLevel is set to 2 which
+                    // indicates the component cares about optimization for built-in interfaces.
+                    (value.properties.isCsWinRTAotOptimizerEnabled && 
+                     (value.properties.isCsWinRTComponent ||
+                      value.properties.isCsWinRTAotOptimizerEnabledForBuiltInInterfaces ||
+                      value.properties.isCsWinRTEmbeddedSupport ||
+                      GeneratorHelper.HasWinRTRuntimeReference(value.compilation, ct)), 
+                     value.properties.isCsWinRTComponent,
+                     value.properties.Item3));
```
Sergio0694, line 43: Why are we returning the token..?

src/Authoring/WinRT.SourceGenerator/Helper.cs
```diff
@@ -11,6 +11,7 @@
 using System.IO;
 using System.Linq;
 using System.Text.RegularExpressions;
+using System.Threading;
```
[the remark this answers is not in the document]
manodasanW: This is from CancellationToken.

src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
manodasanW: which token?  Item3 is another bool which I am using to reconstruct the properties that is used later.

Sergio0694: Oh, sorry, misread. Too many tuple items ahah

src/Authoring/WinRT.SourceGenerator/Helper.cs
Sergio0694, line 14: Leftover now? It seems you removed the cancellation token?

Sergio0694 (approved):
LGTM!

PR #1815, opened 2024-10-09
Add 'WinRTManagedOnlyTypeDetails' type

Sergio0694:
This PR adds the `WinRTManagedOnlyTypeDetails`, which can be used with `WinRTExposedTypeAttribute` to correctly suppress warnings from the AOT generator and also suppress the codegen for a type (regardless of whether it's marked as partial). I also added a bunch of XML docs for all of those types involved here, and made `WinRTExposedTypeAttribute` visible.


src/WinRT.Runtime/TypeNameSupport.cs
```diff
@@ -473,6 +473,16 @@ private static bool TryAppendWinRTInterfaceNameForType(Type type, StringBuilder
                 return true;
             }
 
+#if NET8_0_OR_GREATER
```
manodasanW, line 476: What scenario will actually hit this?

Sergio0694: It was mostly for consistency. I was thinking perhaps there could be some weird case where you could manage to end up calling the name resolution logic on such a type, so might as well check and throw here too? 🤔

manodasanW: This should only be called for GetRuntimeClassName.  That shouldn't be possible given we won't marshal it.

manodasanW (approved):

src/WinRT.Runtime/Attributes.cs
```diff
@@ -131,30 +153,80 @@ interface IWinRTExposedTypeDetails
...
+            ThrowNotSupportedException();
+
+            return Array.Empty<ComWrappers.ComInterfaceEntry>();
+        }
+
+        /// <summary>
+        /// Throws a <see cref="NotSupportedException"/> to signal that a blocked type is being marshalled.
+        /// </summary>
+        [DoesNotReturn]
+        private static void ThrowNotSupportedException()
```
manodasanW, line 220: Can we add a test to the ccw functional test that confirms an exception is thrown.

Sergio0694: Done in https://github.com/microsoft/CsWinRT/pull/1815/commits/9846e0f7650583b805a3585e77a39be8f98e808c!

PR #1817, opened 2024-10-10
Don't emit default ctor for non activatable types

Sergio0694:
This PR fixes the .winmd generation logic to avoid the default constructor for types that do have some non public constructor that takes any parameters. This bug is blocking the migration of one of our WinRT component projects in the Microsoft Store.

### DIFF
--- a/src/Tests/AuthoringTest/Program.cs
+++ b/src/Tests/AuthoringTest/Program.cs
@@ -1889,6 +1889,26 @@ namespace AuthoringTest
             return new("Test123");
         }
     }
+
+    public sealed class TypeOnlyActivatableViaItsOwnFactory
+    {
+        private readonly string _text;
+
+        private TypeOnlyActivatableViaItsOwnFactory(string text)
+        {
+            _text = text;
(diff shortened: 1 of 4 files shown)


Sergio0694:
Checked the .winmd manually, looks just fine:

image

src/Tests/AuthoringTest/Program.cs
```diff
@@ -1865,6 +1865,30 @@ public sealed class CustomInterfaceGuidClass : ICustomInterfaceGuid
...
+            _text = text;
+        }
+
+        public string GetText()
+        {
+            return _text;
+        }
+    }
+
+    public static class NonActivatableFactory
```
manodasanW, line 1885: Can you add another test case or modify this one such that it is a sealed class with an internal constructor but also a static method that returns the object from the internal constructor.

Sergio0694: Done! 🙂

manodasanW (approved):

PR #1818, opened 2024-10-10
Fix CsWinRT component support when source generator gets updated but not CsWinRT version

manodasanW:
It is now possible for there to be scenarios where the source generator is newer but the CsWinRT.exe version is still older due to an older CsWinRT package still being referenced.  To handle these scenarios, the source generator should detect this and not generate sources.


src/Authoring/WinRT.SourceGenerator/Generator.cs
```diff
@@ -349,6 +349,15 @@ public static void GenerateWinRTExportsType(GeneratorExecutionContext context)
                 return;
             }
 
+            // If an older CsWinRT version is referenced, it is used to generate the projection.
+            // The projection generated by it won't be marked partial to generate the attribute on it
+            // and we also don't support it with the new scenarios without updating CsWinRT package
+            // So skip generating them.
+            if (GeneratorHelper.IsOldCsWinRTExe(context))
```
Sergio0694, line 356: Will this result in a projection just generating an empty .dll silently? Should we emit a warning at least?

manodasanW: No, it shouldn't be empty given we are still generating / adding the cswinrt.exe generated component projection sources as part of this generator.

Sergio0694: Mmh what will this skip then? And should we emit some diagnostic if we're skipping part of the source?

manodasanW: This will skip the new ABI GetActivationFactory generated to be referenced during the merged activation factories.  If someone is using the older CsWinRT version to generate their component, they are not on a new version that should be getting that functionality.  They would need to update their CsWinRT version to get that.

Sergio0694 (approved):

Issue #1645, opened 2024-06-20
SkXamlCanvas crashes with new prerelease packages


charlesroddie:
Putting an SkXamlCanvas in a WinUI app crashes with the new prerelease packages.

We are trying to test NativeAOT in WinUI for our app, which is built in SkiaSharp.
We are using SkXamlCanvas because hardware accelerated support for SkiaSharp views in WinUI aren't released yet.

We get the exception:

```
System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
---> System.NotSupportedException: Managed vtable types (ie. containing any reference types) are not supported.
   at WinRT.ObjectReference`1.GetVtable(IntPtr thisPtr)
   at WinRT.ObjectReference`1..ctor(IntPtr thisPtr)
   at WinRT.ObjectReference`1.Attach(IntPtr& thisPtr, Guid iid)
   at WinRT.ObjectReference`1.TryAs(IObjectReference sourceRef, Guid iid, ObjectReference`1& objRef)
   at WinRT.IObjectReference.TryAs[T](Guid iid, ObjectReference`1& objRef)
   at WinRT.IObjectReference.As[T](Guid iid)
   at WinRT.IObjectReference.As[T]()
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
   --- End of inner exception stack trace ---
   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
```

Reproduction:

https://github.com/charlesroddie/WinUITemplate/tree/SkXamlCanvas

Is this a CsWinRT issue or a Skiasharp issue?

Version info:
```
SkiaSharp 2.88.8 and also on 3.0.0-preview.3.1
CsWinRT 2.1.0-prerelease.240602.1
Microsoft.Windows.SDK.BuildTools 10.0.22621.3233
Microsoft.WindowsAppSDK 1.6.240531000-experimental1
```

Two trim warnings are produced in release mode, but the crash happens in debug (non-aot) and release (nativeaot):
```
[...]

PR #1819, opened 2024-10-10
Add fallback code for managed vtables for skia scenarios

manodasanW:
Projects using Skia seem to be hitting our restriction on managed vftbl types that we had put given we couldn't guarantee trimming compat and all generated managed vftbl were in the Windows SDK projection which was updated alongside the change.  But Skia happened to have a manual projection with a managed vftbl and given folks are running into this, adding fallback code to handle it.

https://github.com/dotnet/maui/issues/25150
https://github.com/microsoft/CsWinRT/issues/1645

### DIFF
--- a/src/WinRT.Runtime/ObjectReference.cs
+++ b/src/WinRT.Runtime/ObjectReference.cs
@@ -674,7 +674,19 @@ namespace WinRT
 #if NET
             if (RuntimeHelpers.IsReferenceOrContainsReferences<T>())
             {
-                throw new NotSupportedException("Managed vtable types (ie. containing any reference types) are not supported.");
+                if (!RuntimeFeature.IsDynamicCodeCompiled)
+                {
+                    throw new NotSupportedException("Managed vtable types (ie. containing any reference types) are not supported.");
+                }
+
+                return GetVtableForJitEnvironment(thisPtr);
+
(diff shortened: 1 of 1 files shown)


src/WinRT.Runtime/ObjectReference.cs
```diff
@@ -673,8 +673,20 @@ private static unsafe T GetVtable(IntPtr thisPtr)
...
+                if (!RuntimeFeature.IsDynamicCodeCompiled)
+                {
+                    throw new NotSupportedException("Managed vtable types (ie. containing any reference types) are not supported.");
+                }
+
+                return GetVtableForJitEnvironment(thisPtr);
+
+                [UnconditionalSuppressMessage("Trimming", "IL2090", Justification = "Fallback method for JIT environments that is not trim-safe by design.")]
+                [MethodImpl(MethodImplOptions.NoInlining)]
+                static T GetVtableForJitEnvironment(IntPtr thisPtr)
```
Sergio0694, line 686: I'm fine with the change, given it doesn't impact AOT, but technically speaking this introduces undefined behavior for trimming, since we'll now generate no warnings, but code using this path with trimming will break with no warnings. I don't really have a good solution, just pointing this out.

Sergio0694 (approved):

manodasanW:
Confirmed fix with repro.

PR #1820, opened 2024-10-11
Fix couple issues

manodasanW:
- Fix warning in embedded scenario and ensure trailing slash in authoring scenario
- Fix source generator to support nullable collections and fix marshaling logic for it
- Fix source generator for non-blittable structs such as HttpProgress where wrong marshaling function was used


Sergio0694 (approved):
LGTM. Have we checked any size regressions on NAOT with this change?

manodasanW:
> LGTM. Have we checked any size regressions on NAOT with this change?

Based on the ccw functional test, there is a 2kb diff from a func getting added for EventRegistrationToken for some reason.  But hard coding EventRegistrationToken similar to the other blittables seems to bring a decent improvement as a bunch of funcs get removed.  I will do that improvement as a separate PR.

PR #1824, opened 2024-10-14
Minor improvements to 'ExceptionHelpers.SetErrorInfo'

Sergio0694:
This PR includes just a couple of tweaks to `ExceptionHelpers.SetErrorInfo`:
- Use `WindowsCreateStringReference`, since we can just pin the source text.
- Avoid allocating a temporary `IObjectReference` object.

### DIFF
--- a/src/WinRT.Runtime/ExceptionHelpers.cs
+++ b/src/WinRT.Runtime/ExceptionHelpers.cs
@@ -388,8 +388,21 @@ See https://aka.ms/cswinrt/interop#windows-sdk",
                             hstring = IntPtr.Zero;
                         }
 
+#if NET
+                        IntPtr managedExceptionWrapper = ComWrappersSupport.CreateCCWForObjectUnsafe(ex);
+
+                        try
+                        {
+                            roOriginateLanguageException(GetHRForException(ex), hstring, managedExceptionWrapper);
+                        }
+                        finally
+                        {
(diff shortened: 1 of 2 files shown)


src/WinRT.Runtime/ExceptionHelpers.cs
```diff
@@ -374,18 +374,36 @@ public static unsafe void SetErrorInfo(Exception ex)
...
+                    IntPtr hstringHeader = IntPtr.Zero;
+                    IntPtr hstring = IntPtr.Zero;
 
                     fixed (char* lpMessage = message)
                     {
-                        if (Platform.WindowsCreateString((ushort*)lpMessage, message.Length, &hstring) != 0)
+                        if (Platform.WindowsCreateStringReference(
+                            sourceString: (ushort*)lpMessage,
+                            length: message.Length,
+                            hstring_header: &hstringHeader,
```
manodasanW, line 385: does the header need to be allocated?  See [here](https://github.com/microsoft/CsWinRT/blob/578f89272718974d140402f9d4e82d550ca082f5/src/WinRT.Runtime/Marshalers.cs#L134).

Sergio0694: Oh, oops yes you're right 😅

manodasanW (approved):

PR #1826, opened 2024-10-14
Improve exception messages for failed casts

Sergio0694:
This PR improves the exception message for failed `Cast<TInterface>` calls. Fix following feedback from internal folks.


michael-hawker (commented):
Thanks Sergio, new error message looks great, would have definitely helped me right away pick up quicker on what was going on with mismatched types between XAML data template and my data structure! 🦙❤️

manodasanW (approved):

PR #1827, opened 2024-10-15
Fix enum attribute argument.

### LINKED ISSUE #1825: MarshalingBehaviorAttribute and ThreadingAttribute Emitted into WinMD invalid
**Describe the bug**
If I add `MarshalingBehaviorAttribute` or `ThreadingAttribute` to a type I want to expose from a WinRT Component, the generated WinMD has invalid metadata for the attributes.

**To Reproduce**
Project File:
```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <TargetFramework>net8.0-windows10.0.22621.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
    <CsWinRTComponent>true</CsWinRTComponent>
    <ImplicitUsings>disable</ImplicitUsings>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Microsoft.Windows.CsWinRT" Version="2.1.5" PrivateAssets="all" />
  </ItemGroup>
</Project>
```

Class1.cs
```cs
using System;
using System.Threading.Tasks;
using Windows.Foundation.Metadata;
using GuidAttribute = System.Runtime.InteropServices.GuidAttribute;

namespace ClassLibrary1;

[Guid("DF622368-A7E4-4937-84C6-A375384B6067")]
public interface IDesktopActions
{
    Task<string> DesktopActionAsync(string options);
}


[Threading(ThreadingModel.Both)]
[MarshalingBehavior(MarshalingType.Agile)]
[Guid("9C480054-6397-4126-BDFE-8012BD4667D8")]
public sealed partial class DesktopActions : IDesktopActions
{
    public async Task<string> DesktopActionAsync(string options)
    {
        await Task.Delay(TimeSpan.FromSeconds(1));
        return options;
    }
}
```

**Expected behavior**
The WinMD to contain valid metadata for `MarshalingBehaviorAttribute` or `ThreadingAttribute`

**Version Info**
CsWinRT: 2.1.5
VS: 17.13.0 Preview
.NET SDK: 8.0.401

**Additional context**
Only one of the two attributes is needed to create the invalid metadata.

dongle-the-gadget:
Fixes #1825.

Even though `TaggedScalar` would look more correct, it looks like the actual C# compiler use `Scalar` instead. This is mentioned on win32metadata: https://github.com/microsoft/win32metadata/blob/527806d20d83d3abd43d16cd3fa8795d8deba343/sources/MetadataUtils/EncodeHelpers.cs#L125-L129

### DIFF
--- a/src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
+++ b/src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
@@ -1283,10 +1283,12 @@ namespace Generator
                     encoder.Scalar().Constant(constant.Value);
                     break;
                 case TypedConstantKind.Enum:
-                    encoder.TaggedScalar(
-                        type => type.Enum(constant.Type.ToString()),
-                        scalar => scalar.Constant(constant.Value)
-                    );
+                    encoder.Scalar().Constant(constant.Value);
+
+                    //encoder.TaggedScalar(
+                    //    type => type.Enum(constant.Type.ToString()),
+                    //    scalar => scalar.Constant(constant.Value)
+                    //);
                     break;
                 case TypedConstantKind.Type:
                     encoder.Scalar().SystemType(constant.Type.ToString());
--- a/src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
+++ b/src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
@@ -1285,10 +1285,13 @@ namespace Generator
                 case TypedConstantKind.Enum:
                     encoder.Scalar().Constant(constant.Value);
 
-                    //encoder.TaggedScalar(
-                    //    type => type.Enum(constant.Type.ToString()),
-                    //    scalar => scalar.Constant(constant.Value)
-                    //);
+                    // This looks more correct, but the code above matches what the C# compiler produces.
+                    // Eg. win32metadata does the same (see 'MetadataUtils.EncodeHelpers' in https://github.com/microsoft/win32metadata).
+                    // encoder.TaggedScalar(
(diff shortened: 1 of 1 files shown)


src/Authoring/WinRT.SourceGenerator/WinRTTypeWriter.cs
```diff
@@ -1283,10 +1283,12 @@ private void EncodeTypedConstant(TypedConstant constant, LiteralEncoder encoder)
...
-                    encoder.TaggedScalar(
-                        type => type.Enum(constant.Type.ToString()),
-                        scalar => scalar.Constant(constant.Value)
-                    );
+                    encoder.Scalar().Constant(constant.Value);
+
+                    //encoder.TaggedScalar(
+                    //    type => type.Enum(constant.Type.ToString()),
+                    //    scalar => scalar.Constant(constant.Value)
+                    //);
```
Sergio0694, line 1291: Let's include a comment for context, also let's leave a space after a comment.
```suggestion
                    // This looks more correct, but the code above matches what the C# compiler produces.
                    // Eg. win32metadata does the same (see 'MetadataUtils.EncodeHelpers' in https://github.com/microsoft/win32metadata).
                    // encoder.TaggedScalar(
                    //     type => type.Enum(constant.Type.ToString()),
                    //     scalar => scalar.Constant(constant.Value)
                    // );

```

Sergio0694 (approved):

manodasanW (approved):

manodasanW:
Thanks, confirmed things work with the change and the CI pipeline passed.

PR #1835, opened 2024-10-16
Add 'CsWinRTDynamicallyInterfaceCastableExclusiveTo'

Sergio0694:
This PR adds the "CsWinRTDynamicallyInterfaceCastableExclusiveTo" option, which allows generating IDIC supporting code for `[exclusive_to]` WinRT interfaces. This enables casting existing projected types to them and using their APIs normally.

### DIFF
--- a/src/cswinrt/code_writers.h
+++ b/src/cswinrt/code_writers.h
@@ -8092,7 +8092,7 @@ internal unsafe interface % : %
 %%%%}
 )",
 // Interface abi implementation
-            is_exclusive_to(type) ? "" : "[DynamicInterfaceCastableImplementation]",
+            is_exclusive_to(type) && !settings.idic_exclusiveto ? "" : "[DynamicInterfaceCastableImplementation]",
             bind<write_guid_attribute>(type),
             type_name,
             bind<write_type_name>(type, typedef_name_type::CCW, false),
@@ -8223,7 +8223,7 @@ AbiToProjectionVftablePtr = ComWrappersSupport.AllocateVtableMemory(typeof(@), s
             bind<write_interface_members>(type),
             "",
             [&](writer& w) {
-                if (!is_exclusive_to(type))
+                if (!is_exclusive_to(type) || settings.idic_exclusiveto)
                 {
                     for (auto required_interface : required_interfaces)
                     {
--- a/src/cswinrt/code_writers.h
+++ b/src/cswinrt/code_writers.h
@@ -4997,7 +4997,7 @@ remove => %.Unsubscribe(value);
 
     void write_interface_members(writer& w, TypeDef const& type)
     {
-        if (is_exclusive_to(type))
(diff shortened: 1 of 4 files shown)


Sergio0694:
Tested this locally in the Store, can confirm I'm seeing the right code generated 🙂

src/cswinrt/main.cpp
```diff
@@ -41,6 +41,7 @@ namespace cswinrt
         { "embedded", 0, 0, {}, "Generates an embedded projection."},
         { "public_enums", 0, 0, {}, "Used with embedded option to generate enums as public"},
         { "public_exclusiveto", 0, 0, {}, "Make exclusiveto interfaces public in the projection (default is internal)"},
+        { "idic_exclusiveto", 0, 0, {}, "Make exclusiveto interfaces support IDIC for RCW scenarios (default is false)"},
```
manodasanW, line 44:
```suggestion
        { "idic_exclusiveto", 0, 0, {}, "Make exclusiveto interfaces support IDynamicInterfaceCastable (IDIC) for RCW scenarios (default is false)"},
```

src/cswinrt/code_writers.h
```diff
@@ -8066,15 +8069,27 @@ NativeMemory.Free((void*)abiToProjectionVftablePtr);
 
             if (!hasOverridableAttribute)
             {
-                w.write(R"(%
+                // Under normal conditions, we would stop here and just emit a minimal amount of code.
+                // However, if IDIC is requested, just continue normally, but omit the CCW generation.
+                // We know we can safely omit that because it wouldn't have normally be generated.
```
manodasanW, line 8074: >We know we can safely omit that because it wouldn't have normally be generated

Can you clarify this.  Is this because we removed an exclusive to check or because we don't need it for the idic exclusive to scenario?

Sergio0694: It's because it's not needed, because the IDIC support is only about allowing to do IDIC casts to `exclusive_to` interfaces on RCWs. It doesn't allow _implementing_ those interfaces on managed types to then marshal them to native as a CCW. Nothing's stopping us from also doing that, but I just didn't include it here because (1) it increases codegen and (2) it's technically a separate feature and not needed for this scenario. Thoughts?

manodasanW: Makes sense, for this scenario if this flag is being used to project an interface, it doesn't need it.  Just wanted to confirm if you were thinking the same.

Sergio0694: Yeah, the rationale being that if the interface is `exclusive_to`, it doesn't really make sense for anyone else to implement it (because that's the whole point of it being exclusive after all). But it's still completely valid to just do an IDIC cast to it on an object (which _can_ implement it natively), so you can invoke its methods (which is what we're doing) 🙂

manodasanW (approved):

PR #1843, opened 2024-10-24
Add .NET 9 target for WinRT.Runtime.dll

Sergio0694:
- Enables building CsWinRT with .NET 9 locally
- Fixes some new trim warnings
- Makes `MarshalNonBlittable` work on AOT

### DIFF
--- a/build/AzurePipelineTemplates/CsWinRT-PublishToNuget-Stage.yml
+++ b/build/AzurePipelineTemplates/CsWinRT-PublishToNuget-Stage.yml
@@ -65,7 +65,7 @@ stages:
         itemPattern: ''
         targetPath: $(Build.SourcesDirectory)\net8.0
 
-# Download Net8.0
+# Download Net9.0
     - task: DownloadPipelineArtifact@2
       displayName: 'Download Net9.0'
       inputs:
--- a/build/AzurePipelineTemplates/CsWinRT-PublishToNuget-Stage.yml
+++ b/build/AzurePipelineTemplates/CsWinRT-PublishToNuget-Stage.yml
@@ -65,6 +65,14 @@ stages:
         itemPattern: ''
         targetPath: $(Build.SourcesDirectory)\net8.0
 
+# Download Net8.0
+    - task: DownloadPipelineArtifact@2
+      displayName: 'Download Net9.0'
+      inputs:
+        artifactName: net9.0
+        itemPattern: ''
+        targetPath: $(Build.SourcesDirectory)\net9.0
+
 # Stage Binaries
     - task: CmdLine@2
       displayName: Stage Binaries
@@ -94,6 +102,7 @@ stages:
           netstandard2.0\roslyn4120\WinRT.SourceGenerator.dll
           net8.0\WinRT.Host.Shim.dll
           net8.0\WinRT.Runtime.dll
+          net9.0\WinRT.Runtime.dll
           release_x64\WinRT.Host.dll
           release_x64\WinRT.Host.dll.mui
           release_x86\WinRT.Host.dll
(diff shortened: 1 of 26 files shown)


build/AzurePipelineTemplates/CsWinRT-PublishToNuget-Stage.yml
```diff
@@ -65,6 +65,14 @@ stages:
         itemPattern: ''
         targetPath: $(Build.SourcesDirectory)\net8.0
 
+# Download Net8.0
```
manodasanW, line 68:
```suggestion
# Download Net9.0

```

src/Directory.Build.props
```diff
@@ -28,9 +28,9 @@
     <IsTargetFrameworkNet8OrGreater Condition="$([MSBuild]::GetTargetFrameworkVersion('$(TargetFramework)')) >= 8">true</IsTargetFrameworkNet8OrGreater>
     <AppBuildTFMs>netcoreapp3.1;net8.0</AppBuildTFMs>
     <AppBuildTFMs Condition="'$(CIBuildReason)' == 'CI'">netcoreapp3.1;net8.0</AppBuildTFMs>
-    <LibBuildTFMs>netstandard2.0;net8.0</LibBuildTFMs>
+    <LibBuildTFMs>netstandard2.0;net8.0;net9.0</LibBuildTFMs>
     <LibBuildTFMs Condition="'$(CIBuildReason)' == 'CI'">netstandard2.0;net8.0</LibBuildTFMs>
```
manodasanW, line 32: Don't you also want to enable it for CI?

src/Directory.Build.props
```diff
@@ -28,9 +28,9 @@
...
-    <LibBuildTFMsNoNetStandard>net8.0</LibBuildTFMsNoNetStandard>
+    <LibBuildTFMsNoNetStandard>net8.0;net9.0</LibBuildTFMsNoNetStandard>
     <LibBuildTFMsNoNetStandard Condition="'$(CIBuildReason)' == 'CI'">net8.0</LibBuildTFMsNoNetStandard>
     <TestsBuildTFMs>net8.0-windows10.0.19041.0</TestsBuildTFMs>
     <TestsBuildTFMs Condition="'$(CIBuildReason)' == 'CI'">net8.0-windows10.0.19041.0</TestsBuildTFMs>
```
manodasanW, line 36: We probably also want to run our object lifetime tests on .NET 9 too and functional?

src/WinRT.Runtime/Marshalers.cs
```diff
@@ -1092,7 +1092,7 @@ public void Dispose()
 
         public static new unsafe MarshalerArray CreateMarshalerArray(T[] array)
         {
-#if NET
+#if NET && !NET9_0_OR_GREATER
```
manodasanW, line 1095: We probably should add some non blittable array marshaling test in functional tests for when aot is enabled and are running for .NET 9.

And we should at least manually re-enable the below authoring tests we disabled and confirm it works now on .NET 9.  Note vcxproj will need to be changed to build for .NET 9 which is why I am not sure if we want to do that change in the pipeline yet.

https://github.com/microsoft/CsWinRT/blob/b1733e95c6d35b551fc8cf6fe04e2a0c287346dd/src/Tests/AuthoringConsumptionTest/test.cpp#L263

https://github.com/microsoft/CsWinRT/blob/b1733e95c6d35b551fc8cf6fe04e2a0c287346dd/src/Tests/AuthoringConsumptionTest/test.cpp#L351

src/Directory.Build.props
Sergio0694: Ah, yup! I think this was from before the CI images had .NET 9 😅

build/AzurePipelineTemplates/CsWinRT-Variables.yml
```diff
@@ -12,6 +12,8 @@ variables:
   value: 'https://dotnetcli.blob.core.windows.net/dotnet'
 - name: Net8.SDK.Version
   value: '8.0.303'
+- name: Net9.SDK.Version
+  value: '9.0.100'
```
manodasanW, line 16:
```suggestion
  value: '9.0.201'

```

src/Directory.Build.props
manodasanW: Will address in follow up PR with testing.

src/WinRT.Runtime/Marshalers.cs
manodasanW: Will address in follow up PR with testing.

src/build.cmd
```diff
@@ -2,7 +2,7 @@
 if /i "%cswinrt_echo%" == "on" @echo on
 
 set CsWinRTBuildNetSDKVersion=8.0.303
-set CsWinRTBuildNet8SDKVersion=8.0.303
+set CsWinRTBuildNet9SDKVersion=9.0.100
```
manodasanW, line 5:
```suggestion
set CsWinRTBuildNet9SDKVersion=9.0.201

```

PR #1849, opened 2024-10-26
Add new 'IObjectReference' tracking APIs

Sergio0694:
Another take on #1474. This ports and adapts [`ReferenceTracker` from ComputeSharp](https://github.com/Sergio0694/ComputeSharp/blob/main/src/ComputeSharp/Graphics/Interop/ReferenceTracker.cs).
Main benefits:
- Safe handling of native resources
- Modern API surface matching .NET naming conventions for unsafe APIs (ie. "Unsafe" suffix)
- Enable any number of concurrent callers as well as multiple/concurrent `Dispose()` calls

This PR only adds the new APIs. We can update all internal and generated uses in a follow up.

### DIFF
--- a/src/WinRT.Runtime/ObjectReference.ReferenceTracking.cs
+++ b/src/WinRT.Runtime/ObjectReference.ReferenceTracking.cs
@@ -41,7 +41,7 @@ namespace WinRT
         /// </para>
         /// </remarks>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public IntPtr GetThisRefUnsafe()
+        public IntPtr GetThisPtrUnsafe()
         {
             return GetThisPtrForCurrentContextUnsafe();
         }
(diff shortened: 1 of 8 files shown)


src/WinRT.Runtime/ObjectReference.ReferenceTracking.cs
```diff
@@ -0,0 +1,240 @@
...
+        /// <summary>
+        /// Gets the underlying pointer owned by the current instance.
+        /// </summary>
+        /// <returns>The underlying pointer owned by the current instance.</returns>
+        /// <remarks>
+        /// This method does not check for disposal, nor does it increment the managed reference count of
+        /// the current object. Callers must call <see cref="AddRefUnsafe"/> and <see cref="ReleaseUnsafe"/>.
+        /// </remarks>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public IntPtr GetRefUnsafe()
```
jlaanstra, line 38: Given that we have apis like Unsafe.AsRef, probably want to use a different name here than "Ref".

manodasanW:
Probably want to retarget staging/2.3

Sergio0694:
Yup! Will retarget and rebase after #1843 is merged 🙂

Sergio0694:
This is actually not dependent on the .NET 9 target, so I've just rebased, retargeted and published it 😄

Sergio0694:
Closing this for now, moved to 3.0.

PR #1854, opened 2024-10-29
Support [GeneratedBindableCustomProperty] on structs and records

Sergio0694:
This PR enables using `[GeneratedBindableCustomProperty]` on record and struct types as well.
For struct types, it allows binding to boxed value types coming from some properties in XAML.

### DIFF
--- a/src/Tests/AuthoringTest/Program.cs
+++ b/src/Tests/AuthoringTest/Program.cs
@@ -222,10 +222,15 @@ namespace AuthoringTest
     }
 
     [GeneratedBindableCustomProperty]
-    public partial record struct CustomPropertyRecordStructType
+    internal partial record struct CustomPropertyRecordStructType
     {
         public int Number => 4;
         public string Value => "CsWinRTFromRecordStructType";
+    }
+
+    public static class CustomPropertyRecordStructTypeFactory
+    {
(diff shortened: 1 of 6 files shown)


src/Tests/AuthoringTest/Program.cs
```diff
@@ -207,6 +207,13 @@ public sealed partial class CustomProperty
         public string Value => "CsWinRT";
     }
 
+    [GeneratedBindableCustomProperty]
+    public sealed partial struct CustomPropertyStructType
```
manodasanW, line 211: Should we add a private / internal record which we return as a boxed value which we then do ICustomProperty on?  Or at least a compile time record test case.

Sergio0694: WhyThoGIF

manodasanW:
[  FAILED  ] 3 tests, listed below:
[  FAILED  ] AuthoringTest.GeneratedCustomPropertyStructType
[  FAILED  ] AuthoringTest.GeneratedCustomPropertyRecordType
[  FAILED  ] AuthoringTest.CustomPropertyRecordStructTypeFactoryAndICPP

manodasanW (approved):

PR #1855, opened 2024-10-29
Exclude System GuidAttribute from WinMD.

### LINKED ISSUE #1852: Bug: Usage of System.Runtime.InteropServices.GuidAttribute in WinMD
### Description

When setting a GUID on public types, CsWinRT will include the GUID on the type in the WinMD. While a very nice feature that IDL based WinMDs can't do, because the GUID is added using the `System.Runtime.InteropServices.GuidAttribute` type, the WinMD isn't valid and causes warnings when referenced by things that don't understand it (like .NET Native). I would vote for keeping the GUID, just using `Windows.Foundation.Metadata.GuidAttribute` instead.

### Steps To Reproduce

1. Create a public type in a CsWinRT Authoring project
2. Add an explicit GUID for the public class
3. Reference the generated WinMD in a UWP project

### Expected Behavior

Either:
1. Don't include the GUID
2. Use `Windows.Foundation.Metadata.GuidAttribute` instead of `System.Runtime.InteropServices.GuidAttribute`

### Version Info

CsWinRT: 2.1.6
.NET SDK: 9.0.100-rc.2.24474.11
Windows SDK: 10.0.26100.0

### Additional Context

_No response_

dongle-the-gadget:
Fixes #1852.

Though we probably should be differentiating WinRT attributes and exclude non-WinRT attributes directly rather than hardcoding specific names like this.


hez2010:
I think we should map `System.Runtime.InteropServices.GuidAttribute` to `Windows.Foundation.Metadata.GuidAttribute` instead of dropping it.

dongle-the-gadget:
We already did in another PR, however that one didn't strip away the System GuidAttribute from the WinMD, causing us to generate an invalid WinMD.

manodasanW (approved):

manodasanW:
CI passed, thanks for this fix.

PR #1856, opened 2024-10-29
Fix exception propagation

### LINKED ISSUE #1845: Bug: ObjectDisposedException thrown during crash, corrupting crash report
### Description

There's an exception being thrown in the [exception handling](https://github.com/microsoft/CsWinRT/blob/43e8172ebcb71cb1ee3ed342165bebf8d2f25e08/src/WinRT.Runtime/Projections/EventHandler.cs#L295):
```
System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'ObjectReference'.
  ?, in void IObjectReference.ThrowIfDisposed()
  ?, in T IObjectReference.AsType<T>()
  ?, in int ExceptionHelpers.GetHRForException(Exception ex)
  ?, in int EventHandler<T>.Do_Abi_Invoke<TAbi>(Void* thisPtr, IntPtr sender, TAbi args)
```

While handling the original exception this ObjectDisposedException is being thrown. The original exception is a LayoutCycleException.

We've made a very hacky and risky workaround in order to get the original exception:
Image

Essentially, in [WinRT.Runtime's ExceptionHelper.cs line 364](https://github.com/microsoft/CsWinRT/blob/43e8172ebcb71cb1ee3ed342165bebf8d2f25e08/src/WinRT.Runtime/ExceptionHelpers.cs#L364) an object is read from the `Exception.Data` collection and is used in a using, which **disposes** it. The same object is read again later and since it was disposed, it throws the `ObjectDisposedException`.

I believe the using statement in [ExceptionHelper.cs at line 364](https://github.com/microsoft/CsWinRT/blob/43e8172ebcb71cb1ee3ed342165bebf8d2f25e08/src/WinRT.Runtime/ExceptionHelpers.cs#L364) should be removed. While I don't know why it would be needed, it probably should not be disposing this object, as it doesn't own it, and something may later read it (as proved by this bug).

It also may be wise to look into preventing the exception handling from throwing another exception.

### Steps To Reproduce

Unfortunately, we do not have a simple repro. It's a very complex app with very specific steps.

### Expected Behavior

The original exception should crash the app not an ObjectDisposedException.

### Version Info

.NET 8 (and 7) with WinUI (Microsoft.WindowsAppSDK 1.6.240923002) on Windows 11

### Additional Context

_No response_

manodasanW:
- Fixing `ObjectDisposedException` thrown during exception propagation when the language exception in the error info is a non C# exception.
- Added various test cases to test the various exception propagation scenarios

Putting this in draft as I plan to have another change to improve error propagation

Fixes #1845


src/WinRT.Runtime/ExceptionHelpers.cs
```diff
@@ -512,14 +507,14 @@ public static Exception AttachRestrictedErrorInfo(Exception e)
 
                     if (restrictedErrorInfoPtr != IntPtr.Zero)
                     {
-                        ObjectReference<IUnknownVftbl> restrictedErrorInfoRef = ObjectReference<IUnknownVftbl>.Attach(ref restrictedErrorInfoPtr, IID.IID_IRestrictedErrorInfo);
+                        using ObjectReference<IUnknownVftbl> restrictedErrorInfoRef = ObjectReference<IUnknownVftbl>.Attach(ref restrictedErrorInfoPtr, IID.IID_IRestrictedErrorInfo);
```
Sergio0694, line 510: Can we skip allocating this object reference entirely and just pass `restrictedErrorInfoPtr` to `GetErrorDetails` below? I mean it seems this object reference is only ever used in this scope anyway, so perhaps it's unnecessary?

src/WinRT.Runtime/ExceptionHelpers.cs
```diff
@@ -403,6 +433,20 @@ public static unsafe void SetErrorInfo(Exception ex)
...
                         }
                     }
+                    else if (isLanguageException)
+                    {
+                        if (ex.Data.Contains("__RestrictedLanguageExceptionObjectReference") &&
+                            ex.Data["__RestrictedLanguageExceptionObjectReference"] is ObjectReference<IUnknownVftbl> objRef)
+                        {
+                            // Add ref as we are going to pass the responsibility for keeping the exception alive
+                            // to the error info.
+                            Marshal.AddRef(objRef.ThisPtr);
```
Sergio0694, line 443: This isn't clear to me, why do we need to addref here? Wouldn't the error info have already done an addref on the same pointer when we originally passed it to it? Am I missing some context?

manodasanW: - The way this works is we first originate an error with the language exception.  At this point, the error info has a ref on the CCW of the exception.
- The error then propagates to the native side.
- Eventually lets say it comes back to C#.  We retrieve the error info and get the exception from it.  We then throw it which can go across threads and so on.  At the same time, if this exception were to go back to native code, we want to be able to set the same error info again so we store a reference to it in the exception data.  Now we got a cycle.  The error info keeps the exception CCW ref counted and the exception keeps the error info alive.
- To fix this, we release the ref held on the CCW of the exception by the error info.  Given the exception is now alive in C# code and is being thrown, that keeps it alive.
- Now when we come back to the ABI and need to set the error info again, we add ref again to make sure the error info keeps the exception alive.  At the same time, we remove our stored error info to remove the cycle.

manodasanW: Removed the cycle handling logic as there is the scenario at the end where error info is freed for a caught exception, it will try to free the language exception and fail to.  We can in theory maybe implement our own IUnknown and try to handle this, but for now I won't try to handle the cycle as it is just an exception object in certain cases.

src/WinRT.Runtime/ExceptionHelpers.cs
manodasanW: Addressed along with a couple other allocations.

jamers99:
Once this is in master, how soon will this be available in a nuget package?

src/WinRT.Runtime/ExceptionHelpers.cs
```diff
@@ -151,68 +151,75 @@ private static Exception GetExceptionForHR(int hr, bool useGlobalErrorState, out
...
+                                errorMessage += restrictedError;
+                            }
                         }
                     }
                 }
+                catch (Exception e)
+                {
+                    // If we fail to get the error info or the exception from it,
+                    // we fallback to using the hresult to create the exception.
+                    Debug.Assert(false, e.Message, e.StackTrace);
```
Sergio0694, line 217: Is this catch (and the one in the other helper below) expected to only come from a specific native call? I'm wondering if, assuming that's the case, we can just check the HRESULT instead and go to the fallback path, rather than having it throw and catch it, which would be much less efficient. Kinda like the optimization we did in `CreateCcwForMarshalingForABI` (or whatever that was called). Thoughts?

manodasanW: This is a catch all for we shouldn't accidentally throw a diff exception from what we are actually trying to throw.  i.e. previously when we were trying to get ThisPtr, we ran into ObjectDisposedExcepton and that replaced the exception we were throwing.  Instead this should be best-effort and if we fail, we still throw the original exception based on hr.

Sergio0694: That makes sense. I'm just wondering what cases could possibly still be there? Like, if we did end up in such a scenario, would we not watch to detect that as soon as possible, as that would indicate a bug in CsWinRT? Or, could that not indicate some potentially bigger issue (like, something breaking somewhere) that we wouldn't want to hide?

manodasanW: yea I was debating that which is why I put in a Debug.Assert.  At the same time, we still probably shouldn't replace the original exception as that just makes it harder to diagnose the app issue.  The worst case here is app may get less exception context and I think we would get feedback on that as we have in the past if something goes wrong.

Sergio0694: I guess my concerns are:
- How can we be sure if we got some other exception here that we just ignore, the process is in a valid state?
- How would we ever detect such scenarios if all release builds of CsWinRT will just ignore the exception?

manodasanW: Given this function is specific to error propagation rather than a core functionality, we are saying that even if it is due to the process being in a bad state, we shouldn't hide the original issue which could be what leads to that.  This is technically similar to how the OS error propagation APIs work as they are best effort for diagnostics.

We are saying that if there is an actual issue in error propagation implementation of CsWinRT, we would need to catch it with our tests which is now more thorough, and we would also learn from crash dumps losing diagnostic context.

manodasanW: Also the other thing is when we throw other unrelated exceptions during exception propagation, it can become misleading and confusing what is happening.

Sergio0694: That seems fair. Idea, could we perhaps add the inner exception to a well-known item in `Data` in these cases? So that it's not lost entirely, and it'd be able to get info out of it if needed? Rather than just completely swallowing it.

manodasanW: I think, we might be able to do that.

manodasanW: Added it for this scenario.  The others we probably can't do much as that happens when converting to a native error.  But those are less complicated too and less likely to be an issue that we don't catch in our tests.

PR #1865, opened 2024-11-09
Stop relying on UB in 'IContextCallback' dispatch logic

Sergio0694:
See https://github.com/dotnet/runtime/blob/main/docs/design/specs/Memory-model.md#cross-thread-access-to-local-variables. This PR fixes accessing the callback state (which is a managed object) from the native callback invoked on the target thread.

### DIFF
--- a/src/WinRT.Runtime/Interop/IContextCallback.cs
+++ b/src/WinRT.Runtime/Interop/IContextCallback.cs
@@ -70,6 +70,11 @@ namespace ABI.WinRT.Interop
 
             int hresult;
 
+            // We use a thread local static field to efficiently store the state that's used by the callback. Note that this
+            // is safe with respect to reentrancy, as the target callback will never try to switch back on the original thread.
+            // We're only ever switching once on the original context, only to release the object reference that is passed as
+            // state. There is no way for that to possibly switch back on the starting thread. As such, using a thread static
+            // field to pass the state to the target context (we need to store it somewhere on the managed heap) is fine.
             fixed (object* statePtr = &CallbackData.PerThreadObject)
             {
                 CallbackData callbackData;
--- a/src/WinRT.Runtime/Interop/IContextCallback.cs
+++ b/src/WinRT.Runtime/Interop/IContextCallback.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using WinRT;
 using WinRT.Interop;
@@ -18,11 +19,36 @@ namespace ABI.WinRT.Interop
     }
 
 #if NET && CsWinRT_LANG_11_FEATURES
-    internal unsafe struct CallbackData
+    internal sealed unsafe class CallbackData
     {
+        [ThreadStatic]
+        private static CallbackData TlsInstance;
+
         public delegate*<object, void> Callback;
         public object State;
+        public GCHandle Handle;
+
+        private CallbackData()
+        {
+            // Create a handle to access the object from a native callback invoked on another thread.
+            // The handle is weak to ensure that the object does not leak (or it would keep itself
+            // alive). The target is guaranteed to be alive because callers will use 'GC.KeepAlive'.
+            Handle = GCHandle.Alloc(this, GCHandleType.Weak);
+        }
+
+        ~CallbackData()
+        {
+            Handle.Free();
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static CallbackData GetOrCreate()
+        {
+            return TlsInstance ??= new CallbackData();
+        }
     }
+
+
 #endif
 
 #if NET && CsWinRT_LANG_11_FEATURES
@@ -39,25 +65,18 @@ namespace ABI.WinRT.Interop
             comCallData.dwDispid = 0;
             comCallData.dwReserved = 0;
 
-            CallbackData callbackData;
-            callbackData.Callback = callback;
-            callbackData.State = state;
+            CallbackData callbackData = CallbackData.GetOrCreate();
 
-            // We can just store a pointer to the callback to invoke in the context,
-            // so we don't need to allocate another closure or anything. The callback
-            // will be kept alive automatically, because 'comCallData' is address exposed.
-            // We only do this if we can use C# 11, and if we're on modern .NET, to be safe.
-            // In the callback below, we can then just retrieve the Action again to invoke it.
-            comCallData.pUserDefined = (IntPtr)(void*)&callbackData;
-            
+            comCallData.pUserDefined = GCHandle.ToIntPtr(callbackData.Handle);
+
             [UnmanagedCallersOnly]
             static int InvokeCallback(ComCallData* comCallData)
             {
                 try
                 {
-                    CallbackData* callbackData = (CallbackData*)comCallData->pUserDefined;
+                    CallbackData callbackData = Unsafe.As<CallbackData>(GCHandle.FromIntPtr(comCallData->pUserDefined).Target);
 
-                    callbackData->Callback(callbackData->State);
+                    callbackData.Callback(callbackData.State);
 
                     return 0; // S_OK
                 }
@@ -76,6 +95,11 @@ namespace ABI.WinRT.Interop
                 &iid,
                 /* iMethod */ 5,
                 IntPtr.Zero);
+
+            // This call is critical to ensure that the callback data is kept alive until we get here.
+            // This prevents its finalizer to run (that finalizer would free the GC handle used in the
+            // native callback to get back the target callback data that contains the dispatch parameters).
+            GC.KeepAlive(callbackData);
 
             if (hresult < 0)
             {
--- a/src/WinRT.Runtime/Interop/IContextCallback.cs
+++ b/src/WinRT.Runtime/Interop/IContextCallback.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
+using System.Threading;
 using WinRT;
 using WinRT.Interop;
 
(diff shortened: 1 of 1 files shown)


src/WinRT.Runtime/Interop/IContextCallback.cs
```diff
@@ -39,25 +65,18 @@ public static void ContextCallback(IntPtr contextCallbackPtr, delegate*<object,
...
+            CallbackData callbackData = CallbackData.GetOrCreate();
 
-            // We can just store a pointer to the callback to invoke in the context,
-            // so we don't need to allocate another closure or anything. The callback
-            // will be kept alive automatically, because 'comCallData' is address exposed.
-            // We only do this if we can use C# 11, and if we're on modern .NET, to be safe.
-            // In the callback below, we can then just retrieve the Action again to invoke it.
-            comCallData.pUserDefined = (IntPtr)(void*)&callbackData;
-            
+            comCallData.pUserDefined = GCHandle.ToIntPtr(callbackData.Handle);
```
hamarb123, line 70: As discussed in Discord, you should be able to do it without GCHandle like so (you can fix up the comments, etc.):
```diff
#if NET && CsWinRT_LANG_11_FEATURES
    internal unsafe struct CallbackData
    {
+       [ThreadStatic] public static object PerThreadObject = null!;
        public delegate*<object, void> Callback;
-       public object State;
+       public object* StatePtr;
    }
#endif

#if NET && CsWinRT_LANG_11_FEATURES
    internal unsafe struct IContextCallbackVftbl
    {
#pragma warning disable CS0649 // Native layout
        private global::WinRT.Interop.IUnknownVftbl IUnknownVftbl;
        private delegate* unmanaged[Stdcall]<IntPtr, IntPtr, ComCallData*, Guid*, int, IntPtr, int> ContextCallback_4;
#pragma warning restore CS0649

        public static void ContextCallback(IntPtr contextCallbackPtr, delegate*<object, void> callback, delegate*<object, void> onFailCallback, object state)
        {
            ComCallData comCallData;
            comCallData.dwDispid = 0;
            comCallData.dwReserved = 0;

+           CallbackData.PerThreadObject = state; //you may need some sort of barrier/volatile here depending on what memory ordering the later call provides
+           fixed (object* ptr = &CallbackData.PerThreadObject)
+           {
            CallbackData callbackData;
            callbackData.Callback = callback;
+           callbackData.StatePtr = ptr;
-           callbackData.State = state;

            // We can just store a pointer to the callback to invoke in the context,
            // so we don't need to allocate another closure or anything. The callback
            // will be kept alive automatically, because 'comCallData' is address exposed.
            // We only do this if we can use C# 11, and if we're on modern .NET, to be safe.
            // In the callback below, we can then just retrieve the Action again to invoke it.
            comCallData.pUserDefined = (IntPtr)(void*)&callbackData;
            
            [UnmanagedCallersOnly]
            static int InvokeCallback(ComCallData* comCallData)
            {
                try
                {
                    CallbackData* callbackData = (CallbackData*)comCallData->pUserDefined;

-                   callbackData->Callback(callbackData->State);
+                   callbackData->Callback(*callbackData->StatePtr);

                    return 0; // S_OK
                }
                catch (Exception e)
                {
                    return e.HResult;
                }
            }

            Guid iid = IID.IID_ICallbackWithNoReentrancyToApplicationSTA;

            int hresult = (*(IContextCallbackVftbl**)contextCallbackPtr)->ContextCallback_4(
                contextCallbackPtr,
                (IntPtr)(delegate* unmanaged<ComCallData*, int>)&InvokeCallback,
                &comCallData,
                &iid,
                /* iMethod */ 5,
                IntPtr.Zero);
+           }
+           CallbackData.PerThreadObject = null!; //you may need some sort of barrier/volatile here depending on what memory ordering the above call provides

            if (hresult < 0)
            {
                if (onFailCallback is not null)
                {
                    onFailCallback(state);
                }
            }
        }
    }
```

Sergio0694: Mmh I mean I like the idea of not needing a `GCHandle` and the finalizer and the `GC.KeepAlive` call, but this feels even trickier and I'd like someone else to double check that it would be correct and not relying on some other UB of some kind. @jkotas if you have a minute would you mind taking a look and sharing your thoughts? Thank you! 😅

src/WinRT.Runtime/Interop/IContextCallback.cs
```diff
@@ -18,11 +19,36 @@ internal struct ComCallData
...
+            Handle = GCHandle.Alloc(this, GCHandleType.Weak);
+        }
+
+        ~CallbackData()
+        {
+            Handle.Free();
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static CallbackData GetOrCreate()
```
jkotas, line 45: Are you sure that there cannot ever be recursion that would cause multiple recursive uses of CallbackData on the same thread?

Calls between STA apartments always come with a lot of surprises and corner-case bugs...

src/WinRT.Runtime/Interop/IContextCallback.cs
jkotas: I do not see a problem with this suggestion.

jkotas:
> Fix a GC hole

Nit: The potential undefined behavior that this is fixing does not fit the definition of GC hole (https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/clr-code-guide.md#211-how-gc-holes-are-created).

src/WinRT.Runtime/Interop/IContextCallback.cs
Sergio0694: Applied the suggested changes and also left some notes about reentrancy. I talked with @manodasanW as well and in this scenario this should be fine. We're only ever doing this context switch to release an object reference on its original context, and nothing else. There is no way that should/could then somehow switch back to this starting thread (it wouldn't even know which one that was) to then try to schedule another context callback recursively. So this should be fine. Also in ASTA scenarios this would also just be flat out impossible by construction (because ASTA blocks reentrant calls into it, so even if hypothetically we somehow ended up there, that second context callback would just throw before being scheduled). But then again either case should just never happen. Definitely a good point though and worth writing some additional comments about this! 😄

src/WinRT.Runtime/Interop/IContextCallback.cs
```diff
@@ -67,15 +62,42 @@ static int InvokeCallback(ComCallData* comCallData)
...
+            comCallData.dwDispid = 0;
+            comCallData.dwReserved = 0;
+
+            CallbackData.PerThreadObject = state;
+
+            int hresult;
+
+            // We use a thread local static field to efficiently store the state that's used by the callback. Note that this
+            // is safe with respect to reentrancy, as the target callback will never try to switch back on the original thread.
+            // We're only ever switching once on the original context, only to release the object reference that is passed as
```
jkotas, line 75: Release can run arbitrary code, is that right? I do not see what guarantees that the arbitrary code cannot switch threads at will.

Also, this seems to be used for more than just releasing the object reference: https://github.com/microsoft/CsWinRT/blob/93482af78df5c153ddc709b027c342b90a83ea86/src/WinRT.Runtime/ObjectReference.cs#L751-L760

It would be more correct by construction to implement this as usual pool - clear the cached instance when it is rented, so that there is no way for one instance to be rented multiple times at the same time.

hamarb123: I would have assumed that `IContextCallback::ContextCallback` with `ICallbackWithNoReentrancyToApplicationSTA` means it shouldn't be able to run any arbitrary code on this thread as a result of that call, and nothing else from when we set the field to when we clear it seems problematic. If that thing doesn't guarantee this however, then something should be done to allow multiple current things at once potentially existing (e.g., using a `List<object>` or similar would do the job).

hamarb123: Or do you mean the callback might call back into this thread & somehow end up back here? That would make sense I suppose. Is that possible @Sergio0694?

Sergio0694: > "as long as the callback is called before anything else silly happens, since it immediately reads the field

That's actually a good point. Even if we somehow recursively ended up here (which I don't believe is possible), each callback would've already read the target state before invoking the user-provided callback anyway.

> "Release can run arbitrary code"

Not really, I mean yes from the point of view of this API, but the context callback is only ever used internally to release `IObjectReference` objects, which are implemented in CsWinRT only. And we're only ever using these to pass the `Release` pointers which simply do a release on the tracker ref and the native object on the target context:

https://github.com/microsoft/CsWinRT/blob/93482af78df5c153ddc709b027c342b90a83ea86/src/WinRT.Runtime/ObjectReference.cs#L955-L967

hamarb123 (approved):
This looks fine now (assuming we don't need volatile read in the callback), assuming it's not possible to come back into `ContextCallback` while another one is executing on the same thread.

Sergio0694:
Closing this for now, since the current behavior is technically not a GC hole and we're not actually sure the new code is safe (or even saf_er_) either, since technically once we wait for the callback COM might very well start pumping arbitrary messages on our thread too, in theory. We can revisit this for 3.0 if we notice this is causing any problems or find a better way.

hamarb123:
Yeah, it's not a gc hole, but it is UB which iirc can produce problematic codegen in real cases. Ito handling potentially pumping arbitrary messages, you could use the static field if it's `null` & fallback to just using a normal array otherwise & pin it for the duration of the time (i.e., take the current logic, but add extra handling for "can use static field" based on if it's null, and just allocate & use a `new object[1]` if it's not `null` instead of the static field).

Sergio0694:
I'll port this to the 3.0 branch and first run it with some assertions to see if we ever hit that case. I could also make a build with an assertion for 2.x and use it in the Store to get a better real world idea, I wonder how common it is. As in, if it never hits in our tests nor in the Store, then perhaps the fact the fallback is a bit slower wouldn't matter.

I do have another question though: is there no way to just keep using the current code structure (i.e. with the value being in a local) and just make that safe? If the issue is just that writing that to a local doesn't trigger the GC memory barrier, is there a way to just force it? Then after that once you have a pointer, it shouldn't matter where the location is, no?

hamarb123:
> If the issue is just that writing that to a local doesn't trigger the GC memory barrier, is there a way to just force it?

The issue is that it's UB, which means today it might do one thing, and tomorrow a new optimization could mean it does something else. You could probably make it "work", just like how people can make `Unsafe.As<IncorrectType>(...)` "work" in some cases, but it would still be UB to rely on it doing what you think it does & could break at any point in theory (although some setups are obviously less likely to be able to break than others).

The obvious way around this is to just make it not UB & actually allow it (and get the runtime to handle it properly) for .NET 11/12+, but that's ofc not up to me lol (but I would like to see it personally).

PR #1866, opened 2024-11-09
Replace null pointer function pointer call with helpful exception.

manodasanW:
Previously in these scenarios you would have got a null reference exception which may not have been obvious why when looking at how the stack is reported in VS.  Now you will instead get a helpful exception.


Sergio0694 (approved):
Pushed just some minor tweaks, but love this change! Thank you 🙂

PR #1868, opened 2024-11-12
Fix AnalyzerReleases.Shipped typo

dotMorten:
Code wasn't correctly escaped.

Before:
image

After:
image


Sergio0694 (approved):
Thank you! 🙂

manodasanW (approved):
Good catch, thanks.

PR #1873, opened 2024-11-18
Preserve type metadata for casts to WinRT runtime classes.

### LINKED ISSUE #1860: Bug: Cannot cast some of the UWP activation arguments using dotnet as operator
### Description

When publishing a UWP project using the preview .net 9 support with AOT, I noticed a crash when an appservice from a win32 helper tries to connect. It turns out that casting the argument (`args.TaskInstance.TriggerDetails`) in the `BackgroundActivated(BackgroundActivatedEventArgs args)` activation method to `AppServiceTriggerDetails` fails using the dotnet `as` operator. However, it works using `args.TaskInstance.TriggerDetails.As<AppServiceTriggerDetails>()`.

### Steps To Reproduce

A project that uses background activation through an appservice with a win32 helper is needed. The project also has to be published with AOT in release mode by creating an msix package (it sometimes worked in Debug configuration).

### Expected Behavior

Casting using dotnet operator should work

### Version Info

cswinrt 2.1.16
dotnet 9 rc2

### Additional Context

_No response_

dongle-the-gadget:
Fixes #1860.

> [!NOTE]
> You should disable whitespaces before looking at the changes.

### DIFF
--- a/src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
+++ b/src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
@@ -223,7 +223,7 @@ namespace Generator
                     }
 
                     // Avoid cases where the type to be cast from is unknown, or can be done purely through static metadata.
-                    // That is, the type to be cast to inherits from the type of the expression to be cast,
+                    // That is, the type of the expression inherits from the type to be cast to,
                     // as we know the cast will always work.
                     var sourceType = context.SemanticModel.GetTypeInfo(expression).Type;
                     if (sourceType == null || GeneratorHelper.IsDerivedFromType(sourceType, namedTypeSymbol))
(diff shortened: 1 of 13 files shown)


dongle-the-gadget:
I added a test in, though I didn't actually check it no further than making sure the C++ generates the correct WinMD (since my machine is very limited in spec and I don't want to slam it).
I did manage to test the source generator using another project by casting a XAML resource into a `Style` using cast expressions and that did work.

hez2010:
I don't think this can be easily handled in a source generator. 
The type cast can happen in anywhere especially now we have pattern matching, for example,

```cs
if (obj is { A: Foo { B: Bar(_, Baz { C: [WinRTType x, ..] }) })
{
    // use x here
}
```

We should instead doing this at IL level so that we only need to look for `isinst`.

Sergio0694:
Closing this, we want to support this at the linker/ILC level in .NET 10, via the new interop type mapping system.
Tracking this specific scenario here: https://github.com/dotnet/runtime/issues/112287.

PR #1875, opened 2024-11-21
Remove .NET 6 target from all projects

Sergio0694:
This PR removes .NET 6 from all projects in the CsWinRT repo.


src/build.cmd
```diff
@@ -1,7 +1,7 @@
 @echo off
 if /i "%cswinrt_echo%" == "on" @echo on
 
-set CsWinRTBuildNetSDKVersion=6.0.424
+set CsWinRTBuildNetSDKVersion=8.0.303
 set CsWinRTBuildNet8SDKVersion=8.0.303
```
manodasanW, line 5: technically we are installing it twice now, but if that is later going to change to .NET 9 that might be fine.

src/cswinrt/main.cpp
```diff
@@ -34,7 +34,7 @@ namespace cswinrt
         { "include", 0, option::no_max, "<prefix>", "One or more prefixes to include in projection" },
         { "exclude", 0, option::no_max, "<prefix>", "One or more prefixes to exclude from projection" },
         { "addition_exclude", 0, option::no_max, "<prefix>", "One or more namespace prefixes to exclude from the projection additions" },
-        { "target", 0, 1, "<net8.0|net7.0|net6.0|netstandard2.0>", "Target TFM for projection. Omit for compatibility with .NET 6." },
+        { "target", 0, 1, "<net8.0|net7.0|net6.0|netstandard2.0>", "Target TFM for projection. Omit for compatibility with .NET 8." },
```
manodasanW, line 37: We should drop .NET 7 and 6 here right?

manodasanW:
Looks like both host tests and wux tests are failing.

I think the json in both needed to be updated: https://github.com/microsoft/CsWinRT/blob/e028e15955579f041f4726c8fea6c280b157b5c0/src/Tests/AuthoringWuxConsumptionTest/WinRT.Host.runtimeconfig.json#L7

src/build.cmd
Sergio0694: Yeah I'm going to merge the .NET 9 PR right after this, so I figured we could just leave it 🙂

src/cswinrt/main.cpp
Sergio0694: Done! I've also removed `net7_0_or_greater`, since it's now redundant 👍

src/cswinrt/main.cpp
```diff
@@ -34,7 +34,7 @@ namespace cswinrt
         { "include", 0, option::no_max, "<prefix>", "One or more prefixes to include in projection" },
         { "exclude", 0, option::no_max, "<prefix>", "One or more prefixes to exclude from projection" },
         { "addition_exclude", 0, option::no_max, "<prefix>", "One or more namespace prefixes to exclude from the projection additions" },
-        { "target", 0, 1, "<net8.0|net7.0|net6.0|netstandard2.0>", "Target TFM for projection. Omit for compatibility with .NET 6." },
+        { "target", 0, 1, "<net8.0|net7.0|net6.0|netstandard2.0>", "Target TFM for projection (.NET 8 is the default)" },
```
manodasanW, line 37: I was actually referring to this before.

```suggestion
        { "target", 0, 1, "<net8.0|netstandard2.0>", "Target TFM for projection (.NET 8 is the default)" },
```

src/cswinrt/main.cpp
```diff
@@ -94,12 +94,16 @@ Where <spec> is one or more of:
 
         settings.verbose = args.exists("verbose");
         auto target = args.value("target");
-        if (!target.empty() && target != "netstandard2.0" && !starts_with(target, "net5.0") && !starts_with(target, "net6.0") && !starts_with(target, "net7.0") && !starts_with(target, "net8.0"))
+        if (!target.empty() && target != "netstandard2.0" && !starts_with(target, "net8.0"))
```
manodasanW, line 97: I was debating whether we wanted to do this given for some reason we left net5 before which I don't recall why.  It was probably to stage dropping support.  But due to the other simplifications to it led to in code_writers, it is probably fine.

src/cswinrt/main.cpp
Sergio0694: Oops I'm blind ahah

manodasanW:
This still says .NET 6

https://github.com/microsoft/CsWinRT/blob/5af8521002929f39939d51c6010681517aacfbe9/src/Tests/HostTest/RuntimeFrameworkVersion.csproj#L11

manodasanW:
The above might explain

>[  FAILED  ] HostTest.ProbeByClass
[  FAILED  ] HostTest.ProbeByHost
[  FAILED  ] HostTest.MappedTarget

manodasanW:
Those 3 still seem to be failing.

src/WinRT.Runtime/WinRT.Runtime.csproj
```diff
@@ -101,7 +97,7 @@
     <Service Include="{508349b6-6b84-4df5-91f0-309beebad82d}" />
   </ItemGroup>
 
-  <Target Name="_ResolveResolvedMatchingContract" BeforeTargets="ValidateApiCompatForSrc;RunMatchingRefApiCompat" Condition="'$(TargetFramework)' == 'net6.0' or '$(TargetFramework)' == 'net8.0'">
+  <Target Name="_ResolveResolvedMatchingContract" BeforeTargets="ValidateApiCompatForSrc;RunMatchingRefApiCompat" Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net8.0'))">
```
manodasanW, line 100: This needs to be an exact check to avoid issues when we add a new TFM in the future.

```suggestion
  <Target Name="_ResolveResolvedMatchingContract" BeforeTargets="ValidateApiCompatForSrc;RunMatchingRefApiCompat" Condition="'$(TargetFramework)' == 'net8.0'">
```

manodasanW (approved):

PR #1883, opened 2024-12-04
Add .NET 6 EOL message

dongle-the-gadget:
This is to prevent users from accidentally using the .NET Standard 2.0 WinRT.Runtime.

### DIFF
--- a/nuget/Microsoft.Windows.CsWinRT.targets
+++ b/nuget/Microsoft.Windows.CsWinRT.targets
@@ -56,8 +56,8 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
   <Import Project="$(MSBuildThisFileDirectory)Microsoft.Windows.CsWinRT.Embedded.targets" Condition="'$(CsWinRTEmbedded)' == 'true'"/>
   
-  <Target Name="CsWinRTNet5EOL" Condition="$([MSBuild]::GetTargetFrameworkVersion('$(TargetFramework)')) == 5" BeforeTargets="CsWinRTPrepareProjection">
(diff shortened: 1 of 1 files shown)


nuget/Microsoft.Windows.CsWinRT.targets
```diff
@@ -56,8 +56,8 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
   <Import Project="$(MSBuildThisFileDirectory)Microsoft.Windows.CsWinRT.Embedded.targets" Condition="'$(CsWinRTEmbedded)' == 'true'"/>
   
-  <Target Name="CsWinRTNet5EOL" Condition="$([MSBuild]::GetTargetFrameworkVersion('$(TargetFramework)')) == 5" BeforeTargets="CsWinRTPrepareProjection">
-    <Error Text="Support for .NET 5 ended with C#/WinRT 2.0. For .NET 5 support, use C#/WinRT version 1.6.5. See https://github.com/microsoft/CsWinRT/discussions/1232" />
+  <Target Name="CsWinRTNetEOL" Condition="$([MSBuild]::GetTargetFrameworkVersion('$(TargetFramework)')) < 8" BeforeTargets="CsWinRTPrepareProjection">
```
manodasanW, line 59: Was this PR intended to target the [staging/2.3](https://github.com/microsoft/CsWinRT/tree/staging/2.3) branch or is the message already there as seen [here](https://github.com/microsoft/CsWinRT/blob/cea0c3a5919697d99c949a5d539882669c5a7163/nuget/Microsoft.Windows.CsWinRT.targets#L59-L65) what you were also trying to do?

PR #1887, opened 2024-12-13
Add 'CsWinRTEnableManifestFreeActivation' feature switch

Sergio0694:
This PR adds the new "CsWinRTEnableManifestFreeActivation" feature switch (enabled by default), which allows disabling the manifest-free fallback path for WinRT activation. When disabled, all fallback code will be removed, and trying to activate a type that is not registered will throw `NotSupportedException`. This can be useful to:
- Trim more code when manifest free is not needed
- Make it easy to identify ACIDs that are unintentionally missing from the APPX manifest

Ideally, packaged apps should strive to be able to run with this feature switch disabled, for best performance.

### DIFF
--- a/nuget/Microsoft.Windows.CsWinRT.targets
+++ b/nuget/Microsoft.Windows.CsWinRT.targets
@@ -383,7 +383,7 @@ $(CsWinRTInternalProjection)
                                     Value="$(CsWinRTEnableManifestFreeActivation)"
                                     Trim="true" />
 
-    <!-- CSWINRT_ENABLE_MANIFEST_FREE_ACTIVATION switch -->
+    <!-- CSWINRT_MANIFEST_FREE_ACTIVATION_REPORT_ORIGINAL_EXCEPTION switch -->
     <RuntimeHostConfigurationOption Include="CSWINRT_MANIFEST_FREE_ACTIVATION_REPORT_ORIGINAL_EXCEPTION"
                                     Value="$(CsWinRTManifestFreeActivationReportOriginalException)"
                                     Trim="true" />
--- a/nuget/Microsoft.Windows.CsWinRT.targets
+++ b/nuget/Microsoft.Windows.CsWinRT.targets
@@ -328,6 +328,7 @@ $(CsWinRTInternalProjection)
     <CsWinRTEnableICustomPropertyProviderSupport Condition="'$(CsWinRTEnableICustomPropertyProviderSupport)' == ''">true</CsWinRTEnableICustomPropertyProviderSupport>
     <CsWinRTEnableIReferenceSupport Condition="'$(CsWinRTEnableIReferenceSupport)' == ''">true</CsWinRTEnableIReferenceSupport>
     <CsWinRTEnableIDynamicInterfaceCastableSupport Condition="'$(CsWinRTEnableIDynamicInterfaceCastableSupport)' == ''">true</CsWinRTEnableIDynamicInterfaceCastableSupport>
+    <CsWinRTEnableManifestFreeActivation Condition="'$(CsWinRTEnableManifestFreeActivation)' == ''">true</CsWinRTEnableManifestFreeActivation>
 
     <!--
       Note: the 'CsWinRTUseWindowsUIXamlProjections' property and associated 'CSWINRT_USE_WINDOWS_UI_XAML_PROJECTIONS' feature
@@ -374,6 +375,11 @@ $(CsWinRTInternalProjection)
     <!-- CSWINRT_ENABLE_IDYNAMICINTERFACECASTABLE switch -->
     <RuntimeHostConfigurationOption Include="CSWINRT_ENABLE_IDYNAMICINTERFACECASTABLE"
                                     Value="$(CsWinRTEnableIDynamicInterfaceCastableSupport)"
+                                    Trim="true" />
+
+    <!-- CSWINRT_ENABLE_MANIFEST_FREE_ACTIVATION switch -->
+    <RuntimeHostConfigurationOption Include="CSWINRT_ENABLE_MANIFEST_FREE_ACTIVATION"
+                                    Value="$(CsWinRTEnableManifestFreeActivation)"
                                     Trim="true" />
   </ItemGroup>
 
--- a/nuget/Microsoft.Windows.CsWinRT.targets
+++ b/nuget/Microsoft.Windows.CsWinRT.targets
@@ -108,7 +108,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       <CsWinRTCcwLookupTableGeneratorEnabled Condition="'$(CsWinRTCcwLookupTableGeneratorEnabled)' == ''">true</CsWinRTCcwLookupTableGeneratorEnabled>
 
       <!--
-	    CsWinRTAotWarningLevel decides the level of warning from the AOT analyzer.
+        CsWinRTAotWarningLevel decides the level of warning from the AOT analyzer.
             0: No warnings
            *1: Warnings for scenarios involving WinRT types that are not built-in system types mapped to WinRT.
             2: Level 1 warnings and warnings for scenarios involving built-in system types mapped to WinRT.
@@ -122,7 +122,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
                                                         '$(OutputType)' == 'Library' and
                                                         '$(CsWinRTHasAnyIncludes)' == 'true'">0</CsWinRTAotWarningLevel>
       <CsWinRTAotWarningLevel Condition="'$(CsWinRTAotWarningLevel)' == ''">1</CsWinRTAotWarningLevel>
-	</PropertyGroup>
+    </PropertyGroup>
   </Target>
 
   <!-- Remove WinRT.Host.dll and WinRT.Host.Shim.dll references -->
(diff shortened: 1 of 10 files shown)


Sergio0694:
Can confirm this trims everything as expected:

image

Exceptions also work as expected:

image

src/WinRT.Runtime/ApiCompatBaseline.net8.0.txt
```diff
@@ -1,4 +1,6 @@
 Compat issues with assembly WinRT.Runtime:
 CannotChangeAttribute : Attribute 'System.AttributeUsageAttribute' on 'WinRT.GeneratedBindableCustomPropertyAttribute' changed from '[AttributeUsageAttribute(AttributeTargets.Class, Inherited=false, AllowMultiple=false)]' in the contract to '[AttributeUsageAttribute(AttributeTargets.Class | AttributeTargets.Struct, Inherited=false, AllowMultiple=false)]' in the implementation.
 CannotRemoveAttribute : Attribute 'System.ComponentModel.EditorBrowsableAttribute' exists on 'WinRT.WinRTExposedTypeAttribute' in the contract but not the implementation.
-Total Issues: 2
+ParameterNamesCannotChange : Parameter name on member 'WinRT.ActivationFactory.Get(System.String)' is 'activatableClassId' in the implementation but 'typeName' in the contract.
```
manodasanW, line 4: Are these parameter changes really needed if they are technically API compat concerns?

src/WinRT.Runtime/ActivationFactory.cs
```diff
@@ -315,8 +366,37 @@ public static ObjectReference<I> Get<I>(string typeName, Guid iid)
...
+            }
+
+            [MethodImpl(MethodImplOptions.NoInlining)]
+            static Exception GetException(string typeName, int hr)
+            {
+                Exception exception = Marshal.GetExceptionForHR(hr)!;
+
+                if (hr == ExceptionHelpers.REGDB_E_CLASSNOTREG)
+                {
+                    throw new NotSupportedException(
```
manodasanW, line 385: Hmm, I wonder if we still want to instead still report our typical exception for not registered.  What if someone is instantiating a class that is not registered intentionally and have code to catch this specific error and handle it in some way.  Just because it failed doesn't mean manifest free activation would have succeeded if it was enabled.

src/WinRT.Runtime/ActivationFactory.cs
```diff
@@ -315,8 +366,37 @@ public static ObjectReference<I> Get<I>(string typeName, Guid iid)
...
+            }
+
+            // Explicit throw so the JIT can see it and correctly optimize for always throwing paths.
+            // The exception instantiation is in a separate method so that codegen remains separate.
+            throw GetException(typeName, hr);
+        }
+
 #if NET
-        private static IObjectReference GetFromActivationHandler(string typeName, Guid iid)
+        private static IObjectReference? GetFromActivationHandler(string typeName, Guid iid)
```
manodasanW, line 399: Just to confirm, adding nullable annotations here and above are not compat concerns right given nullable wasn't enabled before?

src/WinRT.Runtime/ApiCompatBaseline.net8.0.txt
Sergio0694: Not strictly needed, but it was more for consistency. These would really only be breaking if someone is recompiling from source and happened to have an invocation with a named argument. Otherwise there's no concerns with binary compatibility. I can revert if you want, but I was thinking it should be fine? Let me know!

src/WinRT.Runtime/ActivationFactory.cs
Sergio0694: Correct, this was just nullable oblivious before, so this is adding new info to it, but not breaking.

src/WinRT.Runtime/ActivationFactory.cs
Sergio0694: I would imagine those users would simply not be setting this feature switch in the first place? My concern is that throwing that exception after setting the feature switch results in a very poor developer experience. Especially if (which is very likely) someone sets the feature switch and then someone else, later on, stumbles upon this exception. Throwing `NotSupportedException` is standard practice for things which are blocked by design eg. on AOT (eg. we do the same in several other parts of CsWinRT already), and the exception message would help and tell people exactly what happened. Whereas just getting a random Win32 exception when things were working before would be confusing. What do you think?

manodasanW: I get what you mean, but when we are hitting this here, we don't know whether the error is being thrown because they are trying to use manifest free registration and the DLL is there ready to be activated and we blocked it or if the error is being thrown due to a missing registration in the manifest somewhere.  The latter already happens pretty often and is common.  Also, I am thinking of places where this might actually get used and there is a good chance they may be checking for the error code to handle certain scenarios and changing the contract on this function will make them have to also need to change calling code if they wanted to adopt this.  Even if we kept the old error code, we can keep this exception message to inform that is also possible that it is not registered due to this being disabled.

src/WinRT.Runtime/ApiCompatBaseline.net8.0.txt
manodasanW: My thinking is given the old parameter name wasn't incorrect either, I wouldn't add an API compat change just for this.

Sergio0694: Reverted! 👍

src/WinRT.Runtime/ActivationFactory.cs
Sergio0694: > "we don't know whether the error is being thrown **because they are trying to use manifest free registration and the DLL is there ready to be activated and we blocked it** or if the error is being thrown due to a missing registration in the manifest somewhere."

I am not sure I understand this distinction. This exception is only thrown if users have specifically set the feature switch to disable manifest-free activation. Which means they cannot possibly be trying to use manifest-free activation in the first place, or they wouldn't have set that feature switch at all. The second case you mentioned is the only one that's valid in this scenario: they're trying to activate a type, and forgot to register it in the manifest, which means we're going to throw `NSE` by design, because the feature switch is set. Like, to me it seems like any other combination is explicitly not supported, and would just mean they've set the feature switch incorrectly. Additionally, one of the reasons why I do want `NSE` is that eg. we always set those to always break, as it makes it much much easier to find all bugs due to AOT and similar. If we were just throwing a normal exception here, finding these and debugging them would be much clunkier.

Thoughts?

src/WinRT.Runtime/ActivationFactory.cs
```diff
@@ -190,6 +191,11 @@ public static unsafe (ObjectReference<I> obj, int hr) GetActivationFactory<I>(st
         }
     }
 
+#nullable enable
```
manodasanW, line 194: For same reason as other PR, we probably should add nullable restore at the bottom the file.

Sergio0694: Done!

src/WinRT.Runtime/ActivationFactory.cs
Sergio0694: Updated as discussed offline 🙂

nuget/Microsoft.Windows.CsWinRT.targets
```diff
@@ -375,6 +377,16 @@ $(CsWinRTInternalProjection)
     <RuntimeHostConfigurationOption Include="CSWINRT_ENABLE_IDYNAMICINTERFACECASTABLE"
                                     Value="$(CsWinRTEnableIDynamicInterfaceCastableSupport)"
                                     Trim="true" />
+
+    <!-- CSWINRT_ENABLE_MANIFEST_FREE_ACTIVATION switch -->
+    <RuntimeHostConfigurationOption Include="CSWINRT_ENABLE_MANIFEST_FREE_ACTIVATION"
+                                    Value="$(CsWinRTEnableManifestFreeActivation)"
+                                    Trim="true" />
+
+    <!-- CSWINRT_ENABLE_MANIFEST_FREE_ACTIVATION switch -->
```
manodasanW, line 386:
```suggestion
    <!-- CSWINRT_MANIFEST_FREE_ACTIVATION_REPORT_ORIGINAL_EXCEPTION switch -->
```

nuget/readme.md
```diff
@@ -48,6 +48,8 @@ C#/WinRT behavior can be customized with these project properties:
 | CsWinRTRcwFactoryFallbackGeneratorForceOptIn | true \| *false | Forces the RCW factory fallback generator to be enabled (it only runs on .exe projects by default)  |
 | CsWinRTRcwFactoryFallbackGeneratorForceOptOut | true \| *false | Forces the RCW factory fallback generator to be disabled (overrides "ForceOptIn" as well)  |
 | CsWinRTMergeReferencedActivationFactories | true \| *false | Makes the native `DllGetActivationFactory` exported function for AOT scenarios also forward the activation call to all referenced WinRT components, allowing them to all be merged into a single executable or shared library |
+| CsWinRTEnableManifestFreeActivation | *true \| false | Enables the manifest-free WinRT activation path as fallback when `RoGetActivationFactory` fails to resolve a WinRT type |
+| CsWinRTManifestFreeActivationReportOriginalException | true \| **false | If 'CsWinRTEnableManifestFreeActivation' is set and activating a type fails, ensures the original exception is thrown, rather than 'NotSupportedException' |
```
manodasanW, line 52:
```suggestion
| CsWinRTManifestFreeActivationReportOriginalException | true \| *false | If 'CsWinRTEnableManifestFreeActivation' is set and activating a type fails, ensures the original exception is thrown, rather than 'NotSupportedException' |
```

src/WinRT.Runtime/Configuration/FeatureSwitches.cs
```diff
@@ -145,6 +165,24 @@ public static bool EnableIDynamicInterfaceCastableSupport
...
+            get => GetConfigurationValue(EnableManifestFreeActivationPropertyName, ref _enableManifestFreeActivation, true);
+        }
+
+        /// <summary>
+        /// Gets a value indicating whether or not the original exception should be thrown if activation fails when <see cref="EnableManifestFreeActivation"/> is disabled (defaults to <see langword="false"/>).
+        /// </summary>
+        public static bool ManifestFreeActivationReportOriginalException
+        {
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            get => GetConfigurationValue(ManifestFreeActivationReportOriginalExceptionPropertyName, ref _manifestFreeActivationReportOriginalException, true);
```
manodasanW, line 183:
```suggestion
            get => GetConfigurationValue(ManifestFreeActivationReportOriginalExceptionPropertyName, ref _manifestFreeActivationReportOriginalException, false);
```

manodasanW (approved):

PR #1889, opened 2024-12-20
Make 'CreateIIDForGenericType' zero-alloc

Sergio0694:
Small performance optimization mostly when getting IIDs of enum types

### DIFF
--- a/src/WinRT.Runtime/GuidGenerator.cs
+++ b/src/WinRT.Runtime/GuidGenerator.cs
@@ -363,7 +363,7 @@ namespace WinRT
             wrt_pinterface_namespace.TryWriteBytes(utf8Bytes);
 
             int encodedUtf8BytesWritten = Encoding.UTF8.GetBytes(signature, utf8Bytes[16..]);
-            Span<byte> sha1Bytes = stackalloc byte[SHA1.HashSizeInBytes];
+            Span<byte> sha1Bytes = stackalloc byte[20]; // 'SHA1.HashSizeInBytes' does not exist on .NET 6, update this when that's dropped
 
             // Hash the encoded signature (the bytes written are guaranteed to always match the span)
             _ = SHA1.HashData(utf8Bytes[..(16 + encodedUtf8BytesWritten)], sha1Bytes);
@@ -377,6 +377,7 @@ namespace WinRT
             }
 
             return iid;
+#nullable restore
 #endif
         }
     }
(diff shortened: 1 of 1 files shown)


src/WinRT.Runtime/GuidGenerator.cs
```diff
@@ -342,15 +346,37 @@ internal static Guid CreateIIDForGenericType(string signature)
...
+            int encodedUtf8BytesWritten = Encoding.UTF8.GetBytes(signature, utf8Bytes[16..]);
+            Span<byte> sha1Bytes = stackalloc byte[SHA1.HashSizeInBytes];
+
+            // Hash the encoded signature (the bytes written are guaranteed to always match the span)
+            _ = SHA1.HashData(utf8Bytes[..(16 + encodedUtf8BytesWritten)], sha1Bytes);
+
+            Guid iid = encode_guid(sha1Bytes);
+
+            // Before exiting, make sure to return the array from the pool, if we rented one
+            if (utf8BytesFromPool is not null)
```
manodasanW, line 374: I assume you didn't put this in a try finally because we assume HashData won't throw as we ensure correct size?

manodasanW:
Given no API surface change, any reason not to target master and then do a merge from master to staging if needed.

manodasanW (approved):

src/WinRT.Runtime/GuidGenerator.cs
Sergio0694: Partly because of that (as in, I don't expect this to ever throw), but also because I'm following the security guidance from the .NET folks about not returning pooled arrays from `finally` blocks. The idea being that in case there's an exception, you shouldn't make assumptions on the state of the program (eg. perhaps you started some work somewhere that took a reference to the rented array and that wasn't correctly stopped, because of an exception, etc.), therefore in those cases you should just throw away the array instead, and only return them in normal circumstances (ie. if no exception is thrown).

src/WinRT.Runtime/GuidGenerator.cs
```diff
@@ -342,15 +346,37 @@ internal static Guid CreateIIDForGenericType(string signature)
...
+            {
+                ArrayPool<byte>.Shared.Return(utf8BytesFromPool);
+            }
+
+            return iid;
```
manodasanW, line 379: Also you want to restore the nullable scope at the end here, otherwise we will run into issues in the scenario / repo that merges these files.

PR #1895, opened 2025-01-12
Implement stub .exe support for apps + WinRT components

Sergio0694:
This PR adds support for generating a stub .exe for an app that's merging WinRT components.
We can split the general "stub .exe" feature into two steps:
- This PR covers the stub .exe for the application entry point
- A follow up PR can add support for additional stub .exe-s for custom entry points

This uses a normal binary with standard options.
We can optimize the size to skip CRT in a follow up.

### DIFF
--- a/nuget/Microsoft.Windows.CsWinRT.targets
+++ b/nuget/Microsoft.Windows.CsWinRT.targets
@@ -391,15 +391,6 @@ int wmain(int argc, wchar_t** argv)
     <WriteLinesToFile Lines="$(_StubExeSourceFileLines)" Overwrite="true" WriteOnlyWhenDifferent="true" File="$(_StubExeSourceFilePath)" />
 
     <!--
-      We want to invoke MSVC in the folder where we also want the .exe to be produced in.
-      So we also need to adjust the input paths to be relative to that working directory.
-    -->
-    <PropertyGroup>
-      <_StubExeSourceFilePathRelative>$([MSBuild]::MakeRelative($(_StubExeGeneratedFilesDir), $(_StubExeSourceFilePath)))</_StubExeSourceFilePathRelative>
-      <_StubExeNativeLibraryPathRelative>$([MSBuild]::MakeRelative($(_StubExeGeneratedFilesDir), $(_StubExeNativeLibraryPath)))</_StubExeNativeLibraryPathRelative>
-    </PropertyGroup>
-
-    <!--
       Get the path of 'cl.exe' from the Native AOT tooling, which finds the install folder for MSVC already.
       By doing this, we make sure that this can work from a normal terminal too, not just a VS Developer cmd.
     -->
@@ -489,7 +480,7 @@ int wmain(int argc, wchar_t** argv)
     
     <!-- Invoke MSVC to produce the native executable -->
     <Exec
-      Command='"$(_ClExeFilePath)" "$(_StubExeSourceFilePathRelative)" "$(_StubExeNativeLibraryPathRelative)" $(_StubExeMsvcArgsText)'
+      Command='"$(_ClExeFilePath)" "$(_StubExeSourceFilePath)" "$(_StubExeNativeLibraryPath)" $(_StubExeMsvcArgsText)'
       ConsoleToMsBuild="true"
       WorkingDirectory="$(_StubExeGeneratedFilesDir)" />
     
--- a/nuget/Microsoft.Windows.CsWinRT.targets
+++ b/nuget/Microsoft.Windows.CsWinRT.targets
@@ -316,6 +316,82 @@ $(CsWinRTInternalProjection)
     </ItemGroup>
   </Target>
 
+  <!-- Produces a stub .exe for the application, if requested -->
+  <Target
+    Name="_CsWinRTGenerateStubExe"
+    AfterTargets="Publish"
+    Condition="'$(CsWinRTGenerateStubExe)' == 'true'">
+
+    <PropertyGroup>
+    
+      <!-- Prepare the generated files directory -->
+      <_StubExeFolderName>StubExe</_StubExeFolderName>
+      <_StubExeGeneratedFilesDir Condition="'$(_StubExeGeneratedFilesDir)' == '' AND '$(GeneratedFilesDir)' != ''">$(GeneratedFilesDir)\$(_StubExeFolderName)\</_StubExeGeneratedFilesDir>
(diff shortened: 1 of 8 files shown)


nuget/Microsoft.Windows.CsWinRT.targets
```diff
@@ -316,6 +322,242 @@ $(CsWinRTInternalProjection)
...
+// exported automatically when the 'CustomNativeMain' property is set to 'true'.
+__declspec(dllimport)
+int __stdcall __managed__Main(int argc, wchar_t** argv)%3B
+
+// Our entry point is simply a direct jump into the entry point from Native AOT
+int wmain(int argc, wchar_t** argv)
+{
+    return __managed__Main(argc, argv)%3B
+}]]>
+      </_StubExeSourceFileLines>
```
jkoritzinsky, line 388: Could the CsWinRT package include the stub C file instead of embedding the source in MSBuild?

Sergio0694: Yeah we can do that! I thought we'd need to modify it but now that it's always the same it's fine 🙂

nuget/readme.md
```diff
@@ -48,6 +48,7 @@ C#/WinRT behavior can be customized with these project properties:
 | CsWinRTRcwFactoryFallbackGeneratorForceOptIn | true \| *false | Forces the RCW factory fallback generator to be enabled (it only runs on .exe projects by default)  |
 | CsWinRTRcwFactoryFallbackGeneratorForceOptOut | true \| *false | Forces the RCW factory fallback generator to be disabled (overrides "ForceOptIn" as well)  |
 | CsWinRTMergeReferencedActivationFactories | true \| *false | Makes the native `DllGetActivationFactory` exported function for AOT scenarios also forward the activation call to all referenced WinRT components, allowing them to all be merged into a single executable or shared library |
+| CsWinRTUseEnvironmentalTools | true \| *false | Makes the invocation of MSVC to produce a stub .exe (for Native AOT scenarios) rely on environmental tools, rather than looking up the MSVC install path |
```
manodasanW, line 51: Any reason not to automatically do this as a fallback or so?  Will someone run into this when building from VS or so?

Sergio0694: I'm not sure how to detect whether we're inside a VS Developer Command Prompt. This ensures that things will still build correctly when just doing `dotnet build` too, or `dotnet msbuild`, from a normal terminal. The option matches what Native AOT also has (they have "IlcUseEnvironmentalTools").

jkoritzinsky (approved):

src/Authoring/WinRT.SourceGenerator/Generator.cs
```diff
@@ -295,7 +298,18 @@ public static int DllGetActivationFactory(void* activatableClassId, void** facto
...
             else if (ShouldEmitCallToTryGetDependentActivationFactory(context))
             {
-                builder.AppendLine("IntPtr obj = TryGetDependentActivationFactory(fullyQualifiedTypeName);");
+                builder.AppendLine("""
+                                   IntPtr obj = TryGetDependentActivationFactory(fullyQualifiedTypeName);
+                    """);
+                builder.AppendLine();
+            }
+            else
+            {
```
manodasanW, line 307: Good catch with this, I was actually debugging a scenario earlier today that was resulting in a compiler error here.  Still need to debug separately why it wasn't seeing a winrt component.

nuget/readme.md
manodasanW: I am trying to understand which scenarios when someone will need to set this property.  Is this when they are using the VS command prompt?  When they are building in VS?  or just terminal?

nuget/Microsoft.Windows.CsWinRT.targets
```diff
@@ -316,6 +322,223 @@ $(CsWinRTInternalProjection)
...
+
+      <!-- Skip generating a manifest file, matching Native AOT (https://learn.microsoft.com/cpp/build/reference/manifest-create-side-by-side-assembly-manifest) -->
+      <_StubExeMsvcArgs Include="/MANIFEST:NO" />
+
+      <!--
+        Change the behavior for UCRT to be dynamically linked (the default would be to statically link it, due to '/MT[d]').
+        See: https://learn.microsoft.com/cpp/build/reference/defaultlib-specify-default-library.
+        Also see: https://learn.microsoft.com/cpp/build/reference/nodefaultlib-ignore-libraries.
+      -->
+      <_StubExeMsvcArgs Include="/NODEFAULTLIB:libucrt.lib" />
```
manodasanW, line 473: Do we need to consider debug versions here or no?

src/Authoring/WinRT.SourceGenerator/Generator.cs
```diff
@@ -271,13 +271,16 @@ public static int DllGetActivationFactory(void* activatableClassId, void** facto
 
                 """);
 
-            // There's three possible cases for this method:
+            // There's four possible cases for this method:
```
manodasanW, line 274: Should we add a C# exe project to reference the authoring project and set this property to exercise these changes in the build?

nuget/readme.md
Sergio0694: Yeah, if you build from VS command prompt you'll want to set the property, to ensure that the build environment is the same one that normal VS would use. You wouldn't set this if you want to build from a normal terminal window.

nuget/Microsoft.Windows.CsWinRT.targets
Sergio0694: Not sure. The stub is always built with `\O2` anyway 🤔

nuget/readme.md
manodasanW: Thoughts on checking for VSCMD_VER to detect VS developer command prompt?

https://learn.microsoft.com/en-us/visualstudio/ide/reference/command-prompt-powershell?view=vs-2022#developer-command-prompt

Sergio0694: Updated to use this to set the default! 👍

manodasanW (approved):

nuget/Microsoft.Windows.CsWinRT.targets
Sergio0694: Getting this when publishing in Debug, I assume this is what you meant?

> "LINK : warning LNK4098: defaultlib 'ucrt.lib' conflicts with use of other libs; use /NODEFAULTLIB:library"

PR #1899, opened 2025-01-20
Fix IDIC casts failing during authoring scenarios

manodasanW:
The ABI interfaces in authoring scenarios were previously implementing the ABI impl interfaces.  This means that IDIC casts cannot be done in cross process scenarios of authored types given the IDIC interface doesn't implement the projected interface being used in the cast.  This addresses it by having the ABI interfaces implement the projected interfaces in authoring scenarios.  This is similar to non-authoring scenarios.  The only special case is exclusive interfaces which have no projected interface and will continue to implement the ABI impl interface.

### DIFF
--- a/src/cswinrt/code_writers.h
+++ b/src/cswinrt/code_writers.h
@@ -1105,12 +1105,15 @@ namespace cswinrt
     void write_explicitly_implemented_method_for_abi(writer& w, MethodDef const& method,
         std::string_view return_type, TypeDef const& method_interface, std::string_view method_target)
     {
+        // In authoring scenarios, exclusive interfaces don't exist, so use the CCW impl type.
+        bool implement_ccw_interface = does_abi_interface_implement_ccw_interface(method_interface);
+
         method_signature signature{ method };
         w.write(R"(
 % %.%(%) => %.%(%);
 )",
             return_type,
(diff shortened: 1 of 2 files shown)


Sergio0694 (commented):
Just curious, any particular reason why we're changing the ABI interface to inherit from the projected interface and not the ABI impl interface, rather than just adding the projected interface too when available, and having the ABI interface just implement both? Is that because it'd just not really be useful, and then we'd need to have a DIM for interface methods twice, once per base interface? 🤔

Sergio0694 (approved):

manodasanW:
> Just curious, any particular reason why we're changing the ABI interface to inherit from the projected interface and not the ABI impl interface, rather than just adding the projected interface too when available, and having the ABI interface just implement both? Is that because it'd just not really be useful, and then we'd need to have a DIM for interface methods twice, once per base interface? 🤔

I have been slowly reducing what we need from the impl namespace types over time and this just fits in with that.  It will allow for more trimming of those interfaces especially when implementing those interfaces as DIM doesn't seem to be needed.  And I saw no value in implementing both interfaces as DIM.  Is there any?

Sergio0694:
Yup, no it makes perfect sense to me! Just wanted to make sure I was following, thank you! 🙂

PR #1927, opened 2025-02-19
Add analyzer for '[ComImport]' casts on AOT

Sergio0694:
These will never work on WinRT objects in AOT scenarios, but they're easy to miss and don't currently warn.


src/Authoring/WinRT.SourceGenerator/Analyzers/ComImportInterfaceAnalyzer.cs
```diff
@@ -0,0 +1,117 @@
...
+    /// <inheritdoc/>
+    public override void Initialize(AnalysisContext context)
+    {
+        context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze | GeneratedCodeAnalysisFlags.ReportDiagnostics);
+        context.EnableConcurrentExecution();
+
+        context.RegisterCompilationStartAction(static context =>
+        {
+            // This analyzer is only enabled when 'EnableAotAnalyzer' is set and when the AOT warning level is not set to 0 (ie. disabled)
+            if (GeneratorExecutionContextHelper.GetCsWinRTAotWarningLevel(context.Options.AnalyzerConfigOptionsProvider) == 0 ||
```
manodasanW, line 34: Can we also scope this to auto mode for now to avoid introducing warnings in non CsWinRT scenarios / opt-in scenarios.

`IsCsWinRTAotOptimizerInAutoMode`

Sergio0694: Updated!

src/Authoring/WinRT.SourceGenerator/Analyzers/ComImportInterfaceAnalyzer.cs
```diff
@@ -0,0 +1,116 @@
...
+    /// <inheritdoc/>
+    public override void Initialize(AnalysisContext context)
+    {
+        context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze | GeneratedCodeAnalysisFlags.ReportDiagnostics);
+        context.EnableConcurrentExecution();
+
+        context.RegisterCompilationStartAction(static context =>
+        {
+            // We only need to emit warnings if CsWinRT is in 'auto' mode (same as the collection expressions analyzer)
+            if (!GeneratorExecutionContextHelper.IsCsWinRTAotOptimizerInAutoMode(context.Options.AnalyzerConfigOptionsProvider, context.Compilation))
```
manodasanW, line 34: I feel like you should keep the GetEnableAotAnalyzer check here still as someone may want to have auto mode enabled here while still not publishing for AOT yet.

Sergio0694: But isn't the point that if the AOT optimizer we should still emit all possible AOT warnings? I mean we do the same for the collection expression analyzer, for instance, even though that code would technically work if you don't set `PublishAot`. I'm happy to change this if you prefer, I guess I'm just a bit confused by the exact matrix of possible combinations for all the available AOT analyzer settings we have 😅

manodasanW: I see the collection expression analyzer a bit different as trimming scenarios can also be impacted by it given we won't generate the vtables and interfaces can get trimmed and so on.  Whereas this, I can see this working fine on trimming while being in auto mode.  So far, all our aot optimizer scenarios have been also useful with just trimming enabled.  I think this is the first one that is going to differ in that behavior.

Sergio0694: I see, fixed! 🙂

manodasanW (approved):

PR #1933, opened 2025-02-24
Enable adding GC pressure on Native AOT

Sergio0694:
Removing a condition that no longer applies, the bug is fixed now.

### DIFF
--- a/src/WinRT.Runtime/ObjectReference.cs
+++ b/src/WinRT.Runtime/ObjectReference.cs
@@ -385,15 +385,9 @@ namespace WinRT
                 }
 
                 DisposeTrackerSource();
-#if NET
-                // Disable on AOT for now until dotnet runtime #104583 is addressed to avoid making the issue happen more often.
-                if (RuntimeFeature.IsDynamicCodeCompiled)
-                {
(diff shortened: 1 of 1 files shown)


src/WinRT.Runtime/ObjectReference.cs
```diff
@@ -142,15 +142,7 @@ protected IObjectReference(IntPtr thisPtr)
...
-#if NET
-            // Disable on AOT for now until dotnet runtime #104583 is addressed to avoid making the issue happen more often.
-            if (RuntimeFeature.IsDynamicCodeCompiled)
-            {
-                GC.AddMemoryPressure(ComWrappersSupport.GC_PRESSURE_BASE);
-            }
-#else
             GC.AddMemoryPressure(ComWrappersSupport.GC_PRESSURE_BASE);
-#endif
         }
```
manodasanW, line 146: Also the remove: https://github.com/microsoft/CsWinRT/blob/1844e3ec4ffad4defe6e3d5664d63333e22c49f4/src/WinRT.Runtime/ObjectReference.cs#L398

Sergio0694: Ah, fixed, thank you!

manodasanW (approved):

PR #1934, opened 2025-02-24
Enable propagating WinRTRuntimeClassName value to activation factory lookup

### LINKED ISSUE #1872: Bug: Command Line Help Prints Nothing
### Description

When I run `cswinrt.exe -?`, no help is printed

### Steps To Reproduce

Run `cswinrt.exe -?`

### Expected Behavior

To see help printed

### Version Info

CsWinRT: 2.2.0

### Additional Context

_No response_

### LINKED ISSUE #1929: Bug: When using WinRTRuntimeClassName attribute, we should be able to also activate that type using that name
### Description

Types can use the `WinRTRuntimeClassName` attribute to override the runtime class name reported for `GetRuntimeClassName`.  This is only used in that scenario today but if this were in an authoring component, we should also allow for that type to be activated with that runtime class name in addition to the one based on implementation.

### Steps To Reproduce

1. Use WinRTRuntimeClassNameAttribute
2. Activate type based on name specified

### Expected Behavior

Runtime class name specified in WinRTRuntimeClassNameAttribute can also be used for activation.

### Version Info

CsWinRT 2.2.0

### Additional Context

_No response_

manodasanW:
Today, we support overriding the `runtime class name` returned for `IInspectable` via the `WinRTRuntimeClassName` attribute.  But if someone asks for the activation factory with that name, it fails.  This enables an activation factory being generating with those runtime class names when `CsWinRTGenerateOverridedClassNameActivationFactory` is set.  This is enabled via the generated projection also calling an partial implementation of the activation factory when `partial_factory` is set which allows the source generator side to implement the activation factory with remapped names.

We also now generate `Module.GetActivationFactory` when activation factories are merged in a non cswinrt component project.  This allows the JIT scenario to be simlar to the AOT scenario which had `DllGetActivationFactory` generated and thereby be able to be activated via WinRT.Host.

Also fixed the issue with help not displaying by increasing the buffer size.

Fixes #1929 
Fixes #1872


Sergio0694 (approved):
Looks good! I assume the new method will just be trimmed anyway when publishing with NAOT and merging activation factories, so I don't expect any binary size diffs.

manodasanW:
> Looks good! I assume the new method will just be trimmed anyway when publishing with NAOT and merging activation factories, so I don't expect any binary size diffs.

Right, I expect it to be trimmed away.  I was debating making DllGetActivationFactory call this function but then didn't just to avoid indrection which was going to do the same thing and it also allows it to be trimmed away.

PR #1953, opened 2025-03-25
Add 'NoInlining' to marshalling stubs

Sergio0694:
This PR adds `NoInlining` to all marshalling stubs. This matches .NET Native and should reduce binary size.

### DIFF
--- a/src/WinRT.Runtime/IInspectable.cs
+++ b/src/WinRT.Runtime/IInspectable.cs
@@ -4,6 +4,7 @@
 using System;
 using System.ComponentModel;
 using System.Diagnostics.CodeAnalysis;
+using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using WinRT.Interop;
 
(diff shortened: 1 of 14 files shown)


jkotas:
Is it really profitable to mark close to everything with `[MethodImpl(MethodImplOptions.NoInlining)]`?

I would expect it to be profitable only for small methods that are called from many places for which you may not be happy about the codesize / perf tradeoff done by the JIT. There should be relatively a few of those.

`[MethodImpl(MethodImplOptions.NoInlining)]` on larger methods is unnecessary. JIT won't inline them unless it has evidence based on profile data that it is worth to inline.

`[MethodImpl(MethodImplOptions.NoInlining)]` on small methods that have one or a few callsites may even regress both code size and performance.

Sergio0694:
We're not entirely sure yet, I was going to try this out in the Store and check the size diff. @MichalStrehovsky suggested making this change to match what .NET Native is doing. Looking at some diffs in the Store, it seems we're currently inlining a ton of these marshalling stubs, and he estimated we might save ~5MB just with this change. There's more details from Michal earlier on in the email thread about .NET Native vs Native AOT regression tracking that Andy added you the other day, if you're curious 🙂

I'll report back when I get numbers from this change and then we can reevaluate, I assume?

jkotas:
It would be interesting to find the top 100 `[MethodImpl(MethodImplOptions.NoInlining)]` annotations that help to reduce the size the most, and how much the rest leaves on the table.

Issue #1342, opened 2022-08-26
Getting LaunchActivatedEventArgs crashes in release mode


lhak:
### Describe the bug

See the attached repro project. When getting the activation arguments in Main (App.xaml.cs), the application will crash with a System.NullReferenceException. Please note that this only happens in Release (x64) mode with PublishReadyToRun enabled (the default value).
[ActivationArgsRepro.zip](https://github.com/microsoft/WindowsAppSDK/files/9430591/ActivationArgsRepro.zip)
 
The following change seems to somehow workaround this issue

```csharp
AppActivationArguments activatedArgs = AppInstance.GetCurrent().GetActivatedEventArgs();
Windows.ApplicationModel.Activation.LaunchActivatedEventArgs launchArgs = activatedArgs.Data as Windows.ApplicationModel.Activation.LaunchActivatedEventArgs;
launchArgs.Kind.ToString(); // Add this line to avoid a crash
if (launchArgs.Arguments.StartsWith("argument"))
{

}
```

### Steps to reproduce the bug

1. Run the attached repro project in Release mode (x64)

### NuGet package version

1.1.4

### Packaging type

Packaged (MSIX)

### Windows version

Windows 11 version 21H2 (22000)

### IDE

Visual Studio 2022

PR #1956, opened 2025-03-27
Add analyzer for trim-unsafe casts to projected runtime class types

### LINKED ISSUE #1860: Bug: Cannot cast some of the UWP activation arguments using dotnet as operator

Sergio0694:
### Closes #1860, #1342

This PR adds a new analyzer to warn when doing trim-unsafe casts to projected runtime classes.


Lamparter (approved):
Thank you!!!

src/Authoring/WinRT.SourceGenerator/Analyzers/RuntimeClassCastAnalyzer.cs
```diff
@@ -0,0 +1,184 @@
...
+{
+    /// <inheritdoc/>
+    public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = [WinRTRules.RuntimeClassCast];
+
+    /// <inheritdoc/>
+    public override void Initialize(AnalysisContext context)
+    {
+        // We're intentionally not analyzing generated code, as we don't want to generate warnings for XAML-generated code.
+        // Casts in code-behind files are fine, for two main reasons:
+        //   - Many code paths generated byt he XAML compiler will use 'Cast<T>', which will root metadata
```
manodasanW, line 31:
```suggestion
        //   - Many code paths generated by the XAML compiler will use 'Cast<T>', which will root metadata
```

src/Authoring/WinRT.SourceGenerator/CsWinRTDiagnosticStrings.resx
```diff
@@ -369,4 +369,16 @@
   <data name="ComImportInterfaceCast_Text" xml:space="preserve">
     <value>The type '{0}' used in a cast operation is a [ComImport] interface, which is not compatible with CsWinRT objects in AOT scenarios. Consider using the COM generators to define the interface, or manually handling the interface query on the underlying native object.</value>
   </data>
+  <data name="RuntimeClassCast_Brief" xml:space="preserve">
+    <value>Casts to WinRT runtime class types are not trim-safe</value>
+  </data>
+  <data name="RuntimeClassCast_Text" xml:space="preserve">
```
manodasanW, line 375: I feel like both error messages, but especially this one, is missing the context on when it is not trim-safe / might be trimmed which is when they are declared as object / lacking static type information.

Sergio0694: Well the problem is that it's more complicated than that. It's not just when there's no static type information. Casting to a derived class is also not trim-safe, even if the source type is not just `object`. I can update the messages and include this info too, was just not sure how verbose and long we wanted the messages to be? 😅

manodasanW: Well I would still consider the latter to be lack of static type information but with respect to not being aware of derived type.  We can probably generalize it with something along the lines of can't be determined with static type information or so?

manodasanW: How about

Casting to type '{0}' where the type being cast to is never instantiated in the C# implementation is not trim-safe as the type can be trimmed.

or

Casting to type '{0}' where the type being cast to was statically not known about when the C# wrapper was created is not trim-safe as the type can be trimmed.

src/Authoring/WinRT.SourceGenerator/CsWinRTDiagnosticStrings.resx
```diff
@@ -369,4 +369,16 @@
...
+    <value>Casting to type '{0}' is not trim-safe, as it is a WinRT runtime class type. Consider using the 'WinRT.CastExtensions.Cast&lt;T&gt;' extension method instead, which will ensure the necessary metadata is kept. Alternatively, you can use '[DynamicDependency]' to manually root the necessary metadata, and suppress the warning via '[UnconditionalSuppressMessage]'.</value>
+  </data>
+  <data name="IReferenceTypeCast_Brief" xml:space="preserve">
+    <value>Casts to WinRT 'IReference`1&lt;T&gt;' unboxed values are not trim-safe</value>
+  </data>
+  <data name="IReferenceTypeCast_Text" xml:space="preserve">
+    <value>Casting to type '{0}' is not trim-safe, as it is unboxing a WinRT 'IReference`1&lt;T&gt;' value type. Consider using the 'WinRT.CastExtensions.Cast&lt;T&gt;' extension method instead, which will ensure the necessary metadata is kept. Alternatively, you can use '[DynamicDependency]' to manually root the necessary metadata, and suppress the warning via '[UnconditionalSuppressMessage]'.</value>
```
manodasanW, line 382:
```suggestion
    <value>Casting to type '{0}' is not trim-safe, as it is unboxing a WinRT 'IReference`1&lt;T&gt;' value type that was statically not known. Consider using the 'WinRT.CastExtensions.Cast&lt;T&gt;' extension method instead, which will ensure the necessary metadata is kept. Alternatively, you can use '[DynamicDependency]' to manually root the necessary metadata, and suppress the warning via '[UnconditionalSuppressMessage]'.</value>
```

manodasanW (approved):

PR #1957, opened 2025-03-31
Migrate pipeline

manodasanW:
Moving to OneBranch for official builds.


DefaultRyan:
@microsoft-github-policy-service rerun

DefaultRyan:
Prefixing all of these yml files with "CsWinRT" is redundant, and impedes readability. Not blocking for this PR, but strongly consider trying to clean some of this up in the future.

src/Directory.Build.props
```diff
@@ -18,6 +18,7 @@
     <GenerateTestProjection Condition="'$(GenerateTestProjection)' == ''">false</GenerateTestProjection>
     <CsWinRTGenerateProjection>$(GenerateTestProjection)</CsWinRTGenerateProjection>
     <CsWinRTWindowsMetadata Condition="'$(CsWinRTWindowsMetadata)' == ''">$(WindowsSDKVersion.TrimEnd('\'))</CsWinRTWindowsMetadata>
+    <CsWinRTWindowsMetadata Condition="'$(CsWinRTWindowsMetadata)' == ''">sdk</CsWinRTWindowsMetadata>
```
DefaultRyan, line 21: Is this only possible if `WindowsSDKVersion` was empty?

build/AzurePipelineTemplates/CsWinRT-Benchmarks-Steps.yml
```diff
@@ -82,13 +58,4 @@ steps:
 
 # Component Detection
 - task: ComponentGovernanceComponentDetection@0
```
DefaultRyan, line 60: I'm not sure a manual ComponentGovernance step is needed when running in OneBranch.

build/AzurePipelineTemplates/CsWinRT-BuildAndTest-Stage-GitHub.yml
```diff
@@ -0,0 +1,106 @@
+stages:
+- stage: BuildAndTest
+  displayName: Build/Test Stage
+  jobs: 
+  - job: BuildAndTest
+    timeoutInMinutes: 270
```
DefaultRyan, line 6: 3 hours?😮

DefaultRyan (approved):

src/Directory.Build.props
manodasanW: Right, I believe WindowsSDKVersion is only set in a VS developer command prompt.  And outside of it, it doesn't seem to be set.  So I had to fallback to our logic to handle this.

build/AzurePipelineTemplates/CsWinRT-BuildAndTest-Stage-GitHub.yml
manodasanW: At some point, we used to hit that when we built from TFMs at the same time.  Lately it has been faster.  At some point, might make sense to reduce.

build/AzurePipelineTemplates/CsWinRT-Benchmarks-Steps.yml
manodasanW: Yea I wasn't sure if I should keep this or not given this was the one running on Azure pipelines rather than OneBranch.

build/AzurePipelineTemplates/CsWinRT-PublishToNuGet-Steps.yml
```diff
@@ -0,0 +1,149 @@
...
+  # Download arm64 
+  - task: DownloadPipelineArtifact@2
+    displayName: Download arm64
+    inputs:
+      artifactName: drop_BuildAndTest_release_arm64
+      itemPattern: ''
+      targetPath: $(Build.SourcesDirectory)\release_arm64
+
+  # Stage Binaries
+  - task: CmdLine@2
```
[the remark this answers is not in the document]
manodasanW: Will address this when I also do the rename in separate PR.

PR #1973, opened 2025-04-08
Add option to generate managed DllGetActivationFactory

manodasanW:
Add an option to generate a managed version of `DllGetActivationFactory` under the `ABI.Exports` namespace when using the `CsWinRTGenerateManagedDllGetActivationFactory` property.


src/Authoring/WinRT.SourceGenerator/Generator.cs
```diff
@@ -463,6 +462,52 @@ public static IntPtr GetActivationFactory(ReadOnlySpan<char> runtimeClassId)
...
+                        /// <summary>
+                        /// Retrieves the activation factory from a DLL that contains activatable Windows Runtime classes.
+                        /// </summary>
+                        /// <param name="activatableClassId">The class identifier that is associated with an activatable runtime class.</param>
+                        /// <param name="factory">A pointer to the activation factory that corresponds with the class specified by <paramref name="activatableClassId"/>.</param>
+                        /// <returns>The <c>HRESULT</c> for the operation.</returns>
+                        /// <seealso href="https://learn.microsoft.com/en-us/previous-versions/br205771(v=vs.85)"/>
+                        public unsafe static int DllGetActivationFactory(void* activatableClassId, void** factory)
+                        {
+                             {{GenerateNativeDllGetActivationFactoryImpl(context)}}
```
Sergio0694, line 506: Double checking: we don't care about duplicating this code since this method is opt-in anyway, right?

Sergio0694 (approved):

src/Authoring/WinRT.SourceGenerator/Generator.cs
manodasanW: Right.

PR #1989, opened 2025-05-15
Add technical docs on generic interfaces codegen

Sergio0694:
This PR adds docs explaining in detail all codegen to handle generic collection interfaces in CsWinRT 3.0.

### DIFF
--- a/docs/generic-interfaces-infrastructure.md
+++ b/docs/generic-interfaces-infrastructure.md
@@ -1,0 +1,423 @@
+# CsWinRT 3.0 generic interfaces infrastructure
+
+This document outlines the design of Windows Runtime generic collection interfaces in CsWinRT 3.0.
+
+There are multiple scenarios to handle for each generic interface:
+- Marshalling with static type information (eg. `IEnumerator<string>` return)
+- Marshalling without static type information (ie. `object` return)
+  - Both of these apply to either marshalling direction
+  - RCW marshalling can be both via anonymous types, and projected RCW types
+- Projection methods (accounting for semantics differences, if needed)
+- CCW vtable implementation
+- IDIC interface implementation
+- IDIC interface implementation lookup
+
+Supporting all of these scenarios, and doing so as efficiently as possible while also respecting all of CsWinRT 3.0 design goals (eg. proper trimming support), requires several additional types being generated per generic interface instantiation, along with some related APIs exposed from the CsWinRT runtime.
+
+We will use `IEnumerator<T>` (`IIterator<T>`) as reference to document all of these types. There are some differences with other generic interface types, but they all follow the same general principles. This document will also mention such differences where relevant.
+
+> [!NOTE]
+> This document specifically only covers generic **collection** interfaces. Other generic type instantiations, such as delegate types (and associated event source infrastructure, when applicable) and `KeyValuePair<TKey, TValue>`, are handled differently, and are already implemented in `cswinrtgen`. See code there for reference.
+
+## RCW (in `WinRT.Runtime.dll`)
+
+The starting point for RCW scenarios is an RCW class that can be used to marshal interfaces to native. The Windows Runtime interfaces are mapped to different interfaces in the .NET world (eg. `IIterator<T>` -> `IEnumerator<T>`), and sometimes there might be subtle semantic differences, which CsWinRT takes care of hiding. Here is the API surface for the `IEnumerator<T>` RCW:
+
+```csharp
+public abstract class WindowsRuntimeEnumerator<T> : WindowsRuntimeObject, IEnumerator<T>, IWindowsRuntimeInterface<IEnumerator<T>>
+{
+    public WindowsRuntimeEnumerator(WindowsRuntimeObjectReference nativeObjectReference);
+
+    protected sealed override bool IsOverridableInterface(in Guid iid);
+
+    // These will be implemented to invoke the underlying 'IIterator<T>' methods directly
+    protected abstract T CurrentNative { get; }
+    protected abstract bool HasCurrentNative { get; }
+    protected abstract bool MoveNextNative();
+
+    // These will implement the managed 'IEnumerator<T>' interface, with the expected .NET semantics
+    public T Current { get; }
+    object IEnumerator.Current { get; }
+    public bool MoveNext();
+    public void Dispose();
+    public void Reset();
+
+    public WindowsRuntimeObjectReferenceValue GetInterface();
+}
+```
+
+This type derives from `WindowsRuntimeObject`, which provides all the usual support for IDIC, generated COM interfaces, and more. If needed, these RCW types (one per generic interface) can also contain additional state necessary to properly map the Windows Runtime interface behavior to the managed one. This is the case for `IEnumerator<T>`, which is slightly different than `IIterator<T>`, for instance.
+
+Next, an interface is needed for each "Methods" implementation, for each generic interface. For `IIterator<T>`, it looks like this:
+
+```csharp
+public interface IIteratorMethods<T>
+{
+    static abstract T get_Current(WindowsRuntimeObjectReference objectReference);
+
+    static abstract bool get_HasCurrent(WindowsRuntimeObjectReference objectReference);
+
+    static abstract bool MoveNext(WindowsRuntimeObjectReference obj);
+
+    static abstract uint GetMany(WindowsRuntimeObjectReference obj, Span<T> items);
+}
+```
+
+This exactly matches the shape that any "Methods" type would have for any given projected interface.
+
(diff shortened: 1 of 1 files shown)


docs/generic-interfaces-infrastructure.md
```diff
@@ -0,0 +1,423 @@
...
+    public bool MoveNext()
+    {
+        // Adapted implementation
+    }
+
+    // Other methods...
+}
+```
+
+This is a simple thing wrapper around a given interface instance, with additional state, and implementations of all necessary methods. It is not required for this wrapper to actually implement the interface itself, nor to implement all interface methods. Only the ones that require additional semantics modifications have to be implemented, the other methods can just be called directly by the marshalled CCWs. Also note that this adapter type would be implementing the native Windows Runtime interface (which is not projected), not the managed .NET interface.
```
jkoritzinsky, line 289: How many of these adapters have additional state?

I think this could be done more efficiently for stateless wrappers.

docs/generic-interfaces-infrastructure.md
```diff
@@ -0,0 +1,423 @@
...
+        {
+            var unboxedValue = ComInterfaceDispatch.GetInstance<IEnumerator<string>>((ComInterfaceDispatch*)thisPtr);
+
+            *result = IEnumerator_stringImplAdapterTable.GetAdapter(unboxedValue).get_Current();
+
+            return WellKnownErrorCodes.S_OK;
+        }
+        catch (global::System.Exception ex)
+        {
+            return RestrictedErrorInfoExceptionMarshaller.ConvertToUnmanaged(ex);
```
jkoritzinsky, line 386: Now that we have the new exception marshalling support (just added), you should be able to switch these implementations to be based on `[GeneratedComInterface]` if you wanted to do so.

dongle-the-gadget: I believe this code should be generated in the sidecar? In that case `[GeneratedComInterface]` won't be available.

Sergio0694: Yeah this is all code in the interop .dll, so we can't leverage the COM generators here 😅
However that will be very useful for custom mixed classic COM scenarios!

docs/generic-interfaces-infrastructure.md
Sergio0694: I _think_ it might just be enumerators. I just wrote the docs in a generalized way so that if it turns out there's more, this can easily be extended to those types too. For all cases where we don't need any stateful adapter logic, then yeah the entire `Adapter` type would be skipped, there would be no generated table type to associate adapters, etc.

So we can definitely do things more efficiently for all stateless wrappers, absolutely 🙂

docs/generic-interfaces-infrastructure.md
jkoritzinsky: Right, this is just a C# representation of the IL you're generating into the sidecar.

Would be worth explicitly stating that somewhere in the doc if it isn't already.

docs/generic-interfaces-infrastructure.md
jkoritzinsky: I'd recommend also adding a sketched-out example for a stateless wrapper as well in some doc about the designs.

Sergio0694: Added some remarks! For stateless CCWs, we'd just completely omit the "AdapterImpl" and "AdapterImplTable" types, and call the stateless adapter "Methods" APIs directly, in the normal marshalling stub. Is this the same thing you had in mind, or did you have an idea on how to optimize that even further? 🤔

jkoritzinsky: No that's about what I was thinking.

Anything else would be ergonomics in the shape of the APIs, but you're emitting IL so that doesn't really matter that much.

Sergio0694: Awesome, thank you for confirming! 🙂

docs/generic-interfaces-infrastructure.md
```diff
@@ -0,0 +1,426 @@
...
+    public void Dispose();
+    public void Reset();
+
+    public WindowsRuntimeObjectReferenceValue GetInterface();
+}
+```
+
+This type derives from `WindowsRuntimeObject`, which provides all the usual support for IDIC, generated COM interfaces, and more. If needed, these RCW types (one per generic interface) can also contain additional state necessary to properly map the Windows Runtime interface behavior to the managed one. This is the case for `IEnumerator<T>`, which is slightly different than `IIterator<T>`, for instance.
+
+Next, an interface is needed for each "Methods" implementation, for each generic interface. For `IIterator<T>`, it looks like this:
```
jevansaks: I was mostly with you until this point. "An interface is needed" -- why? How does this connect to the scenarios above? I assume there's some general-purpose C#/WinRT machinery that is trying to reach out for this, but when/where?

docs/generic-interfaces-infrastructure.md
```diff
@@ -0,0 +1,426 @@
...
+public interface IWindowsRuntimeUnsealedComWrappersCallback
+{
+    static abstract bool TryCreateObject(
+        void* value,
+        ReadOnlySpan<char> runtimeClassName,
+        [NotNullWhen(true)] out object? result);
+}
+```
+
+This is conceptually similar to `IWindowsRuntimeComWrappersCallback` (which is used for delegates and sealed runtime class names), with the main difference being that it is specialized for unsealed types (and interfaces). In this case, implementations need to know the runtime class name of the native object, to determine whether they can marshal it via the fast path using a statically visible RCW type (see below), or whether they should fallback to the logic to marshal arbitrary opaque object that is used when no static type information is available (which involves going through the interop type lookup to get the marshalling info, etc.).
```
jevansaks: Pretend I don't know anything about the current C#/WinRT implementation in this regard. What do I need to know/care about to evaluate this proposal? :)

jevansaks (commented):
Posting my comments but I got a little lost in this writeup so I think we need an in-person review of some of this if you want thoughtful feedback.

docs/cswinrtgen/marshalling-arrays.md
```diff
@@ -0,0 +1,164 @@
+# CsWinRT 3.0 array infrastructure
+
+This document builds on top of the [generic interface marshalling infrastructure document](marshalling-generic-interfaces.md) and describes the design of the CsWinRT 3.0 array marshalling infrastructure. The same basic concepts outlined in the other document apply here, with the amin difference being a different API surface for marshaller types. This will be covered in more detail below.
```
jkoritzinsky:
```suggestion
This document builds on top of the [generic interface marshalling infrastructure document](marshalling-generic-interfaces.md) and describes the design of the CsWinRT 3.0 array marshalling infrastructure. The same basic concepts outlined in the other document apply here, with the main difference being a different API surface for marshaller types. This will be covered in more detail below.
```

docs/cswinrtgen/marshalling-arrays.md
```diff
@@ -0,0 +1,164 @@
...
+```
+
+The `ConvertToManaged` method will perform a `QueryInterface` on the input `value` pointer for the specified IID (as this method is only used to unbox arrays that are passed as opaque objects). It will then call `Value` on the ```IReferenceArray`1<T>``` interface pointer, and invoke the supplied callback to return the unboxed managed array instance.
+
+## Native -> managed (in `WinRT.Interop.dll`)
+
+Each array type will have a corresponding marshaller callback type, implementing `IWindowsRuntimeArrayComWrappersCallback`. Note that inputs to the `CreateArray` method will always be not-null. The generated `<string>ArrayMarshaller` type is detailed at the end of this document. This is what a generated callback type would look like for a `string` array:
+
+```csharp
+public abstract class <string>ArrayComWrappersCallback : IWindowsRuntimeArrayComWrappersCallback
```
jkoritzinsky, line 35: Generate as `static` instead of `abstract` if it isn't supposed to be implemented? 
```suggestion
public static class <string>ArrayComWrappersCallback : IWindowsRuntimeArrayComWrappersCallback
```

Sergio0694: I would love to do that, but I can't make it static since I'm passing it as a type argument 🥲

Making it statis was a suggestion from @MichalStrehovsky to minimize the metadata being kept, as at least this way ILC can guarantee the type will never be constructed. I think it's the closest I can get for now until .NET lifts this very annoying restriction. _In theory_ I could emit it as static given I'm emitting IL, but I would prefer to not emit IL that Roslyn would never generate, to avoid issues with CoreCLR/NAOT seeing stuff it doesn't expect. Does that make sense?

jkoritzinsky: Right, C# never added the "static classes in generics" feature. Nevermind then.

PR #1995, opened 2025-05-20
Integrate Shmueli's ComServer into CsWinRT

beervoley:
Moving https://github.com/shmuelie/Shmuelie.WinRTServer to CsWinRT repo.


nuget/Microsoft.Windows.CsWinRT.nuspec
```diff
@@ -73,6 +75,8 @@
 
     <file src="..\src\WinRT.Runtime\Configuration\*.cs" exclude="..\src\WinRT.Runtime\Configuration\*.net*.cs"  target="embedded\any\" />
 
+    <!-- -->
```
manodasanW, line 78: nit: uneeded?

src/ComServerHelpers/BaseActivationFactory.cs
```diff
@@ -0,0 +1,40 @@
...
+using System;
+using System.Runtime.Versioning;
+
+namespace ComServerHelpers;
+
+/// <summary>
+/// Base for a WinRT Activation Factory for a .NET type.
+/// </summary>
+[SupportedOSPlatform("windows8.0")]
+public abstract class BaseActivationFactory
```
manodasanW, line 13: Can these be internal or no?

src/ComServerHelpers/ComServerHelpers.csproj
```diff
@@ -0,0 +1,79 @@
...
+    <RootNamespace>ComServerHelpers</RootNamespace>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <DebugType>full</DebugType>
+    <DebugSymbols>true</DebugSymbols>
+    <Authors>Microsoft Corporation</Authors>
+    <Company>Microsoft Corporation</Company>
+    <Product>C#/WinRT</Product>
+    <PackageId>ComServerHelpers</PackageId>
+    <FileVersion>$(VersionNumber)</FileVersion>
+    <Version>$(VersionNumber)</Version>
```
manodasanW, line 13: Wonder if we should adopt our own assembly version for this binary rather than use the WinRT.Runtime one.

src/ComServerHelpers/ComServerHelpers.csproj
```diff
@@ -0,0 +1,79 @@
...
+    <!-- Set remote for source link. -->
+    <AdoMirrorBuild Condition="'$(AdoMirrorBuild)' == ''">false</AdoMirrorBuild>
+    <GitRepositoryRemoteName Condition="$(AdoMirrorBuild)">github</GitRepositoryRemoteName>
+    <DisableRuntimeMarshalling>True</DisableRuntimeMarshalling>
+    <RuntimeIdentifiers>win-x64;win-x86;win-arm64;$(RuntimeIdentifiers)</RuntimeIdentifiers>
+    <Nullable>enable</Nullable>
+    <NoWarn>CA1416</NoWarn>
+    <AnalysisLevel>9.0-all</AnalysisLevel>
+  </PropertyGroup>
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
```
manodasanW, line 33: We can remove this and the below block as I assume we don't use the net5.cs and the netstandard2.0.cs pattern here that we do in WinRT.Runtime.

src/ComServerHelpers/ComServerHelpers.csproj
```diff
@@ -0,0 +1,79 @@
...
+  </ItemGroup>
+
+  <ItemGroup Condition="$(IsTargetFrameworkNet5OrGreater)">
+    <Compile Remove="**/*.netstandard2.0.cs" />
+    <None Include="**/*.netstandard2.0.cs" />
+  </ItemGroup>
+
+    <!-- API Compat -->
+  <!-- <PropertyGroup>
+     Checks for both breaking changes and for new API additions.
```
manodasanW, line 45: This is fine for now, we should enable once we got a package published.

src/ComServerHelpers/ComServerHelpers.csproj
```diff
@@ -0,0 +1,79 @@
...
+    <ItemGroup>
+      <_ReferencePathDirectories Include="@(ReferencePath -> '%(RootDir)%(Directory)')" />
+      <ResolvedMatchingContract Include="$(NuGetPackageRoot)\Microsoft.Windows.CsWinRT\$(CsWinRTApiCompatVersion)\lib\$(TargetFramework)\$(TargetFileName)">
+        <DependencyPaths>@(_ReferencePathDirectories-&gt;Distinct(), ',')</DependencyPaths>
+      </ResolvedMatchingContract>
+    </ItemGroup>
+  </Target> -->
+
+  <!-- NativeAOT specific options, only starting from .NET 8 -->
+  <PropertyGroup Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net8.0'))">
```
manodasanW, line 68: We should just set this at the top given we target net8.0

src/ComServerHelpers/ComServerHelpers.csproj
beervoley: Left TODO comment there instead.

src/ComServerHelpers/BaseActivationFactory.cs
beervoley: Don't think so -> all the factory classes are used in public extension methods like https://github.com/shmuelie/Shmuelie.WinRTServer/blob/12fe717ff4ce3e74e677f0bf4578f73f15ba24df/src/Shmuelie.WinRTServer/WinRtServerExtensions.cs#L61
@shmuelie to confirm

src/ComServerHelpers/ComServerHelpers.csproj
beervoley: Up to you :)
If you want a separate version - I'll set it up

manodasanW: Yea we should probably do separate versioning to reduce friction when we add new APIs here.

beervoley: Added separate versioning.

src/Projections/Test/Test.csproj
```diff
@@ -8,11 +8,12 @@
...
     <ProjectReference Include="..\..\Tests\TestComponentCSharp\TestComponentCSharp.vcxproj" />
     <ProjectReference Include="..\..\TestWinRT\TestComponent\TestComponent.vcxproj" />
-	<ProjectReference Include="..\..\TestWinRT\test_component_base\test_component_base.vcxproj" />
-	<ProjectReference Include="..\..\TestWinRT\test_component_derived\test_component_derived.vcxproj" />
+    <ProjectReference Include="..\..\TestWinRT\test_component_base\test_component_base.vcxproj" />
+    <ProjectReference Include="..\..\TestWinRT\test_component_derived\test_component_derived.vcxproj" />
     <ProjectReference Include="..\..\TestWinRT\test_component_fast\test_component_fast.vcxproj" />
     <PackageReference Include="System.Numerics.Vectors" />
     <ProjectReference Include="..\..\WinRT.Runtime\WinRT.Runtime.csproj" />
+    <ProjectReference Include="..\..\ComServerHelpers\ComServerHelpers.csproj"/>
```
manodasanW, line 16: This probably should go on the unit test project

hez2010:
I don't get why we would need yet another standalone WinRT server implementation when it can already be achieved by calling several methods of CsWinRT. 
See https://github.com/hez2010/WinRTServer/blob/master/WinRTServer%2FProgram.cs

src/ComServerHelpers/BaseActivationFactory.cs
shmuelie: They're public so that you're not forced to use the factories we provide. Using it would be a more advanced scenario

dongle-the-gadget:
> I don't get why we would need yet another standalone WinRT server implementation

My guess is that it has to do with App Actions, announced in Build.

> additional ComWrappers

Separation of concerns?

src/ComServerHelpers/README.md
```diff
@@ -0,0 +1,136 @@
...
+[this blog post](https://devblogs.microsoft.com/ifdef-windows/the-journey-of-moving-from-cpp-winrt-to-csharp-in-the-microsoft-store/).
+
+A full trust client can use WinRT activation, which allows you to create the
+remote instances simply by `new SomeType()`, like you would for any other type.
+The sample WPF application shows this in action (using with WinForms would be
+similar).
+
+# Sample
+
+To help understand usage and show what can be done samples can be found under
```
manodasanW, line 114: I don't believe these were brought over right?  Doesn't need to happen in this PR, but we should add some of these samples or equivalent and also use them as tests that we build and run or so.

src/ComServerHelpers/WinRtServer.cs
```diff
@@ -0,0 +1,291 @@
...
+        {
+            return;
+        }
+
+        string[] managedActivatableClassIds = [.. factories.Keys];
+        HSTRING* activatableClassIds = null;
+        delegate* unmanaged[Stdcall]<HSTRING, IActivationFactory**, HRESULT>* activationFactoryCallbacks = null;
+        try
+        {
+            activatableClassIds = (HSTRING*)Marshal.AllocHGlobal(sizeof(HSTRING) * managedActivatableClassIds.Length);
```
manodasanW, line 196: Instead of Alloc from Marshal, we typically make use of `NativeMemory.Alloc` in our .NET 8 and later projections given we seen it being faster.  Should we update our 3 uses here?

src/ComServerHelpers/GlobalSuppressions.cs
```diff
@@ -0,0 +1,11 @@
...
+// Licensed under the MIT License.
+
+// This file is used by Code Analysis to maintain SuppressMessage
+// attributes that are applied to this project.
+// Project-level suppressions either have no target or are given
+// a specific target and scoped to a namespace, type, member, etc.
+
+using System.Diagnostics.CodeAnalysis;
+
+[assembly: SuppressMessage("Performance", "CA1805:Do not initialize unnecessarily", Justification = "<Pending>", Scope = "member", Target = "~F:ComServerHelpers.Internal.TestComWrappers.fpQueryInterface")]
```
manodasanW, line 11: unused?

manodasanW:
> I don't get why we would need yet another standalone WinRT server implementation and additional ComWrappers when it can already be achieved by passing the generated `GetActivationFactory` method to `RoRegisterActivationFactories` and using the ComWrappers provided by CsWinRT. See https://github.com/hez2010/WinRTServer/blob/master/WinRTServer%2FProgram.cs

The goal behind this PR is to provide a higher-level abstraction around this and also add features to it such as lifetime management as there are going to be more scenarios relying on this support.  The lifetime management piece also sounds like it is going to need to be able to keep track of the instances for which CCWs are created for in the server, so it is going to achieve that by deriving from `DefaultComWrappers`.  But that will be initialized as the one CsWinRT's uses.  That part isn't there yet but will probably be added in a follow up.  Otherwise, yes, it would have made sense just to use CsWinRT's ComWrappers rather than creating a new instance,

src/ComServerHelpers/BaseActivationFactory.cs
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Runtime.Versioning;
+
+namespace ComServerHelpers;
```
Sergio0694, line 7: We should really think about what actual namespace names to use. And type names.

src/ComServerHelpers/BaseActivationFactory.cs
```diff
@@ -0,0 +1,40 @@
...
+[SupportedOSPlatform("windows8.0")]
+public abstract class BaseActivationFactory
+{
+    /// <inheritdoc/>
+    public abstract object ActivateInstance();
+
+    /// <summary>
+    /// Gets the Activatable Class ID.
+    /// </summary>
+    public abstract string ActivatableClassId
```
Sergio0694, line 21: Move the getter inline

src/ComServerHelpers/BaseActivationFactory.cs
```diff
@@ -0,0 +1,40 @@
...
+    {
+        get;
+    }
+
+    /// <summary>
+    /// Occurs when a new instance is created.
+    /// </summary>
+    public event EventHandler<InstanceCreatedEventArgs>? InstanceCreated;
```
Sergio0694, line 29: Sort member types (eg. fields before properties, methods after properties, sorted by accessibility, etc.).
Same applies to all other files too.

src/ComServerHelpers/GeneralClassFactory.cs
```diff
@@ -0,0 +1,29 @@
...
+/// General class factory for .NET types using generics.
+/// </summary>
+/// <typeparam name="T">Type the factory creates.</typeparam>
+/// <typeparam name="TInterface">Interface that <typeparamref name="T"/> implements.</typeparam>
+/// <seealso cref="BaseClassFactory"/>
+[SupportedOSPlatform("windows6.0.6000")]
+public sealed class GeneralClassFactory<T, TInterface> : BaseClassFactory where T : class, TInterface, new()
+{
+    /// <inheritdoc/>
+    protected internal override Guid Clsid => typeof(T).GUID;
```
Sergio0694, line 19: This is not really AOT-friendly, you should avoid doing `Type.GUID`.

src/ComServerHelpers/Internal/Windows/Com/Marshalling/TrustLevelMarshaller.cs
```diff
@@ -0,0 +1,18 @@
...
+using System.Runtime.InteropServices.Marshalling;
+using Windows.Win32.System.WinRT;
+
+namespace ComServerHelpers.Internal.Windows.Com.Marshalling;
+
+[CustomMarshaller(typeof(TrustLevel), MarshalMode.UnmanagedToManagedOut, typeof(TrustLevelMarshaller))]
+[CustomMarshaller(typeof(TrustLevel), MarshalMode.UnmanagedToManagedIn, typeof(TrustLevelMarshaller))]
+[CustomMarshaller(typeof(TrustLevel), MarshalMode.ManagedToUnmanagedOut, typeof(TrustLevelMarshaller))]
+[CustomMarshaller(typeof(TrustLevel), MarshalMode.ManagedToUnmanagedIn, typeof(TrustLevelMarshaller))]
+internal static class TrustLevelMarshaller
```
Sergio0694, line 13: `TrustLevel` should be an enum, is this marshaller actually needed?

src/ComServerHelpers/Internal/Windows/DllGetActivationFactory.cs
```diff
@@ -0,0 +1,50 @@
...
+///     </item>
+///     <item>
+///         <description><c>REGDB_E_CLASSNOTREG</c></description>
+///         <description>The class specified by <paramref name="activatableClassId"/> is not supported.</description>
+///     </item>
+/// </list>
+/// </returns>
+/// <seealso href="https://learn.microsoft.com/en-us/previous-versions/br205771(v=vs.85)">DllGetActivationFactory entry point</seealso>
+[UnmanagedFunctionPointer(CallingConvention.StdCall)]
+internal unsafe delegate HRESULT DllGetActivationFactory(HSTRING activatableClassId, IActivationFactory** factory);
```
Sergio0694, line 50: We should not be marshalling function pointers from delegates. Can we just use `[UnmanagedCallersOnly]`?

src/ComServerHelpers/Internal/Windows/Win32/PInvoke.cs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+
+namespace Windows.Win32;
+
+internal partial class PInvoke
+{
+    public static readonly Guid CLSID_GlobalOptions = new(0x0000034b, 0x0000, 0x0000, 0xC0, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x46);
```
Sergio0694, line 10: Make this use an RVA field (copy the pattern from CsWinRT).

src/ComServerHelpers/WinRtServer.cs
```diff
@@ -0,0 +1,291 @@
...
+///     await server.WaitForFirstObjectAsync();
+/// }
+/// ]]>
+/// </code>
+/// </remarks>
+/// <see cref="IDisposable"/>
+/// <threadsafety static="true" instance="false"/>
+[SupportedOSPlatform("windows8.0")]
+[System.Diagnostics.CodeAnalysis.SuppressMessage("Naming", "CA1724", Justification = "No better idea")]
+public sealed class WinRtServer : IDisposable
```
Sergio0694, line 41: Definitely let's not use "WinRt" as a type name. Should at the very least be "WinRT".
Ideally this should eventually be "WindowsRuntimeServer" in 3.0. We can do that now or later,

src/ComServerHelpers/WinRtServer.cs
```diff
@@ -0,0 +1,291 @@
...
+
+    private RO_REGISTRATION_COOKIE registrationCookie = (RO_REGISTRATION_COOKIE)0;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="WinRtServer"/> class.
+    /// </summary>
+    public unsafe WinRtServer()
+    {
+        activationFactoryCallbackWrapper = ActivationFactoryCallback;
+        activationFactoryCallbackPointer = (delegate* unmanaged[Stdcall]<HSTRING, IActivationFactory**, HRESULT>)Marshal.GetFunctionPointerForDelegate(activationFactoryCallbackWrapper);
```
Sergio0694, line 66: Yeah definitely let's not do this. Use a function pointer directly instead.

Sergio0694 (changes requested):

src/ComServerHelpers/Internal/BaseActivationFactoryWrapper.cs
```diff
@@ -0,0 +1,91 @@
...
+        {
+            return HRESULT.E_INVALIDARG;
+        }
+
+        nint unknown = 0;
+        try
+        {
+            object managedInstance = factory.ActivateInstance();
+            unknown = comWrappers.GetOrCreateComInterfaceForObject(managedInstance, CreateComInterfaceFlags.None);
+            *instance = (void*)unknown;
```
AaronRobinsonMSFT, line 29: If the `OnInstanceCreated()` call can throw, this should be placed after that. An `HRESULT` returning function shouldn't update any outs if an error is returned. Imagine the following caller, which is very common in COM infrastructure code.

```cpp
IUnknown* pUnk;
HRESULT hr = ActivateInstance(&pUnk);
if (FAILED(hr))
{
    return hr;
}

// Store the pUnk in a smart pointer here or in the caller.
```

The exceptional case should also `Marshal.Release()` on the non-null `unknown`.

src/ComServerHelpers/Internal/Windows/DllGetActivationFactory.cs
manodasanW: I was going to comment this, but they got local state that they are maintaining.

src/ComServerHelpers/Internal/BaseActivationFactoryWrapper.cs
AaronRobinsonMSFT: Note that fixing the scenario so the `Release` is called makes setting the out an A/V waiting to happen if someone uses a smart COM pointer and will then double release.

src/ComServerHelpers/Internal/BaseActivationFactoryWrapper.cs
```diff
@@ -0,0 +1,91 @@
...
+            return (HRESULT)Marshal.GetHRForException(e);
+        }
+    }
+
+    public unsafe HRESULT GetTrustLevel(global::Windows.Win32.System.WinRT.TrustLevel* trustLevel)
+    {
+        if (trustLevel is null)
+        {
+            return HRESULT.E_INVALIDARG;
+        }
```
AaronRobinsonMSFT, line 86: From a COM perspective, these should really be `E_POINTER`.

src/ComServerHelpers/Internal/BaseClassFactoryWrapper.cs
```diff
@@ -0,0 +1,88 @@
...
+                }
+                *ppvObject = (void*)ppv;
+            }
+
+            factory.OnInstanceCreated(instance);
+        }
+        catch (Exception e)
+        {
+            return (HRESULT)Marshal.GetHRForException(e);
+        }
```
AaronRobinsonMSFT, line 57: Same issue with the previous case. The `unknown`, if non-null should be freed.

src/ComServerHelpers/ComServer.cs
```diff
@@ -0,0 +1,239 @@
...
+        }
+
+        factory.InstanceCreated += Factory_InstanceCreated;
+
+        nint wrapper = this.comWrappers.GetOrCreateComInterfaceForObject(new BaseClassFactoryWrapper(factory, comWrappers), CreateComInterfaceFlags.None);
+
+        uint cookie;
+        CoRegisterClassObject(&clsid, (IUnknown*)wrapper, CLSCTX.CLSCTX_LOCAL_SERVER, (REGCLS.REGCLS_MULTIPLEUSE | REGCLS.REGCLS_SUSPENDED), &cookie).ThrowOnFailure();
+
+        factories.Add(clsid, (factory, cookie));
```
AaronRobinsonMSFT, line 101: `wrapper` will leak if either of these throw.

src/ComServerHelpers/Internal/Windows/ComPtr.cs
```diff
@@ -0,0 +1,287 @@
...
+
+    /// <summary>
+    /// Gets the address of the current <see cref="ComPtr{T}"/> instance as a raw <typeparamref name="T"/> double pointer.
+    /// This method is only valid when the current <see cref="ComPtr{T}"/> instance is on the stack or pinned.
+    /// </summary>
+    /// <returns>The raw pointer to the current <see cref="ComPtr{T}"/> instance.</returns>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public readonly T** GetAddressOf()
+    {
+        return (T**)Unsafe.AsPointer(ref Unsafe.AsRef(in this));
```
AaronRobinsonMSFT, line 194: This is a GC hole waiting to happen. The only way this is ever correct is if the `ComPtr` is forced to always be on the stack, but there is nothing here that guarantees that. Creating an `ComPtr<>[]` makes this code invalid.

src/ComServerHelpers/WinRtServer.cs
```diff
@@ -0,0 +1,291 @@
...
+{
+    /// <summary>
+    /// Mapping of Activatable Class IDs to activation factories and their <see cref="ComWrappers"/> implementation.
+    /// </summary>
+    private readonly Dictionary<string, (BaseActivationFactory Factory, ComWrappers Wrapper)> factories = [];
+
+    private readonly unsafe DllGetActivationFactory activationFactoryCallbackWrapper;
+
+    private unsafe readonly delegate* unmanaged[Stdcall]<HSTRING, IActivationFactory**, HRESULT> activationFactoryCallbackPointer;
+    private readonly StrategyBasedComWrappers comWrappers = new();
```
AaronRobinsonMSFT, line 51: I see many instantiations of this type. Are they all needed? Can they be collapsed to a single for this library?

src/ComServerHelpers/Internal/Windows/ComPtr.cs
dongle-the-gadget: Is this a new addition? I remember the compiler giving errors when I'm using `in this` in a ref struct.

src/ComServerHelpers/WinRtServer.cs
beervoley: Updated.

src/ComServerHelpers/GlobalSuppressions.cs
beervoley: Yep, deleted.

src/ComServerHelpers/Internal/Windows/Win32/PInvoke.cs
beervoley: done

src/ComServerHelpers/WinRtServer.cs
beervoley: I'm leaving typenames/namespace changing to be very last changes.

src/ComServerHelpers/Internal/BaseActivationFactoryWrapper.cs
beervoley: I don't expect `OnInstanceCreated` to throw ever since it's simply invoking raising an event on `EventHandler`.
But I moved it regardless and add release to unknown if it's non-null.

src/ComServerHelpers/ComServer.cs
beervoley: added try/catch/rethrow + free

src/ComServerHelpers/Internal/Windows/ComPtr.cs
beervoley: Updates usages and removed `ComPtr` altogether.

src/ComServerHelpers/Internal/BaseActivationFactoryWrapper.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        nint unknown = 0;
+        try
+        {
+            object managedInstance = factory.ActivateInstance();
+            factory.OnInstanceCreated(managedInstance);
+
+            unknown = comWrappers.GetOrCreateComInterfaceForObject(managedInstance, CreateComInterfaceFlags.None);
+            *instance = (void*)unknown;
```
AaronRobinsonMSFT, line 31: This still suffers from the stale `out` problem. The following code will fail badly. The invariant is that out parameters with lifetime rules can't be modified if an error code is being returned.

```cpp
CComPtr<IUnknown> pUnk;
HRESULT hr = ActivateInstance(&pUnk);
if (FAILED(hr))
{
    return hr;
}

// When pUnk goes out of scope, the destructor will release the pointer,
// but it was already released in the callee.

File: src/ComServerHelpers/Internal/BaseClassFactoryWrapper.cs [hunk 27]
```diff
@@ -0,0 +1,90 @@
...
+                var hr = (HRESULT)Marshal.QueryInterface(unknown, ref *riid, out nint ppv);
+                shouldReleaseUnknown = true;
+                if (hr.Failed)
+                {
+                    return hr;
+                }
+                *ppvObject = (void*)ppv;
+            }
+
+            factory.OnInstanceCreated(instance);
```
AaronRobinsonMSFT, line 52: I would move this up to where `instance` is created and avoid the spoiling of the out.

src/ComServerHelpers/GeneralClassFactory.cs
jkoritzinsky: Heads up that `typeof(T).GUID` can return different results on NAOT vs CoreCLR for types without the `GuidAttribute` applied. This is especially important for any usage of built-in COM.

This is why we have the `IIUnknownInterfaceDetails` abstraction for source-generated COM.

src/ComServerHelpers/Internal/BaseActivationFactoryWrapper.cs
```diff
@@ -0,0 +1,98 @@
...
+
+    public unsafe HRESULT GetIids(uint* iidCount, Guid** iids)
+    {
+        if (iidCount is null || iids is null)
+        {
+            return HRESULT.E_POINTER;
+        }
+
+        *iidCount = 1;
+        *iids = (Guid*)NativeMemory.Alloc((nuint)sizeof(Guid));
```
jkoritzinsky, line 54: What allocator does the callee expect for this to be allocated with? I'd be surprised if it expects it to be allocated with malloc/free.

src/ComServerHelpers/Internal/BaseActivationFactoryWrapper.cs
```diff
@@ -0,0 +1,98 @@
...
+            WinRT.ExceptionHelpers.SetErrorInfo(e);
+            return (HRESULT)WinRT.ExceptionHelpers.GetHRForException(e);
+        }
+    }
+
+    public unsafe HRESULT GetTrustLevel(global::Windows.Win32.System.WinRT.TrustLevel* trustLevel)
+    {
+        if (trustLevel is null)
+        {
+            return HRESULT.E_INVALIDARG;
```
jkoritzinsky, line 92: This one is still not `E_POINTER`.

src/ComServerHelpers/Internal/Windows/Com/Marshalling/TrustLevelMarshaller.cs
jkoritzinsky: If `TrustLevel` is defined in another assembly, technically yes.

src/ComServerHelpers/WinRtServer.cs
```diff
@@ -0,0 +1,295 @@
...
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="WinRtServer"/> class.
+    /// </summary>
+    public unsafe WinRtServer()
+    {
+        activationFactoryCallbackWrapper = ActivationFactoryCallback;
+        activationFactoryCallbackPointer = (delegate* unmanaged[Stdcall]<HSTRING, IActivationFactory**, HRESULT>)Marshal.GetFunctionPointerForDelegate(activationFactoryCallbackWrapper);
+
+        HRESULT result = RoInitialize(RO_INIT_TYPE.RO_INIT_MULTITHREADED);
```
jkoritzinsky, line 69: `RoInitialize` is called by the .NET runtime at thread startup. Why do you need to call it here?

Also, this will fail if someone puts `[STAThread]` on the Main method and calls this from the main thread or starts a thread as an STA thread. Is that intentional? If so, I think this should specifically check the apartment and throw a better exception or have at least a comment about it.

src/ComServerHelpers/Internal/Windows/Com/Marshalling/TrustLevelMarshaller.cs
Sergio0694: Right, but I meant this is just an internal enum defined by CsWin32, right? 🤔

jkoritzinsky: Yeah but if it's defined in a different assembly (I don't think so because it's CsWin32) the source generator can't guarantee that the underlying type isn't changed out from under it to something non-blittable.

If the code compiles without it, it shouldn't be added (as it's not required at that point).

beervoley: I will see if it compiles without it, I suppose it does.

beervoley: Removed.

src/ComServerHelpers/WinRtServer.cs
beervoley: I don't think we need this at all.

src/ComServerHelpers/GeneralClassFactory.cs
shmuelie: I would note that this is just one implementation of `BaseClassFactory`. If we want an AOT-friendly factory, there's no reason we can't make one that uses Source Generators to read from `[Guid]`

beervoley: I'm going to leave it as is for the first iteration and make it `a requirement` to annotate the server class with `Guid` attribute.
I agree with Shmuelie that source generator is a better approach and I will adopt it in the next iteration.

src/ComServerHelpers/Internal/Windows/DllGetActivationFactory.cs
beervoley: @Sergio0694 any way around this while maintaining local state?

Sergio0694: Can you have more than one instance of a given COM server class at the same time?

src/ComServerHelpers/Internal/BaseClassFactoryWrapper.cs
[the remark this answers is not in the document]
beervoley: I prefer the first solution - it looks cleaner to me :) and it doesn't struggle from double free since `ppvObject` is assigned as last step in both if and else clauses.

src/ComServerHelpers/BaseActivationFactory.cs
beervoley: Done:)

src/ComServerHelpers/Internal/BaseClassFactoryWrapper.cs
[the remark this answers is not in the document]
AaronRobinsonMSFT: I don't know what "cleaner" means in this case, but in its current state, it is wrong. Regardless of what is intended here the last statement shouldn't be `factory.OnInstanceCreated(instance);`.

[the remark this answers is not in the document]
beervoley: It's not anymore, I've pushed the changes :)

Sergio0694:
Closing this -- we discussed this offline and we're going to move this to a separate repo and NuGet package, rather than merging it into the CsWinRT repo and shipping it as part of the CsWinRT package. This follows a similar approach to what WindowsAppSDK has also done with the new package split, making the various components more modular and a-la-carte. In particular:
- Maintaining the codebase gets simpler for everyone (both on the CsWinRT side and ComServer side)
- Each component can have its own (semantic) versioning, without a need to align every release (and breaking changes)
  - Even more so with CsWinRT 3.0 on the horizon, having two independent packages is especially convenient
- Folks that prefer to use their own custom infrastructure for OOP servers can do so, without taking a dependency on this at all

We'll be moving the code from this PR to continue that work, so thank you everyone for the reviews, those still helped! 😄

PR #2003, opened 2025-05-31
Fix scenario where one authoring component project references another

manodasanW:
It is possible to have a CsWinRT component project reference another CsWinRT component project and for one to define interfaces while the other implements them.  We were previously not treating the types from the referenced component project as a CsWinRT or WinRT component and were redefining them or not putting them on the vtable.  This addresses those issues by looking for if our authoring metadata type where we define all our attributes for authoring types (ABI.Impl.<TypeName>) exists or not.  If it does exist, we treat that as an existing projected WinRT type.

### DIFF
--- a/src/Authoring/WinRT.SourceGenerator/Helper.cs
+++ b/src/Authoring/WinRT.SourceGenerator/Helper.cs
@@ -444,7 +444,7 @@ namespace Generator
                 // Check if CsWinRT component projected type from another project.
                 if (!isProjectedType)
                 {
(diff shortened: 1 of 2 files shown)


Sergio0694 (approved):

gmgallo:
Will this fix allow one authoring component fully instantiate anothere one?, I need this to interface SDKs from different vendors to move large shared_ptr<> arrarys from component A to component B with zero copy (std::move(large_arrary) thre components that I have created are: producer, transport, and consumer. both producer and consumer can access the implementation of transport to place and retrive the large array. My version of consumer works just fine, but producer relies on a C# arbitrer app to hand over an instance of transport to populate and return it.

PR #2005, opened 2025-06-03
Report unhandled errors in sync context to WinUI 3

### LINKED ISSUE #1983: Bug: DispatcherQueueSynchronizationContext changes break UnhandledException event in winui apps
### Description

The UnhandledException event handler broken since appsdk 1.7 was released (see https://github.com/microsoft/microsoft-ui-xaml/issues/10447#issuecomment-2800107924) and the issue is caused by a change in the DispatcherQueueSynchronizationContext class. I see that this file also exists in this repo (see [here](https://github.com/microsoft/CsWinRT/blob/3d85a632afcf91bc44bf9e64f002aa920e7d435b/src/Projections/WinAppSDK/Microsoft.UI.Dispatching.DispatcherQueueSynchronizationContext.cs)) and has the same bug.

### Steps To Reproduce

1. Register for UnhandledException handler in a winui app
2. Throw an exception in the application
3. App crashes without the event being invoked

### Expected Behavior

UnhandledException handler should be invoked

### Version Info

cswinrt 2.2
appsdk 1.7

### Additional Context

_No response_

dongle-the-gadget:
Closes #1983
Would also close https://github.com/microsoft/microsoft-ui-xaml/issues/10447

### DIFF
--- a/src/Projections/WinAppSDK/Microsoft.UI.Dispatching.DispatcherQueueSynchronizationContext.cs
+++ b/src/Projections/WinAppSDK/Microsoft.UI.Dispatching.DispatcherQueueSynchronizationContext.cs
@@ -22,7 +22,18 @@ namespace Microsoft.UI.Dispatching
             if (d == null)
                 throw new ArgumentNullException(nameof(d));
 
-            m_dispatcherQueue.TryEnqueue(() => d(state));
+            m_dispatcherQueue.TryEnqueue(() =>
+            {
+                try
+                {
+                    d(state);
+                }
+                catch (Exception ex)
+                {
+                    ExceptionHelpers.ReportUnhandledError(ex);
+                    throw;
+                }
+            });
         }
 
         public override void Send(SendOrPostCallback d, object state)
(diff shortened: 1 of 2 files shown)


src/Projections/WinAppSDK/Microsoft.UI.Dispatching.DispatcherQueueSynchronizationContext.cs
```diff
@@ -22,7 +22,18 @@ public override void Post(SendOrPostCallback d, object state)
             if (d == null)
                 throw new ArgumentNullException(nameof(d));
 
-            m_dispatcherQueue.TryEnqueue(() => d(state));
+            m_dispatcherQueue.TryEnqueue(() =>
```
Sergio0694, line 25: @manodasanW should we consider just sharing the synchronization context from UWP XAML with WinUI 3 instead (just changing the IID of course)? That one already has this fix, plus it's also more efficient. Thoughts?

Sergio0694:
Closing this, the implementation in the IXP package that WinUI 3 currently uses already has this fix.
And we ported it into CsWinRT 3.0 as a built-in implementation, and that has the fix as well. Thank you! 🙂

PR #2016, opened 2025-07-15
Fix non-blittable array marshaling

manodasanW:
- Added AOT tests for more array scenarios including non-blittable arrays
- Enabled them for .NET 9 to validate non-blittable array support
- Enabled the non-blittable array support and made it not rely on Marshaler<T>

### DIFF
--- a/build/AzurePipelineTemplates/CsWinRT-FunctionalTest-Steps.yml
+++ b/build/AzurePipelineTemplates/CsWinRT-FunctionalTest-Steps.yml
@@ -62,20 +62,20 @@ steps:
         TargetFolder: $(StagingFolder)\FunctionalTests\${{ functionalTest }}\
 
     - task: MSBuild@1
-      displayName: Publish ${{ functionalTest }} for AOT (net8.0)
+      displayName: Publish ${{ functionalTest }} for AOT (net9.0)
       condition: and(succeeded(), and(eq(variables['BuildConfiguration'], 'release'), eq(variables['BuildPlatform'], 'x64')))
       inputs:
         solution: $(Build.SourcesDirectory)\src\Tests\FunctionalTests\${{ functionalTest }}\${{ functionalTest }}.csproj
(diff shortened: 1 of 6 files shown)


build/AzurePipelineTemplates/CsWinRT-FunctionalTest-Steps.yml
```diff
@@ -62,20 +62,20 @@ steps:
         TargetFolder: $(StagingFolder)\FunctionalTests\${{ functionalTest }}\
 
     - task: MSBuild@1
-      displayName: Publish ${{ functionalTest }} for AOT (net8.0)
+      displayName: Publish ${{ functionalTest }} for AOT (net9.0)
```
Sergio0694, line 65: Should we also keep the .NET 8 tests, since that's still the supported LTS release?

src/Tests/FunctionalTests/JsonValueFunctionCalls/Program.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System;
 using Windows.Foundation;
 
+
```
Sergio0694: Ultra nit (feel free to ignore): accidental extra blank line 😆

src/WinRT.Runtime/Marshalers.cs
```diff
@@ -1088,7 +1114,7 @@ public void Dispose()
                 {
                     foreach (var marshaler in _marshalers)
                     {
-                        Marshaler<T>.DisposeMarshaler(marshaler);
+                        MarshalNonBlittable<T>.DisposeMarshaler(marshaler);
```
Sergio0694, line 1120: Can you maybe leave a comment to explain why it's guaranteed that `T` isn't blittable here? It's not obvious 🤔

Sergio0694 (approved):
Nice!! 😄

src/WinRT.Runtime/Marshalers.cs
manodasanW: Well the class here is `MarshalNonBlittable` and we only use it in the non-blittable scenario other than for enum arrays.  But will add a comment.

build/AzurePipelineTemplates/CsWinRT-FunctionalTest-Steps.yml
manodasanW: I was thinking about that, but for AOT, I thought .NET 9 tests are basically similar if not stricter and we had some .NET 9 AOT fixes.  We still have .NET 8 trimming tests.

Sergio0694 (approved):

PR #2022, opened 2025-07-23
Use FormatMessage for exception messages in cases where the stowed/language exception doesn't contain a message

ahmed605:
Updated HRESULT to Exception conversion logic to fall back to system-formatted error messages via the Win32 FormatMessage API when custom error text is not available or is blank. This ensures more consistent and informative exception messages by leveraging Windows' built-in error descriptions.

Example:
image

### DIFF
--- a/src/WinRT.Runtime/ExceptionHelpers.cs
+++ b/src/WinRT.Runtime/ExceptionHelpers.cs
@@ -227,7 +227,7 @@ namespace WinRT
                 }
             }
 
-            if (string.IsNullOrEmpty(errorMessage))
+            if (string.IsNullOrWhiteSpace(errorMessage))
             {
                 char* message = default;
                 if (Platform.FormatMessageW(0x13FF /* FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS | FORMAT_MESSAGE_MAX_WIDTH_MASK */,
--- a/src/WinRT.Runtime/ExceptionHelpers.cs
+++ b/src/WinRT.Runtime/ExceptionHelpers.cs
@@ -227,6 +227,27 @@ namespace WinRT
                 }
             }
 
+            if (string.IsNullOrEmpty(errorMessage))
+            {
+                char* message = default;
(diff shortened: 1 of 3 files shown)


ahmed605:
@riverar

manodasanW:
Thanks for the change, kicked off CI.

ahmed605:
are the CI failures related to the PR? GitHub doesn't allow me to see the details of them

src/WinRT.Runtime/ExceptionHelpers.cs
```diff
@@ -227,6 +227,27 @@ private static Exception GetExceptionForHR(int hr, bool useGlobalErrorState, boo
                 }
             }
 
+            if (string.IsNullOrEmpty(errorMessage))
```
riverar: Nit: May want to consider `string.IsNullOrWhiteSpace` here to further filter garbage messages reaching users.

src/WinRT.Runtime/Platform.cs
```diff
@@ -48,6 +48,12 @@ public static extern unsafe int WindowsCreateStringReference(
 
         [DllImport("api-ms-win-core-com-l1-1-0.dll")]
         public static extern unsafe int CoCreateFreeThreadedMarshaler(IntPtr outer, IntPtr* marshalerPtr);
+
+        [DllImport("Kernel32.dll")]
```
riverar: Nit: Lowercase kernel32.dll

manodasanW:
> are the CI failures related to the PR? GitHub doesn't allow me to see the details of them

No, it is unrelated.  Have a PR to fix it, after it is merged, you can rebase your changes.

riverar:
I'd still check the error message when possible to ensure we're not getting garbage. Can wrap it in a conditional so that non en-US developers don't break. If that's not a concern for this project, I'd delete the en-US check.

```csharp
[Fact]
public void TestGetExceptionForHR_WithValidHResult_ReturnsSystemFormattedException()
{
    // Arrange
    const int RPC_E_WRONG_THREAD = unchecked((int)0x8001010E);

    // Act
    Exception exception = ExceptionHelpers.GetExceptionForHR(RPC_E_WRONG_THREAD);

    // Assert
    Assert.NotNull(exception);
    Assert.NotNull(exception.Message);
    if (CultureInfo.CurrentUICulture.Name == "en-US")
    {
        Assert.Equal("The application called an interface that was marshalled for a different thread. (0x8001010E)", exception.Message);
    }
}
```

I can't test this for you because the build fails despite following the on-boarding instructions.

manodasanW (approved):

PR #2029, opened 2025-08-07
Add CsWinRT 3.0 overview spec

Sergio0694:
This pull request adds a new specification document outlining the goals, architecture, and adoption plan for CsWinRT 3.0. The document provides background context, details on technical improvements, and describes several breaking changes and enhancements compared to previous versions. The main focus is on modernizing the WinRT interop stack for C#, improving performance, security, and developer experience, and leveraging new .NET 10 features.

### DIFF
--- a/docs/cswinrt3.0-spec.md
+++ b/docs/cswinrt3.0-spec.md
@@ -1,0 +1,129 @@
+# CsWinRT 3.0 overview
+
+## Background context
+
+CsWinRT provides the WinRT interop stack for C# applications. It was built to replace the built-in WinRT interop support that .NET dropped starting from .NET 5. The initial version of CsWinRT was built with specific limitations and constraints (many of which no longer applies), and as a result:
+- It also had to support .NET Standard 2.0 code
+- It also had to support an "embedded", source-only mode
+- Trimming wasn't a consideration at all (it didn't exist yet)
+- AOT wasn't a consideration at all (it didn't exist yet)
+- No projection changes were allowed (decision made at that time)
+- The initial release was also under a lot of time pressure to ship (to align with .NET 5 and WindowsAppSDK 1.0)
+  - The CsWinRT 3.0 effort is driven by internal partner teams adopting C# for Windows components
+  - Unlike CsWinRT 1.0, which had to be published for developers to be able to use WinRT APIs at all on modern .NET, CsWinRT 3.0 doesn't have an urgent need to ship to unblock consumers. Existing C# developers can continue using CsWinRT 2.x for the time being, which allows CsWinRT 3.0 to take the time to "get things right".
+
+All of these original constraints resulted in CsWinRT accumulating technical debt over time, and being structured in a way that fundamentally makes it impossible for it to leverage all the advantages of trimming and AOT in the modern .NET world. Support for trimming and AOT did get added over time, but with it being an afterthought and not something designed into the product from the start, the support for it is not perfect, some things are just such that they cannot be fixed (eg. making runtime class casts trim-compatible and trim-friendly), and a lot of performance is left on the table.
+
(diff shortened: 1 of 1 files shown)


docs/cswinrt3.0-spec.md
```diff
@@ -0,0 +1,129 @@
...
+
+### Multi-targeting support
+
+Because CsWinRT 3.0 is fundamentally incompatible with CsWinRT 2.x, we need a way for NuGet package authors to keep supporting CsWinRT 2.x in parallel with CsWinRT 3.0, so that developers that can't upgrade their apps and libraries to CsWinRT 3.0 (e.g. because they're blocked by one or more dependencies not having updated just yet) will be able to still get new updates for those packages. The solution for this is to leverage the multi-targeting feature from NuGet. Specifically, the .NET SDK will add support for multi-targeting the CsWinRT version based on the revision number of the Windows SDK version indicated in the TFM (see [dotnet/sdk#49721](https://github.com/dotnet/sdk/pull/49721)).
+
+Using a revision value of `.1` will include CsWinRT 3.0 and enable all its tooling. A revision of `.0` will use CsWinRT 2.x, e.g.:
+- `net10.0-windows10.0.22621.0` ---> CsWinRT 2.x
+- `net10.0-windows10.0.22621.1` ---> CsWinRT 3.0
```
riverar, line 125: Nit: Wonder if you should leave a bigger gap there to give each framework some breathing room for their own unplanned revisions. Like perhaps `net10.0-windows10.0.22621.1000`.

Otherwise you could end up in a situation where we have targets .0, .1, and .2 which would be mega confusing.

dongle-the-gadget: I don't believe that .NET already does this today, but my concern would be if someone omitted the revision number entirely (this is supported by .NET), such as `net10.0-windows10.0.26100`.

gus33000: I think @riverar's point has merit, the fact .NET/SDK nuget owners do not apparently already do this today only re-enforces that argument. This would be to account for unplanned unexpected revisions that have to be made, I see a world where this would be a thing that has to occur for critical bug fixes and indeed, not having breathing room for supporting that would not enable this to work properly.

As for whenever its being done already or not, I think the answer is yes, take a look at this ref nuget package for example, there are indeed revisions binding to the windows OS revisions themselves in here, and we had sdk updates in the past too for the same baseline windows in market stable release: https://www.nuget.org/packages/Microsoft.Windows.SDK.NET.Ref

Sergio0694: I think there is some confusion here. There's 3 different "Windows SDK" version numbers that exist:
- The [Windows SDK build](https://developer.microsoft.com/en-us/windows/downloads/windows-sdk/). It uses the revision number to indicate new revisions (which may or may not include new APIs).
- The [Windows SDK projection targeting pack build](https://www.nuget.org/packages/Microsoft.Windows.SDK.NET.Ref). Here the revision number is just an incremental number used to indicate new NuGet package versions. The revision number **does not match** the one of the Windows SDK build that was used to generate the projections from.
- The .NET Windows TFM. Here (currently) the revision number **is always 0**. It doesn't match the Windows SDK build nor the Windows SDK targeting pack version.

It's the .NET SDK that automatically selects the latest known Windows SDK targeting pack (it's hardcoded in the .NET SDK itself). You can also manually indicate it by setting `MicrosoftWindowsSDKPackageVersion`. But the TFM will still always have 0 as the revision, it doesn't matter what actual SDK version is being used.

> "This would be to account for unplanned unexpected revisions that have to be made, I see a world where this would be a thing that has to occur for critical bug fixes and indeed, not having breathing room for supporting that would not enable this to work properly."

Basically this is all to say that this is not correct. The revision of the TFM is unrelated to the Windows SDK revision number, or to the package version of the Windows SDK projection targeting pack. Hence just having .1 for CsWinRT 3.0 is fine, just like so far we've always had just .0 to indicate the normal configuration, and it never caused any issues.

hez2010: Maybe use `net10.0-windows10.0.22621.3` for CsWinRT 3.0 in case we have CsWinRT 4.0 in the future.
It would be clearer if the version number matches.

PR #2034, opened 2025-08-13
Add initial CsWinRT 3.0 prototype (runtime + cswinrtgen)

Sergio0694:
This PR brings over the initial prototype for CsWinRT 3.0 from the hackathon branch. It includes:
- Almost complete WinRT.Runtime.dll, with all the new API surface area, and custom mapped types
- Prototype for cswinrtgen

The new runtime is temporarily using a separate project. We will replace the original one in a follow up. The runtime is partially usable on its own, e.g. for most custom mapped types. The cswinrtgen tool is partially implemented and work in progress.


docs/winrt-interop-dll-spec.md
```diff
@@ -0,0 +1,152 @@
...
+
+All `.` characters in the final mangled name are replaced with `-` characters. Additionally, all `` ` `` characters (backticks) in the final mangled name are replaced with `'` characters (apostrophes).
+
+> [!NOTE]
+> The `.` character replacement is not strictly required, as the `.` character is a valid character for an identifier as per ECMA-335. However, using that character can be inconvenient when using reflection APIs to inspect such types, as it makes it not possible to easily distinguish the namespace from the actual type name. So to account for this, we just do this substitution, given the final length of the mangled name remains the same anyway when doing so.
+> 
+> The `` ` `` character replacement is done to avoid issues with tooling, such as ILSpy, which assume that all types with `` ` `` in their name are generic types. The generated types for marshalling generic types are not themselves generic, which causes these tools to incorrectly interpret the type metadata.
+
+These are the well-known assemblies and their compact identifiers:
```
manodasanW, line 43: Any reason we need compact for the last 2?  For instance, what is the benefit of compact?

docs/winrt-interop-dll-spec.md
```diff
@@ -0,0 +1,152 @@
...
+
+The **mangled namespace** for a given type is defined as follows:
+
+1. **No namespace**: if the input type has no namespace, the generated namespace defaults to `ABI`.
+2. **Existing namespace**: if the input type has a namespace, the generated namespace prepends `ABI.` to the original namespace.
+
+The **mangled type name** for a given type is defined as follows:
+
+1. **Primitive types**: well-known primitive types (e.g., `int`, `string`) are mapped to their corresponding identifiers.
+2. **User-defined types**: the type name is prefixed with the assembly name (or a compact identifier for well-known assemblies) within angle brackets (i.e. `<>`), and suffixed with the type name. The namespace is not included, as it matches the containing namespace for the generated type (without the `ABI[.]` prefix).
```
manodasanW, line 32: As discussed before, we may need a separate rule for projected RCW types if cswinrt.exe is expected to refer to it.  This is because, cswinrt.exe won't know the what the projection assembly is, only the winmd would be known.  We can address this later but calling it out to not loose track.

src/WinRT.Interop.Generator/Builders/InteropTypeDefinitionBuilder.Delegate.cs
```diff
@@ -0,0 +1,793 @@
...
+                baseType: module.CorLibTypeFactory.Object.ToTypeDefOrRef())
+            {
+                Interfaces = { new InterfaceImplementation(interopReferences.IWindowsRuntimeComWrappersCallback.Import(module)) }
+            };
+
+            module.TopLevelTypes.Add(callbackType);
+
+            // Define the 'CreateObject' method as follows:
+            //
+            // public static object CreateObject(void* value)
```
manodasanW: The signature in the comment doesn't match with below, does it?

src/WinRT.Interop.Generator/Builders/InteropTypeDefinitionBuilder.Delegate.cs
```diff
@@ -0,0 +1,793 @@
...
+            // Get the method body for the 'Invoke' method
+            CilMethodBody invokeBody = invokeMethod.CilMethodBody;
+            CilInstructionCollection invokeInstructions = invokeBody.Instructions;
+
+            // Import 'WindowsRuntimeObjectReferenceValue', compute it just once
+            TypeSignature windowsRuntimeObjectReferenceValueType = interopReferences.WindowsRuntimeObjectReferenceValue
+                .Import(module)
+                .ToValueTypeSignature();
+
+            // Declare 3 variables:
```
manodasanW:
```suggestion
            // Declare 4 variables:
```

src/WinRT.Interop.Generator/Builders/InteropTypeDefinitionBuilder.Delegate.cs
```diff
@@ -0,0 +1,793 @@
...
+
+            _ = ctor.CilMethodBody!.Instructions.Insert(0, Ldarg_0);
+            _ = ctor.CilMethodBody!.Instructions.Insert(1, Call, interopReferences.WindowsRuntimeComWrappersMarshallerAttribute_ctor.Import(module));
+
+            // The 'ComputeVtables' method returns the 'ComWrappers.ComInterfaceEntry*' type
+            PointerTypeSignature computeVtablesReturnType = interopReferences.ComInterfaceEntry.Import(module).MakePointerType();
+
+            // Define the 'ComputeVtables' method as follows:
+            //
+            // public static ComInterfaceEntry* ComputeVtables(out int count)
```
manodasanW: static or virtual?

src/WinRT.Interop.Generator/Builders/InteropTypeDefinitionBuilder.EventSource.cs
```diff
@@ -0,0 +1,149 @@
...
+            // We're declaring an 'internal sealed class' type
+            eventSourceType = new(
+                ns: "ABI.WindowsRuntime.InteropServices"u8,
+                name: InteropUtf8NameFactory.TypeName(baseEventSourceType),
+                attributes: TypeAttributes.AutoLayout | TypeAttributes.Sealed | TypeAttributes.BeforeFieldInit,
+                baseType: baseEventSourceType.Import(module).ToTypeDefOrRef());
+
+            module.TopLevelTypes.Add(eventSourceType);
+
+            // Define the constructor
```
manodasanW: Can we put the ctor in the comment too like with the others.

src/WinRT.Interop.Generator/Builders/InteropTypeDefinitionBuilder.IEnumerator1.cs
```diff
@@ -0,0 +1,623 @@
...
+                currentMethod.CilMethodBody.LocalVariables.Add(loc_3_current);
+
+                // Jump labels for the 'try/finally' blocks
+                CilInstruction ldloc_2_tryStart = new(Ldloc_2);
+                CilInstruction ldloc_2_finallyStart = new(Ldloc_2);
+                CilInstruction ldloc_3_finallyEnd = new(Ldloc_3);
+
+                // We need to marshal the native value to a managed object
+                instructions.Add(ldloc_2_tryStart);
+                _ = instructions.Add(Call, interopReferences.HStringMarshallerConvertToManaged.Import(module));
```
manodasanW, line 174: Do we need a check for if it is a string.

src/WinRT.Interop.Generator/Builders/InteropTypeDefinitionBuilder.IEnumerator1.cs
```diff
@@ -0,0 +1,623 @@
...
+                {
+                    { Ldarg_0 },
+                    { Call, interopReferences.IIteratorMethodsHasCurrent.Import(module) },
+                    { Ret }
+                }
+            };
+
+            // Define the 'MoveNext' method as follows:
+            //
+            // public static bool HasCurrent(WindowsRuntimeObjectReference thisReference)
```
manodasanW:
```suggestion
            // public static bool MoveNext(WindowsRuntimeObjectReference thisReference)
```

src/WinRT.Interop.Generator/Builders/InteropTypeDefinitionBuilder.IEnumerator1.cs
```diff
@@ -0,0 +1,623 @@
...
+
+            interfaceImplType.Properties.Add(enumeratorCurrentProperty);
+
+            // Define the 'System.IEnumerator.MoveNext' method
+            MethodDefinition moveNextMethod = new(
+                name: "System.IEnumerator.MoveNext"u8,
+                attributes: WellKnownMethodAttributesFactory.ExplicitInterfaceImplementationInstanceMethod,
+                signature: MethodSignature.CreateInstance(module.CorLibTypeFactory.Boolean));
+
+            // Add and implement the 'Reset' method
```
manodasanW:
```suggestion
            // Add and implement the 'MoveNext' method
```

src/WinRT.Interop.Generator/Builders/InteropTypeDefinitionBuilder.IList1.cs
```diff
@@ -0,0 +1,1112 @@
...
+                SetMethod = set_ItemMethod
+            };
+
+            interfaceImplType.Properties.Add(itemProperty);
+
+            // Create the 'IndexOf' method
+            MethodDefinition indexOfMethod = new(
+                name: $"System.Collections.Generic.IList<{elementType.FullName}>.IndexOf",
+                attributes: WellKnownMethodAttributesFactory.ExplicitInterfaceImplementationInstanceMethod,
+                signature: MethodSignature.CreateInstance(module.CorLibTypeFactory.Int32, elementType.Import(module)));
```
manodasanW, line 770: Is this right?  Should Int32 be the return type.

manodasanW: Or is the first parameter here return type?

src/WinRT.Interop.Generator/Builders/InteropTypeDefinitionBuilder.IReadOnlyDictionary2.cs
```diff
@@ -0,0 +1,729 @@
...
+        /// <param name="vftblType">The resulting vtable type.</param>
+        public static void Vftbl(
+            GenericInstanceTypeSignature readOnlyDictionaryType,
+            InteropDefinitions interopDefinitions,
+            InteropReferences interopReferences,
+            InteropGeneratorEmitState emitState,
+            ModuleDefinition module,
+            out TypeDefinition vftblType)
+        {
+            TypeSignature keyType = readOnlyDictionaryType.TypeArguments[0];
```
manodasanW, line 66: Don't we need to check the value type too and do below checks based on that like with IDictionary.

src/WinRT.Interop.Generator/Builders/InteropTypeDefinitionBuilder.IReadOnlyList1.cs
```diff
@@ -0,0 +1,652 @@
...
+
+                // Jump labels for the 'try/finally' blocks
+                CilInstruction ldloc_2_tryStart = new(Ldloc_2);
+                CilInstruction ldloc_2_finallyStart = new(Ldloc_2);
+                CilInstruction ldloc_3_finallyEnd = new(Ldloc_3);
+
+                // We need to marshal the native value to a managed object
+                getAtMethod.CilMethodBody.Instructions.ReplaceRange(nop_implementation, [
+                    ldloc_2_tryStart,
+                    new(Call, interopReferences.HStringMarshallerConvertToManaged.Import(module)),
```
manodasanW, line 217: Same here, do we need to check for type to determine marshaler.

src/WinRT.Interop.Generator/Builders/InteropTypeDefinitionBuilder.SzArray.cs
```diff
@@ -0,0 +1,549 @@
...
+                implType: out interfaceEntriesImplType,
+                implTypes: [
+                    (implType.GetMethod("get_IID"u8), implType.GetMethod("get_Vtable"u8)),
+                    (interopReferences.IPropertyValueImplget_IID, interopReferences.IPropertyValueImplget_OtherTypeVtable), // TODO
+                    (interopReferences.IStringableImplget_IID, interopReferences.IStringableImplget_Vtable),
+                    (interopReferences.IWeakReferenceSourceImplget_IID, interopReferences.IWeakReferenceSourceImplget_Vtable),
+                    (interopReferences.IMarshalImplget_IID, interopReferences.IMarshalImplget_Vtable),
+                    (interopReferences.IAgileObjectImplget_IID, interopReferences.IAgileObjectImplget_Vtable),
+                    (interopReferences.IInspectableImplget_IID, interopReferences.IInspectableImplget_Vtable),
+                    (interopReferences.IUnknownImplget_IID, interopReferences.IUnknownImplget_Vtable)]);
```
manodasanW, line 369: TODO for IReferenceArray?

src/WinRT.Interop.Generator/Errors/WellKnownInteropExceptions.cs
```diff
@@ -0,0 +1,337 @@
...
+    /// Failed to discover SZ array types.
+    /// </summary>
+    public static Exception WinRTRuntimeDllVersion2References(IEnumerable<string> names)
+    {
+        string combinedNames = string.Join(", ", names.Select(static name => $"'{name}'"));
+
+        return Exception(35,
+            $"One or more referenced assemblies were compiled against CsWinRT 2.x (i.e. referencing 'WinRT.Runtime.dll' version '2.x'), which is not compatible with CsWinRT 3.0 (i.e. referencing 'WinRT.Runtime.dll' version '3.0'). " +
+            $"CsWinRT 3.0 is not backward-compatible with CsWinRT 2.x, so referencing any assemblies targeting an older version of CsWinRT is not supported. Those assemblies will need to be recompiled using CsWinRT 3.0 (i.e. using" +
+            $"a TFM such as 'net10.0-windows10.0.26100.1', where the '.1' revision number at the end of the target platform version indicates CsWinRT 3.0). The names of the assemblies causing issues are: {combinedNames}.");
```
manodasanW:
```suggestion
            $"a TFM such as 'net10.0-windows10.0.26100.1', where the '.1' revision number at the end of the target platform version indicates CsWinRT 3.0). The following assemblies are compiled with the older version: {combinedNames}.");
```

src/WinRT.Interop.Generator/Extensions/TypeDefinitionExtensions.cs
```diff
@@ -0,0 +1,161 @@
...
+
+            // Verify the provided signature is valid
+            if (!SignatureComparer.IgnoreVersion.Equals(type, typeSignature.GenericType))
+            {
+                throw new ArgumentException("The input type signature does not match the type definition.", nameof(typeSignature));
+            }
+
+            GenericContext genericContext = new(typeSignature, null);
+
+            foreach (InterfaceImplementation interfaceImplementation in type.Interfaces)
```
manodasanW, line 138: What scenarios is this used in?  If for vtable, do this also handle covariance?

src/WinRT.Interop.Generator/Factories/InteropUtf8NameFactory.cs
```diff
@@ -0,0 +1,204 @@
...
+            }
+            else if (typeSignature is SzArrayTypeSignature arrayTypeSignature)
+            {
+                // Same as below (see comments there), but we use the element type name, not the array type name
+                interpolatedStringHandler.AppendFormatted(depth == 0 ? arrayTypeSignature.BaseType.Name! : arrayTypeSignature.BaseType.FullName);
+            }
+            else
+            {
+                // If the type is a type definition, append the name of the type definition.
+                // Just like with generic types, we can skip the namespace if the deth is '0'.
```
manodasanW:
```suggestion
                // Just like with generic types, we can skip the namespace if the depth is '0'.
```

src/WinRT.Interop.Generator/Generation/InteropGenerator.Discover.cs
```diff
@@ -0,0 +1,355 @@
...
+        }
+
+        discoveryState.TrackModuleDefinition(path, module);
+
+        args.Token.ThrowIfCancellationRequested();
+
+        // We're only interested in harvesting .dll-s which reference the Windows SDK projections.
+        // This is true for all .dll-s that were built targeting 'netX.0-windows10.0.XXXX.0'.
+        // So this check effectively lets us filter all .dll-s that were in projects with this TFM.
+        if (!module.IsOrReferencesWindowsSDKProjectionsAssembly)
```
manodasanW: Should we check WinRT.Runtime instead?  What if a projection is just of fundamental types?

src/WinRT.Interop.Generator/Resolvers/PathAssemblyResolver.cs
```diff
@@ -0,0 +1,97 @@
...
+    private readonly ConcurrentDictionary<AssemblyDescriptor, AssemblyDefinition> _cache = new(SignatureComparer.IgnoreVersion);
+
+    /// <summary>
+    /// Creates a new <see cref="PathAssemblyResolver"/> instance with the specified parameters.
+    /// </summary>
+    /// <param name="referencePath">The input .dll paths.</param>
+    public PathAssemblyResolver(string[] referencePath)
+    {
+        _referencePath = referencePath;
+        ReaderParameters = new RuntimeContext(new DotNetRuntimeInfo(".NETCoreApp", new Version(10, 0)), this).DefaultReaderParameters;
```
manodasanW, line 36: How is the .NET version used here?  Should it be hard coded?

src/WinRT.Runtime2/ABI/System/Collections.Specialized/NotifyCollectionChangedEventArgs.cs
```diff
@@ -0,0 +1,263 @@
...
+
+        ref readonly Guid iid = ref WindowsRuntimeFeatureSwitches.UseWindowsUIXamlProjections
+            ? ref WellKnownInterfaceIds.IID_WUX_INotifyCollectionChangedEventArgs
+            : ref WellKnownInterfaceIds.IID_MUX_INotifyCollectionChangedEventArgs;
+
+        IUnknownVftbl.QueryInterfaceUnsafe(value, in iid, out void* result).Assert();
+
+        try
+        {
+            return NotifyCollectionChangedEventArgsMarshaller.ConvertToManaged(value)!;
```
manodasanW:
```suggestion
            return NotifyCollectionChangedEventArgsMarshaller.ConvertToManaged(result)!;
```

src/WinRT.Runtime2/ABI/System/Collections.Specialized/NotifyCollectionChangedEventHandler.cs
```diff
@@ -0,0 +1,344 @@
...
+    {
+        count = sizeof(NotifyCollectionChangedEventHandlerInterfaceEntries) / sizeof(ComInterfaceEntry);
+
+        return (ComInterfaceEntry*)Unsafe.AsPointer(in NotifyCollectionChangedEventHandlerInterfaceEntriesImpl.Entries);
+    }
+
+    /// <inheritdoc/>
+    public override object CreateObject(void* value, out CreatedWrapperFlags wrapperFlags)
+    {
+        wrapperFlags = CreatedWrapperFlags.NonWrapping;
```
manodasanW, line 194: Would this be NonWrapping?  Technically invoke has the objref right?

src/WinRT.Runtime2/ABI/System/ComponentModel/DataErrorsChangedEventArgs.cs
```diff
@@ -0,0 +1,130 @@
...
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    public override object CreateObject(void* value, out CreatedWrapperFlags wrapperFlags)
+    {
+        wrapperFlags = CreatedWrapperFlags.NonWrapping;
+
+        IUnknownVftbl.QueryInterfaceUnsafe(value, in WellKnownInterfaceIds.IID_DataErrorsChangedEventArgs, out void* result).Assert();
+
+        try
+        {
+            return DataErrorsChangedEventArgsMarshaller.ConvertToManaged(value)!;
```
manodasanW:
```suggestion
            return DataErrorsChangedEventArgsMarshaller.ConvertToManaged(result)!;
```

src/WinRT.Runtime2/ABI/System/ComponentModel/INotifyDataErrorInfo.cs
```diff
@@ -0,0 +1,331 @@
...
+
+            *token = ErrorsChanged.GetOrCreateValue(unboxedValue).AddEventHandler(managedHandler);
+
+            unboxedValue.ErrorsChanged += managedHandler;
+
+            return WellKnownErrorCodes.S_OK;
+        }
+        catch (global::System.Exception e)
+        {
+            return e.HResult;
```
manodasanW, line 231: Probably want to call RestrictedErrorInfoExceptionMarshaller

src/WinRT.Runtime2/ABI/System/ComponentModel/INotifyPropertyChanged.cs
```diff
@@ -0,0 +1,204 @@
...
+
+        Vftbl.add_PropertyChanged = &add_PropertyChanged;
+        Vftbl.remove_PropertyChanged = &remove_PropertyChanged;
+    }
+
+    /// <summary>
+    /// Gets the IID for <see cref="global::System.ComponentModel.INotifyPropertyChanged"/>.
+    /// </summary>
+    public static ref readonly Guid IID => ref WindowsRuntimeFeatureSwitches.UseWindowsUIXamlProjections
+        ? ref WellKnownInterfaceIds.IID_WUX_INotifyCollectionChanged
```
manodasanW:
```suggestion
        ? ref WellKnownInterfaceIds.IID_WUX_INotifyPropertyChanged
```

src/WinRT.Runtime2/ABI/System/ComponentModel/INotifyPropertyChanged.cs
```diff
@@ -0,0 +1,204 @@
...
+        : ref WellKnownInterfaceIds.IID_MUX_INotifyCollectionChanged;
```
manodasanW:
```suggestion
        : ref WellKnownInterfaceIds.IID_MUX_INotifyPropertyChanged;
```

src/WinRT.Runtime2/ABI/System/ComponentModel/PropertyChangedEventArgs.cs
```diff
@@ -0,0 +1,153 @@
...
+
+        ref readonly Guid iid = ref WindowsRuntimeFeatureSwitches.UseWindowsUIXamlProjections
+            ? ref WellKnownInterfaceIds.IID_WUX_PropertyChangedEventArgs
+            : ref WellKnownInterfaceIds.IID_MUX_PropertyChangedEventArgs;
+
+        IUnknownVftbl.QueryInterfaceUnsafe(value, in iid, out void* result).Assert();
+
+        try
+        {
+            return PropertyChangedEventArgsMarshaller.ConvertToManaged(value)!;
```
manodasanW:
```suggestion
            return PropertyChangedEventArgsMarshaller.ConvertToManaged(result)!;
```

src/WinRT.Runtime2/ABI/System/Collections.Specialized/NotifyCollectionChangedEventHandler.cs
manodasanW: Oh, is this due to this is IReference.

src/WinRT.Runtime2/ABI/System/Object.cs
```diff
@@ -0,0 +1,87 @@
...
+    {
+        count = sizeof(ObjectInterfaceEntries) / sizeof(ComInterfaceEntry);
+
+        return (ComInterfaceEntry*)Unsafe.AsPointer(in ObjectInterfaceEntriesImpl.Entries);
+    }
+
+    /// <inheritdoc/>
+    public override unsafe object CreateObject(void* value, out CreatedWrapperFlags wrapperFlags)
+    {
+        // Marshalling 'object' instances is not supported, and it should just never happen
```
manodasanW, line 84: Is this more that ComWrappers will handle this path separately without calling here or?

src/WinRT.Runtime2/ABI/System/Uri.cs
```diff
@@ -0,0 +1,126 @@
...
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    public override object CreateObject(void* value, out CreatedWrapperFlags wrapperFlags)
+    {
+        wrapperFlags = CreatedWrapperFlags.NonWrapping;
+
+        IUnknownVftbl.QueryInterfaceUnsafe(value, in WellKnownInterfaceIds.IID_UriRuntimeClass, out void* result).Assert();
+
+        try
+        {
+            return UriMarshaller.ConvertToManaged(value)!;
```
manodasanW:
```suggestion
            return UriMarshaller.ConvertToManaged(result)!;
```

src/WinRT.Runtime2/WindowsRuntimeObject.cs
```diff
@@ -0,0 +1,753 @@
...
+    /// Creates a <see cref="WindowsRuntimeObject"/> instance with the specified parameters for sealed scenarios.
+    /// </summary>
+    /// <param name="_">Marker parameter used to select this constructor for sealed types (unused).</param>
+    /// <param name="activationFactoryObjectReference">The <see cref="WindowsRuntimeObjectReference"/> for the <c>IActivationFactory</c> instance.</param>
+    /// <param name="iid">The IID of the default interface for the Windows Runtime class being constructed.</param>
+    /// <exception cref="ArgumentNullException">Thrown if <paramref name="activationFactoryObjectReference"/> is <see langword="null"/>.</exception>
+    /// <exception cref="ObjectDisposedException">Thrown if <paramref name="activationFactoryObjectReference"/> has been disposed.</exception>
+    /// <exception cref="Exception">Thrown if there's any errors when activating the underlying native object.</exception>
+    /// <remarks>
+    /// This constructor should only be used when activating seald types (both projected and user-defined types).
```
manodasanW:
```suggestion
    /// This constructor should only be used when activating sealed types (both projected and user-defined types).
```

src/WinRT.Runtime2/InteropServices/TypeMapInfo/WindowsRuntimeVtableInfo.cs
```diff
@@ -0,0 +1,33 @@
...
+        Count = count;
+    }
+
+    /// <summary>
+    /// Gets the computed vtable entries, allocated in native memory.
+    /// </summary>
+    public ComWrappers.ComInterfaceEntry* VtableEntries { get; }
+
+    /// <summary>
+    /// Gets tt\he number of elements in <see cref="VtableEntries"/>.
```
manodasanW:
```suggestion
    /// Gets the number of elements in <see cref="VtableEntries"/>.
```

src/WinRT.Runtime2/InteropServices/TypeMapInfo/WindowsRuntimeVtableInfo.cs
```diff
@@ -0,0 +1,33 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Runtime.InteropServices;
+
+namespace WindowsRuntime.InteropServices;
+
+/// <summary>
+/// A type holding infor on a computed vtable for a given managed type.
```
manodasanW:
```suggestion
/// A type holding info on a computed vtable for a given managed type.
```

src/WinRT.Runtime2/InteropServices/ProjectionImpls/IPropertyValueImpl.cs
```diff
@@ -0,0 +1,29 @@
...
+/// <see href="https://learn.microsoft.com/uwp/api/windows.foundation.ipropertyvalue"/>
+public static partial class IPropertyValueImpl
+{
+    /// <summary>
+    /// Gets the IID for the <c>IPropertyValue</c> interface.
+    /// </summary>
+    public static ref readonly Guid IID
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        get => ref WellKnownInterfaceIds.IID_IInspectable;
```
manodasanW: I assume this is a TODO?

src/WinRT.Runtime2/Collections/WindowsRuntimeEnumerable{T}.cs
```diff
@@ -0,0 +1,64 @@
...
+/// </summary>
+/// <typeparam name="T">The type of objects to enumerate.</typeparam>
+/// <typeparam name="TIIterableMethods">The <c>Windows.Foundation.Collections.IIterable&lt;T&gt;</c> implementation type.</typeparam>
+/// <remarks>
+/// This type should only be used as a base type by generated generic instantiations.
+/// </remarks>
+/// <see href="https://learn.microsoft.com/uwp/api/windows.foundation.collections.iiterable-1"/>
+[Obsolete("This type is an implementation detail, and it's only meant to be consumed by 'cswinrtgen'")]
+[EditorBrowsable(EditorBrowsableState.Never)]
+public abstract class WindowsRuntimeEnumerable<T, TIIterableMethods> : WindowsRuntimeObject, IEnumerable<T>, IWindowsRuntimeInterface<IEnumerable<T>>
```
manodasanW, line 25: If all types in this folder are used by cswinrtgen, should they be internal or is there a reason to be public.

src/WinRT.Runtime2/InteropServices/Activation/WindowsRuntimeActivationFactory.cs
```diff
@@ -0,0 +1,357 @@
...
+        if (!WellKnownErrorCodes.Succeeded(hresult))
+        {
+            return hresult;
+        }
+
+        // At this point we have an activation factory, and we need to 'QueryInterface' for the exact IID
+        hresult = IUnknownVftbl.QueryInterfaceUnsafe(activationFactoryUnknown, in iid, out activationFactory);
+
+        // We can always release the original activation factory at this point
+        _ = IUnknownVftbl.ReleaseUnsafe(activationFactory);
```
manodasanW:
```suggestion
        _ = IUnknownVftbl.ReleaseUnsafe(activationFactoryUnknown);
```

src/WinRT.Runtime2/InteropServices/Activation/WindowsRuntimeActivationFactoryCallback.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+    /// <summary>
+    /// A callback to activate a composed Windows Runtime object with custom logic.
+    /// </summary>
+    /// <param name="additionalParameters">The additional parameters to provide to the activation callback.</param>
+    /// <param name="baseInterface">The <see cref="WindowsRuntimeObject"/> instance being constructed.</param>
+    /// <param name="innerInterface">The resulting non-delegating <c>IInspectable</c> object.</param>
+    /// <param name="defaultInterface">The resulting default interface pointer.</param>
+    /// <see href="https://learn.microsoft.com/uwp/winrt-cref/winrt-type-system#composable-activation"/>
+    public unsafe delegate void DerivedComposed(
```
manodasanW, line 21: Maybe we should call these Composed and Sealed dropping the Derived prefix.  We can figure this out in a later PR too, just noting it.

src/WinRT.Runtime2/InteropServices/Activation/WindowsRuntimeActivationHelper.cs
```diff
@@ -0,0 +1,186 @@
...
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    public static unsafe void ActivateInstanceUnsafe(WindowsRuntimeObjectReference activationFactoryObjectReference, out void* defaultInterface)
+    {
+        using WindowsRuntimeObjectReferenceValue activationFactoryValue = activationFactoryObjectReference.AsValue();
+
+        fixed (void** defaultInterfacePtr = &defaultInterface)
+        {
+            HRESULT hresult = IActivationFactoryVftbl.ActivateInstanceUnsafe(
+                thisPtr: activationFactoryValue.GetThisPtrUnsafe(),
+                baseInterface: null,
```
manodasanW: If this is for parameter less constructor, then should it be calling the one that doesn't have baseInterface and innerInterface.

src/WinRT.Runtime2/InteropServices/Marshalling/WindowsRuntimeArrayMarshaller.cs
```diff
@@ -0,0 +1,60 @@
...
+        // First, make sure we have the right 'IReferenceArray<T>' interface on 'value'
+        IUnknownVftbl.QueryInterfaceUnsafe(value, in iid, out void* referencePtr).Assert();
+
+        try
+        {
+            uint count;
+            void* result;
+
+            // Unbox the underlying array (we always just discard the outer reference)
+            IReferenceArrayVftbl.get_ValueUnsafe(value, &count, &result).Assert();
```
manodasanW:
```suggestion
            IReferenceArrayVftbl.get_ValueUnsafe(referencePtr, &count, &result).Assert();
```

src/WinRT.Runtime2/InteropServices/Marshalling/WindowsRuntimeValueTypeMarshaller.cs
```diff
@@ -0,0 +1,135 @@
...
+    /// <exception cref="NullReferenceException">Thrown if <paramref name="value"/> is <see langword="null"/>.</exception>
+    /// <exception cref="Exception">Thrown if <paramref name="value"/> cannot be marshalled.</exception>
+    public static T UnboxToManagedUnsafe<T>(void* value, in Guid iid)
+        where T : unmanaged
+    {
+        IUnknownVftbl.QueryInterfaceUnsafe(value, in iid, out void* referencePtr).Assert();
+
+        T result;
+
+        IReferenceVftbl.get_ValueUnsafe(referencePtr, &result).Assert();
```
manodasanW: release referencePtr?

src/WinRT.Runtime2/InteropServices/Marshalling/WindowsRuntimeInterfaceMarshaller.cs
```diff
@@ -0,0 +1,95 @@
...
+            return default;
+        }
+
+        // If 'value' is a projected runtime class, get the cached object reference for the interface.
+        // This is a critical optimization that avoids 'QueryInterface' for interface method parameters.
+        if (value is IWindowsRuntimeInterface<T> windowsRuntimeInterface)
+        {
+            return windowsRuntimeInterface.GetInterface();
+        }
+
```
manodasanW, line 52: Would we still need a fallback or can we say in all scenarios that this is called, IWindowsRuntimeInterface<T> should be implemented for T.

        if (value is WindowsRuntimeObject { HasUnwrappableNativeObjectReference: true } windowsRuntimeObject)

src/WinRT.Runtime2/InteropServices/ObjectReference/WindowsRuntimeObjectReference.Initialization.cs
```diff
@@ -0,0 +1,317 @@
...
+            // The runtime has already done 'AddRefFromTrackerSource' for this instance, so it would also handle doing
+            // 'ReleaseFromTrackerSource' upon finalization. Because of that, we prevent it in the object reference.
+            createObjectReferenceFlags |= CreateObjectReferenceFlags.PreventReleaseFromTrackerSourceOnDispose;
+
+            // To balance the overall reference count on the inner instance (see notes at the start of the method),
+            // we need to release it here. This would've been handled in a 'finally' block in caller methods otherwise.
+            // We can only avoid this in the aggregation scenario, because we balance it out with the initial 'AddRef'
+            // that we can skip. But when not aggregating, that 'AddRef' is on the new instance, so the inner instance
+            // would have an extra reference count by the end of this method. We can fix that here to balance it again.
+            _ = IUnknownVftbl.ReleaseUnsafe(acquiredInnerInstanceUnknown);
```
manodasanW, line 221: Would this be newInstance or Inner?  Do we have inner, if there is no aggregation here?

src/WinRT.Runtime2/InteropServices/WindowsRuntimeComWrappers.cs
```diff
@@ -0,0 +1,345 @@
...
+        nint externalComObject,
+        WindowsRuntimeObjectComWrappersCallback? objectComWrappersCallback,
+        WindowsRuntimeUnsealedObjectComWrappersCallback? unsealedObjectComWrappersCallback)
+    {
+        ObjectComWrappersCallback = objectComWrappersCallback;
+        UnsealedObjectComWrappersCallback = unsealedObjectComWrappersCallback;
+        CreateObjectTargetInterfacePointer = (void*)externalComObject;
+
+        // We always pass no flags, as our implementation will use 'CreatedWrapperFlags' to signal the
+        // appropriate flags back from the marshalling stubs. We do pass a user state so that we can
```
manodasanW, line 210: >We do pass a user state

Is that accurate if the below is null?

docs/winrt-interop-dll-spec.md
Sergio0694: > "Any reason we need compact for the last 2?"

I figured they were just common enough. We can add more as well (e.g. for WinUI 3 and whatnot).

> "what is the benefit of compact?"

Makes the type names much shorter and more readable. Especially for very common assemblies, it makes inspecting the generated .dll-s much more convenient as well in ILSpy, as the assembly name doesn't take up so much space.

docs/winrt-interop-dll-spec.md
Sergio0694: Yeah good point. I think we should go back to this after we figure out the ref assembly story, as I think that would also likely end up affecting how we decide to handle this. In theory we also have that option we discussed of passing a list of winmd names to interop .dll-s as arguments to cswinrtgen. I'm sure we can figure out some good solution 🙂

src/WinRT.Interop.Generator/Builders/InteropTypeDefinitionBuilder.Delegate.cs
Sergio0694: Ah yes, forgot to add the new `wrapperFlags` parameter. Fixed, thank you! 🙂

src/WinRT.Interop.Generator/Builders/InteropTypeDefinitionBuilder.IEnumerator1.cs
Sergio0694: Oh, yeah this whole thing is a TODO, I just use the `HSTRING` marshaller as a placeholder everywhere 😅

src/WinRT.Interop.Generator/Builders/InteropTypeDefinitionBuilder.IList1.cs
Sergio0694: Yup, return type is first and then there's `params` for any parameter types 👍

src/WinRT.Interop.Generator/Builders/InteropTypeDefinitionBuilder.IReadOnlyDictionary2.cs
Sergio0694: It's not needed here because for readonly dictionaries we never have the value type passed by value in the ABI. See comments the lines right below this that explain this better. It's basically just a size optimization we can do 🙂

src/WinRT.Interop.Generator/Builders/InteropTypeDefinitionBuilder.IReadOnlyList1.cs
Sergio0694: Same as above, the `HSTRING` marshaller is just a placeholder here for now 😅

src/WinRT.Interop.Generator/Builders/InteropTypeDefinitionBuilder.SzArray.cs
Sergio0694: `IReferenceArray` is already plugged in (first entry), the TODO is to plug the right `IPropertyValue` vtables for arrays of projected types. I'll probably implement them in the runtime and then expose them so we can wire them up here.

src/WinRT.Interop.Generator/Extensions/TypeDefinitionExtensions.cs
Sergio0694: We use this in `DiscoverGenericTypeInstantiations`. It's needed to ensure we discover all generic interface instantiations even if they're e.g. from other assemblies and used as base interfaces. E.g. say we see a type that implements `IFoo`, but then `IFoo : IEnumerable<int>`. We also want to make sure that `IEnumerable<int>` is discovered. I added this later as without this, we were missing a bunch of generic type instantiations specifically for base interfaces. For instance, we'd be seeing a bunch of `IList<T>` or `IDictionary<K, V>` instantiations, but then we'd miss the associated constructed base interfaces (e.g. `IEnumerable<T>`, `IEnumerable<KeyValuePair<K, V>>`, so we'd fail during the emit phase when trying to look them up. This method ensures we always see the full transitive closure of all generic interfaces we care about.

src/WinRT.Runtime2/ABI/System/ComponentModel/DataErrorsChangedEventArgs.cs
Sergio0694: Good catch! 😅

src/WinRT.Runtime2/ABI/System/Collections.Specialized/NotifyCollectionChangedEventHandler.cs
Sergio0694: Yup! Here we're unboxing and marshalling the inner value, not the original input native object 👍

src/WinRT.Runtime2/ABI/System/Object.cs
Sergio0694: The reason is we'll never be hitting this path, because it's not possible to be creating an RCW for an object that reports "System.Object" as runtime class name, which is the only way we could possibly get here. Also I've deleted this method entirely in a follow-up PR, as I've made it just virtual and throwing-by-default in the base attribute.

src/WinRT.Runtime2/InteropServices/ProjectionImpls/IPropertyValueImpl.cs
Sergio0694: Oops, forgot to update this, thanks! 😅

src/WinRT.Runtime2/Collections/WindowsRuntimeEnumerable{T}.cs
Sergio0694: I consider all of these to be part of the public API contract for cswinrtgen. Yes they could technically be internal, but I'd like to avoid explicitly designing this to have all generated code depend on internal types from another .dll (only exception being for marshalled types we generate CCWs for, that's different and a special case). Doing so just feels more like a hack to me. This way instead they're effectively the same, like, they're hidden and also explicitly marked as obsolete, but at least they're technically public. In theory we could even make `[Obsolete]` be an error instead of a warning, to further enforce this, so the types would be completely unusable from source. What do you think?

src/WinRT.Runtime2/InteropServices/Marshalling/WindowsRuntimeInterfaceMarshaller.cs
Sergio0694: Mmh I'm thinking we might need a fallback maybe to account for cases where you get a `T` instance from some RCW via `IDynamicInterfaceCastable` and pass that? In that case we'd want to unwrap and QI, not create a CCW, right? 🤔

manodasanW: Right, we want to unwrap and QI or get the objref from IDIC, not create a CCW.

src/WinRT.Runtime2/Collections/WindowsRuntimeEnumerable{T}.cs
manodasanW: I see, for now, what we have is fine.

src/WinRT.Interop.Generator/Generation/InteropGenerator.Discover.cs
Sergio0694: I think I didn't do that as it was causing issues for assemblies that didn't have an explicit reference to WinRT.Runtime.dll, but only to a projection .dll, though I've updated to also just explicitly check for that too now 👍

src/WinRT.Interop.Generator/Resolvers/PathAssemblyResolver.cs
Sergio0694: For our purposes, it's only used so that AsmResolver will use the .NET 10 reference corlib when resolving types from corlib, which is what we want. It shouldn't matter even if you tried running this against a .dll compiled against .NET 11 or and older .NET version I think. @windows10ce can you confirm this is fine? 🤔

src/WinRT.Runtime2/ABI/System/ComponentModel/INotifyDataErrorInfo.cs
Sergio0694: Do we have some exact guidance on when one should use that marshaller vs. just returning the HRESULT? I remember I was a bit confused because in CsWinRT 2.x we sometimes just do the latter, but it's not clear to me whether it's intentional or an oversight. For instance, this is the same in 2.x: if you look at e.g. `Do_Abi_add_ErrorsChanged_1`, we're also just returning the HRESULT directly in the exports for the `add`/`remove` methods. Is this by design?

src/WinRT.Runtime2/InteropServices/Activation/WindowsRuntimeActivationHelper.cs
Sergio0694: Fixed!

src/WinRT.Runtime2/InteropServices/Marshalling/WindowsRuntimeInterfaceMarshaller.cs
Sergio0694: Done! Also improved the exception messages.

src/WinRT.Runtime2/InteropServices/WindowsRuntimeComWrappers.cs
Sergio0694: Yeah it's still accurate. We're not actually passing a user state object, but we still pass that parameter so that the new overload in .NET 10 gets called, instead of the old one. We need that because the new overload is the only one with the `CreatedWrapperFlags`, which we need. In the future we may or may not decide to actually leverage the user state object too to move away from thread statics, but we already need the flags either way 👍

src/WinRT.Runtime2/InteropServices/ObjectReference/WindowsRuntimeObjectReference.Initialization.cs
Sergio0694: Let's postpone this. I need to go back and check again that it 1:1 matches the original implementation from 2.x. I might make a separate PR just for that, as I could use another pair of eyes (yours) on this, given how tricky it is ahah

src/WinRT.Runtime2/InteropServices/Marshalling/WindowsRuntimeValueTypeMarshaller.cs
Sergio0694: Fixed, thank you! 🙂

src/WinRT.Interop.Generator/Resolvers/PathAssemblyResolver.cs
Windows10CE: Yeah it's fine, the only thing it would affect since you're supplying an assembly resolver is what it thinks the "runtime" corlib is, which is only used for resolving types in custom attributes with no scope, which shouldn't ever be the case in the assemblies CsWinRT looks at anyway.

src/WinRT.Interop.Generator/Extensions/TypeDefinitionExtensions.cs
manodasanW: In addition to base interfaces you mention, does this also include the covariant interfaces?

manodasanW (approved):

PR #2035, opened 2025-08-13
Fix 'Initialize' typo in several files

Lightczx:
'Initialize' wording in several files (including cswinrt generated projections) are misspelled as 'Initalize'


Lightczx:
@microsoft-github-policy-service agree

Sergio0694 (approved):

PR #2038, opened 2025-08-21
Initial prototype for user-defined types handling

Sergio0694:
- Initial prototype for emitting code for user-defined types
  - Note: very approximated discovery logic, also not deterministic yet
- Fix handling of nested types
- Fix handling of some generic delegate types
- Minor bug fixes
- Updated documentation

### DIFF
--- a/src/WinRT.Interop.Generator/Extensions/WindowsRuntimeExtensions.cs
+++ b/src/WinRT.Interop.Generator/Extensions/WindowsRuntimeExtensions.cs
@@ -99,8 +99,9 @@ internal static class WindowsRuntimeExtensions
                     return false;
                 }
 
-                // Only user-defined class types and struct types (not enums) can be exposed types
-                return type is { IsClass: true } or { IsValueType: true, IsEnum: false };
+                // Only user-defined class types and struct types (not enums) can be exposed types.
+                // We ignore delegates too, as generic delegates are already processed separately.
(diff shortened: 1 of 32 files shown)


src/WinRT.Interop.Generator/Extensions/WindowsRuntimeExtensions.cs
```diff
@@ -48,6 +81,40 @@ public bool IsProjectedWindowsRuntimeClassType
...
 
+        /// <summary>
+        /// Checks whether a <see cref="TypeDefinition"/> represents a type that can be constructed (i.e. instantiated).
+        /// </summary>
+        public bool IsConstructibleType => type is { IsInterface: false, IsAbstract: false };
+
+        /// <summary>
+        /// Checks whether a <see cref="TypeDefinition"/> represents a type that can possibly be exposed to Windows Runtime.
+        /// </summary>
+        public bool IsPossiblyWindowsRuntimeExposedType
```
manodasanW, line 92: How is this used?

Sergio0694: It's used when discovering user-defined types that should be exposed, to quickly filter out types that can't possibly be exposed (e.g. static types, ref struct types, etc.). It's just to avoid the more expensive work in those scenarios.

PR #2049, opened 2025-09-08
Update write_struct and write_abi_struct in code_writers.h

kythant:
Updating write_struct and write_abi_struct in code_writers.h. 

guid_generator.h was taken from
https://github.com/microsoft/cppwinrt/blob/master/natvis/type_resolver.cpp#L269
helpers from
https://github.com/microsoft/cppwinrt/blob/master/strings/base_identity.h


src/cswinrt/strings/InspectableVftbl.cs
```diff
@@ -0,0 +1,20 @@
...
+
+using System;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using Windows.Foundation;
+
+namespace WindowsRuntime.InteropServices;
+
+[StructLayout(LayoutKind.Sequential)]
+internal unsafe struct IInspectableVftbl
```
Sergio0694, line 12: Wondering, can we just link this file from the one in the runtime, instead of copying it?

src/cswinrt/strings/InitalizeProjection.cs
```diff
@@ -28,7 +28,7 @@ internal static class ProjectionInitializer
         [ModuleInitializer]
         public static void InitalizeProjection()
```
Sergio0694, line 29: I assume we'll delete this file entirely in a follow up PR?

src/cswinrt/code_writers.h
```diff
@@ -3549,6 +3550,473 @@ private % AsInternal(InterfaceTag<%> _) => % ?? Make_%();
...
+    public static int get_Value(void* thisPtr, %* result)
+    {
+        if (result is null)
+        {
+            return E_POINTER;
+        }
+
+        try
+        {
+            *result = (%)(ComInterfaceDispatch.GetInstance<object>((ComInterfaceDispatch*)thisPtr));
```
Sergio0694: Note, this is only valid if the type is blittable, otherwise you'll need to marshal it here.

src/cswinrt/code_writers.h
```diff
@@ -3549,6 +3550,473 @@ private % AsInternal(InterfaceTag<%> _) => % ?? Make_%();
...
+    }
+
+    void write_reference_vftbl_impl(writer& w, TypeDef const& type)
+    {
+        auto name = w.write_temp("%", bind<write_type_name>(type, typedef_name_type::ABI, false));
+        auto projection_name = w.write_temp("%", bind<write_projection_type>(type));
+        auto abi_name = w.write_temp("%", bind<write_abi_type>(type));
+        w.write(
+R"([StructLayout(LayoutKind.Sequential)]
+file unsafe struct %ReferenceVftbl
```
Sergio0694: You can share a single vtable type for all of these, if you just use `void*`. Just to minimize code size.

src/cswinrt/code_writers.h
```diff
@@ -3549,6 +3550,473 @@ private % AsInternal(InterfaceTag<%> _) => % ?? Make_%();
...
+)", name, name, name, name, name, name, name, name);
+    }
+
+    void write_com_interface_entries(writer& w, TypeDef const& type)
+    {
+        auto name = w.write_temp("%", bind<write_typedef_name>(type, typedef_name_type::ABI, false));
+
+        w.write(
+R"([StructLayout(LayoutKind.Sequential)]
+file struct %InterfaceEntries
```
Sergio0694: I think you should be able to use a single copy of this type as well.

src/cswinrt/code_writers.h
kythant: Fixed

src/cswinrt/strings/InspectableVftbl.cs
kythant: If I link it from the runtime, I can't access it because it is internal

Sergio0694: I'm confused... But this type is also defined as internal..? 😅

kythant: Actually, the real issue was that it did not appear in the Generated Files folder once I have gotten it to link. And when I did make it appear in the Generated Files by updating Strings.props, we are going to require some changes to the HSTRING declarations in the file and what not because we don't define it that in the cswinrt project. 

I will not block on this and Mano and I will tackle it in another PR.

src/cswinrt/code_writers.h
kythant: Done

src/cswinrt/code_writers.h
kythant: Still figuring this one out

kythant: Done

src/cswinrt/strings/InterfaceEntries.cs
```diff
@@ -0,0 +1,23 @@
...
+using System;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using Windows.Foundation;
+using static System.Runtime.InteropServices.ComWrappers;
+
+namespace WindowsRuntime.InteropServices;
+
+[StructLayout(LayoutKind.Sequential)]
+internal struct InterfaceEntries
```
Sergio0694: Should this be named something like `ReferenceInterfaceEntries` or something? That is, this set of interface entries is only valid for some `IReference<T>` value, not for any type.

src/cswinrt/code_writers.h
```diff
@@ -3549,6 +3550,452 @@ private % AsInternal(InterfaceTag<%> _) => % ?? Make_%();
...
+                                }
+                                else
+                                {
+                                    w.write("    value.%", fieldName);
+                                }
+                            });
+                    }, ",\n", type.FieldList()));
+        }
+
+        w.write(
```
Sergio0694, line 3902: Don't we also need the `Dispose` methods for non blittable struct types that contain managed fields?

@manodasanW we said those would go here in each marshaller class, right? 🤔

src/cswinrt/code_writers.h
```diff
@@ -3549,6 +3550,452 @@ private % AsInternal(InterfaceTag<%> _) => % ?? Make_%();
...
+        }
+        catch (Exception e)
+        {
+            return RestrictedErrorInfoExceptionMarshaller.ConvertToUnmanaged(e);
+        }
+    }
+
+    %
+}
+)", bind<write_guid_property_from_signature>(type, ireference_guid_sig));
```
Sergio0694: The IID properties in all generated `Impl` types should just be named `IID`, not `IID_...`.

src/cswinrt/code_writers.h
```diff
@@ -3549,6 +3550,452 @@ private % AsInternal(InterfaceTag<%> _) => % ?? Make_%();
...
+    {
+        auto name = w.write_temp("%", bind<write_type_name>(type, typedef_name_type::ABI, false));
+        auto projection_name = w.write_temp("%", bind<write_projection_type>(type));
+        auto abi_name = w.write_temp("%", bind<write_abi_type>(type));
+        w.write(
+R"(internal sealed unsafe class %ComWrappersMarshallerAttribute : WindowsRuntimeComWrappersMarshallerAttribute
+{
+    public override void* GetOrCreateComInterfaceForObject(object value)
+    {
+        return WindowsRuntimeValueTypeMarshaller.BoxToUnmanaged<%>((%) value, in %ReferenceImpl.IID_IReferenceOf%).DetachThisPtrUnsafe();
```
Sergio0694: To reduce cross dependencies across generated types and help with trimming, we should split the IID generation into two places (this is the same thing that cswinrtgen also does). That is:
- Have some `InterfaceIIDs` type containing RVA fields for all IIDs across the projection
- Forward to that in the `IID` property in the `Impl` type
- Use `Impl.IID` when assigning the IID in the interface entry type slot
- Use it directly everywhere else (instead of using the `Impl` type)

That is, something like this:
```csharp
internal static class InterfaceIIDs
{
    public static ref readonly Guid IID_IFoo => ...;
    public static ref readonly Guid IID_IBar => ...;
    public static ref readonly Guid IID_IReferenceOfFoo => ...;
    // etc.
}

file static unsafe class FooReferenceImpl
{
    // ...
    public static ref readonly Guid IID
    {
        [MethodImpl(MethodImplOptions.AggressiveInlining)]
        get => ref InterfaceIIDs.IID_IReferenceOfFoo;
    }
}
```

Sergio0694 (changes requested):

src/cswinrt/strings/InterfaceEntries.cs
kythant: Good point!

src/cswinrt/code_writers.h
kythant: Ah I see that makes it simpler

src/cswinrt/code_writers.h
kythant: I think this will require substantial changes in main.cpp so I will do this in another PR

src/cswinrt/code_writers.h
kythant: Ah WindowsRuntimeValueTypeMarshaller.BoxToUnmanaged won't handle it?
We would need to handle it here instead?

kythant: Fixed

src/cswinrt/type_writers.h
```diff
@@ -53,8 +53,11 @@ using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using WinRT;
-using WinRT.Interop;
-
+using Windows.Foundation;
```
manodasanW: What did this simplify?  (Windows.Foundation specifically)

src/cswinrt/guid_generator.h
```diff
@@ -0,0 +1,320 @@
+#pragma once
```
manodasanW, line 7: Copyright at top?

manodasanW: and link to where it is from if copy

src/cswinrt/type_writers.h
kythant: Seems like we didn't need this. Removing.

src/cswinrt/guid_generator.h
kythant: Added

src/cswinrt/code_writers.h
```diff
@@ -3549,6 +3550,551 @@ private % AsInternal(InterfaceTag<%> _) => % ?? Make_%();
...
+                                w.write("    % = WindowsRuntimeDelegateMarshaller.ConvertToUnmanaged(value.%)", fieldName, fieldName);
+                                break;
+                            case category::enum_type:
+                                w.write("    % = value.%", fieldName, fieldName);
+                                // TODO: consider arrays
+                                break;
+                            case category::struct_type:
+                                if (!is_type_blittable(td))
+                                {
+                                    w.write("    % = %Marshaller.ConvertToUnmanaged(value.%)", fieldName, td.TypeName(), fieldName);
```
manodasanW: is this going to work for different namespaces, might want to use bind<write_type_name>(...)

src/cswinrt/code_writers.h
```diff
@@ -3549,6 +3550,551 @@ private % AsInternal(InterfaceTag<%> _) => % ?? Make_%();
...
+                                w.write("    WindowsRuntimeDelegateMarshaller.ConvertToManaged(value.%)", fieldName);
+                                break;
+                            case category::enum_type:
+                                w.write("    value.%", fieldName);
+                                // TODO: array case
+                                break;
+                            case category::struct_type:
+                                if (!is_type_blittable(td))
+                                {
+                                    w.write("    %Marshaller.ConvertToManaged(value.%)", td.TypeName(), fieldName);
```
manodasanW: same here

src/cswinrt/code_writers.h
```diff
@@ -3549,6 +3550,551 @@ private % AsInternal(InterfaceTag<%> _) => % ?? Make_%();
...
+                            if (type == fundamental_type::String)
+                            {
+                                w.write("HStringMarshaller.Free((void*)value.%);\n", fieldName);
+                            }
+                        });
+                }, "", type.FieldList()));
+
+    }
+
+    void write_marshaller_class(writer& w, TypeDef const& type)
```
manodasanW: write_struct_marshaller_class

src/cswinrt/code_writers.h
```diff
@@ -3549,6 +3550,551 @@ private % AsInternal(InterfaceTag<%> _) => % ?? Make_%();
...
+    {
+        auto name = w.write_temp("%", bind<write_type_name>(type, typedef_name_type::ABI, false));
```
manodasanW: nit: we should probably make this abi_name and the other abi_type.  The main reason for that is we typically consider name to be the projected name.

src/cswinrt/code_writers.h
```diff
@@ -3549,6 +3550,551 @@ private % AsInternal(InterfaceTag<%> _) => % ?? Make_%();
...
+        auto projection_name = w.write_temp("%", bind<write_projection_type>(type));
+        auto abi_name = w.write_temp("%", bind<write_abi_type>(type));
+
+        w.write(
+R"([global::System.ComponentModel.EditorBrowsable(global::System.ComponentModel.EditorBrowsableState.Never)]
+public static unsafe class %Marshaller
+{
+)", name);
```
manodasanW: The way we are using this, it should just be type.TypeName()

src/cswinrt/code_writers.h
```diff
@@ -3549,6 +3550,551 @@ private % AsInternal(InterfaceTag<%> _) => % ?? Make_%();
...
+    [UnmanagedCallersOnly(CallConvs = [typeof(CallConvMemberFunction)])]
+    public static int get_Value(void* thisPtr, void* result)
+    {
+        if (result is null)
+        {
+            return E_POINTER;
+        }
+
+        try
+        {)", name, name);
```
manodasanW: Same thing here, I would typically just use TypeName directly rather than going through write_type_name.

src/cswinrt/code_writers.h
```diff
@@ -3549,6 +3550,551 @@ private % AsInternal(InterfaceTag<%> _) => % ?? Make_%();
...
+        
+        if (is_type_blittable(type))
+        {
+            w.write(
+R"(            
+            var value = (%)(ComInterfaceDispatch.GetInstance<object>((ComInterfaceDispatch*)thisPtr));
+            *(%*)result = value; // Explicit cast back to the correct type)"
```
manodasanW: I would not put the comment in generated code, but outside in codewriters if desired.

src/cswinrt/code_writers.h
```diff
@@ -3549,6 +3550,551 @@ private % AsInternal(InterfaceTag<%> _) => % ?? Make_%();
...
+            return RestrictedErrorInfoExceptionMarshaller.ConvertToUnmanaged(e);
+        }
+    }
+
+    %
+}
+)", bind<write_guid_property_from_signature>(type, ireference_guid_sig));
+    }
+
+    void write_com_wrappers_marshaller_attribute_impl(writer& w, TypeDef const& type)
```
manodasanW: write_struct_com_wrappers_marshaller_attribute_impl

manodasanW: We may later generalize this code to work for all scenarios based on how different it is

src/cswinrt/code_writers.h
```diff
@@ -3549,6 +3550,551 @@ private % AsInternal(InterfaceTag<%> _) => % ?? Make_%();
...
+
+    }
+
+    void write_winrt_metadata_attribute(writer& w, TypeDef const& type)
+    {
+        std::filesystem::path db_path(type.get_database().path());
+        w.write("[WindowsRuntimeMetadata(\"%\")]\n", db_path.stem().string());
+    }
+
+    void write_winrt_classname_attribute(writer& w, TypeDef const& type)
```
manodasanW: write_struct_winrt_classname_attribute

src/cswinrt/code_writers.h
```diff
@@ -3549,6 +3550,551 @@ private % AsInternal(InterfaceTag<%> _) => % ?? Make_%();
...
+    {
+        w.write("[WindowsRuntimeClassName(\"Windows.Foundation.IReference<%.%>\")]\n",
+            type.TypeNamespace(), type.TypeName());
+    }
+
+    void write_comwrapper_marshaller_attribute(writer& w, TypeDef const& type)
+    {
+        auto ns = type.TypeNamespace();
```
manodasanW: nit: unused?

src/cswinrt/code_writers.h
```diff
@@ -3549,6 +3550,551 @@ private % AsInternal(InterfaceTag<%> _) => % ?? Make_%();
...
+    target: typeof(%),
+    trimTarget: typeof(%))]
+)", ns, name, projection_name, projection_name);
+
+    }
+
+    void write_winrt_metadata_attribute(writer& w, TypeDef const& type)
+    {
+        std::filesystem::path db_path(type.get_database().path());
+        w.write("[WindowsRuntimeMetadata(\"%\")]\n", db_path.stem().string());
```
manodasanW, line 4068: If you do R(.., you won't need to worry about escaping.

src/cswinrt/code_writers.h
kythant: I wondering if there is something for just "name"

Because it's just being used to determine that comes infront for the marshaller class name.

kythant: Oh wait, I could just use type.TypeName()

src/cswinrt/code_writers.h
kythant: The enums are using these the same way in my enum branch but I will change it to write_struct_enum_com_wrappers_marshaller_attribute_impl later accordingly

src/cswinrt/code_writers.h
kythant: If I use literal here, then I would have to deal with the \n by adding an actual new line and that looks weird.

src/cswinrt/code_writers.h
manodasanW: write_value_type_com...?

src/cswinrt/code_writers.h
kythant: Fixed using  bind<write_type_name>(td, typedef_name_type::ABI, false)

src/cswinrt/code_writers.h
kythant: Fixed using  bind<write_type_name>(td, typedef_name_type::ABI, false)

src/cswinrt/code_writers.h
```diff
@@ -4060,20 +4060,6 @@ R"(file static class %InterfaceEntriesImpl
 )", name, name, name, name);
     }
 
-    void write_winrt_typemapgroup_assembly_attribute(writer& w, TypeDef const& type)
```
manodasanW: You do need to call this somewhere right to emit typemap or are you doing it elsewhere?

kythant: ahh wait yea these are required for non-blittable types?

Sergio0694: The type map attribute is needed for any struct type (doesn't matter if blittable or not), because it's what we use to marshal them as `IReference<T>` values when we get such an object as an opaque `IInspectable*` value.

kythant: Leaving this out for this PR because it will require code changes to split the write_begin() in text_writer.h so we can fit in the assembly attributes in between from the caller in main.cpp

manodasanW (approved):

PR #2060, opened 2025-09-25
Implementing WindowsRuntimeComWrappersTypeMapGroup CodeGen

kythant:
Introducing write_winrt_typemapgroup_assembly_attribute

In type_writers.h, I've split up write_begin into two separate calls, write_begin and write_begin_projected to allow for code injection in between.


src/cswinrt/code_writers.h
```diff
@@ -4062,6 +4062,22 @@ R"(file static class %InterfaceEntriesImpl
 )", name, name, name, name);
     }
 
+    void write_winrt_typemapgroup_assembly_attribute(writer& w, TypeDef const& type)
+    {
+        auto projection_name = w.write_temp("%", bind<write_type_name>(type, typedef_name_type::Projected, true));
+        auto abi_name = w.write_temp("%", bind<write_type_name>(type, typedef_name_type::ABI, true));
+        auto target = is_type_blittable(type) ? projection_name : abi_name;
```
manodasanW: target doesn't change right, its always projection name if not custom type mapping

Sergio0694: Yeah the trim target can't be the ABI type, as you'd never box it (you'd never see that type at all).

kythant: Hmm so in the datetimeoffset example, target was using the ABI name
https://github.com/microsoft/CsWinRT/blob/staging/3.0/src/WinRT.Runtime2/ABI/System/DateTimeOffset.cs

kythant: But let me know if that's not necessary and we can just use the Projection Name

Sergio0694: Sorry I thought you meant the trim target. But even then, remember we said `DateTime` had to do that just because it's a custom mapped type, so we can't have the marshaller attribute on the projected type directly. This is not the case for projected types, you can always annotate them directly. So the target of the type map can always be that type.

kythant: Thanks for the explanation! Fixed.

manodasanW (approved):

Sergio0694 (approved):

PR #2061, opened 2025-09-26
Update write_enum and write_abi_enum in code_writers.h

kythant:


src/cswinrt/code_writers.h
```diff
@@ -10256,9 +10253,18 @@ return true;
                 }
             }
         }
-        w.write("}\n");
+        w.write("}\n\n");
     }
 
+    void write_abi_enum(writer& w, TypeDef const& type)
+    {
+        w.write("%\n%\n%\n%\n",
```
manodasanW: Instead of bind, I would probably do `write_struct_and_enum_marshaller_class(w, type);` and similar for the others.  This is due to they are all different classes and also should implicitly have new line too.

src/cswinrt/code_writers.h
```diff
@@ -10230,19 +10230,16 @@ return true;
...
-        {
-            w.write("[FlagsAttribute]\n");
-        }
-
         auto enum_underlying_type = is_flags_enum(type) ? "uint" : "int";
 
-        w.write(R"(%%%% enum % : %
+        w.write(
+R"(
+%[WindowsRuntimeMetadata("Windows.Foundation.FoundationContract")]
```
manodasanW: Not always foundation contract, will need to call that metadata function to write it.

src/cswinrt/code_writers.h
```diff
@@ -10230,19 +10230,16 @@ return true;
...
+[global::System.ComponentModel.EditorBrowsable(global::System.ComponentModel.EditorBrowsableState.Never)]
+% enum % : %
 {
-)", 
-        bind<write_winrt_attribute>(type),
-        bind<write_winrt_exposed_type_attribute>(type, false),
-        bind<write_type_custom_attributes>(type, true),
+)",             
+        is_flags_enum(type) ? "[FlagsAttribute]\n" : "",
         (settings.internal || settings.embedded) ? (settings.public_enums ? "public" : "internal") : "public",
```
manodasanW: I would remove embedded check while making changes here.

manodasanW: we might want to keep write_type_custom_attributes.  Not sure if it wrote anything relevant for enums or not, but its basically any attribute someone decided to put on the enum in the winmd.

src/cswinrt/code_writers.h
kythant: That's fair. I can do the same for write_abi_struct

PR #2064, opened 2025-09-30
Update projection of classes and interfaces

manodasanW:
Updated codewriters to generate classes and interfaces in new format.
- Currently don't support classes implementing generic interfaces.
- Currently there is a todo to fix the method name for the cswinrtgen static methods.
- There is also a todo to move where IIDs are generated

Attached generated cs files (renamed to txt to attach)

[TestComponent.txt](https://github.com/user-attachments/files/22625437/TestComponent.txt)
[TestComponentCSharp.txt](https://github.com/user-attachments/files/22625438/TestComponentCSharp.txt)

### DIFF
--- a/src/cswinrt/code_writers.h
+++ b/src/cswinrt/code_writers.h
@@ -1059,7 +1059,7 @@ namespace cswinrt
 
     void write_abi_get_property_static_method_call(writer& w, type_semantics const& iface, Property const& prop, std::string const& targetObjRef)
     {
-        w.write("%.get_%(%)",
+        w.write("%.%(%)",
             bind<write_type_name>(iface, typedef_name_type::StaticAbiClass, true),
             prop.Name(),
             targetObjRef);
@@ -1067,7 +1067,7 @@ namespace cswinrt
 
     void write_abi_set_property_static_method_call(writer& w, type_semantics const& iface, Property const& prop, std::string const& targetObjRef)
     {
-        w.write("%.set_%(%, value)",
+        w.write("%.%(%, value)",
             bind<write_type_name>(iface, typedef_name_type::StaticAbiClass, true),
             prop.Name(),
             targetObjRef);
@@ -5339,7 +5339,7 @@ public static unsafe % %(WindowsRuntimeObjectReference thisReference%%)
                 auto marshalers = get_abi_marshalers(w, signature, is_generic, prop.Name());
                 w.write(R"(
 [MethodImpl(MethodImplOptions.NoInlining)]
-public static unsafe % get_%(WindowsRuntimeObjectReference thisReference)
+public static unsafe % %(WindowsRuntimeObjectReference thisReference)
 {%}
 )",
                     write_prop_type(w, prop),
@@ -5358,7 +5358,7 @@ public static unsafe % get_%(WindowsRuntimeObjectReference thisReference)
 
                 w.write(R"(
 [MethodImpl(MethodImplOptions.NoInlining)]
-public static unsafe void set_%(WindowsRuntimeObjectReference thisReference, % value)
+public static unsafe void %(WindowsRuntimeObjectReference thisReference, % value)
 {%}
 )",                 
                     prop.Name(),


src/WinRT.Runtime2/ABI/System/Type.cs
```diff
@@ -144,6 +144,15 @@ public static WindowsRuntimeObjectReferenceValue BoxToUnmanaged(global::System.T
 
         return abi.HasValue ? ConvertToManaged(abi.GetValueOrDefault()) : null;
     }
+
+    /// <summary>
+    /// Frees resources associated with an unmanaged <see cref="Type"/> value.
+    /// </summary>
+    /// <param name="value">The unmanaged <see cref="Type"/> value to free.</param>
+    public static void Free(Type value)
```
Sergio0694: Didn't we call these `Dispose`, not `Free`? 🤔

src/WinRT.Runtime2/Windows.Foundation/FoundationContract.cs
```diff
@@ -1,7 +1,9 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-namespace Windows.Foundation.Metadata;
+using Windows.Foundation.Metadata;
+
+namespace Windows.Foundation;
```
Sergio0694, line 6: Changes not needed?

src/WinRT.Runtime2/Windows.UI.Xaml.Interop/TypeKind.cs
```diff
@@ -3,6 +3,7 @@
 
 using System.ComponentModel;
 using System.Runtime.Versioning;
+using Windows.Foundation;
```
Sergio0694, line 6: Also not needed?

src/cswinrt/code_writers.h
```diff
@@ -2276,160 +2159,147 @@ ComWrappersHelper.Init(_inner, false);
...
+private WindowsRuntimeObjectReference %
 {
+    get
+    {
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        WindowsRuntimeObjectReference MakeObjectReference()
+        {
+            _ = global::System.Threading.Interlocked.CompareExchange(
+                location1: ref field,
+                value: NativeObjectReference.As(%Impl.IID),
```
Sergio0694, line 2171: Leave a TODO here too just to remember to move the IID later?

src/cswinrt/code_writers.h
```diff
@@ -2637,49 +2468,36 @@ IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
...
 
         w.write(R"(
-%bool %MoveNext() => %.MoveNext();
-%void %Reset() => %.Reset();
-%void %Dispose() => %.Dispose();
-%% %Current => %.Current;
+bool global::System.Collections.Generic.IEnumerator<%>.MoveNext() => ((global::System.Collections.Generic.IEnumerator<%>)(WindowsRuntimeObject)this).MoveNext();
+void global::System.Collections.Generic.IEnumerator<%>.Reset() => ((global::System.Collections.Generic.IEnumerator<%>)(WindowsRuntimeObject)this).Reset();
+void global::System.Collections.Generic.IEnumerator<%>.Dispose() => ((global::System.Collections.Generic.IEnumerator<%>)(WindowsRuntimeObject)this).Dispose();
+% global::System.Collections.Generic.IEnumerator<%>.Current => ((global::System.Collections.Generic.IEnumerator<%>)(WindowsRuntimeObject)this).Current;
```
Sergio0694, line 2494: Do we _have_ to go through interface stub dispatch here? Why can't these explicit implementations just call the `MoveNext()` etc. methods on the current type directly?

src/cswinrt/code_writers.h
```diff
@@ -4065,18 +3810,54 @@ R"(file static class %InterfaceEntriesImpl
...
+        if (is_exclusive_to(type) && !settings.idic_exclusiveto)
+        {
+            return;
+        }
+
+        w.write(
+            R"(
+[assembly: TypeMapAssociation<DynamicInterfaceCastableImplementationTypeMapGroup>(
+    typeof(%),
+    typeof(%))]
```
Sergio0694: Nit: name the params just for clarity
```suggestion
    source: typeof(%),
    proxy: typeof(%))]
```

src/cswinrt/code_writers.h
```diff
@@ -4580,44 +4329,79 @@ event % %;)",
                 w.write_temp("__params[%]", param_index);
         }
 
+        void write_convert_to_managed_function(writer& w) const
+        {
+            if (interop_dll_type != "")
+            {
+                w.write(R"(
+[UnsafeAccessor(UnsafeAccessorKind.StaticMethod, Name="ConvertToManaged")]
```
Sergio0694: Nit: formatting (same below)
```suggestion
[UnsafeAccessor(UnsafeAccessorKind.StaticMethod, Name = "ConvertToManaged")]
```

src/cswinrt/code_writers.h
```diff
@@ -4884,31 +4655,52 @@ event % %;)",
...
 
             if (is_out())
             {
-                w.write("%.DisposeAbi%(%);\n",
-                    marshaler_type,
-                    is_array() ? "Array" : "",
-                    get_param_local(w));
+                if (is_array())
+                {
+                    w.write("%.DisposeAbi%(%);\n",
```
Sergio0694, line 4674: Is this a typo? We don't have any methods named "DisposeAbi" in 3.0, right? 🤔

Sergio0694 (commented):
Some early comments, only skimmed the first half or so 🙂

Sergio0694 (commented):
Nit: use a collection expression for the IID bytes:

image

Sergio0694 (approved):

src/WinRT.Runtime2/ABI/System/Type.cs
manodasanW: Right, for structs, it is dispose.  And I assume that is what we want to use here.  I was initially matching this with strings and the pinnable naming conventions.

src/WinRT.Runtime2/Windows.Foundation/FoundationContract.cs
manodasanW: This is moving it to the right namespace it is in based on winmd.

src/WinRT.Runtime2/Windows.UI.Xaml.Interop/TypeKind.cs
manodasanW: This was a result of the above change.

src/WinRT.Runtime2/ABI/System/Type.cs
Sergio0694: Right, yeah I think we're doing `Free` for objects and `Dispose` for structs 👍

src/WinRT.Runtime2/Windows.Foundation/FoundationContract.cs
Sergio0694: Oh sorry, misread. Good catch, thank you! 🙂
Can you also move this to the right folder while at it?

src/cswinrt/code_writers.h
manodasanW: arrays are still a TODO.  Will be addressed as part of array support.

src/cswinrt/code_writers.h
manodasanW: Its already on my list to bulk move all IIDs, there is a couple too many places IIDs are used that I avoided doing todo in each place.

src/cswinrt/code_writers.h
manodasanW: This is used in the IDIC implementation.  The cast is probably to get the right vtable, but if those same functions are implemented, we might be able to use them without the cast.  Will revisit this when I do get classes implementing generic interfaces working.

Sergio0694: Sounds good, yeah let's revisit this later. I'm still convinced we might actually be able to delete all of these methods.

PR #2069, opened 2025-10-07
Update write_delegates and write_abi_delegates in code_writers.h; EventSource Implementation;

kythant:


src/cswinrt/code_writers.h
```diff
@@ -3724,12 +3725,24 @@ R"(
         }
         else
         {
-            w.write(
+            if (get_category(type) == category::delegate_type)
+            {
+                w.write(
+R"(            
+            % unboxedValue = (%)ComInterfaceDispatch.GetInstance<object>((ComInterfaceDispatch*)thisPtr);  
+            var value = %Marshaller.ConvertToUnmanaged(unboxedValue).DetachThisPtrUnsafe();
```
Sergio0694: Nit: better to be explicit
```suggestion
            void* value = %Marshaller.ConvertToUnmanaged(unboxedValue).DetachThisPtrUnsafe();
```

src/cswinrt/code_writers.h
```diff
@@ -7640,26 +7653,203 @@ return false;
...
+        auto method = get_delegate_invoke(type);
+        method_signature signature{ method };
+
+        w.write(R"(
+internal static unsafe class %Impl
+{
+    [FixedAddressValueType]
+    private static readonly %Vftbl Vftbl;
+
+    private const int S_OK = unchecked((int)0x00000000);
```
Sergio0694: Nit: constants should always be above all fields (applies to all other "Impl" types too)

src/cswinrt/code_writers.h
```diff
@@ -7640,26 +7653,203 @@ return false;
...
+    }
+
+
+    void write_delegate_com_wrappers_marshaller_attribute_impl(writer& w, TypeDef const& type)
+    {
+        auto name = type.TypeName();
+        w.write(
+            R"(internal sealed unsafe class %ComWrappersMarshallerAttribute : WindowsRuntimeComWrappersMarshallerAttribute
+{
+    public override void* GetOrCreateComInterfaceForObject(object value)
```
Sergio0694, line 7766: Nit: while at it, add `/// <inheritdoc/>` above both these methods too?

src/cswinrt/code_writers.h
```diff
@@ -7640,26 +7653,203 @@ return false;
...
+    {
+        return WindowsRuntimeMarshal.GetOrCreateComInterfaceForObject(value, CreateComInterfaceFlags.TrackerSupport);
+    }
+
+    public override ComInterfaceEntry* ComputeVtables(out int count)
+    {
+        count = sizeof(ReferenceInterfaceEntries) / sizeof(ComInterfaceEntry);
+        return (ComInterfaceEntry*)Unsafe.AsPointer(in %InterfaceEntriesImpl.Entries);
```
Sergio0694, line 7776: Super nit:
```suggestion
        count = sizeof(ReferenceInterfaceEntries) / sizeof(ComInterfaceEntry);

        return (ComInterfaceEntry*)Unsafe.AsPointer(in %InterfaceEntriesImpl.Entries);
```

src/cswinrt/code_writers.h
```diff
@@ -7640,26 +7653,203 @@ return false;
...
+        auto projected_type = w.write_temp("%", bind<write_type_name>(type, typedef_name_type::Projected, false));
+        w.write(R"(
+[global::System.ComponentModel.EditorBrowsable(global::System.ComponentModel.EditorBrowsableState.Never)]
+public static unsafe class %Marshaller
+{
+    public static WindowsRuntimeObjectReferenceValue ConvertToUnmanaged(% value)
+    {
+        return WindowsRuntimeDelegateMarshaller.ConvertToUnmanaged(value, in %ReferenceImpl.IID);
+    }
+    #nullable enable
```
Sergio0694, line 7799: Why only here? Technically speaking `ConvertToUnmanaged` also can receive a `null` value. But also shouldn't we enable nullability annotations in the whole file? We can also do that as a follow up. We could also skip it given this is all private implementation detail code anyway. @manodasanw thoughts?

src/cswinrt/code_writers.h
```diff
@@ -7640,26 +7653,203 @@ return false;
...
+            [](auto&& pair) { return !pair.second.empty(); }) != generic_abi_types.end();
+
+        w.write(R"(
+public static unsafe class %NativeDelegate
+{
+    public static % Invoke(this WindowsRuntimeObjectReference objectReference%%)
+    {
+        using WindowsRuntimeObjectReferenceValue ThisValue = objectReference.AsValue();
+        void* ThisPtr = ThisValue.GetThisPtrUnsafe();
+        var abiInvoke = ((%Vftbl*)ThisPtr)->Invoke;
```
Sergio0694: Nit: for consistency, name the value `objectValue` (we do the same in the runtime). Also use lowercase for locals, and leave a blank line. Lastly, imho it's cleaner to invoke in one line, also for consistency. We do the same everywhere else.
```suggestion
        using WindowsRuntimeObjectReferenceValue objectValue = objectReference.AsValue();

        void* thisPtr = ThisValue.GetThisPtrUnsafe();

        int hresult = ((%Vftbl*)ThisPtr)->Invoke(...)
```

Sergio0694 (commented):
Nice! Also, can you attach a full .cs file with the generated projections too? 🙂

src/cswinrt/main.cpp
```diff
@@ -332,7 +332,6 @@ Where <spec> is one or more of:
                                     break;
                                 case category::delegate_type:
                                     write_abi_delegate(w, type);
-                                    write_winrt_exposed_type_class(w, type, false);
```
kythant, line 335: Had to remove this to make it compile because write_winrt_exposed_type_class is on the old implementation

src/Projections/Test/Test.csproj
```diff
@@ -49,7 +49,46 @@
...
+      TestComponent.Param5Handler;
+      TestComponent.Param6Handler;
+      TestComponent.Param7Handler;
+      TestComponent.Param8Handler;
+      TestComponent.Param9Handler;
+      TestComponent.Param10Handler;
+      TestComponent.Param11Handler;
+      TestComponent.Param12Handler;
+      TestComponent.Param13Handler;
+      <!--TestComponent.Param14Handler;
```
kythant: The commented-out ones are the ones that doesn't work yet

src/cswinrt/code_writers.h
kythant: Oh what's the reasoning for the extra line? Just style?

src/cswinrt/code_writers.h
kythant: Agree with the lower case for local but write_abi_method_call_marshalers is using uppercase ThisPtr. If I were to change this, I am worried it will revibrate across all the class or interfaces abi imlementations. 

Let's do this fix another time.

src/cswinrt/code_writers.h
kythant: image
I was getting these errors without so thought we needed it here

kythant: Thought doing this to the whole file would be overkill and will hide compile issues that we don't want it to hide. So I made it targeted to this method. The other one didn't have this issue. 

`The annotation for nullable reference types should only be used in code within a '#nullable' annotations context. Auto-generated code requires an explicit '#nullable' directive in source.`

src/cswinrt/code_writers.h
kythant: fixed

src/cswinrt/code_writers.h
manodasanW: Lets hold off on doing it to the entire file and just do scoped for now.  I was looking at if we should do this for the entire generated file, but then that brings the question about what things in the generated projections should be nullable or not.  So, I was going to look at that later after we got everything else working.

src/cswinrt/main.cpp
manodasanW: right, we don't need this anymore.

kythant: Oh nevermind I thought we still needed it

src/cswinrt/code_writers.h
Sergio0694: My idea was to enable in the entire file and disable it over projected APIs, since those are nullable-unaware.

src/cswinrt/code_writers.h
```diff
@@ -560,6 +560,29 @@ namespace cswinrt
             bind<write_parameter_name>(param));
     }
 
+    void write_parmaeters(writer& w, method_signature::param_t const& param)
```
manodasanW, line 563: nit: write_parameters

kythant: Fixing in another PR

PR #2073, opened 2025-10-09
Move all Guid properties to GeneratedInterfaceIIDs.cs

kythant:
To reduce cross dependencies across generated types and help with trimming.

### DIFF
--- a/src/cswinrt/type_writers.h
+++ b/src/cswinrt/type_writers.h
@@ -95,18 +95,17 @@ using System;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 
-namespace ABI
+namespace ABI;
+
+internal static class InterfaceIIDs
 {
-    internal static class InterfaceIIDs
-    {
 )"          
             );
         }
 
         void write_end_interface_iids()
         {
-            write("}\n}\n\n");
-            _in_abi_impl_namespace = false;
+            write("}\n\n");
         }
 
         void write_begin_abi()


src/cswinrt/code_writers.h
```diff
@@ -3318,47 +3318,58 @@ return %.AsValue();
             [&](auto const&) {});
     }
 
-    static void write_guid_property_from_signature(writer& w, std::string const& signature)
+    static void write_iid_guid_property_name(writer& w, TypeDef const& type)
     {
-        GUID guid_value = generate_guid(signature);
+        category category = get_category(type);
+        w.write("IID_%%", type.TypeName(), category == category::enum_type || category == category::struct_type ? "Reference" : "");
```
Sergio0694: Do we need to generate two IIDs for delegates? As in, one for the delegate itself and one for its reference version?

kythant: Would they be the same GUID or two different GUIDs?

Sergio0694: They would be different (different interfaces, one is the delegate, one is the `IReference<TheDelegate>`).

src/cswinrt/type_writers.h
```diff
@@ -80,15 +80,30 @@ namespace %%
...
+//     Changes to this file may cause incorrect behavior and will be lost if
+//     the code is regenerated.
+// </auto-generated>
+//------------------------------------------------------------------------------
+
+using System;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+namespace ABI
```
Sergio0694: Nit: we should consider switching all generated code to file-scoped namespaces, since it's nicer. Now that we're targeting .NET 10 only, we can always rely on that language feature being enabled by default.

src/cswinrt/code_writers.h
```diff
@@ -3926,8 +3926,7 @@ R"(internal sealed unsafe class %ComWrappersMarshallerAttribute : WindowsRuntime
 {
     public override void* GetOrCreateComInterfaceForObject(object value)
     {
-        return WindowsRuntimeValueTypeMarshaller.BoxToUnmanaged<%>((%) value, in %).DetachThisPtrUnsafe();
-    }
+        return WindowsRuntimeComWrappersMarshal.GetOrCreateComInterfaceForObject(value, CreateComInterfaceFlags.TrackerSupport);    }
```
Sergio0694: Nit: closed bracket on next line

src/cswinrt/type_writers.h
kythant: Just to clarify then, if the name of the file was GeneratedInterfaceIIDs.cs, the namespace should be GeneratedInterfaceIIDs?

kythant: Finally got what you meant
https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0160-ide0161

Sergio0694: Oh whoops sorry, completely missed the notification for your reply yesterday 😅
Thank you for tweaking this!

manodasanW (approved):

PR #2075, opened 2025-10-10
Enabling end to end tests and projection / runtime fixes

manodasanW:
- Bringing in part of the .NET SDK targets / task for cswinrtgen until we can consume it here which we will be able to once rc2 is out
- Enabling a functional test to run in the test pass
- Fixing issue discovered during end to end test
- Adding WindowsRuntimeReferenceAssembly and then changing cswinrtgen check to check for WinRT.Runtime instead.  Note we may need to fix this further to be transitive.

### DIFF
--- a/src/Tests/FunctionalTests/ClassActivation/Program.cs
+++ b/src/Tests/FunctionalTests/ClassActivation/Program.cs
@@ -1,9 +1,14 @@
-﻿using System;
+﻿using System.Runtime.InteropServices;
 using TestComponent;
 using TestComponentCSharp;
+using WindowsRuntime.InteropServices;
 
 // TODO: This shouldn't be needed if transitive references are detected correctly.
 [assembly: WindowsRuntime.WindowsRuntimeReferenceAssembly]
+
+[assembly: TypeMapAssemblyTarget<WindowsRuntimeComWrappersTypeMapGroup>("WinRT.Runtime2")]
+[assembly: TypeMapAssemblyTarget<WindowsRuntimeComWrappersTypeMapGroup>("Test")]
+[assembly: TypeMapAssemblyTarget<WindowsRuntimeComWrappersTypeMapGroup>("WinRT.Interop")]
 
 CustomDisposableTest customDisposableTest = new();
 customDisposableTest.Dispose();


src/WinRT.Runtime2/InteropServices/ObjectReference/WindowsRuntimeObjectReference.Initialization.cs
```diff
@@ -202,28 +205,34 @@ internal static WindowsRuntimeObjectReference InitializeFromManagedTypeUnsafe(
...
-            // We can only avoid this in the aggregation scenario, because we balance it out with the initial 'AddRef'
-            // that we can skip. But when not aggregating, that 'AddRef' is on the new instance, so the inner instance
-            // would have an extra reference count by the end of this method. We can fix that here to balance it again.
-            _ = IUnknownVftbl.ReleaseUnsafe(acquiredInnerInstanceUnknown);
-
-            // If we're not in a COM aggregation scenario, we can release the reference tracker object. This is because
-            // it is already considered kept alive, even if we don't hold a reference to it, due to the fact that its
-            // implementation lives in the same object as the target COM object (it will be a native object).
-            _ = IUnknownVftbl.ReleaseUnsafe(referenceTracker);
+            if (referenceTracker != null)
```
Sergio0694, line 210: Just so I understand, how did you make this change? Did you compare with the original code for this in 2.x and was the previous implementation in 3.0 incorrect, and this changes it back to match that I assume?

manodasanW: Yea, this makes it match what we had in 2.0.  2.0 had a try finally which handled this which this doesn't anymore in 3.0.

src/cswinrt/code_writers.h
```diff
@@ -2050,7 +2050,7 @@ private static WindowsRuntimeObjectReference %
         {
             return __%;
         }
-        return field = WindowsRuntimeActivationFactory.GetActivationFactory(%.RuntimeClassName);
+        return field = WindowsRuntimeActivationFactory.GetActivationFactory(%.RuntimeClassName, %Impl.IID);
```
Sergio0694, line 2053: This should take the IID from `InterfaceIIDs`. If it depends on Will's PR, maybe leave a TODO?

manodasanW: FYI @kythant, once my PR is merged I think Will should be able to easily address this in his PR.

kythant: Yup mine will come last

PR #2076, opened 2025-10-10
More 'WinRT.Runtime.dll' tweaks

Sergio0694:
This PR includes a few changes to 'WinRT.Runtime.dll':
- Update `WindowsRuntimeValueTypeMarshaller` to take explicit flags.
- Add `WindowsRuntimeComWrappersMarshal` to avoid double boxing for value types
- Centralized the "private implementation detail" message
- Annotated all actual private implementation detail types, and moved some APIs to separate types for better separation
- Add `WindowsRuntimeMarshal.NativeReferenceEqual`

### DIFF
--- a/src/Tests/FunctionalTests/ClassActivation/Program.cs
+++ b/src/Tests/FunctionalTests/ClassActivation/Program.cs
@@ -3,6 +3,8 @@ using TestComponent;
 using TestComponentCSharp;
 using WindowsRuntime.InteropServices;
 
+#pragma warning disable CSWINRT3001 // TODO: remove this once the attributes below are removed
+
 // TODO: This shouldn't be needed if transitive references are detected correctly.
 [assembly: WindowsRuntime.WindowsRuntimeReferenceAssembly]
 


src/cswinrt/code_writers.h
```diff
@@ -7404,6 +7404,8 @@ public static unsafe class %Marshaller
         bool has_base_class = !std::holds_alternative<object_type>(get_type_semantics(type.Extends()));
         separator s{ w, " || " };
         w.write(R"(
+[Obsolete]
```
manodasanW, line 7407: Is this going to show up as warnings when folks are generating projections.

src/cswinrt/code_writers.h
```diff
@@ -3673,9 +3673,9 @@ public static unsafe class %Marshaller
         w.write(
 R"(public static WindowsRuntimeObjectReferenceValue BoxToUnmanaged(%? value)
 {
-    return WindowsRuntimeValueTypeMarshaller.BoxToUnmanaged(value, in %ReferenceImpl.IID);
+    return WindowsRuntimeValueTypeMarshaller.BoxToUnmanaged(value, CreateComInterfaceFlags.TrackerSupport, in %ReferenceImpl.IID);
```
manodasanW, line 3676: At a minimum. the TODO addressing this should make sure if it is blittable, pass none.  And maybe strings are the only new scenarios that can also pass none?

src/WinRT.Runtime2/ABI/System/ComponentModel/DataErrorsChangedEventArgs.cs
```diff
@@ -32,6 +33,7 @@ namespace ABI.System.ComponentModel;
 /// <summary>
 /// Marshaller for <see cref="global::System.ComponentModel.DataErrorsChangedEventArgs"/>.
 /// </summary>
+[Obsolete(WindowsRuntimeConstants.PrivateImplementationDetailObsoleteMessage)]
```
manodasanW: Generated projections would call this right?  Not sure if we want to just keep EditorBrowsable rather than also obsolete.

src/WinRT.Runtime2/ABI/System/ComponentModel/INotifyDataErrorInfo.cs
```diff
@@ -21,6 +21,7 @@ namespace ABI.System.ComponentModel;
 /// <summary>
 /// Marshaller for <see cref="INotifyDataErrorInfo"/>.
 /// </summary>
+[Obsolete(WindowsRuntimeConstants.PrivateImplementationDetailObsoleteMessage)]
 [EditorBrowsable(EditorBrowsableState.Never)]
```
manodasanW, line 25: same here and for methods and throughout other files.

src/cswinrt/code_writers.h
Sergio0694: It might show up as a warning if we ever use these directly from projections, but don't think we do? If we do, we can just add a `#pragma warning disable CS0614` for that warning in the generated projection file I guess? 🤔

src/cswinrt/code_writers.h
Sergio0694: We should always pass `None` except any field(s) are `IReference<T>` at any nesting depth.

src/WinRT.Runtime2/ABI/System/ComponentModel/DataErrorsChangedEventArgs.cs
Sergio0694: I want the obsolete with the message to enforce that these are private implementation details. Generated projections can just suppress the warning and we can say it's allowed only for them to consume these assuming the WinRT version is exactly in sync (which we also require). But I'm also not entirely sure projections call these directly either.

PR #2087, opened 2025-10-22
Exception Handling Implementation in WinRT.Runtime2/InteropServices/RestrictedErrorInfo.cs

kythant:


src/WinRT.Runtime/IInspectable.net5.cs
```diff
@@ -4,32 +4,30 @@
 using System;
 using System.Collections.Concurrent;
 
-namespace WinRT
-{
+namespace WinRT;
```
Sergio0694: Nit: is this by accident? As in, it's fixing the namespace in the old runtime for CsWinRT 2.x, not 3.0. We might consider not doing this to avoid merge conflicts maybe, but I don't feel strongly about it.

src/WinRT.Runtime2/InteropServices/ExceptionErrorInfo.cs
```diff
@@ -0,0 +1,150 @@
...
+        void* _description = null;
+        void* _restrictedDescription = null;
+        void* _capabilitySid = null;
+
+        try
+        {
+            fixed (int* pError = &error)
+            {
+                // GetErrorDetails
+                Marshal.ThrowExceptionForHR(((delegate* unmanaged[Stdcall]<void*, void**, int*, void**, void**, int>)(*(void***)thisPtr)[3])(
```
Sergio0694: Let's not have inline function pointer invocations. Add a vtable type for all the interfaces we need in this PR in the `Vtables` folder. There's a whole bunch of them already there, you can follow the same exact pattern 🙂

src/WinRT.Runtime2/InteropServices/ExceptionHelpers.cs
```diff
@@ -0,0 +1,451 @@
...
+using System.Runtime.InteropServices;
+using static System.Runtime.InteropServices.ComWrappers;
+
+namespace WindowsRuntime.InteropServices;
+
+#pragma warning disable IDE0051 // TODO
+
+internal static unsafe class ExceptionHelpers
+{
+    internal const int COR_E_OBJECTDISPOSED = unchecked((int)0x80131622);
```
Sergio0694: Move all of these to `WellKnownErrorCodes.tt`. Also add a comment if available, you can find some of them e.g. you can find many by looking up via https://james.darpinian.com/decoder/ or [here](https://learn.microsoft.com/en-us/windows/win32/seccrypto/common-hresult-values).

src/WinRT.Runtime2/InteropServices/ExceptionHelpers.cs
```diff
@@ -0,0 +1,451 @@
...
+                return ref *p;
+            }
+        }
+#endif
+}
+
+// Handcrafted P/Invoke with TFM-specific handling, or thin high-level abstractions (eg. 'TryGetProcAddress'/'GetProcAddress')
+internal partial class Platform
+{
+    public static unsafe void* TryGetProcAddress(void* moduleHandle, ReadOnlySpan<byte> functionName)
```
Sergio0694: Move any new P/Invoke-s to `WindowsRuntimeImports`.

src/WinRT.Runtime2/InteropServices/ExceptionHelpers.cs
```diff
@@ -0,0 +1,451 @@
...
+    internal const int REGDB_E_CLASSNOTREG = unchecked((int)0x80040154);
+
+    public static unsafe delegate* unmanaged[Stdcall]<void**, int> getRestrictedErrorInfo;
+    public static unsafe delegate* unmanaged[Stdcall]<void*, int> setRestrictedErrorInfo;
+    public static unsafe delegate* unmanaged[Stdcall]<int, void*, void*, int> roOriginateLanguageException;
+    public static unsafe delegate* unmanaged[Stdcall]<void*, int> roReportUnhandledError;
+
+    private static unsafe bool Initialize()
+    {
+        void* winRTErrorModule = Platform.LoadLibraryExW("api-ms-win-core-winrt-error-l1-1-1.dll", null, (uint)DllImportSearchPath.System32);
```
Sergio0694: Talked about this with @manodasanW a while back, we should remove the conditional paths here and `Initialize`, as we should be able to rely on all the APIs we need here to always exist on Windows versions we support. That should allow us to delete all those static fields for those function pointers, remove this `Initialize()` method entirely, and just call the various APIs directly when needed.

src/WinRT.Runtime2/InteropServices/RestrictedErrorInfoMethods.cs
```diff
@@ -0,0 +1,150 @@
...
+// Licensed under the MIT License.
+
+using System;
+using System.Runtime.InteropServices;
+using WindowsRuntime.InteropServices;
+using WindowsRuntime.InteropServices.Marshalling;
+
+namespace ABI.WinRT.Interop;
+
+internal static class IRestrictedErrorInfoMethods
```
Sergio0694: Code structure:
- We should have one file per `*Methods` class
- Let's put these under `InteropServices\Exceptions` (that folder doesn't currently exist)
- Namespace should be `WindowsRuntime.InteropServices`

src/WinRT.Runtime2/InteropServices/ExceptionErrorInfo.cs
```diff
@@ -0,0 +1,150 @@
...
+        finally
+        {
+            Marshal.FreeBSTR((nint)__retval);
+        }
+    }
+}
+
+internal static class ILanguageExceptionErrorInfoMethods
+{
+    public static unsafe void* GetLanguageException(void* thisPtr)
```
Sergio0694: Direct stubs like this one that directly just invoke a method with no additional marshalling on an untyped pointer can just be added to the vtable type. E.g. delete this type and move the logic for this method to `ILanguageExceptionErrorInfoVftbl.GetLanguageExceptionUnsafe`. The structure can match what e.g. `IInspectableVftbl.GetRuntimeClassNameUnsafe`.

src/WinRT.Runtime/IInspectable.net5.cs
manodasanW: The changes to this file should hopefully not be needed.

kythant: Oh I think It might just gotten caught up because IntelliSense complains about it anytime I have any of the winrt.runtime1 files opened. I will revert.

src/WinRT.Runtime2/InteropServices/RestrictedErrorInfoMethods.cs
kythant: Would it make sense to just combine all these static classes into one?

Sergio0694: No, because they refer to different interfaces. We have exactly 1 `Methods` class per type (except for generics).

src/WinRT.Runtime2/Properties/WindowsRuntimeFeatureSwitches.cs
```diff
@@ -51,6 +56,11 @@ internal static class WindowsRuntimeFeatureSwitches
     [FeatureSwitchDefinition(UseWindowsUIXamlProjectionsPropertyName)]
     public static bool UseWindowsUIXamlProjections { get; } = GetConfigurationValue(UseWindowsUIXamlProjectionsPropertyName, defaultValue: false);
 
+    /// <summary>
+    /// Gets a value indicating whether or not should initialize all default type mappings automatically (defaults to <see langword="true"/>).
+    /// </summary>
+    public static bool EnableDefaultCustomTypeMappings { get; } = GetConfigurationValue(EnableDefaultCustomTypeMappingsPropertyName, defaultValue: false);
```
Sergio0694: Why are we adding this switch? It shouldn't be applicable anymore in CsWinRT 3.0 (no static initializers)

src/WinRT.Runtime2/InteropServices/ExceptionErrorInfo.net5.cs
```diff
@@ -0,0 +1,53 @@
...
+        return _exception.Source;
+    }
+
+    public string? GetDescription()
+    {
+        string message = _exception.Message;
+        if (string.IsNullOrEmpty(message))
+        {
+            Type exceptionType = _exception.GetType();
+            if (exceptionType != null)
```
Sergio0694: I don't think this can possibly be `null`.

src/WinRT.Runtime2/InteropServices/Vtables/ILanguageExceptionErrorInfo2Vftbl.cs
```diff
@@ -0,0 +1,25 @@
...
+/// Binding type for the <c>IBindableVectorView</c> interface vtable.
+/// </summary>
+/// <remarks>
+/// This interface is equivalent to <see href="https://learn.microsoft.com/en-us/windows/win32/api/restrictederrorinfo/nn-restrictederrorinfo-ilanguageexceptionerrorinfo2"/>.
+/// </remarks>
+/// <see href="https://learn.microsoft.com/en-us/windows/win32/api/restrictederrorinfo/nn-restrictederrorinfo-ilanguageexceptionerrorinfo2"/>
+[StructLayout(LayoutKind.Sequential)]
+internal unsafe struct ILanguageExceptionErrorInfo2
+{
+    public delegate* unmanaged[MemberFunction]<void*, Guid*, void**, int> QueryInterface;
```
Sergio0694: Use `HRESULT` for the return types, not `int` (same for all methods, and all 3 new vtable types)

src/WinRT.Runtime2/InteropServices/Vtables/ILanguageExceptionErrorInfo2Vftbl.cs
```diff
@@ -0,0 +1,25 @@
...
+
+namespace WindowsRuntime.InteropServices;
+
+/// <summary>
+/// Binding type for the <c>IBindableVectorView</c> interface vtable.
+/// </summary>
+/// <remarks>
+/// This interface is equivalent to <see href="https://learn.microsoft.com/en-us/windows/win32/api/restrictederrorinfo/nn-restrictederrorinfo-ilanguageexceptionerrorinfo2"/>.
+/// </remarks>
+/// <see href="https://learn.microsoft.com/en-us/windows/win32/api/restrictederrorinfo/nn-restrictederrorinfo-ilanguageexceptionerrorinfo2"/>
```
Sergio0694: Update the summary and remove the remarks, only leave the `<see>` element (same for all 3 vtable types)

Sergio0694: Remove the `en-us` bit from all docs links (same for all 3 vtable types)

src/WinRT.Runtime2/InteropServices/ExceptionErrorInfo.net5.cs
kythant: I guess without the ? there, it is not

src/WinRT.Runtime2/InteropServices/LanguageExceptionErrorInfoMethods2.cs
```diff
@@ -0,0 +1,26 @@
...
+        try
+        {
+            Marshal.ThrowExceptionForHR(((ILanguageExceptionErrorInfo2Vftbl*)*(void***)thisPtr)->CapturePropagationContext(
+                thisPtr,
+                managedExceptionWrapper.GetThisPtrUnsafe()));
+        }
+        finally
+        {
+            _ = Marshal.Release((nint)managedExceptionWrapper.DetachThisPtrUnsafe());
+        }
```
Sergio0694: You can remove the `try/finally` and just use a `using` statement here.
Look at other `Methods` types for reference 🙂
```suggestion
        using WindowsRuntimeObjectReferenceValue exValue = WindowsRuntimeObjectMarshaller.ConvertToUnmanaged(ex);

        Marshal.ThrowExceptionForHR(((ILanguageExceptionErrorInfo2Vftbl*)*(void***)thisPtr)->CapturePropagationContext(
            thisPtr,
            exValue.GetThisPtrUnsafe()));
        }
```

src/WinRT.Runtime2/InteropServices/LanguageExceptionErrorInfoMethods2.cs
```diff
@@ -0,0 +1,26 @@
...
+
+using System;
+using System.Runtime.InteropServices;
+using WindowsRuntime.InteropServices.Marshalling;
+
+namespace WindowsRuntime.InteropServices;
+
+internal static class ILanguageExceptionErrorInfo2Methods
+{
+    public static unsafe void CapturePropagationContext(void* thisPtr, Exception ex)
```
Sergio0694: Nits, for consistency:
- Rename the exception parameter `exception` (never use abbreviations)
- The reference value should be named the param name + "Value", so here it'd be "exceptionValue"

src/WinRT.Runtime2/InteropServices/Vtables/ILanguageExceptionErrorInfo2Vftbl.cs
```diff
@@ -0,0 +1,40 @@
...
+internal unsafe struct ILanguageExceptionErrorInfo2Vftbl
+{
+    public delegate* unmanaged[MemberFunction]<void*, Guid*, void**, HRESULT> QueryInterface;
+    public delegate* unmanaged[MemberFunction]<void*, uint> AddRef;
+    public delegate* unmanaged[MemberFunction]<void*, uint> Release;
+    public delegate* unmanaged[MemberFunction]<void*, void**, HRESULT> GetPreviousLanguageExceptionErrorInfo;
+    public delegate* unmanaged[MemberFunction]<void*, void*, HRESULT> CapturePropagationContext;
+    public delegate* unmanaged[MemberFunction]<void*, void**, HRESULT> GetPropagationContextHead;
+
+    public static void* GetPreviousLanguageExceptionErrorInfoUnsafe(void* thisPtr)
```
Sergio0694: The `*Unsafe` methods in vtable types are meant to be directly forwarding stubs for vtable slots, just for convenience, but they should not add additional marshalling/error logic. So e.g. here these should return `HRESULT` directly, and callers should do any checks they want. You can use the methods in `IInspectableVftbl` for reference.

Sergio0694: Missing aggressive inlining on these two stubs

kythant: OH shoot yea I didn't notice the ThrowExceptionForHR was actually marshalling

src/WinRT.Runtime2/InteropServices/RestrictedErrorInfo.cs
```diff
@@ -62,26 +294,155 @@ public static void ThrowExceptionForHR(HRESULT errorCode)
     /// <seealso cref="Marshal.GetExceptionForHR(int)"/>
     public static HRESULT GetHRForException(Exception? exception)
     {
-        // TODO
-        return 0;
+        if (exception == null)
+        {
+            throw new ArgumentNullException("exception");
+        }
```
Sergio0694:
```suggestion
        ArgumentNullException.ThrowIfNull(exception);
```

src/WinRT.Runtime2/InteropServices/RestrictedErrorInfo.cs
```diff
@@ -62,26 +294,155 @@ public static void ThrowExceptionForHR(HRESULT errorCode)
...
     /// Stores info on the input exception through the <c>IRestrictedErrorInfo</c> infrastructure, to retrieve it later.
     /// </summary>
     /// <param name="exception">The input <see cref="Exception"/> instance to store.</param>
-    public static void SetErrorInfo(Exception exception)
+    public static unsafe void SetErrorInfo(Exception exception)
     {
-        // TODO
+        try
+        {
+            if (WindowsRuntimeImports.GetRestrictedErrorInfo != null && WindowsRuntimeImports.SetRestrictedErrorInfo != null && WindowsRuntimeImports.RoOriginateLanguageException != null)
```
Sergio0694: I suspect the implicit delegate conversion is tripping you up here, as these checks aren't valid (as in, these are now methods, they can never be null). Basically you can delete any checks like these that we had from before 🙂

src/WinRT.Runtime2/InteropServices/RestrictedErrorInfo.cs
```diff
@@ -62,26 +294,155 @@ public static void ThrowExceptionForHR(HRESULT errorCode)
...
+        }
+        int hr = exception.HResult;
+        try
+        {
+            if (exception.TryGetRestrictedLanguageErrorInfo(out WindowsRuntimeObjectReference? restrictedErrorObject, out bool _))
+            {
+                if (restrictedErrorObject != null)
+                {
+                    IRestrictedErrorInfoMethods.GetErrorDetails(restrictedErrorObject.GetThisPtr(), out hr);
+                    GC.KeepAlive(restrictedErrorObject);
```
Sergio0694: Calling `GetThisPtr()` will leak the pointer here, as we never release it. This part needs some tweaks compared to CsWinRT 2.x, as we changed a lot of the reference tracking logic here. Basically what you can do here is to go through a value, which will get a "reference tracker lease" for the object reference, and pass a pointer from there. It also has the benefit that can entirely skip the native `AddRef` and `Release` pair of calls. Do this:
```csharp
using WindowsRuntimeObjectReferenceValue restrictedErrorObjectValue = restrictedErrorObject.AsValue();
```
Then pass `GetThisPtrUnsafe()` from that value to that `GetErrorDetails` call.
Also don't need `GC.KeepAlive` anymore, the value will take care of things 🙂

src/WinRT.Runtime2/InteropServices/ExceptionHelpers.cs
```diff
@@ -0,0 +1,266 @@
...
+                _ = Marshal.Release(languageErrorInfo2Ptr);
+            }
+        }
+
+        return null;
+    }
+
+    private static unsafe Exception? GetLanguageExceptionInternal(void* languageErrorInfoPtr, int hr)
+    {
+        void* languageExceptionPtr = default;
```
Sergio0694: Nit: always use `null` to initialize null pointers instead of `default`

src/WinRT.Runtime2/InteropServices/ExceptionHelpers.cs
```diff
@@ -0,0 +1,266 @@
...
+        _ = ILanguageExceptionErrorInfoVftbl.GetLanguageExceptionUnsafe(languageErrorInfoPtr, &languageExceptionPtr);
+        if (languageExceptionPtr != default)
```
Sergio0694: Same here, compare with `null`

src/WinRT.Runtime2/InteropServices/FreeThreadedMarshaler.cs
```diff
@@ -100,14 +100,25 @@ static FreeThreadedMarshaler InitializeInstanceForCurrentThread()
...
+                {
+                    HRESULT hr = IUnknownVftbl.QueryInterfaceUnsafe(marshalerPtr, in WellKnownInterfaceIds.IID_IMarshal, out void* pMarshalObject);
 
-                // The returned marshaler is documented to be free-threaded, so we can instantiate 'FreeThreadedObjectReference'
-                // directly. This also should allow inlining all virtual calls to the object in this class, in the stubs below.
-                FreeThreadedObjectReference objectReference = new(marshalerPtr, referenceTrackerPtr: null);
+                    if (!WellKnownErrorCodes.Succeeded(hr))
+                    {
+                        Marshal.ThrowExceptionForHR(hr);
+                    }
```
Sergio0694: Just do `Marshal.ThrowExceptionForHR(hr);` directly, it'll already do the same check

src/WinRT.Runtime2/InteropServices/FreeThreadedMarshaler.cs
```diff
@@ -100,14 +100,25 @@ static FreeThreadedMarshaler InitializeInstanceForCurrentThread()
             {
                 // Create the free-threaded marshaler
                 void* marshalerPtr;
-
                 WindowsRuntimeImports.CoCreateFreeThreadedMarshaler(punkOuter: null, ppunkMarshal: &marshalerPtr).Assert();
+                try
+                {
+                    HRESULT hr = IUnknownVftbl.QueryInterfaceUnsafe(marshalerPtr, in WellKnownInterfaceIds.IID_IMarshal, out void* pMarshalObject);
```
Sergio0694: Don't use abbreviations. This would be `hresult`.

src/WinRT.Runtime2/InteropServices/FreeThreadedMarshaler.cs
```diff
@@ -100,14 +100,25 @@ static FreeThreadedMarshaler InitializeInstanceForCurrentThread()
             {
                 // Create the free-threaded marshaler
                 void* marshalerPtr;
-
                 WindowsRuntimeImports.CoCreateFreeThreadedMarshaler(punkOuter: null, ppunkMarshal: &marshalerPtr).Assert();
```
Sergio0694: For clarity:
- Rename this parameter to `marshalUnknownPtr`
- Rename the one below to `marshalPtr`

src/WinRT.Runtime2/InteropServices/FreeThreadedMarshaler.cs
Sergio0694: If you don't actually need the `HRESULT` separately, you can simplify calls like these by always just using `.Assert()` at the end. Same everywhere else you're doing similar calls immediately followed by `Marshal.ThrowExceptionForHR`

src/WinRT.Runtime2/InteropServices/RestrictedErrorInfo.cs
```diff
@@ -62,26 +293,137 @@ public static void ThrowExceptionForHR(HRESULT errorCode)
...
+                fixed (char* lpMessage = message)
+                {
+                    if (WindowsRuntimeImports.WindowsCreateStringReference(
+                        sourceString: lpMessage,
+                        length: (uint)message.Length,
+                        hstringHeader: header,
+                        @string: hstring) != 0)
+                    {
+                        hstring = null;
+                    }
```
Sergio0694: Don't need to use this API directly. You can simplify this by using `HStringMarshaller.ConvertToUnmanagedUnsafe`

src/WinRT.Runtime2/InteropServices/ExceptionHelpers.cs
```diff
@@ -0,0 +1,264 @@
...
+    private static unsafe Exception? GetLanguageExceptionInternal(void* languageErrorInfoPtr, int hr)
+    {
+        void* languageExceptionPtr = null;
+        try
+        {
+            if (WellKnownErrorCodes.Succeeded(ILanguageExceptionErrorInfoVftbl.GetLanguageExceptionUnsafe(languageErrorInfoPtr, &languageExceptionPtr)) && languageExceptionPtr != null)
+            {
+                if (WindowsRuntimeMarshal.IsReferenceToManagedObject(languageExceptionPtr))
+                {
+                    Exception ex = ComInterfaceDispatch.GetInstance<Exception>((ComInterfaceDispatch*)languageExceptionPtr);
```
Sergio0694: You can simplify this by replacing the `null` check + `IsReferenceToManagedObject` + `ComInterfaceDispatch` call with just a single call to `WindowsRuntimeMarshal.TryGetManagedObject`, it does exactly the same as all this 🙂

src/WinRT.Runtime2/InteropServices/ExceptionHelpers.cs
```diff
@@ -0,0 +1,264 @@
...
+    {
+        int hr = ex.HResult;
+        try
+        {
+            if (ex.TryGetRestrictedLanguageErrorInfo(out WindowsRuntimeObjectReference? restrictedErrorObject, out bool _))
+            {
+                if (restrictedErrorObject != null)
+                {
+                    IRestrictedErrorInfoMethods.GetErrorDetails(restrictedErrorObject.GetThisPtrUnsafe(), out hr);
+                    GC.KeepAlive(restrictedErrorObject);
```
Sergio0694: Same here, we need to replace this with an object reference value to make it safe. Also remove `GC.KeepAlive`.

src/WinRT.Runtime2/InteropServices/ExceptionHelpers.cs
```diff
@@ -0,0 +1,264 @@
...
+
+        return null;
+    }
+
+    public static unsafe int GetHRForException(Exception ex)
+    {
+        int hr = ex.HResult;
+        try
+        {
+            if (ex.TryGetRestrictedLanguageErrorInfo(out WindowsRuntimeObjectReference? restrictedErrorObject, out bool _))
```
Sergio0694: Nit: if you discard the value, you don't need to declare the type either.
```suggestion
            if (ex.TryGetRestrictedLanguageErrorInfo(out WindowsRuntimeObjectReference? restrictedErrorObject, out _))
```
Same everywhere else you do this.

src/WinRT.Runtime2/InteropServices/ExceptionHelpers.cs
```diff
@@ -0,0 +1,264 @@
...
+}
+
+internal static class MarshalExtensions
+{
+    /// <summary>
+    /// Releases a COM object, if not <see langword="null"/>.
+    /// </summary>
+    /// <param name="pUnk">The input COM object to release.</param>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static unsafe void ReleaseIfNotNull(void* pUnk)
```
Sergio0694: Remove this. You can use `WindowsRuntimeObjectMarshaller.Free` instead.

Sergio0694 (changes requested):

src/WinRT.Runtime2/InteropServices/ExceptionHelpers.cs
kythant: Oh I already had it removed before seeing your comment. But I used Marshal.Release() instead. I will switch it to WindowsRuntimeObjectMarshaller.Free

src/WinRT.Runtime2/InteropServices/FreeThreadedMarshaler.cs
kythant: Not sure which one you are referring to by the one below.

src/WinRT.Runtime2/InteropServices/FreeThreadedMarshaler.cs
kythant: Not sure what you mean by assert but I was just going to do it this way. 

Marshal.ThrowExceptionForHR(IUnknownVftbl.QueryInterfaceUnsafe(marshalUnknownPtr, in WellKnownInterfaceIds.IID_IMarshal, out void* pMarshalObject));

src/WinRT.Runtime2/InteropServices/ExceptionHelpers.cs
kythant: I've removed all this code because I realize there is already a duplicate in RestrictedError.cs

https://github.com/microsoft/CsWinRT/pull/2087/commits/081cf64d4a82e8ee905514cea7296b96ff4e7622

src/WinRT.Runtime2/InteropServices/FreeThreadedMarshaler.cs
Sergio0694: I meant `pMarshalObject`, the one you get doing a QI on this initial untyped marshaller.

src/WinRT.Runtime2/InteropServices/FreeThreadedMarshaler.cs
Sergio0694: I meant that unless you need the `HRESULT` explicitly, you can just do this:

```suggestion
                    IUnknownVftbl.QueryInterfaceUnsafe(marshalerPtr, in WellKnownInterfaceIds.IID_IMarshal, out void* pMarshalObject).Assert();
```

Which is more readable 🙂

src/WinRT.Runtime2/InteropServices/FreeThreadedMarshaler.cs
kythant: gotcha! Thanks!

src/WinRT.Runtime2/InteropServices/ErrorInfo/IErrorInfo.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Runtime.InteropServices.Marshalling;
+
+namespace WindowsRuntime;
+
+/// <summary>
+/// IManagedExceptionErrorInfo
+/// </summary>
+[Guid("1CF2B120-547D-101B-8E65-08002B2BD119")]
+[GeneratedComInterface]
+internal unsafe partial interface IErrorInfo
```
Sergio0694: Unused, delete this?

src/WinRT.Runtime2/InteropServices/ExceptionErrorInfo.net5.cs
```diff
@@ -0,0 +1,49 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Runtime.InteropServices.Marshalling;
+
+namespace WindowsRuntime.InteropServices;
+
+[GeneratedComClass]
+internal sealed partial class ManagedExceptionErrorInfo : IErrorInfo
```
Sergio0694: Also unused, delete this type too?

src/WinRT.Runtime2/InteropServices/ExceptionHelpers.Microsoft.cs
```diff
@@ -0,0 +1,92 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using WindowsRuntime.InteropServices;
+
+namespace Microsoft.UI.Xaml;
+
+internal class ElementNotAvailableException : Exception
```
Sergio0694: These exception types should all be sealed

Sergio0694: Move all of these under an `\Exceptions` folder at the top level (next to our RCW wrappers), and then with two subfolders: `\Microsoft.UI.Xaml` and `\Windows.UI.Xaml`. One exception per file, with the file named after the exception type it contains.

Sergio0694: Please add XML docs to all members.

src/WinRT.Runtime2/InteropServices/ExceptionHelpers.Microsoft.cs
```diff
@@ -0,0 +1,92 @@
...
+        HResult = WellKnownErrorCodes.E_ELEMENTNOTAVAILABLE;
+    }
+
+    public ElementNotAvailableException(string message)
+        : base(message)
+    {
+        HResult = WellKnownErrorCodes.E_ELEMENTNOTAVAILABLE;
+    }
+
+    public ElementNotAvailableException(string message, Exception innerException)
```
Sergio0694: Given that these exception types are internal, you can also delete all unused members. For instance, all these constructors taking an inner exception have 0 references, you should be able to delete all of them?

src/WinRT.Runtime2/InteropServices/RestrictedErrorInfoMethods.cs
```diff
@@ -0,0 +1,89 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Runtime.InteropServices;
+
+namespace WindowsRuntime.InteropServices;
+
+internal static class IRestrictedErrorInfoMethods
```
Sergio0694, line 19: Filename should match the interface name

src/WinRT.Runtime2/InteropServices/LanguageExceptionErrorInfoMethods2.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using WindowsRuntime.InteropServices.Marshalling;
+
+namespace WindowsRuntime.InteropServices;
+
+internal static class ILanguageExceptionErrorInfo2Methods
```
Sergio0694: Filename should match the interface name

src/WinRT.Runtime2/InteropServices/RestrictedErrorInfoMethods.cs
Sergio0694, line 19: All these `*Methods` types should be in an `\Exceptions` subfolder under `\InteropServices`

Sergio0694, line 19: Missing XML docs in all of these new `*Methods` types and members

src/WinRT.Runtime2/Exceptions/InteropServices/RestrictedErrorInfoMethods.cs
```diff
@@ -0,0 +1,89 @@
...
+        {
+            Marshal.FreeBSTR((nint)_description);
+            Marshal.FreeBSTR((nint)_restrictedDescription);
+            Marshal.FreeBSTR((nint)_capabilitySid);
+        }
+    }
+
+    public static unsafe string GetReference(void* thisPtr)
+    {
+        void* __retval = default;
```
Sergio0694:
```suggestion
        void* __retval = null;
```

src/WinRT.Runtime2/InteropServices/FreeThreadedMarshaler.cs
```diff
@@ -99,15 +99,20 @@ public static FreeThreadedMarshaler InstanceForCurrentThread
...
+                    // The returned marshaler is documented to be free-threaded, so we can instantiate 'FreeThreadedObjectReference'
+                    // directly. This also should allow inlining all virtual calls to the object in this class, in the stubs below.
+                    FreeThreadedObjectReference objectReference = new(marshalPtr, referenceTrackerPtr: null);
+                    return instanceForCurrentThread = new FreeThreadedMarshaler(objectReference);
+                }
+                finally
+                {
+                    _ = IUnknownVftbl.ReleaseUnsafe(marshalUnknownPtr);
+                }
             }
```
Sergio0694: We should add some clarifying comments, plus (nit) I'd suggest following the style of the rest of the codebase, where we add blank lines between different kinds of lines to make the code more readable (e.g. between declarations and statements, before comments, etc.).

This is how I'd update this bit to address both of the above, for instance:

```csharp
void* marshalUnknownPtr;

// Create the free-threaded marshaler (the return will only be an 'IUnknown' interface pointer)
WindowsRuntimeImports.CoCreateFreeThreadedMarshaler(punkOuter: null, ppunkMarshal: &marshalUnknownPtr).Assert();

try
{
    // We need an explicit 'QueryInterface' call to actually get the 'IMarshal' interface pointer to use
    IUnknownVftbl.QueryInterfaceUnsafe(marshalUnknownPtr, in WellKnownInterfaceIds.IID_IMarshal, out void* marshalPtr).Assert();

    // The returned marshaler is documented to be free-threaded, so we can instantiate 'FreeThreadedObjectReference'
    // directly. This also should allow inlining all virtual calls to the object in this class, in the stubs below.
    FreeThreadedObjectReference objectReference = new(marshalPtr, referenceTrackerPtr: null);

    return instanceForCurrentThread = new FreeThreadedMarshaler(objectReference);
}
finally
{
    _ = IUnknownVftbl.ReleaseUnsafe(marshalUnknownPtr);
}
```

src/WinRT.Runtime2/InteropServices/RestrictedErrorInfo.cs
```diff
@@ -28,10 +34,127 @@ public static unsafe class RestrictedErrorInfo
     /// calls (both in managed and native code). This improves the debugging experience.
     /// </remarks>
     /// <seealso cref="Marshal.GetExceptionForHR(int)"/>
-    public static Exception? GetExceptionForHR(HRESULT errorCode)
+    public static Exception GetExceptionForHR(HRESULT errorCode, out bool restoredExceptionFromGlobalState)
     {
-        // TODO
-        return null;
+        restoredExceptionFromGlobalState = false;
+        Exception? ex;
```
Sergio0694: Never use abbreviations for parameter/local names. E.g. this should be `exception`.

src/WinRT.Runtime2/InteropServices/RestrictedErrorInfo.cs
```diff
@@ -28,10 +34,127 @@ public static unsafe class RestrictedErrorInfo
...
+        string errorMessage = string.Empty;
+        Exception internalGetGlobalErrorStateException;
+
+        WindowsRuntimeObjectReferenceValue restrictedErrorInfoValue = ExceptionHelpers.BorrowRestrictedErrorInfo();
+
+        try
+        {
+            void* restrictedErrorInfoValuePtr = restrictedErrorInfoValue.GetThisPtrUnsafe();
+            if (restrictedErrorInfoValuePtr != default)
```
Sergio0694: Compare pointers with `null`, not `default`

src/WinRT.Runtime2/InteropServices/RestrictedErrorInfo.cs
```diff
@@ -28,10 +34,127 @@ public static unsafe class RestrictedErrorInfo
...
+            {
+                if (Marshal.QueryInterface((IntPtr)restrictedErrorInfoValuePtr, WellKnownInterfaceIds.IID_ILanguageExceptionErrorInfo, out nint languageErrorInfoPtr) >= 0)
```
Sergio0694: For consistency, use `IUnknownVftbl.QueryInterfaceUnsafe` instead of `Marshal.QueryInterface`

Sergio0694: You can check the success with `WellKnownErrorCodes.Succeeded` rather than with `>= 0` manually. Might want to split that into two lines for readability, like:

```csharp
HRESULT hresult = IUnknownVftbl.QueryInterfaceUnsafe(restrictedErrorInfoValuePtr, in WellKnownInterfaceIds.IID_ILanguageExceptionErrorInfo, out void* languageErrorInfoPtr);

if (WellKnownErrorCodes.Succeeded(hresult))
{
    // TODO
}
```

This also avoids needing those casts

src/WinRT.Runtime2/InteropServices/RestrictedErrorInfo.cs
```diff
@@ -28,10 +34,127 @@ public static unsafe class RestrictedErrorInfo
...
+                {
+                    try
+                    {
+                        ex = ExceptionHelpers.GetLanguageException((void*)languageErrorInfoPtr, errorCode);
+                        if (ex is not null)
+                        {
+                            restoredExceptionFromGlobalState = true;
+                            WindowsRuntimeObjectReference? errRef = WindowsRuntimeObjectReference.Create(restrictedErrorInfoValuePtr, WellKnownInterfaceIds.IID_IRestrictedErrorInfo);
```
Sergio0694: Do we need a QI here? If the pointer is guaranteed to match already, use `CreateUnsafe` instead.

src/WinRT.Runtime2/InteropServices/RestrictedErrorInfo.cs
```diff
@@ -28,10 +34,127 @@ public static unsafe class RestrictedErrorInfo
...
+                            return ex;
+                        }
+                    }
+                    finally
+                    {
+                        _ = Marshal.Release(languageErrorInfoPtr);
+                    }
+                }
+
+                WindowsRuntimeObjectReference? restrictedErrorInfoToSave = WindowsRuntimeObjectReference.Create(restrictedErrorInfoValuePtr, WellKnownInterfaceIds.IID_IRestrictedErrorInfo);
```
Sergio0694: Same here for `CreateUnsafe`

src/WinRT.Runtime2/InteropServices/RestrictedErrorInfo.cs
```diff
@@ -28,10 +34,127 @@ public static unsafe class RestrictedErrorInfo
...
+            internalGetGlobalErrorStateException = e;
+        }
+        finally
+        {
+            restrictedErrorInfoValue.Dispose();
+        }
+
+        ex = errorCode switch
+        {
+            WellKnownErrorCodes.E_CHANGED_STATE or WellKnownErrorCodes.E_ILLEGAL_STATE_CHANGE or WellKnownErrorCodes.E_ILLEGAL_METHOD_CALL or WellKnownErrorCodes.E_ILLEGAL_DELEGATE_ASSIGNMENT or WellKnownErrorCodes.APPMODEL_ERROR_NO_PACKAGE or WellKnownErrorCodes.COR_E_INVALIDOPERATION => !string.IsNullOrEmpty(errorMessage) ? new InvalidOperationException(errorMessage) : new InvalidOperationException(),
```
Sergio0694: Nit: go to a new line after each `or` to make this more readable

src/WinRT.Runtime2/InteropServices/RestrictedErrorInfo.cs
```diff
@@ -28,10 +34,127 @@ public static unsafe class RestrictedErrorInfo
...
+            WellKnownErrorCodes.E_XAMLPARSEFAILED => WindowsRuntimeFeatureSwitches.UseWindowsUIXamlProjections
+                                ? !string.IsNullOrEmpty(errorMessage) ? new Windows.UI.Xaml.XamlParseException(errorMessage) : new Windows.UI.Xaml.XamlParseException()
```
Sergio0694: Nit: left align all of these with 4 spaces to the right of the previous line, like this:

image

src/WinRT.Runtime2/InteropServices/RestrictedErrorInfo.cs
```diff
@@ -28,10 +34,127 @@ public static unsafe class RestrictedErrorInfo
...
+                                ? !string.IsNullOrEmpty(errorMessage) ? new Windows.UI.Xaml.ElementNotAvailableException(errorMessage) : new Windows.UI.Xaml.ElementNotAvailableException()
+                                : !string.IsNullOrEmpty(errorMessage) ? new Microsoft.UI.Xaml.ElementNotAvailableException(errorMessage) : new Microsoft.UI.Xaml.ElementNotAvailableException(),
+            WellKnownErrorCodes.E_ELEMENTNOTENABLED => WindowsRuntimeFeatureSwitches.UseWindowsUIXamlProjections
+                                ? !string.IsNullOrEmpty(errorMessage) ? new Windows.UI.Xaml.ElementNotEnabledException(errorMessage) : new Windows.UI.Xaml.ElementNotEnabledException()
+                                : !string.IsNullOrEmpty(errorMessage) ? new Microsoft.UI.Xaml.ElementNotEnabledException(errorMessage) : new Microsoft.UI.Xaml.ElementNotEnabledException(),
+            WellKnownErrorCodes.ERROR_INVALID_WINDOW_HANDLE => new COMException(
+            @"Invalid window handle. (0x80070578)
+Consider WindowNative, InitializeWithWindow
+See https://aka.ms/cswinrt/interop#windows-sdk",
+                                WellKnownErrorCodes.ERROR_INVALID_WINDOW_HANDLE),
```
Sergio0694: Exception messages shouldn't have multiple lines, let's change this:
```suggestion
            WellKnownErrorCodes.ERROR_INVALID_WINDOW_HANDLE => new COMException(
                "Invalid window handle (0x80070578). Consider WindowNative, InitializeWithWindow. See https://aka.ms/cswinrt/interop#windows-sdk.",
                WellKnownErrorCodes.ERROR_INVALID_WINDOW_HANDLE),
```

src/WinRT.Runtime2/InteropServices/RestrictedErrorInfo.cs
```diff
@@ -47,7 +170,26 @@ public static unsafe class RestrictedErrorInfo
...
-        // TODO
+        if (errorCode < 0)
+        {
+            Throw(errorCode);
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        static void Throw(int errorCode)
+        {
+#pragma warning disable CS8600 // Converting null literal or possible null value to non-nullable type.
```
Sergio0694: This suppression isn't needed, and the local should be declared as not nullable

src/WinRT.Runtime2/InteropServices/RestrictedErrorInfo.cs
```diff
@@ -47,7 +170,26 @@ public static unsafe class RestrictedErrorInfo
...
+            Exception? ex = GetExceptionForHR(errorCode, out bool restoredExceptionFromGlobalState);
```
Sergio0694:
```suggestion
            Exception? exception = GetExceptionForHR(errorCode, out bool restoredExceptionFromGlobalState);
```

src/WinRT.Runtime2/InteropServices/RestrictedErrorInfo.cs
```diff
@@ -62,26 +204,125 @@ public static void ThrowExceptionForHR(HRESULT errorCode)
     /// <seealso cref="Marshal.GetExceptionForHR(int)"/>
     public static HRESULT GetHRForException(Exception? exception)
     {
-        // TODO
-        return 0;
+        ArgumentNullException.ThrowIfNull(exception);
+        int hresult = exception.HResult;
```
Sergio0694: Always use `HRESULT` for hresults (it's an alias):
```suggestion
        HRESULT hresult = exception.HResult;
```

src/WinRT.Runtime2/InteropServices/RestrictedErrorInfo.cs
```diff
@@ -62,26 +204,125 @@ public static void ThrowExceptionForHR(HRESULT errorCode)
...
     /// Triggers the global error handler when an unhandled exception occurs.
     /// </summary>
     /// <param name="exception">The input <see cref="Exception"/> instance to flow to the global error handler.</param>
     /// <see href="https://learn.microsoft.com/windows/win32/api/roerrorapi/nf-roerrorapi-roreportunhandlederror"/>
-    public static void ReportUnhandledError(Exception exception)
+    public static unsafe void ReportUnhandledError(Exception exception)
     {
-        // TODO
+        SetErrorInfo(exception);
+        WindowsRuntimeObjectReferenceValue restrictedErrorInfoValue = ExceptionHelpers.BorrowRestrictedErrorInfo();
```
Sergio0694: Use a `using` block here instead of conditionally disposing. If the pointer is `null` the dispose will just be a no-op

src/WinRT.Runtime2/InteropServices/RestrictedErrorInfo.cs
```diff
@@ -62,26 +204,125 @@ public static void ThrowExceptionForHR(HRESULT errorCode)
...
+        void* restrictedErrorInfoValuePtr = restrictedErrorInfoValue.GetThisPtrUnsafe();
+        if (restrictedErrorInfoValuePtr != default)
```
Sergio0694:
```suggestion
        if (restrictedErrorInfoValuePtr != null)
```

src/WinRT.Runtime2/InteropServices/RestrictedErrorInfo.cs
```diff
@@ -62,26 +204,125 @@ public static void ThrowExceptionForHR(HRESULT errorCode)
...
     /// </summary>
     /// <param name="exception">The input <see cref="Exception"/> instance to flow to the global error handler.</param>
     /// <see href="https://learn.microsoft.com/windows/win32/api/roerrorapi/nf-roerrorapi-roreportunhandlederror"/>
-    public static void ReportUnhandledError(Exception exception)
+    public static unsafe void ReportUnhandledError(Exception exception)
     {
-        // TODO
+        SetErrorInfo(exception);
+        WindowsRuntimeObjectReferenceValue restrictedErrorInfoValue = ExceptionHelpers.BorrowRestrictedErrorInfo();
+        void* restrictedErrorInfoValuePtr = restrictedErrorInfoValue.GetThisPtrUnsafe();
```
Sergio0694, line 399: Also same general nit as above about leaving blank line between different kinds of statements, for readability

src/WinRT.Runtime2/InteropServices/ExceptionHelpers.cs
```diff
@@ -0,0 +1,165 @@
...
+using System.Runtime.InteropServices;
+using WindowsRuntime.InteropServices.Marshalling;
+
+namespace WindowsRuntime.InteropServices;
+
+#pragma warning disable IDE0051 // TODO
+
+internal static unsafe class ExceptionHelpers
+{
+    public static unsafe WindowsRuntimeObjectReferenceValue BorrowRestrictedErrorInfo()
```
Sergio0694: Please add XML docs to the members

src/WinRT.Runtime2/InteropServices/ExceptionHelpers.cs
```diff
@@ -0,0 +1,165 @@
...
+        {
+            return default;
+        }
+
+        if (WindowsRuntimeImports.SetRestrictedErrorInfo != null)
+        {
+            WindowsRuntimeImports.SetRestrictedErrorInfo(restrictedErrorInfoPtr).Assert();
+        }
+
+        return new WindowsRuntimeObjectReferenceValue(restrictedErrorInfoPtr);
```
Sergio0694:
```suggestion
        return new (restrictedErrorInfoPtr);
```

src/WinRT.Runtime2/InteropServices/ExceptionHelpers.cs
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    // This is a helper method specifically to be used by exception propagation scenarios where we carefully
+    // manage the lifetime of the CCW for the exception object to avoid cycles and thereby leaking it.
+    internal static unsafe Exception? GetLanguageException(void* languageErrorInfoPtr, int hr)
```
Sergio0694:
```suggestion
    internal static unsafe Exception? GetLanguageException(void* languageErrorInfoPtr, HRESULT hresult)
```

src/WinRT.Runtime2/InteropServices/ExceptionHelpers.cs
```diff
@@ -0,0 +1,165 @@
...
+                        {
+                            WindowsRuntimeObjectMarshaller.Free(previousLanguageExceptionErrorInfo2Ptr);
+                        }
+                    }
+                }
+            }
+            finally
+            {
+                WindowsRuntimeObjectMarshaller.Free(currentLanguageExceptionErrorInfo2Ptr);
+                _ = Marshal.Release(languageErrorInfo2Ptr);
```
Sergio0694: Use `IUnknownVftbl.Release` instead of `Marshal`, for consistency

src/WinRT.Runtime2/InteropServices/ExceptionHelpers.cs
```diff
@@ -0,0 +1,165 @@
...
+                isLanguageException = (bool)dict["__HasRestrictedLanguageErrorObject"]!;
+            }
+
+            return restrictedErrorObject is not null;
+        }
+
+        return false;
+    }
+
+    public static void SetHResult(this Exception ex, int value)
```
Sergio0694: Delete this extension and just always set `HResult`, this method was only for backward-compatibility

src/WinRT.Runtime2/InteropServices/ExceptionHelpers.cs
```diff
@@ -0,0 +1,165 @@
...
+            // using Data["RestrictedErrorReference"]
+            dict["__RestrictedErrorObjectReference"] = restrictedErrorObject;
+            dict["__HasRestrictedLanguageErrorObject"] = hasRestrictedLanguageErrorObject;
+        }
+    }
+}
+
+internal static class ExceptionExtensions
+{
+    public static bool TryGetRestrictedLanguageErrorInfo(
```
Sergio0694: Move this to `ExceptionHelpers`, and it shouldn't be an extension, just a normal method there

src/WinRT.Runtime2/InteropServices/ErrorInfo/IErrorInfo.cs
kythant: Hmmm
image

kythant: > Hmmm. This has 25 references

image

kythant: Nevermind, it's lying to me...
image
Once I click it, it turned into one lolol

src/WinRT.Runtime2/InteropServices/ExceptionHelpers.Microsoft.cs
kythant: Wait. Top level of the solution or top level of the winrt.runtime2 project?
Sorry I could not identify where the RCW wrappers are.

src/WinRT.Runtime2/InteropServices/RestrictedErrorInfoMethods.cs
kythant: Hmmm so because this is a static class and not an interface. I've renamed this class to RestrictedErrorInfoMethods to match the file name.

src/WinRT.Runtime2/InteropServices/LanguageExceptionErrorInfoMethods2.cs
kythant: Deleting this file because there is only one reference.

src/WinRT.Runtime2/InteropServices/FreeThreadedMarshaler.cs
kythant: Clarifying comments on the whole file or just this piece?

src/WinRT.Runtime2/InteropServices/RestrictedErrorInfo.cs
kythant: Fixed it in line 44 too

src/WinRT.Runtime2/InteropServices/ExceptionHelpers.cs
kythant: I thought extension meant the first param had a "this". Hence extending the exception class. 

But I've moved it but kept the "this"

src/WinRT.Runtime2/InteropServices/RestrictedErrorInfo.cs
kythant: I'll ask Mano about this

src/WinRT.Runtime2/InteropServices/ExceptionHelpers.Microsoft.cs
Sergio0694: Sorry I meant top level as in right under WinRT.Runtime2, i.e. not nested in another folder in the project 😅

src/WinRT.Runtime2/InteropServices/RestrictedErrorInfoMethods.cs
Sergio0694: No this is not the pattern that we've been using. Yes in normal .NET you should only use the "I" prefix for interfaces, but this is a bit of a special repo. What we do for all "*Methods" classes is we always use the name of the interface + "Methods", and the type is a class. E.g. you can see lots of types like this under "AsyncInfo", "Bindables" and "Collections" (under "InteropServices").

So e.g. this file should be named "IRestrictedErrorInfomethods", and the class should be named the same.

src/WinRT.Runtime2/InteropServices/FreeThreadedMarshaler.cs
Sergio0694: Oh I just meant this piece, basically what I put in the snippet I shared 🙂

src/WinRT.Runtime2/InteropServices/ExceptionHelpers.cs
Sergio0694: That's correct, but my point is that extension methods are meant to add functionality that's conceptually "on the type" itself, just via the extension. It's for things that would always be in scope whenever the extension type is also in scope. Whereas here this `TryGetRestrictedLanguageErrorInfo` method is specific to the exception helpers logic, like for the other methods doing exception language stuff. So I'm saying it shouldn't just show up in IntelliSense on any exception anywhere, but rather it should be explicit at the callsite:

```csharp
// Not this
bool result = exception.TryGetRestrictedLanguageErrorInfo(...);

// But this, to be explicit
bool result = ExceptionHelpers.TryGetRestrictedLanguageErrorInfo(exception, ...);
```

kythant: Gotcha! I shall move away from using "This" in this instance.

src/WinRT.Runtime2/InteropServices/RestrictedErrorInfoMethods.cs
kythant: Gotcha! I shall make this change.

src/WinRT.Runtime2/Exceptions/InteropServices/RestrictedErrorInfoMethods.cs
```diff
@@ -0,0 +1,130 @@
...
+/// Provides low-level helpers to call <c>IRestrictedErrorInfo</c> methods through a vtable pointer,
+/// and to safely marshal returned BSTR values to managed strings.
+/// </summary>
+/// <remarks>
+/// These helpers assume "thisPtr" param points to a valid COM object implementing
+/// the <c>IRestrictedErrorInfo</c> interface and that HRESULTs are validated via the
+/// <c>Assert()</c> helper on the returned status code. Any BSTRs returned by the COM methods
+/// are converted to <see cref="string"/> and then released with <see cref="Marshal.FreeBSTR(System.IntPtr)"/>.
+/// </remarks>
+internal static class RestrictedErrorInfoMethods
```
Sergio0694: Wrong folder, this should be in `\InteropServices\Exceptions`, not `\Exceptions\InteropServices`.
You could move `ExceptionHelpers` to that same folder too while at it, so they're grouped by feature 🙂

Sergio0694 (changes requested):

src/WinRT.Runtime2/Exceptions/InteropServices/RestrictedErrorInfoMethods.cs
kythant: Whoops my bad. I misunderstood what you meant.

src/WinRT.Runtime2/InteropServices/RestrictedErrorInfo.cs
kythant: Oh so just realize, WindowsRuntimeObjectReference.CreateUnsafe is the one calling the QI. WindowsRuntimeObjectReference.Create is the one that does the AddRefUnsafe. 

I think calling WindowsRuntimeObjectReference.Create is the right call here?

src/WinRT.Runtime2/InteropServices/Exceptions/ExceptionHelpers.cs
```diff
@@ -0,0 +1,207 @@
...
+        WindowsRuntimeImports.GetRestrictedErrorInfo(&restrictedErrorInfoPtr).Assert();
+        if (restrictedErrorInfoPtr == null)
+        {
+            return default;
+        }
+
+        if (WindowsRuntimeImports.SetRestrictedErrorInfo != null)
+        {
+            WindowsRuntimeImports.SetRestrictedErrorInfo(restrictedErrorInfoPtr).Assert();
+        }
```
Sergio0694: This is another accidental comparison with an implicit natural type delegate, you can remove it:
```suggestion
        WindowsRuntimeImports.SetRestrictedErrorInfo(restrictedErrorInfoPtr).Assert();
```

src/WinRT.Runtime2/InteropServices/Exceptions/ExceptionHelpers.cs
```diff
@@ -0,0 +1,207 @@
...
+    /// <returns>
+    /// A <see cref="WindowsRuntimeObjectReferenceValue"/> representing the restricted error info object,
+    /// or <c>default</c> if none is available.
+    /// </returns>
+    internal static unsafe WindowsRuntimeObjectReferenceValue BorrowRestrictedErrorInfo()
+    {
+        if (WindowsRuntimeImports.GetRestrictedErrorInfo == null)
+        {
+            return default;
+        }
```
Sergio0694: This can never be null (Roslyn should really emit a warning here...), you can remove this check entirely:
```suggestion
```

src/WinRT.Runtime2/InteropServices/Exceptions/ExceptionHelpers.cs
```diff
@@ -0,0 +1,207 @@
...
+                {
+                    while (currentLanguageExceptionErrorInfo2Ptr != null)
+                    {
+                        Exception? propagatedException = GetLanguageExceptionInternal(currentLanguageExceptionErrorInfo2Ptr, hr);
+                        if (propagatedException is not null)
+                        {
+                            return propagatedException;
+                        }
+
+                        void* previousLanguageExceptionErrorInfo2Ptr = currentLanguageExceptionErrorInfo2Ptr;
```
Sergio0694, line 85: Reassigning raw pointers like this is extremely dangerous, especially when we're also releasing them from `finally` blocks etc. (e.g. it's easy to accidentally release them twice or stuff like that). Let's just stick to [SSA form](https://en.wikipedia.org/wiki/Static_single-assignment_form) and not reassign `currentLanguageExceptionErrorInfo2Ptr` here.

That is:
- Leave `previousLanguageExceptionErrorInfo2Ptr` set to `null`
- Pass that as parameter to `GetPreviousLanguageExceptionErrorInfoUnsafe` (matching the parameter name)

src/WinRT.Runtime2/InteropServices/Exceptions/ExceptionHelpers.cs
```diff
@@ -0,0 +1,207 @@
...
+
+    /// <summary>
+    /// Internal helper to retrieve a managed language exception from a COM pointer.
+    /// </summary>
+    /// <param name="languageErrorInfoPtr">Pointer to the language error info COM object.</param>
+    /// <param name="hr">The HRESULT associated with the error.</param>
+    /// <returns>
+    /// The managed <see cref="Exception"/> if found; otherwise, <c>null</c>.
+    /// </returns>
+    internal static unsafe Exception? GetLanguageExceptionInternal(void* languageErrorInfoPtr, HRESULT hr)
```
Sergio0694: Don't use abbreviations
```suggestion
    internal static unsafe Exception? GetLanguageExceptionInternal(void* languageErrorInfoPtr, HRESULT hresult)
```

src/WinRT.Runtime2/InteropServices/Exceptions/ExceptionHelpers.cs
```diff
@@ -0,0 +1,207 @@
...
+        }
+
+        void* languageExceptionPtr = null;
+        try
+        {
+            if (WellKnownErrorCodes.Succeeded(ILanguageExceptionErrorInfoVftbl.GetLanguageExceptionUnsafe(languageErrorInfoPtr, &languageExceptionPtr)))
+            {
+                if (WindowsRuntimeMarshal.TryGetManagedObject(languageExceptionPtr, out object? exception))
+                {
+                    if (RestrictedErrorInfo.GetHRForException((Exception)exception) == hr)
```
Sergio0694: Declare an `Exception` local here to assign from the cast and reuse it instead of casting twice.

src/WinRT.Runtime2/InteropServices/Exceptions/ExceptionHelpers.cs
```diff
@@ -0,0 +1,207 @@
...
+                    }
+                }
+            }
+        }
+        finally
+        {
+            if (languageExceptionPtr != null)
+            {
+                WindowsRuntimeObjectMarshaller.Free(languageExceptionPtr);
+            }
```
Sergio0694: Unnecessary check, `WindowsRuntimeObjectMarshaller.Free` already handles this

src/WinRT.Runtime2/InteropServices/Exceptions/IRestrictedErrorInfoMethods.cs
```diff
@@ -0,0 +1,130 @@
...
+                    thisPtr,
+                    &_description,
+                    pError,
+                    &_restrictedDescription,
+                    &_capabilitySid).Assert();
+            }
+
+            description = _description != null ? Marshal.PtrToStringBSTR((nint)_description) : string.Empty;
+            restrictedDescription = _restrictedDescription != null ? Marshal.PtrToStringBSTR((nint)_restrictedDescription) : string.Empty;
+            capabilitySid = _capabilitySid != null ? Marshal.PtrToStringBSTR((nint)_capabilitySid) : string.Empty;
```
Sergio0694: We can avoid this dance and make this simpler by using the new marshaller, same below too to free:
```csharp
description = BStrStringMarshaller.ConvertToManaged(_description);
```
We'll need to make the out parameters nullable as well.

Sergio0694 (changes requested):

src/WinRT.Runtime2/InteropServices/Exceptions/ExceptionHelpers.cs
kythant: Ah right we can guarantee that it's there now.

src/WinRT.Runtime2/InteropServices/RestrictedErrorInfo.cs
Sergio0694: Oh, uhhh... I think I accidentally implemented those two backwards, whoops 😅
You can switch to `CreateUnsafe` here, I'll fix them in my other PR. Thank you for double checking!

src/WinRT.Runtime2/InteropServices/RestrictedErrorInfo.cs
```diff
@@ -28,10 +34,151 @@ public static unsafe class RestrictedErrorInfo
...
+                    {
+                        WindowsRuntimeObjectMarshaller.Free(languageErrorInfoPtr);
+                    }
+                }
+
+                WindowsRuntimeObjectReference? restrictedErrorInfoToSave = WindowsRuntimeObjectReference.Create(restrictedErrorInfoValuePtr, WellKnownInterfaceIds.IID_IRestrictedErrorInfo);
+
+                IRestrictedErrorInfoMethods.GetErrorDetails(restrictedErrorInfoValuePtr, out string? description, out HRESULT hrLocal, out string? restrictedError, out string? restrictedCapabilitySid);
+
+                string restrictedErrorReference = IRestrictedErrorInfoMethods.GetReference(restrictedErrorInfoValuePtr);
```
Sergio0694: This is unused. Did you mean to use it anywhere? If not, we should remove this and delete that method too.

kythant: Oh wait in the original code, it's being used at the end like this

image

manodasanW (approved):

PR #2095, opened 2025-10-28
Rework public marshalling APIs

Sergio0694:
This PR makes some changes to the public API surface for marshallers:
- Marks all `WindowsRuntimeObjectReference*` types as private implementation details
- Marks all marshallers as private implementation details
- Adds new public marshalling primitives to `WindowsRuntimeMarshal`

### DIFF
--- a/src/WinRT.Runtime2/InteropServices/Activation/WindowsRuntimeActivationFactory.cs
+++ b/src/WinRT.Runtime2/InteropServices/Activation/WindowsRuntimeActivationFactory.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using System.ComponentModel;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Runtime.CompilerServices;
@@ -20,7 +21,7 @@ public static unsafe class WindowsRuntimeActivationFactory
     /// <summary>
     /// The registered hook into activation to hook/mock activation of Windows Runtime types.
     /// </summary>
-    private static WindowsRuntimeActivationHandler? activationHandler;
+    private static volatile WindowsRuntimeActivationHandler? activationHandler;
 
     /// <summary>
     /// Set the <see cref="WindowsRuntimeActivationHandler"/> callback for activating Windows Runtime types.
@@ -70,6 +71,10 @@ public static unsafe class WindowsRuntimeActivationFactory
     /// </remarks>
     /// <exception cref="NotSupportedException">Thrown if <paramref name="runtimeClassName"/> is not registered, <c>CsWinRTEnableManifestFreeActivation</c> is disabled, and <c>CsWinRTManifestFreeActivationReportOriginalException</c> is not set.</exception>
     /// <exception cref="Exception">Thrown for any failure to activate the specified type (the exact exception type might be a derived type).</exception>
+    [Obsolete(WindowsRuntimeConstants.PrivateImplementationDetailObsoleteMessage,
+        DiagnosticId = WindowsRuntimeConstants.PrivateImplementationDetailObsoleteDiagnosticId,
+        UrlFormat = WindowsRuntimeConstants.CsWinRTDiagnosticsUrlFormat)]
+    [EditorBrowsable(EditorBrowsableState.Never)]
     public static WindowsRuntimeObjectReference GetActivationFactory(string runtimeClassName)
     {
         void* activationFactory = GetActivationFactoryUnsafe(runtimeClassName);
@@ -79,6 +84,10 @@ public static unsafe class WindowsRuntimeActivationFactory
 
     /// <param name="iid">The IID of the interface pointer (from the resolved activation factory) to wrap in the returned object reference.</param>
     /// <inheritdoc cref="GetActivationFactory(string)"/>
+    [Obsolete(WindowsRuntimeConstants.PrivateImplementationDetailObsoleteMessage,
+        DiagnosticId = WindowsRuntimeConstants.PrivateImplementationDetailObsoleteDiagnosticId,
+        UrlFormat = WindowsRuntimeConstants.CsWinRTDiagnosticsUrlFormat)]
+    [EditorBrowsable(EditorBrowsableState.Never)]
     public static WindowsRuntimeObjectReference GetActivationFactory(string runtimeClassName, in Guid iid)
     {
         void* activationFactory = GetActivationFactoryUnsafe(runtimeClassName, in iid);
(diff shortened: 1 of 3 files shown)


src/WinRT.Runtime2/InteropServices/WindowsRuntimeMarshal.cs
```diff
@@ -89,42 +89,108 @@ public static bool TryGetManagedObject(void* externalComObject, [NotNullWhen(tru
...
-    /// Attempts to extract a <see cref="WindowsRuntimeObjectReference"/> from the specified object.
+    /// Tries to retrieve a native object from a managed object, if it is actually a wrapper of some native object.
     /// </summary>
-    /// <param name="value">The object to attempt to unwrap.</param>
-    /// <param name="objectReference">The unwrapped <see cref="WindowsRuntimeObjectReference"/> object, if successfully retrieved.</param>
-    /// <returns>Whether <paramref name="objectReference"/> was successfully unwrapped.</returns>
+    /// <param name="managedObject">The managed object to try to get a native object from.</param>
+    /// <param name="result">The resulting native object, if successfully retrieved.</param>
+    /// <returns>Whether <paramref name="managedObject"/> was a reference to a native object, and <paramref name="result"/> could be retrieved.</returns>
+    public static bool TryGetNativeObject(object? managedObject, out void* result)
```
dongle-the-gadget, line 97: This API seems redundant, given that `ConvertToUnmanaged` exists.

src/WinRT.Runtime2/InteropServices/WindowsRuntimeMarshal.cs
```diff
@@ -89,42 +89,108 @@ public static bool TryGetManagedObject(void* externalComObject, [NotNullWhen(tru
...
+    /// <seealso cref="System.Runtime.InteropServices.Marshalling.ComInterfaceMarshaller{T}.ConvertToManaged"/>
+    public static object? ConvertToManaged(void* value)
+    {
+        if (value is null)
+        {
+            return null;
+        }
+
+        // If the value is a CCW we recognize, just unwrap it directly
+        if (IsReferenceToManagedObject(value))
```
dongle-the-gadget: Is there a reason why Unwrap is not feasible here?

src/WinRT.Runtime2/InteropServices/WindowsRuntimeMarshal.cs
Sergio0694: They're not the same:
- `TryGetNativeObject` will only unwrap an RCW
- `ConvertToUnmanaged` will also create a CCW for a managed object if needed

src/WinRT.Runtime2/InteropServices/WindowsRuntimeMarshal.cs
Sergio0694: What do you mean?

dongle-the-gadget: Can't you pass in `CreateObjectFlags.Unwrap` in order to make it unwrap automatically instead of having your own check like this?

src/WinRT.Runtime2/InteropServices/WindowsRuntimeMarshal.cs
dongle-the-gadget: Is there a scenario where you wouldn't want a CCW creation?

Sergio0694: ComWrappers also offers a similar extension point, plus it's consistent with `TryGetManagedObject`.

src/WinRT.Runtime2/InteropServices/WindowsRuntimeMarshal.cs
Sergio0694: We looked into this a while back and we concluded it wasn't applicable for us. I think it was about perf.

src/WinRT.Runtime2/InteropServices/Activation/WindowsRuntimeActivationFactory.cs
```diff
@@ -79,6 +84,10 @@ public static WindowsRuntimeObjectReference GetActivationFactory(string runtimeC
 
     /// <param name="iid">The IID of the interface pointer (from the resolved activation factory) to wrap in the returned object reference.</param>
     /// <inheritdoc cref="GetActivationFactory(string)"/>
+    [Obsolete(WindowsRuntimeConstants.PrivateImplementationDetailObsoleteMessage,
+        DiagnosticId = WindowsRuntimeConstants.PrivateImplementationDetailObsoleteDiagnosticId,
+        UrlFormat = WindowsRuntimeConstants.CsWinRTDiagnosticsUrlFormat)]
+    [EditorBrowsable(EditorBrowsableState.Never)]
     public static WindowsRuntimeObjectReference GetActivationFactory(string runtimeClassName, in Guid iid)
```
manodasanW, line 91: We will probably need a solution for interop interfaces that someone may want to QI for and use, but we can address that later with a higher level wrapper.

src/WinRT.Runtime2/InteropServices/ObjectReference/WindowsRuntimeObjectReference.Factory.cs
```diff
@@ -17,39 +17,38 @@ public unsafe partial class WindowsRuntimeObjectReference
...
-    /// already points to the interface represented by <paramref name="iid"/>. It is responsibility of the caller to respect this invariant.
+    /// This method will perform a <c>QueryInterface</c> call on <paramref name="thisPtr"/> to retrieve the requested interface pointer.
     /// </para>
     /// <para>
     /// The resulting <see cref="WindowsRuntimeObjectReference"/> is <see langword="null"/> if <paramref name="thisPtr"/> is <see langword="null"/>.
     /// </para>
     /// </remarks>
-    public static WindowsRuntimeObjectReference? Create(void* thisPtr, in Guid iid)
+    /// <exception cref="Exception">Thrown if the <c>QueryInterface</c> operation fails.</exception>
+    internal static WindowsRuntimeObjectReference? Create(void* thisPtr, in Guid iid)
```
manodasanW, line 27: Given we are switching which does the QI, we will need to go through all uses and make sure the assumptions they called it with holds.

src/WinRT.Runtime2/InteropServices/Activation/WindowsRuntimeActivationFactory.cs
Sergio0694: Yeah for now it's usable already, but callers will need to get the pointer and then marshal with `ComWrappers` on their end. We can definitely think of some higher level API where we'd handle that part ourselves. Like idk, `GetActivationFactoryInterface<T>` or something, we can do that later, agreed 🙂

src/WinRT.Runtime2/InteropServices/ObjectReference/WindowsRuntimeObjectReference.Factory.cs
Sergio0694: I did do a pass and basically we're not really using these anywhere now, except for one use for `IAgileReference` which I did check, and the ones that Will added in his PR, which are also correct. Everywhere else we're now only going through `WindowsRuntimeComWrappersMarshal`, which has its own implementation.

PR #2096, opened 2025-10-28
Project classes implementing generics and Project arrays

manodasanW:
- Added support for projecting classes implementing generics
- Added support for projecting arrays
- Ported all the additions to the new format.
- To avoid having them as all manual projections, I added support for partial additions which basically cswinrt will generate the type with partial and we will add the addition as another partial type.  This is mostly for types which just had a custom ToString that they wanted to add.
- Enabled all the projections and address issues encountered.
- Moved most event sources to respective ABI namespace to avoid hard coding in cswinrt.exe

### DIFF
--- a/src/WinRT.Runtime2/InteropServices/Collections/IObservableMapMethodsImpl{TKey, TValue}.cs
+++ b/src/WinRT.Runtime2/InteropServices/Collections/IObservableMapMethodsImpl{TKey, TValue}.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.ComponentModel;
-using ABI.Windows.Foundation.Collections;
 
 namespace WindowsRuntime.InteropServices;
 
@@ -25,5 +24,5 @@ public interface IObservableMapMethodsImpl<TKey, TValue>
     /// <param name="thisReference">The <see cref="WindowsRuntimeObjectReference"/> instance to use to invoke the native method.</param>
     /// <returns>The <see cref="EventSource{T}"/> instance associated with <see cref="Windows.Foundation.Collections.IObservableMap{K, V}.MapChanged"/>.</returns>
     /// <see href="https://learn.microsoft.com/uwp/api/windows.foundation.collections.iobservablemap-2.mapchanged"/>
-    static abstract MapChangedEventHandlerEventSource<TKey, TValue> MapChanged(WindowsRuntimeObject thisObject, WindowsRuntimeObjectReference thisReference);
+    static abstract global::ABI.Windows.Foundation.Collections.MapChangedEventHandlerEventSource<TKey, TValue> MapChanged(WindowsRuntimeObject thisObject, WindowsRuntimeObjectReference thisReference);
 }
--- a/src/WinRT.Runtime2/InteropServices/Collections/IObservableVectorMethodsImpl{T}.cs
+++ b/src/WinRT.Runtime2/InteropServices/Collections/IObservableVectorMethodsImpl{T}.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.ComponentModel;
-using ABI.Windows.Foundation.Collections;
 
 namespace WindowsRuntime.InteropServices;
 
@@ -24,5 +23,5 @@ public interface IObservableVectorMethodsImpl<T>
     /// <param name="thisReference">The <see cref="WindowsRuntimeObjectReference"/> instance to use to invoke the native method.</param>
     /// <returns>The <see cref="EventSource{T}"/> instance associated with <see cref="Windows.Foundation.Collections.IObservableVector{T}.VectorChanged"/>.</returns>
     /// <see href="https://learn.microsoft.com/uwp/api/windows.foundation.collections.iobservablevector-1.vectorchanged"/>
-    static abstract VectorChangedEventHandlerEventSource<T> VectorChanged(WindowsRuntimeObject thisObject, WindowsRuntimeObjectReference thisReference);
+    static abstract global::ABI.Windows.Foundation.Collections.VectorChangedEventHandlerEventSource<T> VectorChanged(WindowsRuntimeObject thisObject, WindowsRuntimeObjectReference thisReference);
 }


src/Projections/Windows.UI.Xaml/ABI.Windows.System/DispatcherQueueProxyHandler.cs
```diff
@@ -144,14 +143,64 @@ private static ref readonly Guid IID_IDispatcherQueueHandler
             }
         }
 
+        /// <summary>The IID for <c>IUnknown</c> (00000000-0000-0000-C000-000000000046).</summary>
+        private static ref readonly Guid IID_IUnknown
```
Sergio0694: We should be able to use `WindowsRuntime.InteropServices.WellKnownInterfaceIIDs` for these two IIDs, right?

src/cswinrt/strings/additions/Windows.Storage.Streams/WellKnownStreamInterfaceIIDs.cs
```diff
@@ -0,0 +1,45 @@
+namespace ABI.Windows.Storage.Streams
+{
+    internal static class WellKnownStreamInterfaceIIDs
+    {
+        public static ref readonly Guid IID_IBufferByteAccess
```
Sergio0694: All these are already in `WellKnownWindowsInterfaceIIDs`. We should add the missing ones to `WellKnownInterfaceIIDs` (e.g. `IID_IBufferByteAccess`), and then we can delete these here and just use the ones from the runtime instead.

src/WinRT.Runtime2/InteropServices/ProjectionDllExports/TokenizerHelper.cs
```diff
@@ -0,0 +1,39 @@
...
+namespace ABI.Windows.Foundation;
+
+/// <summary>
+/// Helper class for tokenizing used by generated projections.
+/// </summary>
+[Obsolete(WindowsRuntimeConstants.PrivateImplementationDetailObsoleteMessage,
+    DiagnosticId = WindowsRuntimeConstants.PrivateImplementationDetailObsoleteDiagnosticId,
+    UrlFormat = WindowsRuntimeConstants.CsWinRTDiagnosticsUrlFormat)]
+[EditorBrowsable(EditorBrowsableState.Never)]
+public static class TokenizerHelper
```
Sergio0694, line 17: I'm not opposed to having some shared helpers for projections, but I don't think this is the right folder/namespace for them. We already have "\InteropServices\InteropDllExports". Can we move this to a new "\InteropServices\ProjectionDllExports" for consistency and just use "WindowsRuntime.InteropServices" as the namespace? We can reuse the same place for any future helpers for projections too.

src/cswinrt/strings/additions/Microsoft.UI.Xaml.Controls.Primitives/Microsoft.UI.Xaml.Controls.Primitives.GeneratorPosition.cs
```diff
@@ -0,0 +1,13 @@
+
+namespace Microsoft.UI.Xaml.Controls.Primitives
+{
+    using global::Windows.Foundation;
+
+    partial struct GeneratorPosition
+    {
+        public override string ToString()
+        {
+            return string.Concat("GeneratorPosition (", Index.ToString(global::System.Globalization.CultureInfo.InvariantCulture), ",", Offset.ToString(global::System.Globalization.CultureInfo.InvariantCulture), ")");
```
Sergio0694: We can optimize this with an interpolated string builder. I'll do this in a separate PR, want to experiment with exposing a custom handler from the runtime so that we can reuse it in all these places where we just always do invariant culture formatting.

src/cswinrt/strings/additions/Microsoft.UI.Xaml.Media.Animation/Microsoft.UI.Xaml.Media.Animation.RepeatBehavior.cs
```diff
@@ -0,0 +1,186 @@
...
+                case RepeatBehaviorType.Count:
+
+                    global::System.Text.StringBuilder sb = new global::System.Text.StringBuilder();
+
+                    sb.AppendFormat(
+                        formatProvider,
+                        "{0:" + format + "}x",
+                        Count);
+
+                    return sb.ToString();
```
Sergio0694: Should be able to optimize this:
```suggestion
                    DefaultInterpolatedStringHandler handler = new(0, 1, formatProvider, stackalloc char[64]);

                    handler.AppendFormatted(Count, format);

                    return handler.ToStringAndClear();
```

src/cswinrt/strings/additions/Microsoft.UI.Xaml.Media.Animation/Microsoft.UI.Xaml.Media.Animation.RepeatBehavior.cs
```diff
@@ -0,0 +1,186 @@
...
+        {
+            switch (Type)
+            {
+                case RepeatBehaviorType.Forever:
+
+                    return "Forever";
+
+                case RepeatBehaviorType.Count:
+
+                    global::System.Text.StringBuilder sb = new global::System.Text.StringBuilder();
```
Sergio0694: Also we should set `SkipLocalsInit` for all projection .dlls if we don't already. At least the merged .dll, as we own it.

src/cswinrt/strings/additions/Microsoft.UI.Xaml.Media.Media3D/Microsoft.UI.Xaml.Media.Media3D.Matrix3D.cs
```diff
@@ -418,7 +411,7 @@ private string ConvertToString(string format, IFormatProvider provider)
             }
 
             // Helper to get the numeric list separator for a given culture.
-            char separator = TokenizerHelper.GetNumericListSeparator(provider);
+            char separator = global::ABI.Windows.Foundation.TokenizerHelper.GetNumericListSeparator(provider);
             return string.Format(provider,
```
Sergio0694: Use `DefaultInterpolatedStringHandler` here too. Basically everywhere we do `string.Format` and similar.

src/cswinrt/strings/additions/Microsoft.UI.Xaml.Media/Microsoft.UI.Xaml.Media.Matrix.cs
```diff
@@ -0,0 +1,105 @@
...
+namespace Microsoft.UI.Xaml.Media
+{
+    using global::Windows.Foundation;
+
+    partial struct Matrix : IFormattable
+    {
+        // the transform is identity by default
+        private static Matrix s_identity = CreateIdentity();
+
+        public static Matrix Identity
```
Sergio0694, line 11: @tannergooding is there something smart we could do here like creating a BCL matrix and reinterpreting it or something, rather than storing this value in a static field like this?

src/cswinrt/code_writers.h
```diff
@@ -2754,20 +2922,27 @@ IEnumerator IEnumerable.GetEnumerator() => %.GetEnumerator();
         auto element = w.write_temp("%", bind<write_generic_type_name>(0));
         auto self = emit_explicit ? w.write_temp("global::System.Collections.Generic.IEnumerable<%>.", element) : "";
         auto visibility = emit_explicit ? "" : "public ";
-        auto abiClass = w.write_temp("global::ABI.System.Collections.Generic.IEnumerableMethods<%>", element);
+        auto interop_method_name_prefix = w.write_temp("IEnumerableMethods_%_", escape_type_name_for_identifier(element, true));
 
         w.write(R"(
-%IEnumerator<%> %GetEnumerator() => %.GetEnumerator(%);
+[UnsafeAccessor(UnsafeAccessorKind.StaticMethod, Name = "GetEnumerator")]
+static extern IEnumerator<%> %GetEnumerator([UnsafeAccessorType("ABI.System.Collections.Generic.<#CsWinRT>IEnumerableMethods`1<<#corlib>System-ComponentModel-DataErrorsChangedEventArgs>, WinRT.Interop.dll")] object _, WindowsRuntimeObjectReference objRef);
```
Sergio0694: Generic mangled names use `'` for the generic arity prefix, not the backtick (we changed it as it broke ILSpy) 😄

Sergio0694 (changes requested):

src/cswinrt/strings/additions/Microsoft.UI.Xaml.Media.Media3D/Microsoft.UI.Xaml.Media.Media3D.Matrix3D.cs
```diff
@@ -411,16 +411,40 @@ private string ConvertToString(string format, IFormatProvider provider)
...
-                                 "{1:" + format + "}{0}{2:" + format + "}{0}{3:" + format + "}{0}{4:" + format + "}{0}{5:" + format +
-                                 "}{0}{6:" + format + "}{0}{7:" + format + "}{0}{8:" + format + "}{0}{9:" + format + "}{0}{10:" + format +
-                                 "}{0}{11:" + format + "}{0}{12:" + format + "}{0}{13:" + format + "}{0}{14:" + format + "}{0}{15:" + format + "}{0}{16:" + format + "}",
-                                 separator,
-                                 _m11, _m12, _m13, _m14,
-                                 _m21, _m22, _m23, _m24,
-                                 _m31, _m32, _m33, _m34,
-                                 _offsetX, _offsetY, _offsetZ, _m44);
+            char separator = global::WindowsRuntime.InteropServices.TokenizerHelper.GetNumericListSeparator(provider);
+            DefaultInterpolatedStringHandler handler = new(0, 31, provider, stackalloc char[64]);
```
Sergio0694: Not entirely sure 64 is enough, can we double check what's the average length of this text and adjust the scratch buffer? If it's 256 or less we can use that, otherwise we can skip it entirely if it won't be enough anyway, to avoid the extra copy.

src/cswinrt/strings/additions/Microsoft.UI.Xaml.Media/Microsoft.UI.Xaml.Media.Matrix.cs
```diff
@@ -50,16 +50,20 @@ private string ConvertToString(string format, IFormatProvider provider)
...
-                                 "{1:" + format + "}{0}{2:" + format + "}{0}{3:" + format + "}{0}{4:" + format + "}{0}{5:" + format + "}{0}{6:" + format + "}",
-                                 separator,
-                                 M11,
-                                 M12,
-                                 M21,
-                                 M22,
-                                 OffsetX,
-                                 OffsetY);
+            char separator = global::WindowsRuntime.InteropServices.TokenizerHelper.GetNumericListSeparator(provider);
+            DefaultInterpolatedStringHandler handler = new(0, 11, provider, stackalloc char[64]);
```
Sergio0694, line 54: Same here as above, we might want to tweak the scratch buffer size.

src/cswinrt/strings/additions/Microsoft.UI.Xaml.Media.Animation/Microsoft.UI.Xaml.Media.Animation.RepeatBehavior.cs
manodasanW: Will handler later once we have merged dll

src/cswinrt/strings/additions/Microsoft.UI.Xaml.Media.Media3D/Microsoft.UI.Xaml.Media.Media3D.Matrix3D.cs
manodasanW: Increased this one.

src/cswinrt/strings/additions/Microsoft.UI.Xaml.Media/Microsoft.UI.Xaml.Media.Matrix.cs
manodasanW: Looking at typical values, I think this is fine for common scenarios.

src/WinRT.Runtime2/InteropServices/RestrictedErrorInfo.cs
```diff
@@ -75,6 +75,17 @@ public static void SetErrorInfo(Exception exception)
         // TODO
     }
 
+    /// <summary>
+    /// Attaches the error info stored by the <c>IRestrictedErrorInfo</c> infrastructure to the input exception.
+    /// </summary>
+    /// <param name="exception">The input <see cref="Exception"/> instance to attach the error info to.</param>
+    /// <returns>The input <see cref="Exception"/> instance with attached error info.</returns>
+    public static Exception AttachErrorInfo(Exception exception)
```
Sergio0694, line 83: Is this only meant to be used by projections? I'm thinking if so we might want to mark this one and the other ones in `RestrictedErrorInfo` that are not for general use as private implementation detail methods (or move them).

src/WinRT.Runtime2/InteropServices/Collections/IObservableMapMethodsImpl{TKey, TValue}.cs
```diff
@@ -3,6 +3,7 @@
 
 using System;
 using System.ComponentModel;
+using ABI.Windows.Foundation.Collections;
```
Sergio0694: Nit: don't add `ABI.*` to usings, as it can create ambiguties and whatnot (we've removed all of them in another PR). If you need to refer to an ABI type from a non-ABI namespace, use the fully qualified type name instead in each location.

src/WinRT.Runtime2/InteropServices/Collections/IObservableVectorMethodsImpl{T}.cs
```diff
@@ -3,6 +3,7 @@
 
 using System;
 using System.ComponentModel;
+using ABI.Windows.Foundation.Collections;
```
Sergio0694: Same here (and for any other place where we have `using ABI.*`).

src/WinRT.Runtime2/InteropServices/Collections/IObservableMapMethodsImpl{TKey, TValue}.cs
manodasanW: Addressed both.

src/WinRT.Runtime2/InteropServices/RestrictedErrorInfo.cs
manodasanW: Will address in upcoming PR.

Issue #2097, opened 2025-11-01
Bug: IReference<TypeName> mapped to Nullable<System.Type> causes compile error


lgztx96:
### Description

When generating C# projection from a WinMD that contains a property of type
`Windows.Foundation.IReference<Windows.UI.Xaml.Interop.TypeName>`,
CsWinRT generates:

```csharp
public global::System.Nullable<global::System.Type> TargetType { get; set; }
```
<img width="1235" height="192" alt="Image" src="https://github.com/user-attachments/assets/2d4d9b59-d063-4a06-811e-cbbdfb878704" />
This causes a compilation error:

The type 'Type' must be a non-nullable value type in order to use it as parameter 'T' in the generic type or method 'System.Nullable<T>'

### Steps To Reproduce

A minimal IDL to reproduce:
```
namespace Example
{
    runtimeclass TestClass
    {
        Windows.Foundation.IReference<Windows.UI.Xaml.Interop.TypeName> TargetType;
    }
}

```

### Expected Behavior

CsWinRT should generate the property as:

```csharp
public global::System.Type TargetType { get; set; }

### Version Info

Microsoft.Windows.CsWinRT 2.2.0

### Additional Context

_No response_

PR #2103, opened 2025-11-05
Fix IReference for XAML TypeName

dongle-the-gadget:
`XAML.Interop.TypeName` presents a unique challenge that the original type is a struct, but the projected type is a class (`System.Type`). Hence, `Nullable<TypeName>` is valid in WinRT context but not in C#.

This PR fixes this by special casing `IReference` to use `T?` syntax in C#, which works for both value types and reference types.

### DIFF
--- a/src/cswinrt/code_writers.h
+++ b/src/cswinrt/code_writers.h
@@ -451,9 +451,21 @@ namespace cswinrt
             [&](generic_type_instance const& type)
             {
                 auto guard{ w.push_generic_args(type) };
-                w.write("%<%>",
-                    bind<write_projection_type_for_name_type>(type.generic_type, nameType),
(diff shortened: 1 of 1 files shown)


dongle-the-gadget:
Fixes #2097

manodasanW:
Thanks for the fix.  We probably want to add a property that gets / sets a IReference<Type>.  I expect it is going to error about nullable annotations given we don't enable it for the entire projection currently as we haven't done anything to say these reference parameters are nullable vs non nullable.  So I think we might want a scoped nullable enable / disable for this parameter when it is not a value type.

It probably also makes sense to port this fix to master too but I would just project it as just a type without nullable annotations enabled.

Sergio0694:
Closing this one, #2461 fixed it properly for 3.0, and we don't plan on taking this change in 2.x. Thank you! 🙂

PR #2104, opened 2025-11-06
Fix boxing scenarios calling cswinrtgen and update to new factory pattern

manodasanW:
- Boxing scenarios where previously going through the code path to call cswinrtgen functions due to nullable being seen as a generic type.  But we handle those differently given we have unbox / box functions in the individual marshal classes.  So, updating to detect and handle that.
- Also updating the factory pattern with constructors to no longer pass a function as a delegate but rather pass an instance of a sealed read only class.  This seems to improve size on AOT.

### DIFF
--- a/src/cswinrt/code_writers.h
+++ b/src/cswinrt/code_writers.h
@@ -2698,7 +2698,7 @@ private static class _%
 
             w.write(R"(
 %public unsafe %(%)
-  :base(%, %, [%])
+  :base(%.Instance, %, [%])
 {
 %}
 )",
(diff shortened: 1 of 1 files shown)


src/cswinrt/code_writers.h
```diff
@@ -7219,7 +7285,29 @@ CopyToUnmanaged_%(null, __%, __%Size, (void**)%);
                     m.local_type = m.param_type;
                     m.interface_guid = w.write_temp("%", bind<write_iid_guid>(type));
                     m.marshal_by_object_reference_value = true;
-                    if (distance(type.GenericParam()) > 0)
+                    if (is_projected_as_nullable(type))
+                    {
+                        m.is_boxed_value = true;
+                        auto generic_arg = w.get_generic_arg(0);
+                        call(generic_arg,
```
Sergio0694, line 7292: This logic to get the right marshaller seems the same above, I wonder if it should be some shared helper method?

manodasanW: It is basically the same right now for this.  I did think about moving this elsewhere but kept it separate for now as we do the managed marshaler and abi marshaler decisions separately today.

PR #2112, opened 2025-11-12
Add source generator for `[TypeMapAssemblyTarget]` items

Sergio0694:
Title.

### DIFF
--- a/src/Authoring/WinRT.SourceGenerator2/TypeMapAssemblyTargetGenerator.Execute.cs
+++ b/src/Authoring/WinRT.SourceGenerator2/TypeMapAssemblyTargetGenerator.Execute.cs
@@ -2,7 +2,6 @@
 // Licensed under the MIT License.
 
 using System.Collections.Immutable;
-using System.Linq;
 using System.Text;
 using System.Threading;
 using Microsoft.CodeAnalysis;


src/Authoring/WinRT.SourceGenerator2/TypeMapAssemblyTargetGenerator.Execute.cs
```diff
@@ -0,0 +1,131 @@
...
+        /// <returns>All portable executable references for <paramref name="compilation"/>.</returns>
+        public static ImmutableArray<EquatablePortableExecutableReference> GetAllPortableExecutableReferences(Compilation compilation, CancellationToken token)
+        {
+            var executableReferences = ImmutableArray.CreateBuilder<EquatablePortableExecutableReference>();
+
+            foreach (MetadataReference metadataReference in compilation.References)
+            {
+                token.ThrowIfCancellationRequested();
+
+                // We are only interested in PE references (not project references)
```
333fred, line 41: When running on the command line, there are no project references, only PE references.

src/Authoring/WinRT.SourceGenerator2/Helpers/EquatablePortableExecutableReference.cs
```diff
@@ -0,0 +1,62 @@
...
+
+    /// <inheritdoc/>
+    public override bool Equals(object? other)
+    {
+        return Equals(other as EquatablePortableExecutableReference);
+    }
+
+    /// <inheritdoc/>
+    public bool Equals([NotNullWhen(true)] EquatablePortableExecutableReference? other)
+    {
```
manodasanW, line 53: There was an open issue on this, this will need a try catch here I believe as GetMetadataId can apaprently throw.

Sergio0694: Oh, interesting. Do we want to leave this for later? Since this matches what we already have in 2.x anyway 🤔

manodasanW: I plan on fixing it in 2.x, so probably should fix here too.

src/Authoring/WinRT.SourceGenerator2/TypeMapAssemblyTargetGenerator.Execute.cs
```diff
@@ -0,0 +1,131 @@
...
+
+            // We only care about resolved assembly symbols (this should always be the case anyway)
+            if (compilation.GetAssemblyOrModuleSymbol(executableReference.Value.Reference) is not IAssemblySymbol assemblySymbol)
+            {
+                return null;
+            }
+
+            token.ThrowIfCancellationRequested();
+
+            ITypeSymbol attributeSymbol = compilation.GetTypeByMetadataName("WindowsRuntime.InteropServices.WindowsRuntimeReferenceAssemblyAttribute")!;
```
333fred, line 75: This will return `null` if multiple references have this attribute. Is this something you need to handle?

src/Authoring/WinRT.SourceGenerator2/TypeMapAssemblyTargetGenerator.cs
```diff
@@ -0,0 +1,80 @@
...
+        // Get whether the generator should actually run or not
+        IncrementalValueProvider<bool> isGeneratorEnabled =
+            isOutputTypeExe
+            .Combine(isPublishAotLibrary)
+            .Select(static (flags, token) => flags.Left || flags.Right);
+
+        // Bypass all items if the flag is not set
+        IncrementalValuesProvider<(EquatablePortableExecutableReference Value, bool)> enabledExecutableReferences =
+            executableReferences
+            .Combine(isGeneratorEnabled)
```
333fred, line 53: I'm concerned about this `Combine`. In theory, if it turned to `true` while the references didn't change, it could lead to a situation where `Compilation` from the reference was a prior run, and the reference is dead. To make it safe, you'd have to thread through a weak reference to the compilation from all paths involved.

Sergio0694: I've reworked this completely to remove caching PE references entirely. Now I'm just gathering all assembly names directly, so after that there's no concerns about keeping compilations alive. Should also be more efficient, as I've removed a whole bunch of incremental steps. Let me know what you think of the new pipeline! 🙂

src/Authoring/WinRT.SourceGenerator2/TypeMapAssemblyTargetGenerator.Execute.cs
Sergio0694: We should be able to rely on this always being present exactly once. The attribute is defined by CsWinRT itself.

src/Authoring/WinRT.SourceGenerator2/TypeMapAssemblyTargetGenerator.Execute.cs
```diff
@@ -19,66 +21,66 @@ public partial class TypeMapAssemblyTargetGenerator : IIncrementalGenerator
...
+            // Bypass the entire operation if the generator is disabled
+            if (!data.IsEnabled)
+            {
+                return [];
+            }
+
+            ITypeSymbol attributeSymbol = data.Compilation.GetTypeByMetadataName("WindowsRuntime.InteropServices.WindowsRuntimeReferenceAssemblyAttribute")!;
+
+            // Try to get a good initial capacity (if we fail, just use '0', since private projections are not common)
+            if (!Enumerable.TryGetNonEnumeratedCount(data.Compilation.References, out int initialCapacity))
```
333fred: `TryGetNonEnumeratedCount` will never return true.

src/Authoring/WinRT.SourceGenerator2/TypeMapAssemblyTargetGenerator.Execute.cs
```diff
@@ -19,66 +21,66 @@ public partial class TypeMapAssemblyTargetGenerator : IIncrementalGenerator
...
+            {
+                initialCapacity = 0;
+            }
+
+            var assemblyNames = ImmutableArray.CreateBuilder<string>(initialCapacity);
```
333fred: Honestly, I wonder if this is a premature optimization for this. Using the entire count is likely never going to be what you want, given that at least the entire BCL won't have this attribute.

333fred (commented):
No more comments on the generator. Can't really sign off the on the whole PR, but I don't expect issues from the current generator structure here.

PR #2113, opened 2025-11-13
CSWinRTGEN GUID Handling

kythant:


src/Tests/FunctionalTests/ClassActivation/ClassActivation.csproj
```diff
@@ -14,4 +14,10 @@
     <ProjectReference Include="..\..\..\Projections\Windows\Windows.csproj" />
   </ItemGroup>
 
+  <Target Name="TestGatherRefPaths" BeforeTargets="CoreCompile">
```
Sergio0694: Is this and the target in the other project just a leftover from testing I assume?

src/WinRT.Interop.Generator/Builders/InteropTypeDefinitionBuilder.Delegate.cs
```diff
@@ -45,7 +45,7 @@ public static void IIDs(
                 interopDefinitions: interopDefinitions,
                 interopReferences: interopReferences,
                 module: module,
-                iid: Guid.NewGuid(), // TODO
+                iid: GuidGenerator.CreateIID(delegateType, interopReferences), // TODO
```
Sergio0694: Remove the `// TODO` if this line has been replaced with the correct call. Same in the other files.

Sergio0694 (changes requested):

src/Tests/FunctionalTests/ClassActivation/ClassActivation.csproj
kythant: Yea ignore the test leftover. They won't be in final

src/WinRT.Interop.Generator/Extensions/TypeDefinitionExtensions.cs
```diff
@@ -21,13 +21,20 @@ internal static class TypeDefinitionExtensions
         /// </summary>
         public bool IsStatic => type.IsAbstract && type.IsSealed;
 
+
+        /// <summary>
+        /// Gets whether a given type is a System.Guid
+        /// </summary>
+        /// <returns>Whether the type has or inherits an attribute with the specified type.</returns>
+        public bool IsGuid => type.FullName == "System.Guid";
```
Sergio0694: Rather than this, define an extension next to the existing `IsKeyValuePair`, that takes `InteropReferences` and compares with the `TypeReference` for `Guid` instead. You can actually define that under the `ITypeDescriptor` extensions, since it can work on any kind of type representation.

src/WinRT.Interop.Generator/Program.cs
```diff
@@ -4,5 +4,7 @@
 using ConsoleAppFramework;
 using WindowsRuntime.InteropGenerator.Generation;
 
+args = [@"C:\Users\kythant\Documents\staging\computesharp-ClassActivation.rsp"];
```
kythant: Debugging code. Will remove after first review

src/WinRT.Interop.Generator/Helpers/GuidGenerator.cs
```diff
@@ -0,0 +1,343 @@
...
+
+            // encode rfc clock/reserved field
+            data[8] = (byte)((data[8] & 0x3f) | 0x80);
+        }
+        return new Guid(data[0..16]);
+    }
+
+
+    // TODO: Debug code; Will remove later
+#pragma warning disable IDE0044 // Add readonly modifier
```
kythant: Debugging code. Will remove after first review

src/WinRT.Interop.Generator/Helpers/GuidGenerator.cs
```diff
@@ -0,0 +1,343 @@
...
+    private static HashSet<TypeSignature> generatedIIDs = [];
+    private static StreamWriter writer = new(printPath, append: false);
+#pragma warning restore IDE0044 // Add readonly modifier
+
+    public static Guid CreateIID(TypeSignature type, InteropReferences interopReferences, bool useWindowsUIXamlProjections)
+    {
+        string signature = GetSignature(type, interopReferences, useWindowsUIXamlProjections);
+        Guid guid = CreateGuidFromSignature(signature);
+        //TODO: Debug code; Will remove later
+        if (!generatedIIDs.Contains(type))
```
kythant: Debugging code. Will remove after first review

src/WinRT.Interop.Generator/Generation/InteropGenerator.Discover.cs
```diff
@@ -329,7 +329,8 @@ private static void DiscoverGenericTypeInstantiations(
 
                 // Gather all 'KeyValuePair<,>' instances
                 if (typeDefinition.IsValueType &&
-                    SignatureComparer.IgnoreVersion.Equals(typeSignature.GenericType, interopReferences.KeyValuePair2))
+                    SignatureComparer.IgnoreVersion.Equals(typeSignature.GenericType, interopReferences.KeyValuePair2) &&
+                    typeDefinition.IsProjectedWindowsRuntimeType)
```
Sergio0694: This isn't needed. We already know that the type is some `KeyValuePair<,>`, which is custom mapped.

src/WinRT.Interop.Generator/Helpers/GuidGenerator.cs
```diff
@@ -0,0 +1,343 @@
...
+
+
+    // TODO: Debug code; Will remove later
+#pragma warning disable IDE0044 // Add readonly modifier
+    private static readonly string printPath = @"C:\Users\kythant\Documents\staging\GUIDsFromCSWinRTGen.txt";
+    private static HashSet<TypeSignature> generatedIIDs = [];
+    private static StreamWriter writer = new(printPath, append: false);
+#pragma warning restore IDE0044 // Add readonly modifier
+
+    public static Guid CreateIID(TypeSignature type, InteropReferences interopReferences, bool useWindowsUIXamlProjections)
```
Sergio0694: Sort methods so that all public members are first, then private ones. Basically this should be the first method.

src/WinRT.Interop.Generator/Helpers/GuidGenerator.cs
```diff
@@ -0,0 +1,343 @@
...
+                if (first is AsmResolver.Utf8String s)
+                {
+                    return Guid.TryParse(s.Value, out guid);
+                }
+            }
+        }
+        return false;
+    }
+
+    internal static bool GetDefaultInterfaceSignatureFromAttribute(TypeDefinition typeDef, out TypeSignature defaultInterfaceSig)
```
Sergio0694: Are all these helpers just used internally within `GuidGenerator`? If so make them private, not internal.

src/WinRT.Interop.Generator/Helpers/GuidGenerator.cs
```diff
@@ -0,0 +1,343 @@
...
+                {
+                    defaultInterfaceSig = sig;
+                    return true;
+                }
+            }
+        }
+        return false;
+    }
+
+    private static Guid EncodeGuid(Span<byte> data)
```
Sergio0694:
```suggestion
    private static Guid EncodeGuid(ReadOnlySpan<byte> data)
```

src/WinRT.Interop.Generator/Helpers/GuidGenerator.cs
```diff
@@ -0,0 +1,343 @@
...
+                    return Guid.TryParse(s.Value, out guid);
+                }
+            }
+        }
+        return false;
+    }
+
+    internal static bool GetDefaultInterfaceSignatureFromAttribute(TypeDefinition typeDef, out TypeSignature defaultInterfaceSig)
+    {
+        defaultInterfaceSig = null!;
```
Sergio0694: Don't suppress nullability warnings. Make the parameter `[NotNullWhen(true)] out TypeSignature?` instead.

src/WinRT.Interop.Generator/Helpers/GuidGenerator.cs
```diff
@@ -0,0 +1,343 @@
...
+        if (typeDef.TryGetCustomAttribute("WindowsRuntime", "WindowsRuntimeDefaultInterfaceAttribute", out AsmResolver.DotNet.CustomAttribute? customAttribute))
```
Sergio0694: Can we pass `InteropReferences` and get `WindowsRuntimeDefaultInterfaceAttribute` from there instead?

src/WinRT.Interop.Generator/Helpers/GuidGenerator.cs
```diff
@@ -0,0 +1,343 @@
...
+            CustomAttributeSignature? customAttributeSignature = customAttribute.Signature;
+            if (customAttributeSignature is not null && customAttributeSignature.FixedArguments.Count != 0)
+            {
+                object? element = customAttributeSignature.FixedArguments[0].Element;
+                if (element is TypeSignature sig)
+                {
+                    defaultInterfaceSig = sig;
+                    return true;
+                }
+            }
```
Sergio0694: You can simplify this with some pattern magic:
```suggestion
            if (attribute is { Signature.FixedArguments: [{ Element: TypeSignature signature }, ..] })
            {
                defaultInterfaceSig = signature;

                return true;
            }
```

src/WinRT.Interop.Generator/Helpers/GuidGenerator.cs
```diff
@@ -0,0 +1,343 @@
...
+        }
+
+        return Guid.Empty; // TODO: don'turn empty guid but throw instead like below. Currently, not all the types we need to be filtered are not filtered out such as System.Reflection.* types.
+        //throw new ArgumentException("Type does not have a Guid attribute");
+    }
+
+    internal static bool GetGuidFromAttribute(TypeDefinition typeDef, out Guid guid)
+    {
+        guid = Guid.Empty;
+        if (typeDef.TryGetCustomAttribute("System.Runtime.InteropServices", "GuidAttribute", out AsmResolver.DotNet.CustomAttribute? customAttribute))
```
Sergio0694: Same as the other case, can we pass `InteropReferences` and get `Guid` from there?

Sergio0694: Don't think the fully qualified type name is needed, right? Same as below
```suggestion
        if (typeDef.TryGetCustomAttribute("System.Runtime.InteropServices", "GuidAttribute", out CustomAttribute? customAttribute))
```

src/WinRT.Interop.Generator/Helpers/GuidGenerator.cs
```diff
@@ -0,0 +1,343 @@
...
+            case ElementType.FnPtr:
+            case ElementType.MVar:
+            case ElementType.CModReqD:
+            case ElementType.CModOpt:
+            case ElementType.Internal:
+            case ElementType.Modifier:
+            case ElementType.Sentinel:
+            case ElementType.Pinned:
+            case ElementType.Boxed:
+            case ElementType.Enum:
```
Sergio0694: These cases are not needed, just remove them. If there's a warning just add a `#pragma` for it at the top of the file

src/WinRT.Interop.Generator/Helpers/GuidGenerator.cs
```diff
@@ -0,0 +1,343 @@
...
+            case ElementType.Boolean:
+                return "b1";
+            case ElementType.Char:
+                return "c2";
+            case ElementType.Object:
+                return "cinterface(IInspectable)";
+            case ElementType.String:
+                return "string";
+            case ElementType.Type:
+                return "struct(Windows.UI.Xaml.Interop.222TypeName;string;enum(Windows.UI.Xaml.Interop.TypeKind;i4))";
```
Sergio0694: I assume the `222` is a typo?

src/WinRT.Interop.Generator/Helpers/GuidGenerator.cs
Sergio0694: Nit: all these methods are missing XML docs.

src/WinRT.Interop.Generator/Helpers/GuidGenerator.cs
```diff
@@ -0,0 +1,343 @@
...
+            }
+        }
+        return false;
+    }
+
+    private static Guid EncodeGuid(Span<byte> data)
+    {
+        if (BitConverter.IsLittleEndian)
+        {
+            // swap bytes of int a
```
Sergio0694: Super-nit: comments should start with an uppercase letter

src/WinRT.Interop.Generator/Helpers/GuidGenerator.cs
```diff
@@ -0,0 +1,343 @@
...
+        int minimumPooledLength = 16 /* Number of bytes in a GUID */ + maxUtf8ByteCount;
+        byte[]? utf8BytesFromPool = null;
+        Span<byte> utf8Bytes = minimumPooledLength <= 512
+            ? stackalloc byte[512]
+            : (utf8BytesFromPool = ArrayPool<byte>.Shared.Rent(minimumPooledLength));
+
+        _ = wrt_pinterface_namespace.TryWriteBytes(utf8Bytes);
+
+        int encodedUtf8BytesWritten = Encoding.UTF8.GetBytes(signature, utf8Bytes[16..]);
+        Span<byte> sha1Bytes = stackalloc byte[20]; // 'SHA1.HashSizeInBytes' does not exist on .NET 6, update this when that's dropped
```
Sergio0694: This project targets .NET 10, can you address this TODO?

src/WinRT.Interop.Generator/Helpers/TypeMapping.cs
```diff
@@ -0,0 +1,235 @@
...
+using System.Collections.Generic;
+using System.Linq;
+
+namespace WindowsRuntime.InteropGenerator.Helpers;
+
+internal static class TypeMapping
+{
+    internal readonly record struct MappedType(
+         string PublicName,
+         string WinRTNamespace, // WinRT namespace
```
Sergio0694: Write proper XML docs, don't use inline comments like these

Sergio0694: Use "WindowsRuntime", not "WinRT". We're never using that abbreviation.

src/WinRT.Interop.Generator/Helpers/TypeMapping.cs
```diff
@@ -0,0 +1,235 @@
...
+         string? Signature = null
+    );
+
+    internal readonly record struct WindowsUIXamlMappedType(
+         string WinRTNamespace, // WinRT namespace
+         string WinRTName,      // WinRT type name (same as winrt_name)
+         string? Signature = null
+    );
+
+    private static readonly Dictionary<string, MappedType[]> WinRTToABITypeMapping = new(StringComparer.Ordinal)
```
Sergio0694: Use a `FrozenDictionary<string, ImmutableArray<MappedType>>` instead.

src/WinRT.Interop.Generator/Helpers/TypeMapping.cs
```diff
@@ -0,0 +1,235 @@
...
+    {
+        ["System"] =
+        [
+            new MappedType("IServiceProvider", "Microsoft.UI.Xaml", "IXamlServiceProvider"),
+
+            new MappedType("IntPtr", "WinRT.Interop", "HWND"),
```
Sergio0694: This doesn't seem correct.

src/WinRT.Interop.Generator/Helpers/TypeMapping.cs
```diff
@@ -0,0 +1,235 @@
...
+         string WinRTName,      // WinRT type name (same as winrt_name)
+         string? Signature = null
+    );
+
+    private static readonly Dictionary<string, MappedType[]> WinRTToABITypeMapping = new(StringComparer.Ordinal)
+    {
+        ["System"] =
+        [
+            new MappedType("IServiceProvider", "Microsoft.UI.Xaml", "IXamlServiceProvider"),
+
```
Sergio0694: Nit: there's a bunch of blank lines in this file, can you remove them?

src/WinRT.Interop.Generator/Helpers/TypeMapping.cs
```diff
@@ -0,0 +1,235 @@
...
+        [
+            new MappedType("Point", "Windows.Foundation", "Point", "struct(Windows.Foundation.Point;f4;f4)"),
+            new MappedType("Size", "Windows.Foundation", "Size", "struct(Windows.Foundation.Size;f4;f4)"),
+            new MappedType("Rect", "Windows.Foundation", "Rect", "struct(Windows.Foundation.Rect;f4;f4;f4;f4)")
+        ]
+    };
+
+
+
+    private static readonly Dictionary<string, WindowsUIXamlMappedType> WindowsUIXamlProjectionTypeMapping = new(StringComparer.Ordinal)
```
Sergio0694: This should be `FrozenDictionary<string, WindowsUIXamlMappedType>`.

src/WinRT.Interop.Generator/References/InteropReferences.cs
```diff
@@ -128,6 +128,11 @@ public InteropReferences(
     /// </summary>
     public TypeReference Array => field ??= _corLibTypeFactory.CorLibScope.CreateTypeReference("System"u8, "Array"u8);
 
+    /// <summary>
+    /// Gets the <see cref="TypeReference"/> for <see cref="System.Nullable"/>.
+    /// </summary>
+    public TypeReference Nullable => field ??= _corLibTypeFactory.CorLibScope.CreateTypeReference("System"u8, "Nullable`1"u8);
```
Sergio0694: This should be called `Nullable1`.

src/WinRT.Interop.Generator/References/WellKnownInterfaceIIDs.cs
```diff
@@ -67,4 +68,81 @@ _ when SignatureComparer.IgnoreVersion.Equals(interfaceType, interopReferences.I
...
+                => new Guid("75f99e2a-137e-537e-a5b1-0b5a6245fc02"),
+            _ when SignatureComparer.IgnoreVersion.Equals(type, interopReferences.IAsyncOperation1)
+                => new Guid("2bd35ee6-72d9-5c5d-9827-05ebb81487ab"),
+            _ when SignatureComparer.IgnoreVersion.Equals(type, interopReferences.VectorChangedEventHandler1)
+                => new Guid("a1e9acd7-e4df-5a79-aefa-de07934ab0fb"),
+            _ when SignatureComparer.IgnoreVersion.Equals(type, interopReferences.AsyncActionProgressHandler1)
+                => new Guid("c261d8d0-71ba-5f38-a239-872342253a18"),
+            _ when SignatureComparer.IgnoreVersion.Equals(type, interopReferences.IVectorChangedEventArgs)
+                => new Guid("575933df-34fe-4480-af15-07691f3d5d9b"),
+            _ => Guid.Empty
```
Sergio0694, line 144: Don't return an empty `Guid` for no matches. Instead make this method `TryGetGuid`, and return a `bool` with the `Guid` being an `out` parameter.

src/WinRT.Interop.Generator/References/WellKnownInterfaceIIDs.cs
```diff
@@ -67,4 +68,81 @@ _ when SignatureComparer.IgnoreVersion.Equals(interfaceType, interopReferences.I
...
+        {
+            SzArrayTypeSignature szArrayTypeSignature => szArrayTypeSignature.GetUnderlyingType().ToTypeDefOrRef(),
+            GenericInstanceTypeSignature genericSignature => genericSignature.GenericType,
+            _ => signature.ToTypeDefOrRef(),
+        };
+        return type switch
+        {
+            // Shared types
+            _ when SignatureComparer.IgnoreVersion.Equals(type, interopReferences.EventHandler)
+                => new Guid("9de1c535-6ae1-11e0-84e1-18a905bcc53f"),
```
Sergio0694: Nit: make all these GUID strings uppercase.

src/WinRT.Interop.Generator/References/WellKnownInterfaceIIDs.cs
```diff
@@ -67,4 +68,81 @@ _ when SignatureComparer.IgnoreVersion.Equals(interfaceType, interopReferences.I
             memberName: $"get_IID_{nameSuffix}",
             signature: MethodSignature.CreateStatic(WellKnownTypeSignatureFactory.InGuid(interopReferences)));
     }
+
+    public static Guid get_GUID(
+        TypeSignature signature,
+        bool useWindowsUIXamlProjections,
```
Sergio0694: This parameter is never used

src/WinRT.Interop.Generator/References/WellKnownInterfaceIIDs.cs
```diff
@@ -67,4 +68,81 @@ _ when SignatureComparer.IgnoreVersion.Equals(interfaceType, interopReferences.I
...
+    {
+        ITypeDefOrRef type = signature switch
+        {
+            SzArrayTypeSignature szArrayTypeSignature => szArrayTypeSignature.GetUnderlyingType().ToTypeDefOrRef(),
+            GenericInstanceTypeSignature genericSignature => genericSignature.GenericType,
+            _ => signature.ToTypeDefOrRef(),
+        };
+        return type switch
+        {
+            // Shared types
```
Sergio0694: Unnecessary comment, remove it

src/WinRT.Interop.Generator/References/WellKnownInterfaceIIDs.cs
```diff
@@ -67,4 +68,81 @@ _ when SignatureComparer.IgnoreVersion.Equals(interfaceType, interopReferences.I
...
+            _ when SignatureComparer.IgnoreVersion.Equals(type, interopReferences.AsyncOperationCompletedHandler1)
+                => new Guid("9d534225-231f-55e7-a6d0-6c938e2d9160"),
+            _ when SignatureComparer.IgnoreVersion.Equals(type, interopReferences.MapChangedEventHandler2)
+                => new Guid("19046f0b-cf81-5dec-bbb2-7cc250da8b8b"),
+            _ when SignatureComparer.IgnoreVersion.Equals(type, interopReferences.IList)
+                => new Guid("393de7de-6fd0-4c0d-bb71-47244a113e93"), // Unsure of this one
+            _ when SignatureComparer.IgnoreVersion.Equals(type, interopReferences.IList1)
+                => new Guid("0e3f106f-a266-50a1-8043-c90fcf3844f6"), // Unsure of this one
+            _ when SignatureComparer.IgnoreVersion.Equals(type, interopReferences.IReadOnlyList1)
+                => new Guid("5f07498b-8e14-556e-9d2e-2e98d5615da9"), // Unsure of this one
```
Sergio0694: Don't forget to check these 3 and remove the comments after

src/WinRT.Interop.Generator/References/WellKnownInterfaceIIDs.cs
```diff
@@ -67,4 +68,81 @@ _ when SignatureComparer.IgnoreVersion.Equals(interfaceType, interopReferences.I
             memberName: $"get_IID_{nameSuffix}",
             signature: MethodSignature.CreateStatic(WellKnownTypeSignatureFactory.InGuid(interopReferences)));
     }
+
+    public static Guid get_GUID(
+        TypeSignature signature,
+        bool useWindowsUIXamlProjections,
+        InteropReferences interopReferences)
+    {
+        ITypeDefOrRef type = signature switch
```
Sergio0694: I don't think we should be unwrapping constructed generics. Callers should be able to pass generic type definitions if needed, also because otherwise what we're doing here would be incorrect (we're taking some `IFoo<Bar>` but giving the IID for just `IFoo<>` instead). Do this:
- Change the parameter to `ITypeDescriptor interfaceType` like the method above
- Remove this switch
- Update the callers to ensure they pass the right thing they need

src/WinRT.Interop.Generator/References/WellKnownInterfaceIIDs.cs
```diff
@@ -67,4 +68,81 @@ _ when SignatureComparer.IgnoreVersion.Equals(interfaceType, interopReferences.I
             memberName: $"get_IID_{nameSuffix}",
             signature: MethodSignature.CreateStatic(WellKnownTypeSignatureFactory.InGuid(interopReferences)));
     }
+
+    public static Guid get_GUID(
```
Sergio0694: Nit: missing XML docs

Sergio0694 (changes requested):
Looks nice! Did a first review pass for now 🙂

src/WinRT.Interop.Generator/Generation/InteropGenerator.Discover.cs
kythant: This was meant to filter System.WeakReference. Is there another proper way to do this?

Sergio0694: I am not following. This code could've never matched `WeakReference` 🤔
Like, even just the first condition is `IsValueType`, but `WeakReference` is a class.

src/WinRT.Interop.Generator/Helpers/GuidGenerator.cs
kythant: Can't do read only span without creating a buffer for the swtiches that are happening.

Do we want to go with this approach?

src/WinRT.Interop.Generator/Generation/InteropGenerator.Discover.cs
kythant: Sorry to clarify, this was showing up as a generic param to KeyValuePair

System.Collections.Generic.KeyValuePair`2<System.Int32, System.WeakReference`1<System.Object>>

Sergio0694: Sure but even in that case I don't see how this check is doing anything. The `TypeDefinition` here would just be `KeyValuePair<,>` (unconstructed), no?

src/WinRT.Interop.Generator/Helpers/GuidGenerator.cs
Sergio0694: You can just `stackalloc` that buffer. But also we're never going to run on big endian, so not sure we even need that.

src/WinRT.Interop.Generator/Helpers/GuidGenerator.cs
kythant: That's what I originally had haha

src/WinRT.Interop.Generator/References/WellKnownInterfaceIIDs.cs
kythant: This will produce incorrect GUIDs with the Guid generator because it is case sensitive. But I should probably just sanitize it and call toLowerCase()?

src/WinRT.Interop.Generator/Helpers/TypeMapping.cs
kythant: I wanted to divide them by namespaces

src/WinRT.Interop.Generator/References/InteropReferences.cs
kythant: Are you sure?

image
It shows up as Nulllable'1 in my testing.

src/WinRT.Interop.Generator/References/WellKnownInterfaceIIDs.cs
kythant: Gonna double check everything about this at the end. There are probably some missing entries.

src/WinRT.Interop.Generator/References/InteropReferences.cs
Sergio0694: No I mean the property type you defined should be called `Nullable1` 🙂
E.g. just like the one for `KeyValuePair<,>` is called `KeyValuePair2`.

src/WinRT.Interop.Generator/References/WellKnownInterfaceIIDs.cs
Sergio0694: No that's not how this works. This string is just used to create a `Guid` instance here. It doesn't affect how `ToString()` formats the `Guid` value later. A `Guid` itself is just a 128 bit binary value, it doesn't have actual characters. Calling `ToString()` on **any** `Guid` value will always give you a lowercase output, it doesn't matter how the `Guid` value was originally constructed or what `string` it was parsed from.

kythant: Oh gotcha. Yea no I was making sure these strings here was lowercase because the hardcoded ones in the TypeMapping were returning the wrong values.

src/WinRT.Interop.Generator/Helpers/GuidGenerator.cs
```diff
@@ -0,0 +1,342 @@
...
+        ITypeDefOrRef typeDefOrRef = typeSig switch
+        {
+            SzArrayTypeSignature szArrayTypeSignature => szArrayTypeSignature.GetUnderlyingType().ToTypeDefOrRef(),
+            GenericInstanceTypeSignature genericSignature => genericSignature.GenericType,
+            _ => typeSig.ToTypeDefOrRef(),
+        };
+        if (WellKnownInterfaceIIDs.try_GetGUID(typeDefOrRef, interopReferences, out Guid result))
+        {
+            return result;
+        }
```
Sergio0694: Not sure I see how this is correct. Let's say you call this with the type signature being `EventHandler<string>`. Here you're unwrapping it and just calling `try_GetGUID` with the generic type definition. So that will succeed and give you the PIID for `EventHandler<T>`. And this method will return that. But that will **not** be the IID for the (constructed) `EventHandler<string>` signature that we actually have here, no?

src/WinRT.Interop.Generator/Helpers/GuidGenerator.cs
```diff
@@ -0,0 +1,342 @@
...
+    /// <param name="typeDef">The type definition to inspect for the GUID attribute.</param>
+    /// <param name="interopReferences">The <see cref="InteropReferences"/> instance to use.</param>
+    /// <param name="guid">When this method returns <c>true</c>, contains the parsed GUID value.</param>
+    /// <returns><c>true</c> if a valid GUID was found and parsed; otherwise, <c>false</c>.</returns>
+    private static bool GetGuidFromAttribute(TypeDefinition typeDef, InteropReferences interopReferences, out Guid guid)
+    {
+        guid = Guid.Empty;
+        if (typeDef.TryGetCustomAttribute(interopReferences.GuidAttribute, out CustomAttribute? customAttribute))
+        {
+            if (customAttribute.Signature is { FixedArguments: [{ Element: AsmResolver.Utf8String guidString }, ..] })
```
Sergio0694:
```suggestion
            if (customAttribute.Signature is { FixedArguments: [{ Element: Utf8String guidString }, ..] })
```

src/WinRT.Interop.Generator/Helpers/GuidGenerator.cs
```diff
@@ -0,0 +1,342 @@
...
+    /// <summary>
+    /// Attempts to retrieve a GUID from the <c>GuidAttribute</c> applied to the specified type.
+    /// </summary>
+    /// <param name="typeDef">The type definition to inspect for the GUID attribute.</param>
+    /// <param name="interopReferences">The <see cref="InteropReferences"/> instance to use.</param>
+    /// <param name="guid">When this method returns <c>true</c>, contains the parsed GUID value.</param>
+    /// <returns><c>true</c> if a valid GUID was found and parsed; otherwise, <c>false</c>.</returns>
+    private static bool GetGuidFromAttribute(TypeDefinition typeDef, InteropReferences interopReferences, out Guid guid)
+    {
+        guid = Guid.Empty;
```
Sergio0694: Unnecessary here, just move this assignment below, right above `return false`.

src/WinRT.Interop.Generator/Helpers/GuidGenerator.cs
```diff
@@ -0,0 +1,342 @@
...
+    /// Attempts to retrieve the default interface signature from the <c>WindowsRuntimeDefaultInterfaceAttribute</c>
+    /// applied to the specified type.
+    /// </summary>
+    /// <param name="typeDef">The type definition to inspect for the default interface attribute.</param>
+    /// <param name="interopReferences">The <see cref="InteropReferences"/> instance to use.</param>
+    /// <param name="defaultInterfaceSig">When this method returns <c>true</c>, contains the default interface type signature.</param>
+    /// <returns><c>true</c> if the default interface signature was found; otherwise, <c>false</c>.</returns>
+    private static bool GetDefaultInterfaceSignatureFromAttribute(TypeDefinition typeDef, InteropReferences interopReferences, [NotNullWhen(true)] out TypeSignature defaultInterfaceSig)
+    {
+        defaultInterfaceSig = null!;
```
Sergio0694: Same as above, move this above `return false`.

src/WinRT.Interop.Generator/References/InteropReferences.cs
kythant: oh gotcha!

src/WinRT.Interop.Generator/Helpers/TypeMapping.cs
```diff
@@ -0,0 +1,185 @@
...
+internal static class TypeMapping
+{
+    /// <summary>
+    /// Represents a mapping from a CLR/public type name to its Windows Runtime (ABI) counterpart,
+    /// optionally including the hardcoded WinRT type signature used for GUID generation and marshalling.
+    /// </summary>
+    /// <variable name="WindowsRuntimeNamespace">The Windows Runtime namespace.</variable>
+    /// <variable name="WindowsRuntimeName">The Windows Runtime type name.</variable>
+    /// <variable name="Signature">Optional hardcoded WinRT type signature</variable>
+    internal readonly record struct MappedType(
```
Sergio0694: The containing type is already internal, so you can make all these nested internal members just public.

src/WinRT.Interop.Generator/Helpers/TypeMapping.cs
```diff
@@ -0,0 +1,185 @@
...
+        new("System.Exception", new("Windows.Foundation", "HResult", "struct(Windows.Foundation.HResult;i4)")),
+        new("System.IDisposable", new("Windows.Foundation", "IClosable")),
+        new("System.Nullable`1", new("Windows.Foundation", "IReference`1")),
+        new("System.TimeSpan", new("Windows.Foundation", "TimeSpan", "struct(Windows.Foundation.TimeSpan;i8)")),
+        new("System.EventHandler`2", new("Windows.Foundation", "TypedEventHandler`2")),
+        new("System.Uri", new("Windows.Foundation", "Uri", "rc(Windows.Foundation.Uri;{9e365e57-48b2-4160-956f-c7385120bbfc})")),
+        new("System.AttributeTargets", new("Windows.Foundation.Metadata", "AttributeTargets")),
+        new("System.AttributeUsageAttribute", new("Windows.Foundation.Metadata", "AttributeUsageAttribute")),
+        new("System.Type", new("Windows.UI.Xaml.Interop", "TypeName", "struct(Windows.UI.Xaml.Interop.TypeName;string;enum(Windows.UI.Xaml.Interop.TypeKind;i4)")),
+
```
Sergio0694: Let's remove all these blank lines, I don't think they're needed to separate namespaces. All the keys are already sorted alphabetically, so it's easier to go through the list.

src/WinRT.Interop.Generator/Helpers/TypeMapping.cs
```diff
@@ -0,0 +1,185 @@
...
+        new("System.Numerics.Plane", new("Windows.Foundation.Numerics", "Plane", "struct(Windows.Foundation.Numerics.Plane;struct(Windows.Foundation.Numerics.Vector3;f4;f4;f4);f4)")),
+        new("System.Numerics.Quaternion", new("Windows.Foundation.Numerics", "Quaternion", "struct(Windows.Foundation.Numerics.Quaternion;f4;f4;f4;f4)")),
+        new("System.Numerics.Vector2", new("Windows.Foundation.Numerics", "Vector2", "struct(Windows.Foundation.Numerics.Vector2;f4;f4)")),
+        new("System.Numerics.Vector3", new("Windows.Foundation.Numerics", "Vector3", "struct(Windows.Foundation.Numerics.Vector3;f4;f4;f4)")),
+        new("System.Numerics.Vector4", new("Windows.Foundation.Numerics", "Vector4", "struct(Windows.Foundation.Numerics.Vector4;f4;f4;f4;f4)")),
+
+        new("Windows.Foundation.Point", new("Windows.Foundation", "Point", "struct(Windows.Foundation.Point;f4;f4)")),
+        new("Windows.Foundation.Size", new("Windows.Foundation", "Size", "struct(Windows.Foundation.Size;f4;f4)")),
+        new("Windows.Foundation.Rect", new("Windows.Foundation", "Rect", "struct(Windows.Foundation.Rect;f4;f4;f4;f4)"))
+    );
```
Sergio0694: Nit, same as below
```suggestion
        new("Windows.Foundation.Rect", new("Windows.Foundation", "Rect", "struct(Windows.Foundation.Rect;f4;f4;f4;f4)")));
```

src/WinRT.Interop.Generator/Helpers/TypeMapping.cs
```diff
@@ -0,0 +1,185 @@
...
+    /// <param name="useWindowsUIXamlProjections">True to apply Windows.UI.Xaml projection mappings if available.</param>
+    /// <param name="mappedName">Resulting string representing the mapped full name.</param>
+    /// <returns><c>true</c> if mapped name was found; otherwise, <c>false</c>.</returns>
+    internal static bool TryFindMappedWinRTFullName(string? fullName, bool useWindowsUIXamlProjections, out string mappedName)
+    {
+        mappedName = string.Empty;
+        if (fullName is null)
+        {
+            throw new ArgumentNullException("fullName is null");
+        }
```
Sergio0694: If `fullName` can't be `null`, just don't make it nullable, and then you can remove this check.

src/WinRT.Interop.Generator/Helpers/TypeMapping.cs
```diff
@@ -0,0 +1,185 @@
...
+    /// <summary>
+    /// Returns the mapped WinRT full type name for a given CLR fullName.
+    /// </summary>
+    /// <param name="fullName">The fullName of the type.</param>
+    /// <param name="useWindowsUIXamlProjections">True to apply Windows.UI.Xaml projection mappings if available.</param>
+    /// <param name="mappedName">Resulting string representing the mapped full name.</param>
+    /// <returns><c>true</c> if mapped name was found; otherwise, <c>false</c>.</returns>
+    internal static bool TryFindMappedWinRTFullName(string? fullName, bool useWindowsUIXamlProjections, out string mappedName)
+    {
+        mappedName = string.Empty;
```
Sergio0694: Move this just above `return false`.

src/WinRT.Interop.Generator/Helpers/TypeMapping.cs
```diff
@@ -0,0 +1,185 @@
...
+    );
+
+    /// <summary>
+    /// Returns the mapped WinRT full type name for a given CLR fullName.
+    /// </summary>
+    /// <param name="fullName">The fullName of the type.</param>
+    /// <param name="useWindowsUIXamlProjections">True to apply Windows.UI.Xaml projection mappings if available.</param>
+    /// <param name="mappedName">Resulting string representing the mapped full name.</param>
+    /// <returns><c>true</c> if mapped name was found; otherwise, <c>false</c>.</returns>
+    internal static bool TryFindMappedWinRTFullName(string? fullName, bool useWindowsUIXamlProjections, out string mappedName)
```
Sergio0694: Make the returned value `[NotNullWhen(true)] out string?` instead (and update the failure assignment below).

src/WinRT.Interop.Generator/Helpers/TypeMapping.cs
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    /// <summary>
+    /// Retrieves the WinRT signature string for a mapped CLR type.
+    /// </summary>
+    /// <param name="fullName">The fullName of the type.</param>
+    /// <param name="useWindowsUIXamlProjections">True to apply Windows.UI.Xaml projection mappings if available.</param>
+    /// <param name="signature">Resulting string represent the guid signature if found.</param>
+    /// <returns><c>true</c> if hardcoded GUID signature was found; otherwise, <c>false</c>.</returns>
+    internal static bool TryFindGuidSignatureForMappedType(string? fullName, bool useWindowsUIXamlProjections, out string signature)
```
Sergio0694: Same notes here as the method above

src/WinRT.Interop.Generator/Helpers/TypeMapping.cs
```diff
@@ -0,0 +1,185 @@
...
+                string? projSignature = WindowsUIXamlProjectionTypeMapping[resultFullName].Signature;
+                if (projSignature is not null)
+                {
+                    signature = projSignature;
+                    return true;
+                }
+                else
+                {
+                    throw new InvalidOperationException($"Signature missing for Windows UI Xaml projection mapping for type {resultFullName}");
+                }
```
Sergio0694: The `else` is unnecessary, since the `if` always return. You can inline the block.
```suggestion

                throw new InvalidOperationException($"Signature missing for Windows UI Xaml projection mapping for type {resultFullName}");
```

src/WinRT.Interop.Generator/References/WellKnownInterfaceIIDs.cs
```diff
@@ -67,4 +69,83 @@ _ when SignatureComparer.IgnoreVersion.Equals(interfaceType, interopReferences.I
...
\ No newline at end of file
+
+    /// <summary>
+    /// Attempts to resolve a well-known Windows Runtime interface GUID for the specified type signature.
+    /// </summary>
+    /// <param name="interfaceType"> The <see cref="ITypeDescriptor"/> representing the managed type to inspect</param>
+    /// <param name="interopReferences">The <see cref="InteropReferences"/> instance to use.</param>
+    /// <param name="guid">Out parameter for the resolved <see cref="Guid"/> of the type. </param>
+    /// <returns><c>true</c> if a matching GUID was found; otherwise, <c>false</c>.</returns>
+    public static bool try_GetGUID(
```
Sergio0694:
```suggestion
    public static bool TryGetGuid(
```

src/WinRT.Interop.Generator/References/WellKnownInterfaceIIDs.cs
```diff
@@ -67,4 +69,83 @@ _ when SignatureComparer.IgnoreVersion.Equals(interfaceType, interopReferences.I
...
+                => new Guid("575933df-34fe-4480-af15-07691f3d5d9b"),
+            _ => Guid.Empty
+        };
+
+        return guid != Guid.Empty;
+    }
+}
+
+
+
```
Sergio0694:
```suggestion
```

src/WinRT.Interop.Generator/Helpers/GuidGenerator.cs
kythant: I thought we would be getting the IID of the String separately in the caller so for example this line on line 150 would be handling it; typeArgumentSignatures would contain the IID of the String here. 

 return "pinterface({" + GetGuid(typeSignature, interopReferences) + "};" + string.Join(";", typeArgumentSignatures) + ")";

src/WinRT.Interop.Generator/Helpers/TypeMapping.cs
kythant: Do you mean make MappedType public?

src/WinRT.Interop.Generator/Helpers/TypeMapping.cs
kythant: Shoot I forgot those

src/WinRT.Interop.Generator/References/WellKnownInterfaceIIDs.cs
kythant: Can I also change get_IID in this file to GetIID?

src/WinRT.Interop.Generator/Helpers/GuidGenerator.cs
Sergio0694: Ok but then shouldn't it be _that_ caller passing the right `ITypeDescriptor` directly? Since presumably it would already know exactly whether it's dealing with a constructed generic, or an array, or a simple signature?

src/WinRT.Interop.Generator/Helpers/TypeMapping.cs
Sergio0694: Yeah, make it public. Since it's inside an internal type, it's effective accessibility will still be internal anyway.

src/WinRT.Interop.Generator/References/WellKnownInterfaceIIDs.cs
Sergio0694: No leave that one. It's a bit special in that it specifically refers to the `get_IID` method name in IL which it's returning.

kythant: got it

src/WinRT.Interop.Generator/Helpers/GuidGenerator.cs
kythant: Hmmm. I see what you mean

src/WinRT.Interop.Generator/Generation/InteropGenerator.Discover.cs
kythant: So this was just a coincidence that KeyValuePair`2 below

{System.Collections.Generic.KeyValuePair`2<System.Int32, System.WeakReference`1<System.Object>>}

didn't have the WindowsRuntime Attribute and it didn't get added.
Plus, there is the only example of a TypeSig containing a WeakReference.
The check wasn't accurate/precise. I will remove it. Thanks!

src/WinRT.Interop.Generator/Helpers/GuidGenerator.cs
```diff
@@ -0,0 +1,341 @@
...
+using AsmResolver.DotNet;
+using AsmResolver.DotNet.Signatures;
+using AsmResolver.PE.DotNet.Metadata.Tables;
+using WindowsRuntime.InteropGenerator.References;
+
+namespace WindowsRuntime.InteropGenerator.Helpers;
+
+internal static class GuidGenerator
+{
+    private static readonly Guid wrt_pinterface_namespace = new(0xd57af411, 0x737b, 0xc042, 0xab, 0xae, 0x87, 0x8b, 0x1e, 0x16, 0xad, 0xee);
```
Sergio0694: Let's rename this to something more C#-like. Perhaps `WindowsRuntimePIIDNamespace`.

Sergio0694: Nit: make the hex values here uppercase

src/WinRT.Interop.Generator/Helpers/GuidGenerator.cs
```diff
@@ -0,0 +1,341 @@
...
+                return "b1";
+            case ElementType.Char:
+                return "c2";
+            case ElementType.Object:
+                return "cinterface(IInspectable)";
+            case ElementType.String:
+                return "string";
+            case ElementType.Type:
+                return "struct(Windows.UI.Xaml.Interop.TypeName;string;enum(Windows.UI.Xaml.Interop.TypeKind;i4))";
+            case ElementType.ValueType:
```
Sergio0694: Let's avoid nested `if`-s inside switch cases. Can you split all these into cases with conditions? That is, like:
```csharp
case ElementType.ValueType when typeDefinition.IsEnum:
    // ...
case ElementType.ValueType when typeDefinition.IsGuid:
    // ...
case ElementType.ValueType:
    // ...
```
Same below for the other cases with nested `if`-s.

src/WinRT.Interop.Generator/Helpers/GuidGenerator.cs
```diff
@@ -0,0 +1,341 @@
...
+    {
+        if (TypeMapping.TryFindGuidSignatureForMappedType(typeSignature.FullName, useWindowsUIXamlProjections, out string? mappedSignature))
+        {
+            return mappedSignature;
+        }
+
+        TypeDefinition? typeDefinition = typeSignature.Resolve()! ?? throw new ArgumentException("TypeDefinition could not be resolved for type signature: " + typeSignature.FullName);
+        string typeFullName = TypeMapping.TryFindMappedWinRTFullName(typeDefinition.FullName, useWindowsUIXamlProjections, out string? typeFullNameMapped) ? typeFullNameMapped : typeDefinition.FullName;
+
+#pragma warning disable IDE0010 // Add missing cases
```
Sergio0694: Move this at the top of the file, above the namespace (leave a blank line) and remove this here and the restore below

src/WinRT.Interop.Generator/References/WellKnownInterfaceIIDs.cs
```diff
@@ -67,4 +68,80 @@ _ when SignatureComparer.IgnoreVersion.Equals(interfaceType, interopReferences.I
...
+    /// <param name="interfaceType"> The <see cref="ITypeDescriptor"/> representing the managed type to inspect</param>
+    /// <param name="interopReferences">The <see cref="InteropReferences"/> instance to use.</param>
+    /// <param name="guid">Out parameter for the resolved <see cref="Guid"/> of the type. </param>
+    /// <returns><c>true</c> if a matching GUID was found; otherwise, <c>false</c>.</returns>
+    public static bool TryGetGUID(
+        ITypeDescriptor interfaceType,
+        InteropReferences interopReferences,
+        out Guid guid)
+    {
+        guid = interfaceType switch
```
Sergio0694, line 84: Should we also have `IDisposable` (`IClosable`) here?

src/WinRT.Interop.Generator/Helpers/GuidGenerator.cs
```diff
@@ -0,0 +1,322 @@
...
+            buffer[8] = (byte)((buffer[8] & 0x3F) | 0x80);
+        }
+
+        return new Guid(buffer);
+    }
+
+    /// <summary>Computes a deterministic GUID (IID) from a WinRT signature</summary>
+    /// <param name="signature">WinRT signature string.</param>
+    /// <returns>The derived <see cref="Guid"/>.</returns>
+    private static Guid CreateGuidFromSignature(string signature)
```
Sergio0694: The parameter here could be `ReadOnlySpan<char>`
```suggestion
    private static Guid CreateGuidFromSignature(ReadOnlySpan<char> signature)
```

src/WinRT.Interop.Generator/Helpers/GuidGenerator.cs
```diff
@@ -0,0 +1,322 @@
...
+            (buffer[5], buffer[4]) = (buffer[4], buffer[5]);
+
+            // Swap bytes of short c and encode RFC time/version field
+            (buffer[7], buffer[6]) = ((byte)((buffer[6] & 0x0F) | (5 << 4)), buffer[7]);
+
+            // Encode RFC clock/reserved field
+            buffer[8] = (byte)((buffer[8] & 0x3F) | 0x80);
+        }
+
+        return new Guid(buffer);
```
Sergio0694: Nit:
```suggestion
        return new(buffer);
```

PR #2115, opened 2025-11-13
Plug in correct ABI types in cswinrtgen

Sergio0694:
Title. Also adds the logic to check for blittability.


src/WinRT.Interop.Generator/References/InteropReferences.cs
```diff
@@ -1651,7 +1661,7 @@ public InteropReferences(
             returnType: _corLibTypeFactory.Void,
             parameterTypes: [
                 _corLibTypeFactory.UInt32,
-                ABIType.MakePointerType()]));
+                AbiType.MakePointerType()]));
```
manodasanW: Do you need ToValueTypeSignature here?

Sergio0694: Mmh this code was the same before, but yeah doesn't hurt to fix it, thank you! 🙂

manodasanW (approved):

PR #2117, opened 2025-11-14
Port unit tests and add Vector extensions

manodasanW:


src/WinRT.Runtime2/System.Numerics/VectorExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+
+using System.Runtime.Versioning;
+using Windows.Foundation;
+
+namespace System.Numerics;
+
+/// <summary>
+/// Provides extension methods for converting between <see cref="Vector2"/> and Windows Foundation types.
+/// </summary>
+[SupportedOSPlatform("Windows10.0.10240.0")]
```
Sergio0694, line 12: This attribute seems unnecessary, these extensions will always work.

src/WinRT.Runtime2/System.Numerics/VectorExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+public static class VectorExtensions
+{
+    /// <summary>
+    /// Converts a <see cref="Vector2"/> to a <see cref="Point"/>.
+    /// </summary>
+    /// <param name="vector">The <see cref="Vector2"/> to convert.</param>
+    /// <returns>A <see cref="Point"/> with X and Y coordinates from the vector's X and Y components.</returns>
+    public static Point ToPoint(this Vector2 vector)
+    {
+        return new Point(vector.X, vector.Y);
```
Sergio0694: Nit: use target-typing in all extensions
```suggestion
        return new(vector.X, vector.Y);
```

src/WinRT.Runtime2/System.Numerics/VectorExtensions.cs
```diff
@@ -0,0 +1,54 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Runtime.Versioning;
+using Windows.Foundation;
+
+namespace System.Numerics;
```
Sergio0694, line 7: Do we want to move this to `Windows.Foundation`, or will we just move everything together later?

src/WinRT.Runtime2/System.Numerics/VectorExtensions.cs
manodasanW: Yea this was just to suppress the warnings from using point which had that attribute.

src/WinRT.Runtime2/System.Numerics/VectorExtensions.cs
manodasanW: Lets discuss on namespaces and then move them all together.

PR #2127, opened 2025-11-24
Refactor IID generation logic in 'cswinrtgen'

Sergio0694:
This PR includes some changes to the IID generation logic in 'cswinrtgen':
- Removed `[Guid]` from delegates, switched to reading IIDs from RVA fields
- Refactored the signature generation code for future maintainability
- Streamlined exceptions for signature generation failures, added a well known error code
- Centralized all `IID` methods for all interface types to remove duplication
- Fixed the IID generation for `IReference<T>` for delegate types

### DIFF
--- a/src/WinRT.Interop.Generator/Resolvers/InterfaceIIDResolver.cs
+++ b/src/WinRT.Interop.Generator/Resolvers/InterfaceIIDResolver.cs
@@ -24,24 +24,35 @@ internal static class InterfaceIIDResolver
     /// <returns>The IID for <paramref name="type"/>.</returns>
     public static Guid GetIID(TypeDefinition type)
     {
+        // The IID of projected types that have a fixed one (e.g. interfaces, delegates, etc.)
+        // is always in an 'IID_<INTERFACE_NAME>' property in the 'ABI.InterfaceIIDs' type.
         TypeDefinition interfaceIIDsType = type.DeclaringModule!.GetType("ABI"u8, "InterfaceIIDs"u8);
 
         DefaultInterpolatedStringHandler handler = new(0, 0, null, stackalloc char[256]);
 
+        // Prepare the name of the property accessor method that we need to analyze
         handler.AppendLiteral("get_IID_");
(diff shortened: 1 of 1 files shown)


src/WinRT.Interop.Generator/Helpers/GuidGenerator.cs
```diff
@@ -3,276 +3,144 @@
...
     /// </summary>
-    /// <param name="typeDef">The type definition to inspect for the default interface attribute.</param>
+    /// <param name="type">The type definition to inspect for <see cref="System.Runtime.InteropServices.GuidAttribute"/>.</param>
     /// <param name="interopReferences">The <see cref="InteropReferences"/> instance to use.</param>
-    /// <param name="defaultInterfaceSig">When this method returns <c>true</c>, contains the default interface type signature.</param>
-    /// <returns><c>true</c> if the default interface signature was found; otherwise, <c>false</c>.</returns>
-    private static bool TryGetDefaultInterfaceSignatureFromAttribute(TypeDefinition typeDef, InteropReferences interopReferences, [NotNullWhen(true)] out TypeSignature defaultInterfaceSig)
+    /// <param name="iid">The resulting <see cref="Guid"/> value, if found.</param>
+    /// <returns>Whether <paramref name="iid"/> was succesfully retrieved.</returns>
+    private static bool TryGetIIDFromAttribute(TypeDefinition type, InteropReferences interopReferences, out Guid iid)
```
kythant, line 89: Nit: Should TryGetIIDFromAttribute and TryGetDefaultInterfaceFromAttribute be in the same file? Seems like they are the same category.

Sergio0694: I've split them based on use. This one specifically gets the IID, whereas the other one only gets a type, and is only needed by the signature generator. To be honest I'm kinda thinking they should all be moved out to some extension type 🤔

manodasanW (approved):

PR #2128, opened 2025-11-24
Optimize 'NoMetadataTypeInfo' cache usage

Sergio0694:
Refactored `NoMetadataTypeInfo` to use a `ConcurrentDictionary` with `concurrencyLevel: 1` and `capacity: 32`, reducing memory usage for XAML scenarios. Improved the cache lookup to avoid unnecessary `string` allocations by using alternate lookup with `ReadOnlySpan<char>` and only allocating when needed.

### DIFF
--- a/src/WinRT.Runtime2/ABI/System/Type.cs
+++ b/src/WinRT.Runtime2/ABI/System/Type.cs
@@ -17,7 +17,7 @@ using WindowsRuntime.InteropServices;
 using WindowsRuntime.InteropServices.Marshalling;
 using static System.Runtime.InteropServices.ComWrappers;
 
-#pragma warning disable IDE1006, CA1416
+#pragma warning disable IDE0008, IDE1006, CA1416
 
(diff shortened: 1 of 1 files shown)


kythant (approved):

src/WinRT.Runtime2/ABI/System/Type.cs
```diff
@@ -346,7 +346,17 @@ private static HRESULT get_Value(void* thisPtr, Type* result)
 /// </summary>
 file sealed class NoMetadataTypeInfo : TypeInfo
 {
-    private static readonly ConcurrentDictionary<string, NoMetadataTypeInfo> FakeMetadataTypeCache = new(StringComparer.Ordinal);
+    /// <summary>
+    /// The cache of <see cref="NoMetadataTypeInfo"/> instances.
+    /// </summary>
+    /// <remarks>
+    /// This cache is mostly only used by XAML, meaning it should pretty much always be accessed from the UI thread.
```
kythant, line 353: Insane work!

PR #2132, opened 2025-11-26
Fix some ABI types in 'cswinrtgen'

Sergio0694:
This PR fixes some incorrect ABI types in 'cswinrtgen' and updates one incorrectly declared local variable.

### DIFF
--- a/src/WinRT.Interop.Generator/Extensions/WindowsRuntimeExtensions.cs
+++ b/src/WinRT.Interop.Generator/Extensions/WindowsRuntimeExtensions.cs
@@ -198,7 +198,7 @@ internal static class WindowsRuntimeExtensions
                     return interopReferences.AbiType.ToValueTypeSignature();
                 }
 
-                // 'Exception' is also a class, but is custom-mapped to the 'HResult' struc type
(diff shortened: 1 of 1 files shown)


src/WinRT.Interop.Generator/Extensions/WindowsRuntimeExtensions.cs
```diff
@@ -171,12 +171,40 @@ public TypeSignature GetAbiType(InteropReferences interopReferences)
...
+            // We have a few special cases to handle for custom-mapped types that are reference types in C#
+            if (typeDefinition.IsClass)
+            {
+                // 'Type' is a class, but is custom-mapped to the 'TypeName' struct type
+                if (SignatureComparer.IgnoreVersion.Equals(type, interopReferences.Type))
+                {
+                    return interopReferences.AbiType.ToValueTypeSignature();
+                }
+
+                // 'Exception' is also a class, but is custom-mapped to the 'HResult' struc type
```
manodasanW:
```suggestion
                // 'Exception' is also a class, but is custom-mapped to the 'HResult' struct type
```

manodasanW (approved):

PR #2136, opened 2025-11-29
Emit marshaller types for `KeyValuePair<,>` types in 'cswinrtgen'

Sergio0694:
This PR implements codegen for the "Marshaller" type for `KeyValuePair<,>` types. It also adds an extension to check whether a managed type requires tracker support when marshaller as a CCW, which is used to optimize CCWs for `KeyValuePair<,>`.

The codegen looks like this (e.g. for `KeyValuePair<int, IList<EnumValue>>`):

```csharp
// Shared type for all accessors for all type arguments
internal static class <#corlib>KeyValuePairMethods
{
    // Shared accessor for 'int' keys
    [MethodImpl(MethodImplOptions.NoInlining)]
    public unsafe static int get_Key(int)(void* P_0)
    {
	    Unsafe.SkipInit(out int result);
	    RestrictedErrorInfo.ThrowExceptionForHR((*(<IKeyValuePairVftbl>**)P_0)->get_Key(P_0, &result));
	    return result;
    }
    
    // Shared accessor for 'IList<EnumValue>' values
    [MethodImpl(MethodImplOptions.NoInlining)]
    public unsafe static IList<EnumValue> get_Value(<#corlib>IList'1<<TestComponentCSharp>TestComponentCSharp-EnumValue>)(void* P_0)
    {
	    Unsafe.SkipInit(out void* ptr);
	    RestrictedErrorInfo.ThrowExceptionForHR((*(<IKeyValuePairVftbl>**)P_0)->get_Value(P_0, &ptr));
	    try
	    {
		    return <#corlib>IList'1<<TestComponentCSharp>TestComponentCSharp-EnumValue>Marshaller.ConvertToManaged(ptr);
	    }
	    finally
	    {
		    WindowsRuntimeUnknownMarshaller.Free(ptr);
	    }
    }
}

// Marshaller for 'KeyValuePair<int, IList<EnumValue>>'
internal static class <#corlib>KeyValuePair'2<int|<#corlib>System-Collections-Generic-IList'1<<TestComponentCSharp>TestComponentCSharp-EnumValue>>Marshaller
{
	public static WindowsRuntimeObjectReferenceValue ConvertToUnmanaged(KeyValuePair<int, IList<EnumValue>> P_0)
	{
		return WindowsRuntimeValueTypeMarshaller.ConvertToUnmanagedUnsafe((object)P_0, CreateComInterfaceFlags.TrackerSupport, ref <InterfaceIIDs>.IID_<#corlib>KeyValuePair'2<int|<#corlib>System-Collections-Generic-IList'1<<TestComponentCSharp>TestComponentCSharp-EnumValue>>);
	}

	public unsafe static KeyValuePair<int, IList<EnumValue>> ConvertToManaged(void* P_0)
	{
		if (P_0 == null)
		{
			return default(KeyValuePair<int, IList<EnumValue>>);
		}
		return new KeyValuePair<int, IList<EnumValue>>(<#corlib>KeyValuePairMethods.get_Key(int)(P_0), <#corlib>KeyValuePairMethods.get_Value(<#corlib>IList'1<<TestComponentCSharp>TestComponentCSharp-EnumValue>)(P_0));
	}
}



```

### DIFF
--- a/src/WinRT.Interop.Generator/Extensions/WindowsRuntimeExtensions.cs
+++ b/src/WinRT.Interop.Generator/Extensions/WindowsRuntimeExtensions.cs
@@ -300,7 +300,7 @@ internal static class WindowsRuntimeExtensions
         }
 
         /// <summary>
-        /// Checks whether a given type is needs tracker support (when marshalled as a CCW).
+        /// Checks whether a given type needs tracker support (when marshalled as a CCW).
         /// </summary>
         /// <param name="interopReferences">The <see cref="InteropReferences"/> instance to use.</param>
         /// <returns>Whether the type requires tracker support.</returns>
@@ -355,7 +355,7 @@ internal static class WindowsRuntimeExtensions
                 return false;
             }
 
-            // The 'TimeSpan' and 'DateTimeOffset' are not blittable, but they're also unmanaged
+            // The 'TimeSpan' and 'DateTimeOffset' types are not blittable, but they're also unmanaged
             if (SignatureComparer.IgnoreVersion.Equals(type, interopReferences.TimeSpan) ||
                 SignatureComparer.IgnoreVersion.Equals(type, interopReferences.DateTimeOffset))
             {


src/WinRT.Interop.Generator/Extensions/WindowsRuntimeExtensions.cs
```diff
@@ -299,6 +299,87 @@ public bool IsManagedValueType(InteropReferences interopReferences)
...
+                if (type is SzArrayTypeSignature arrayType)
+                {
+                    return arrayType.BaseType.IsTrackerSupportRequired(interopReferences);
+                }
+
+                // For all other cases, we assume tracker support is required, to be safe
+                return true;
+            }
+
+            // For generic value types (i.e. 'Nullable<T>' or 'KeyValuePair<,>'), we only need
```
manodasanW, line 329: Is the code path here scoped to custom mapped scenarios?

manodasanW: or to constructed / activatable types?

Sergio0694: Mmh what do you mean? This method can be called with any type (and it assumes that it'd only be called with Windows Runtime types, though it technically doesn't really matter). This code path here handles all generic struct types. So in practice we'll only ever hit this for `Nullable<T>` and `KeyValuePair<TKey, TValue>` instantiations, since those are the only ones that would get past the Windows Runtime checks. Does that make sense?

manodasanW: I was thinking about like `IObservableVector<T>` and if we will ever see that here.

https://learn.microsoft.com/en-us/uwp/api/windows.foundation.collections.iobservablevector-1?view=winrt-26100

Sergio0694: Oh gotcha. That case will go through the branch above (since it's not a value type), and will always just return `true`, since we can't know if the type implementing that interface has any other references or not. This part here is specifically only for constructed generic value types, not any constructed generics.

PR #2143, opened 2025-12-07
Add support for reference projections

manodasanW:
- With this change, projection projects need `CsWinRTGenerateReferenceProjection` set to true.  When this is set, a reference assembly and a forwarder assembly is generated (replaces the output DLL that gets built).
- Components would distribute both these binaries and with project references they would just flow through.
- On the final app project which consumes all the reference projection assemblies, cswinrt is triggered to generate a merged projection of all of them.  The forwarders are also included in the output folder which means any references to the projection DLL are forwarded to the merged projection DLL.
- In the reference projections, we exclude the entire ABI namespace.  This was achieved via changes to codewriters based on detecting a reference projection is being generated.  The changes here could have been more scoped if we just built only a reference assembly with `ProduceOnlyReferenceAssembly` which is possible but that relied on an assumption that if we generated sources which referenced ABI namespaces that aren't there that they will continue to build like they do today and also have more msbuild target changes to enable (seen [here](https://github.com/microsoft/CsWinRT/tree/2e7c611b2f3ffcc231c4450b768cacafd40aaa20)).  So, to keep it straight forward, we do the equivalent in codewriters instead.


src/WinRT.Impl.Generator/Generation/ImplGenerator.cs
```diff
@@ -268,13 +268,40 @@ private static void EmitTypeForwards(ModuleDefinition inputModule, ModuleDefinit
                     Attributes = AsmResolver.PE.DotNet.Metadata.Tables.TypeAttributes.Forwarder
                 });
             }
+
+            TypeReference interfaceIIDsTypeReference = inputModule.CreateTypeReference("ABI"u8, "InterfaceIIDs"u8);
```
Sergio0694: Can you clarify why this is needed? This type should not be in the ref assembly, right?

src/WinRT.Impl.Generator/Generation/ImplGenerator.cs
```diff
@@ -268,13 +268,40 @@ private static void EmitTypeForwards(ModuleDefinition inputModule, ModuleDefinit
...
         }
     }
 
+    /// <summary>
+    /// There are a couple types in the Windows SDK projection which are not public and vtables are generated for.
+    /// This checks whether it is one of those by checking the namespace.
+    /// </summary>
+    /// <param name="exportedType">The type to check.</param>
+    /// <returns>True if the namespace is one of the forwarded internal types; otherwise, false.</returns>
+    private static bool IsForwardedInternalType(TypeDefinition exportedType)
```
Sergio0694, line 284: These should all be temporary, let's see once I move the stream stuff if we can remove this too.

src/WinRT.Projection.Generator/Generation/ProjectionGenerator.Emit.cs
```diff
@@ -0,0 +1,94 @@
...
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.Emit;
+using WindowsRuntime.ProjectionGenerator.Errors;
+
+namespace WindowsRuntime.ProjectionGenerator.Generation;
+
+/// <inheritdoc cref="ProjectionGenerator"/>
+internal partial class ProjectionGenerator
+{
+    private static readonly string ProjectionAssemblyName = "WinRT.Projection";
```
Sergio0694:
```suggestion
    private const string ProjectionAssemblyName = "WinRT.Projection";
```

src/WinRT.Projection.Generator/Generation/ProjectionGenerator.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+        catch (Exception e) when (!e.IsWellKnown)
+        {
+            throw new UnhandledProjectionGeneratorException("emit", e);
+        }
+
+        args.Token.ThrowIfCancellationRequested();
+
+        // Notify the user that generation was successful
+        ConsoleApp.Log($"Projection code generated -> {Path.Combine(args.GeneratedAssemblyDirectory, ProjectionAssemblyName)}");
```
Sergio0694, line 52: Should we add a couple logs before too matching what we do in 'cswinrtgen'? Something like:
- "Processing {count} reference assemblies"
- "Generating projection code"

src/WinRT.Projection.Generator/WinRT.Projection.Generator.csproj
```diff
@@ -0,0 +1,45 @@
...
+    <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
+    <Features>strict</Features>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+
+    <!--
+      Workaround for: "CSC(0,0): Error AD0001: Analyzer 'ILLink.RoslynAnalyzer.DynamicallyAccessedMembersAnalyzer' threw
+      an exception of type 'System.InvalidCastException' with message 'Unable to cast object of type
+      'Microsoft.CodeAnalysis.CSharp.Symbols.PublicModel.NonErrorNamedTypeSymbol' to type 'Microsoft.CodeAnalysis.IMethodSymbol'.'.".
+    -->
+    <NoWarn>$(NoWarn);AD0001</NoWarn>
```
Sergio0694: This should not be needed anymore

src/cswinrt/code_writers.h
```diff
@@ -2611,6 +2647,22 @@ private static % _% = new %("%.%", %.IID);
     void write_activation_factory_objref_definition(writer& w, TypeDef const& classType)
     {
         auto objrefname = w.write_temp("%", bind<write_objref_type_name>(classType));
+        if (settings.reference_projection)
+        {
+            w.write(R"(
+private static WindowsRuntimeObjectReference %
```
Sergio0694, line 2653: `WindowsRuntimeObjectReference` is a private implementation detail type, so we shouldn't use it in the ref assembly. Can we skip emitting this property entirely, since it's private anyway? Like, we shouldn't need to emit any non-public members in general.

src/cswinrt/code_writers.h
```diff
@@ -2636,6 +2688,22 @@ private static WindowsRuntimeObjectReference %
     void write_static_objref_definition(writer& w, TypeDef const& staticsType, TypeDef const& classType)
     {
         auto objrefname = w.write_temp("%", bind<write_objref_type_name>(staticsType));
+        if (settings.reference_projection)
+        {
+            w.write(R"(
+private static WindowsRuntimeObjectReference %
```
Sergio0694, line 2694: Same here as above.

src/WinRT.Impl.Generator/Generation/ImplGenerator.cs
manodasanW: Yea I wanted to have at least something working right now and we can figure out later if we keep it or not.

src/WinRT.Impl.Generator/Generation/ImplGenerator.cs
manodasanW: I believe this was from before I fixed the IID referenced by CsWinRTGen.  Seems to be working without this.

src/cswinrt/code_writers.h
manodasanW: My goal was to leave as much as to the .NET reference assembly thing to remove things that are not on the public API surface.  So these should still be removed in the reference assembly.  But I will note that the protected constructors do show up which do reference WindowsRuntimeObjectReference.  I wonder if we should remove those in ref assemblies or not.

src/cswinrt/code_writers.h
manodasanW: This should also get removed in the .NET reference assembly.

src/cswinrt/code_writers.h
Sergio0694: Mmmhhhhh. I would _think_ we should be able to remove them, since at runtime the method would still be there anyway. The only difference is that user code would not be able to see them. Which honestly is even better. @jkoritzinsky can you confirm that it's fine for us to omit those constructors in the ref assemblies?

src/cswinrt/strings/additions/Microsoft.UI.Xaml.Media.Animation/Microsoft.UI.Xaml.Media.Animation.KeyTime.cs
```diff
@@ -5,7 +5,9 @@ namespace Microsoft.UI.Xaml.Media.Animation
 
     [WindowsRuntimeMetadata("Microsoft.UI")]
     [WindowsRuntimeClassName("Windows.Foundation.IReference<Microsoft.UI.Xaml.Media.Animation.KeyTime>")]
```
Sergio0694: All of these attributes should also be hidden from reference assemblies, right?

src/WinRT.Projection.Generator/Generation/ProjectionGenerator.Emit.cs
```diff
@@ -0,0 +1,99 @@
...
+
+    private static CSharpCompilation CreateCompilationForProjection(string sourcesFolder, string[] referencePaths)
+    {
+        try
+        {
+            // Parse the source files into a syntax tree
+            List<SyntaxTree> syntaxTrees = [];
+            foreach (string file in Directory.GetFiles(sourcesFolder, "*.cs"))
+            {
+                syntaxTrees.Add(CSharpSyntaxTree.ParseText(File.ReadAllText(file), path: file));
```
Sergio0694: Nit: you should be able to skip reading the file to a `string` if you do `ParseText` with the overload taking a `Stream`, so you can directly call that over a readable stream for the file. So the whole `string` will never materialize.

src/WinRT.Projection.Generator/Generation/ProjectionGenerator.Emit.cs
```diff
@@ -0,0 +1,99 @@
...
+
+    /// <summary>
+    /// Runs the emit logic for the generator.
+    /// </summary>
+    /// <param name="args">The arguments for this invocation.</param>
+    private static void Emit(ProjectionGeneratorArgs args)
+    {
+        args.Token.ThrowIfCancellationRequested();
+
+        ConsoleApp.Log("Generating merged projection sources");
```
Sergio0694: Nit: remove this, since it'd be printed right after the message in the caller (there's no work in between).

src/cswinrt/strings/additions/Microsoft.UI.Xaml.Media.Animation/Microsoft.UI.Xaml.Media.Animation.KeyTime.cs
manodasanW: Technically we use these in cswinrtgen right?

src/WinRT.Interop.Generator/References/InteropReferences.cs
```diff
@@ -88,6 +88,15 @@ public InteropReferences(
         publicKey: false,
         publicKeyOrToken: WellKnownPublicKeyTokens.SystemMemory);
 
+    /// <summary>
+    /// Gets the <see cref="AssemblyReference"/> for <c>WinRT.Projection.dll</c>.
+    /// </summary>
+    public AssemblyReference WinRTProjection => field ??= new AssemblyReference(
+        name: "WinRT.Projection"u8,
+        version: new Version(0, 0, 0, 0),
```
Sergio0694, line 96: Thinking this should rather match the .dll version with WinRT.Interop.dll? You can copy the logic we use for that.

src/WinRT.Interop.Generator/Helpers/GuidGenerator.cs
```diff
@@ -58,15 +57,6 @@ public static bool TryGetIIDFromWellKnownInterfaceIIDsOrAttribute(ITypeDescripto
...
         {
-            // For delegate types, we resolve the IID from their generated RVA field.
-            // Only interface types have the '[Guid]' attribute on them, not delegates.
-            if (typeDefinition.IsClass && typeDefinition.IsDelegate)
-            {
-                iid = InterfaceIIDResolver.GetIID(typeDefinition);
-
-                return true;
-            }
-
```
Sergio0694, line 77: Why was this removed? Don't we still need this for delegates?

manodasanW: This was looking at interface iids to get the value, but with reference assemblies, it is not implemented anymore and thereby this was failing.  So I instead put guid attributes for delegates in reference assemblies.

Sergio0694: Mmh wait I don't think this will work, because reference assemblies will also **not** have a bunch of other attributes that we need (e.g. `[WindowsRuntimeMetadata]`). We'll need to figure out what changes to do to 'cswinrtgen' so that we can correctly follow the type forwards and resolve `TypeDefinition`-s from the merged projection .dll I think 🤔

src/cswinrt/code_writers.h
manodasanW: Addressed.

src/WinRT.Interop.Generator/References/InteropReferences.cs
manodasanW: Technically we can, but I don't pass the project dll to the projection generator today and it is generated differently from WinRT.Interop due to it uses the CSharpCompilation to do so.  We can technically pass a parameter for the version and then add a version attribute which is passed to the CSharpCompilation if we want, but does it add any benefit?

src/cswinrt/strings/additions/Microsoft.UI.Xaml.Media.Animation/Microsoft.UI.Xaml.Media.Animation.KeyTime.cs
Sergio0694: Yes but I feel like we should still find a way to hide them. In theory both the metadata and the runtime class name can be derived at compile time anyway. We could check if a type is a projected type just by checking if the defining assembly is a projection assembly, and the runtime class name we can just compute in 'cswinrtgen' easily, we already have most of the logic for it anyway. That would allow us to just keep these attributes in the merged projection and hide them from here.

manodasanW: Removed WindowsRuntimeClassName.  Discussed offline and agreed to keep metadata as it could still be helpful to have in a reference assembly to say which winmd the type comes from.

src/WinRT.Interop.Generator/References/InteropReferences.cs
Sergio0694: I suppose it seems "more correct", and could further enforce the fact the two .dll-s must be in sync?

src/cswinrt/code_writers.h
Sergio0694, line 8558: Shouldn't we also skip the IDIC implementation types for ref assemblies?

src/WinRT.Interop.Generator/Helpers/GuidGenerator.cs
```diff
@@ -66,15 +65,6 @@ public static bool TryGetIIDFromWellKnownInterfaceIIDsOrAttribute(
 
         if (type.Resolve() is TypeDefinition typeDefinition)
         {
-            // For delegate types, we resolve the IID from their generated RVA field.
-            // Only interface types have the '[Guid]' attribute on them, not delegates.
-            if (typeDefinition.IsClass && typeDefinition.IsDelegate)
-            {
-                iid = InterfaceIIDResolver.GetIID(typeDefinition);
```
Sergio0694, line 73: If we're removing this, you should also delete the `InterfaceIIDResolver` type then.

src/WinRT.Interop.Generator/References/InteropReferences.cs
```diff
@@ -88,6 +88,15 @@ public InteropReferences(
...
+        publicKey: false,
+        publicKeyOrToken: default);
```
Sergio0694, line 98: Don't we have the public key token for it? We'd always be using our own, right?

src/cswinrt/code_writers.h
manodasanW: The whole ABI namespace writing part is skipped at a higher level, so it doesn't even get called.

src/WinRT.Interop.Generator/References/InteropReferences.cs
manodasanW: The two dlls are not exactly in sync - one is just based on projections / winmds referenced while other is based on the entire app and everything that is used by the app.

Also we probably don't want the user changing the version in the app to cause for a rebuild of the merged projections if the winmds / referenced projections haven't changed with respect to incremental builds.

Sergio0694: I see. I guess we can leave it as is for now then. What version will the merged projection use, just 0.0.0.0?

manodasanW: Yea right now it is 0.0.0.0.

src/WinRT.Interop.Generator/References/InteropReferences.cs
manodasanW: Right now, we are not strong name signing the dll.  If we do decide to, then this would need to be updated.  But I will defer strong name signing it to a separate PR due to it does seem a bit involved and also to figure out if we really should if it won't be validated.

Sergio0694: Yeah let's do that in another PR, that's fine 🙂

> "if it won't be validated"

Well, .NET won't, but we can validate it from cswinrtgen to ensure we are picking up the right binary.

PR #2152, opened 2025-12-11
Fix checks for manually-projected Windows Runtime types

Sergio0694:
We accidentally broke all async interfaces and observable interfaces in 'cswinrtgen'. This enables them again.

### DIFF
--- a/src/WinRT.Runtime2/Windows.Foundation/PropertyType.cs
+++ b/src/WinRT.Runtime2/Windows.Foundation/PropertyType.cs
@@ -1,7 +1,6 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-using System;
 using System.ComponentModel;
 using System.Runtime.Versioning;
 using Windows.Foundation.Metadata;
@@ -19,9 +18,6 @@ namespace Windows.Foundation;
 [WindowsRuntimeMetadata("Windows.Foundation.FoundationContract")]
 [SupportedOSPlatform("Windows10.0.10240.0")]
 [ContractVersion(typeof(FoundationContract), 65536u)]
-[Obsolete(WindowsRuntimeConstants.PrivateImplementationDetailObsoleteMessage,
-    DiagnosticId = WindowsRuntimeConstants.PrivateImplementationDetailObsoleteDiagnosticId,
-    UrlFormat = WindowsRuntimeConstants.CsWinRTDiagnosticsUrlFormat)]
 [EditorBrowsable(EditorBrowsableState.Never)]
 public enum PropertyType : uint
 {


src/WinRT.Runtime2/Windows.Foundation/PropertyType.cs
```diff
@@ -18,6 +19,9 @@ namespace Windows.Foundation;
 [WindowsRuntimeMetadata("Windows.Foundation.FoundationContract")]
 [SupportedOSPlatform("Windows10.0.10240.0")]
 [ContractVersion(typeof(FoundationContract), 65536u)]
+[Obsolete(WindowsRuntimeConstants.PrivateImplementationDetailObsoleteMessage,
+    DiagnosticId = WindowsRuntimeConstants.PrivateImplementationDetailObsoleteDiagnosticId,
+    UrlFormat = WindowsRuntimeConstants.CsWinRTDiagnosticsUrlFormat)]
```
manodasanW: This is a projected type, so probably don't want to mark this as implementation detail.

Sergio0694: It's a special projected type, it's always been hidden (even in the past) and should never be used directly. It's the same as `IPropertyValue`, which is not projected at all. The only reason this exists is because it appears in vtable signatures.

manodasanW: There was a version of it that was projected in the Windows SDK projection.  It is used by https://learn.microsoft.com/en-us/uwp/api/windows.graphics.imaging.bitmaptypedvalue.type?view=winrt-26100#windows-graphics-imaging-bitmaptypedvalue-type.

manodasanW: The IPropertyValue use case was indeed hidden within WinRT.Runtime before but the enum was still projected in the Windows SDK proection.

Sergio0694: Oh interesting. In that case then we should probably support this for real, not like before. As in, supporting marshalling it too like any other enum type. I'll just revert that change here and can make another PR for it.

manodasanW: We should probably also remove EditorBrowsable.

Sergio0694: Sure, but I was thinking I can address all of that in the other PR?
Like, the changes in this one are unrelated in theory.

manodasanW: Before, we just had 2 versions of this enum projection (one internal to WinRT.Runtime for our uses with IPropertyValue and one actually in the Windows SDK projection projected like any other enum type).

Sergio0694: Why do we need 2 though? Like, why not just have the real public one? It's already in WinRT.Runtime.dll anyway. Wouldn't that be the same as any other manually-projected types we have here too that we use in internal places too?

PR #2157, opened 2025-12-16
Add exclusion list for types in 'cswinrtgen

Sergio0694:
This PR adds an extensible way to special-case types to exclude in 'cswinrtgen', such as `Task<T>` instantiations.

### DIFF
--- a/src/WinRT.Interop.Generator/Helpers/TypeExclusions.cs
+++ b/src/WinRT.Interop.Generator/Helpers/TypeExclusions.cs
@@ -34,15 +34,17 @@ internal static class TypeExclusions
             return IsExcluded(arraySignature.BaseType, interopReferences);
         }
 
-        // Check if the input type matches any of our exclusions
-        foreach (TypeReference excludedType in (ReadOnlySpan<TypeReference>)[interopReferences.Task1])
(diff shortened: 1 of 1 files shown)


src/WinRT.Interop.Generator/Helpers/TypeExclusions.cs
```diff
@@ -0,0 +1,48 @@
...
+        }
+
+        // Also handle SZ arrays and check their element type
+        if (type is SzArrayTypeSignature arraySignature)
+        {
+            return IsExcluded(arraySignature.BaseType, interopReferences);
+        }
+
+        // Check if the input type matches any of our exclusions
+        foreach (TypeReference excludedType in (ReadOnlySpan<TypeReference>)[interopReferences.Task1])
```
manodasanW: Should this be in a separate global variable or so if we are planning on adding to it or is it just this?

Sergio0694: We can add more over time (e.g. just added `ConditionalWeakTable<,>` too).
I've refactored this and added some comments too to explain why we need it here 🙂

PR #2177, opened 2026-01-14
Mapping RuntimeClassNames of Instantiated Generic Types 

kythant:
Map System.* to Windows.Foundation.* in Instantiated Generic Types. Added a new helper in CSWinRTGen called RuntimeClassNameGenerator.cs to faciliate this. 

For fundamental WinRT types such as System.Int32, it needs to map to just Int32. Updated TypeMapping.cs to reflect this case.


src/WinRT.Interop.Generator/Helpers/RuntimeClassNameMapping.cs
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Runtime.CompilerServices;
+using AsmResolver.DotNet.Signatures;
+
+namespace WindowsRuntime.InteropGenerator.Helpers;
+
+internal class RuntimeClassNameMapping
```
Sergio0694: Make this static
```suggestion
internal static class RuntimeClassNameMapping
```

Sergio0694: Can we name this `RuntimeClassNameGenerator` to be consistent with `GuidGenerator`? Also update the method name in the same way for consistency, since the two are also next to each other.

src/WinRT.Interop.Generator/Helpers/RuntimeClassNameMapping.cs
```diff
@@ -0,0 +1,57 @@
...
+    /// </returns>
+    public static string GetMappedGenericInstanceRuntimeClassName(TypeSignature type, bool useWindowsUIXamlProjections)
+    {
+        if (type is GenericInstanceTypeSignature genericInstanceType
+            && TypeMapping.TryFindMappedTypeName(genericInstanceType.GenericType.FullName, useWindowsUIXamlProjections, out string? mappedTypeName))
+        {
+            if (genericInstanceType.TypeArguments.Count == 0)
+            {
+                return mappedTypeName;
+            }
```
Sergio0694: This shouldn't be needed, you can't have a generic type with 0 type arguments.

src/WinRT.Interop.Generator/Helpers/TypeMapping.cs
```diff
@@ -117,6 +135,12 @@ public static bool TryFindMappedTypeName(
         bool useWindowsUIXamlProjections,
         [NotNullWhen(true)] out string? mappedName)
     {
+        if (FundamentalTypeMapping.GetAlternateLookup<ReadOnlySpan<char>>().TryGetValue(fullName, out string? primitiveMappedResult))
+        {
+            mappedName = primitiveMappedResult;
+            return true;
```
Sergio0694: Nit: leave a blank line like below
```suggestion
            mappedName = primitiveMappedResult;

            return true;
```

src/WinRT.Interop.Generator/Helpers/TypeMapping.cs
```diff
@@ -117,6 +135,12 @@ public static bool TryFindMappedTypeName(
         bool useWindowsUIXamlProjections,
         [NotNullWhen(true)] out string? mappedName)
     {
+        if (FundamentalTypeMapping.GetAlternateLookup<ReadOnlySpan<char>>().TryGetValue(fullName, out string? primitiveMappedResult))
```
Sergio0694: Nit: rename the value for consistency
```suggestion
        if (FundamentalTypeMapping.GetAlternateLookup<ReadOnlySpan<char>>().TryGetValue(fullName, out string? fundamentalTypeName))
```

src/WinRT.Interop.Generator/Helpers/RuntimeClassNameMapping.cs
```diff
@@ -0,0 +1,57 @@
...
+
+            DefaultInterpolatedStringHandler handler = $"{mappedTypeName}<";
+
+            handler.AppendLiteral(GetMappedGenericInstanceRuntimeClassName(genericInstanceType.TypeArguments[0], useWindowsUIXamlProjections));
```
Sergio0694: This isn't super efficient, as we're instantiating all strings for the runtime class name and then appending them. Could we rather make this method use a private recursive helper that takes the interpolated string handler and just writes directy to it? You can see an example of this same pattern in `InteropUtf8NameFactory.TypeName`, see all the various `Append___` helpers.

src/WinRT.Interop.Generator/Helpers/RuntimeClassNameMapping.cs
```diff
@@ -0,0 +1,57 @@
...
+    /// <param name="useWindowsUIXamlProjections">Whether to use <c>Windows.UI.Xaml</c> projections.</param>
+    /// <returns>
+    /// The mapped runtime class name. For generic instances with a mapped generic type,
+    /// returns the mapped name with type arguments (e.g., <c>Namespace.Type&lt;TArg&gt;</c>).
+    /// If no mapping exists, returns <paramref name="type"/>.<see cref="TypeSignature.FullName"/>.
+    /// </returns>
+    public static string GetMappedGenericInstanceRuntimeClassName(TypeSignature type, bool useWindowsUIXamlProjections)
+    {
+        if (type is GenericInstanceTypeSignature genericInstanceType
+            && TypeMapping.TryFindMappedTypeName(genericInstanceType.GenericType.FullName, useWindowsUIXamlProjections, out string? mappedTypeName))
```
Sergio0694: I don't think this check is correct. There's generic interfaces (and delegate types) that are not custom mapped, so presumably they wouldn't get a mapped type name here. Which means they'd go to the path below and end up being interpolated like a non-generic type, which would likely cause the type arguments to be formatted incorrectly. I think you could fix this along with the refactor to make this recursive: you just interpolate the current part of the type and then just delegate to recursive calls for any type arguments, and they'll all recursively do the right thing.

Sergio0694 (changes requested):

src/WinRT.Interop.Generator/Helpers/RuntimeClassNameMapping.cs
kythant: It didn't feel right to call it a Generator because everything in there was about mapping.

src/WinRT.Interop.Generator/Helpers/RuntimeClassNameMapping.cs
kythant: That makes sense.

kythant: Should I leave an assert?

src/WinRT.Interop.Generator/Helpers/RuntimeClassNameMapping.cs
Sergio0694: I mean it's generating a name, much like the IID generator is. One could say the IID is also "mapped" in theory.

src/WinRT.Interop.Generator/Helpers/RuntimeClassNameMapping.cs
Sergio0694: No need, it can't happen or we wouldn't have a `GenericInstanceTypeSignature`

src/WinRT.Interop.Generator/Helpers/RuntimeClassNameMapping.cs
kythant: Wait so if we are given a type that is just System.Int32, we don't convert that to Int32?

Sergio0694: We do, but that wouldn't be a `GenericInstanceTypeSignature`.

kythant: Right I think this code is already converting it to Int32 right? Unless we don't want it to convert to Int32 if that's what you are saying.

kythant: Oh wait so are you saying if it is something like 
TestComponent<System.Int32> 

That we want it to be 
TestComponent<Int32>?

Since TestComponent isn't part of the Custom mapped type that this code wouldn't do the right thing.

src/WinRT.Interop.Generator/Helpers/RuntimeClassNameGenerator.cs
```diff
@@ -0,0 +1,75 @@
...
+    /// </param>
+    /// <param name="useWindowsUIXamlProjections">Whether to use <c>Windows.UI.Xaml</c> projections.</param>
+    /// <returns>
+    /// The mapped runtime class name. For generic instances with a mapped generic type,
+    /// returns the mapped name with type arguments (e.g., <c>Namespace.Type&lt;TArg&gt;</c>).
+    /// If no mapping exists, returns <paramref name="type"/>.<see cref="TypeSignature.FullName"/>.
+    /// </returns>
+    public static string GetMappedGenericInstanceRuntimeClassName(TypeSignature type, bool useWindowsUIXamlProjections)
+    {
+        DefaultInterpolatedStringHandler handler = new(0, 0, null, stackalloc char[type.FullName.Length * 2]);
```
Sergio0694: Never use dynamic size for `stackalloc`, it produces suboptimal code because the JIT can't unroll it. Just allocate a fixed amount, like `stackalloc char[256]` instead.

src/WinRT.Interop.Generator/Helpers/RuntimeClassNameGenerator.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        GetGenericInstanceRuntimeClassNameHelper(type, useWindowsUIXamlProjections, ref handler);
+
+        return handler.ToStringAndClear();
+    }
+
+    private static void GetGenericInstanceRuntimeClassNameHelper(TypeSignature type, bool useWindowsUIXamlProjections, ref DefaultInterpolatedStringHandler interpolatedStringHandler)
+    {
+        if (type is GenericInstanceTypeSignature genericInstanceTypeSignature)
```
Sergio0694: We should also check for SZ arrays and return `IReferenceArray<...>`, right? Since we can have CCWs for those too.

src/WinRT.Interop.Generator/Helpers/RuntimeClassNameGenerator.cs
kythant: Good thing to know! Thanks!

src/WinRT.Interop.Generator/Helpers/TypeMapping.cs
```diff
@@ -117,6 +135,13 @@ public static bool TryFindMappedTypeName(
         bool useWindowsUIXamlProjections,
         [NotNullWhen(true)] out string? mappedName)
     {
+        if (FundamentalTypeMapping.GetAlternateLookup<ReadOnlySpan<char>>().TryGetValue(fullName, out string? fundamentalMappedResult))
+        {
+            mappedName = fundamentalMappedResult;
+
+            return true;
+        }
```
Sergio0694: Just realized you can simplify this
```suggestion
        if (FundamentalTypeMapping.GetAlternateLookup<ReadOnlySpan<char>>().TryGetValue(fullName, out mappedName))
        {
            return true;
        }
```

src/WinRT.Interop.Generator/Helpers/TypeMapping.cs
```diff
@@ -83,6 +83,24 @@ private readonly record struct MappedType(
         new("Windows.Foundation.Size", new("Windows.Foundation", "Size", "struct(Windows.Foundation.Size;f4;f4)")),
         new("Windows.Foundation.Rect", new("Windows.Foundation", "Rect", "struct(Windows.Foundation.Rect;f4;f4;f4;f4)")));
 
+    /// <summary>
+    /// Mapping of System.* to their corresponding WinRT Fundamental Types
```
Sergio0694: Nit
```suggestion
    /// Mapping of built-in <c>System.*</c> to their corresponding Windows Runtime fundamental types.
```

PR #2188, opened 2026-01-22
Fixes to get more tests passing

manodasanW:
- Update the UnsafeAccessor signatures in projections to make up with what cswinrtgen is generating for arrays
- Fix IReference / IReferenceArray missing `1
- Fix issue where MarshallingInfo was getting reused across different calls
- Fix signature of ReadOnlySpan
- Treat bool as blittable as projections will be compiled with marshaling disabled.


src/cswinrt/code_writers.h
```diff
@@ -1212,10 +1210,6 @@ namespace cswinrt
                 }
                 else
                 {
-                    if (type == fundamental_type::Boolean)
-                    {
-                        type = fundamental_type::UInt8;
-                    }
                     if (type == fundamental_type::Char)
                     {
                         type = fundamental_type::UInt16;
```
Sergio0694: We should also consider `char` as blittable, should we update that in this same PR?

src/WinRT.Runtime2/InteropServices/WindowsRuntimeComWrappers.cs
```diff
@@ -164,6 +164,8 @@ public nint GetOrCreateComInterfaceForObject(object instance, in Guid iid)
     /// <seealso cref="ComWrappers.GetOrCreateComInterfaceForObject"/>
     public nint GetOrCreateComInterfaceForObject(object instance, CreateComInterfaceFlags flags, in Guid iid)
     {
+        MarshallingInfo = null;
```
Sergio0694, line 167: Why is this needed here? I thought we were setting the right things on the caller side. Don't we have callers for this method that are setting that field? 🤔

manodasanW: None of the callers seem to be setting it.  The other variants of GetOrCreateComInterfaceForObject we have do set it themselves based on type.  But the comment below makes it seem like it is expected in this scenario there is no marshaling info set and is left to CreateObject, so I set it to null.

kythant (approved):

src/cswinrt/code_writers.h
manodasanW: Addressed

PR #2189, opened 2026-01-23
Implement 'GetMany' methods for all collection types in 'cswinrtgen'

Sergio0694:
Title. This implements `GetMany` for:
- `IEnumerator<T>`
- `IList<T>`
- `IReadOnlyList<T>`

### DIFF
--- a/src/WinRT.Interop.Generator/References/InteropReferences.cs
+++ b/src/WinRT.Interop.Generator/References/InteropReferences.cs
@@ -4543,10 +4543,10 @@ internal sealed class InteropReferences
     }
 
     /// <summary>
-    /// Gets the <see cref="MemberReference"/> for <c>WindowsRuntime.InteropServices.IListAdapterBlittableValueTypeExtensions.GetMany&lt;T&gt;</c>.
+    /// Gets the <see cref="MethodSpecification"/> for <c>WindowsRuntime.InteropServices.IListAdapterBlittableValueTypeExtensions.GetMany&lt;T&gt;</c>.
     /// </summary>
     /// <param name="elementType">The input element type.</param>
-    public MemberReference IListAdapterBlittableValueTypeGetMany(TypeSignature elementType)
+    public MethodSpecification IListAdapterBlittableValueTypeGetMany(TypeSignature elementType)
     {
         return IListAdapterBlittableValueTypeExtensions
             .CreateMemberReference("GetMany"u8, MethodSignature.CreateStatic(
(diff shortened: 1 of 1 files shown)


src/WinRT.Interop.Generator/Factories/InteropTypeDefinitionFactory.IEnumeratorElementMarshaller.cs
```diff
@@ -0,0 +1,261 @@
...
+            TypeSignature elementType,
+            TypeSignature interfaceType,
+            MemberReference convertToUnmanagedInterfaceMethod,
+            bool isValueType,
+            InteropReferences interopReferences,
+            InteropGeneratorEmitState emitState,
+            ModuleDefinition module)
+        {
+            // Select the attributes and base type depending on whether we want a value type or not
+            (TypeAttributes attributes, ITypeDefOrRef baseType) = isValueType
```
manodasanW, line 212: Why does the ElementMarshaller type base type switch when value type?  What does it look like?

Sergio0694: The `isValueType` parameter refers to the element marshaller type itself, not the element type. That is, it indicates whether we want to declare the element marshaller type as a `struct` or as a `class`. The reason why we need to also change the base type is that a `struct` in .NET is just a class that derives from `System.ValueType` instead of `object`, so we're toggling that here along with the type attributes 🙂

src/WinRT.Runtime2/InteropServices/Collections/IEnumeratorAdapterExtensions.cs
```diff
@@ -0,0 +1,389 @@
...
+            if (itemsSize == 0)
+            {
+                return 0;
+            }
+
+            ArgumentNullException.ThrowIfNull(items);
+
+            uint index = 0;
+
+            for (; index < itemsSize & adapter.HasCurrent; index++)
```
kythant, line 120: This doesn't need a try catch w/ dispose or free?

Sergio0694: It doesn't, because `Exception` is special (that's also why it has a specialized implementation). That is, `Exception` is a class in C#, but its ABI type is `HResult`, which is a blittable value type, so we can just marshal values without worrying about disposing anything, as there's no unmanaged resources we allocated.

PR #2194, opened 2026-01-24
Support enumerators without marshalling info in 'cswinrtgen'

Sergio0694:
This brings them to parity with CsWinRT 2.x, especially given in some cases (e.g. arrays) there's no way to ever see the types.


src/WinRT.Interop.Generator/References/InteropReferences.cs
```diff
@@ -3886,6 +3896,40 @@ public MemberReference IAsyncOperationWithProgressMethodsImpl2GetResults(TypeSig
...
+    /// <param name="elementType">The input element type.</param>
+    /// <param name="iterableMethods">The <see cref="IIterableMethodsImpl1"/> type.</param>
+    public MethodSpecification IEnumerableMethods1GetEnumerator(TypeSignature elementType, TypeSignature iterableMethods)
+    {
+        return IEnumerableMethods1
+            .MakeGenericReferenceType(elementType)
+            .ToTypeDefOrRef()
+            .CreateMemberReference("GetEnumerator"u8, MethodSignature.CreateStatic(
+                returnType: IEnumerator1.MakeGenericReferenceType(new GenericParameterSignature(GenericParameterType.Type, 0)),
+                genericParameterCount: 1,
```
manodasanW, line 3928: is that right?

manodasanW: Looks like it is from below.

PR #2198, opened 2026-01-26
Fix covariant expansion for recursive type definitions in 'cswinrtgen'

Sergio0694:
Title. Review PR with **whitespaces off**.

### DIFF
--- a/src/WinRT.Interop.Generator/Discovery/InteropTypeDiscovery.cs
+++ b/src/WinRT.Interop.Generator/Discovery/InteropTypeDiscovery.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Concurrent;
+using System.Linq;
 using AsmResolver.DotNet;
 using AsmResolver.DotNet.Signatures;
 using WindowsRuntime.InteropGenerator.Errors;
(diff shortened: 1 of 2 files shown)


src/WinRT.Interop.Generator/Discovery/InteropTypeDiscovery.cs
```diff
@@ -181,7 +180,7 @@ public static void TryTrackExposedUserDefinedType(
             // - 'IEnumerable<IEnumerable<object>>'
             // - 'IEnumerable<IEnumerable>'
             // - 'IEnumerable<IDisposable>'
-            foreach (TypeSignature covariantInterfaceSignature in WindowsRuntimeTypeAnalyzer.EnumerateCovariantInterfaceTypes(interfaceSignature, interopReferences).Concat([interfaceSignature]))
+            foreach (TypeSignature covariantInterfaceSignature in WindowsRuntimeTypeAnalyzer.EnumerateCovarianceExpandedInterfaceTypes(interfaceSignature, interopReferences))
```
manodasanW: When does interfaceSignature get added given you removed the concat?  EnumerateCovarianceExpandedInterfaceTypes filters it just to covariant interfaces right, so if it isn't one, it doesn't get included all or am I missing something?

Sergio0694: Yeah you're actually bad, it's my fault for trusting a Copilot PR review comment too much 🙃

PR #2200, opened 2026-01-26
Interop fixes

manodasanW:
- Fixed unsafe accessor for IID and mapped types
- Went through a bunch of unit test failures and fixed a couple interop fixes

### DIFF
--- a/src/WinRT.Interop.Generator/Factories/InteropMethodDefinitionFactory.IKeyValuePair2Impl.cs
+++ b/src/WinRT.Interop.Generator/Factories/InteropMethodDefinitionFactory.IKeyValuePair2Impl.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System;
 using AsmResolver;
 using AsmResolver.DotNet;
 using AsmResolver.DotNet.Code.Cil;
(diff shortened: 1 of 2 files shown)


src/WinRT.Interop.Generator/Factories/InteropMethodDefinitionFactory.IKeyValuePair2Impl.cs
```diff
@@ -108,7 +108,7 @@ private static MethodDefinition get_KeyOrValue(
                     memberName: methodName,
                     signature: MethodSignature.CreateInstance(new GenericParameterSignature(
                         parameterType: GenericParameterType.Type,
-                        index: keyValuePairType.TypeArguments.IndexOf(keyOrValueType))));
+                        index: methodName == "get_Key" ? 0 : 1)));
```
Sergio0694: Mmh why is this needed? Is it because I forgot to pass the signature comparer to `IndexOf`?

manodasanW: When both were the same type argument (i.e string), it was choosing the wrong index (the first one).  Would passing SignatureComparer resolve that?

Sergio0694: Ooh gotcha, yeah no that makes sense. Good find! 🙂

PR #2209, opened 2026-01-30
Fix vtable for async and fix ordering of catch handler

manodasanW:
- Fix ordering of catch handle causing invalid IL error
- Fix vtable ordering for async interfaces causing nullreferenceexception due to wrong method being called
- Fix version being used to look up WinRT.Runtime2 as it was causing runtime errors with FileNotFound given it doesn't seem we are using 3.0.0 right now.

### DIFF
--- a/src/WinRT.Interop.Generator/Factories/InteropMethodDefinitionFactory.IAsyncInfoMethods.cs
+++ b/src/WinRT.Interop.Generator/Factories/InteropMethodDefinitionFactory.IAsyncInfoMethods.cs
@@ -224,6 +224,7 @@ internal partial class InteropMethodDefinitionFactory
                 },
                 ExceptionHandlers =
                 {
+                    // TODO: Reverse order once AsmResolver ordering issue is resolved
                     new CilExceptionHandler
                     {
                         HandlerType = CilExceptionHandlerType.Finally,


src/WinRT.Interop.Generator/Factories/InteropMethodDefinitionFactory.IAsyncInfoMethods.cs
```diff
@@ -224,21 +224,21 @@ public static MethodDefinition set_Handler(
...
-                        HandlerEnd = ret_finallyEnd.CreateLabel()
-                    },
                     new CilExceptionHandler
                     {
                         HandlerType = CilExceptionHandlerType.Finally,
                         TryStart = ldloca_s_0_tryStart.CreateLabel(),
                         TryEnd = ldloca_s_1_finallyStart.CreateLabel(),
                         HandlerStart = ldloca_s_1_finallyStart.CreateLabel(),
                         HandlerEnd = ldloca_s_0_finallyStart.CreateLabel()
+                    },
```
Sergio0694, line 235: Can you add a TODO here to revert this once we get the new AsmResolver version?

src/WinRT.Interop.Generator/Generation/InteropGenerator.Discover.cs
```diff
@@ -312,12 +312,12 @@ private static void ValidateWinRTRuntimeDllVersion2References(InteropGeneratorAr
     private static InteropReferences CreateDiscoveryInteropReferences(ModuleDefinition module)
     {
         // Create the interop references scoped to this module, which we need to lookup some references from
-        // the 'WinRT.Runtime.dll' assembly. We haven't loaded it just here here, so we can't use the real
+        // the 'WinRT.Runtime.dll' assembly. We haven't loaded it just here, so we can't use the real
```
Sergio0694: I think I made a typo here, can you fix it while you're at it?
```suggestion
        // the 'WinRT.Runtime.dll' assembly. We haven't loaded it here just yet, so we can't use the real
```

Sergio0694 (approved):

PR #2213, opened 2026-01-31
Fix some native parameter two-pass rewriting in 'cswinrtgen'

Sergio0694:
Title. This fixes two issues:
- Bad IL when unpinning `string` and `Type` parameters
- Incorrect call to `Dispose` for unmanaged value types


src/WinRT.Interop.Generator/Rewriters/InteropMethodRewriter.NativeParameter.cs
```diff
@@ -328,16 +344,37 @@ private static void RewriteBodyForTypeOfString(
                 ldloca_s_getHStringReference,
                 new CilInstruction(Call, interopReferences.HStringMarshallerConvertToUnmanagedUnsafe.Import(module))]);
 
-            // Get the 'HString' value from the reference and pass it as a parameter
+            // Get the 'HString' value from the reference and pass it as a parause a protected region for to unpinmeter
```
manodasanW: typo?

Sergio0694: Oops, yeah 😅

PR #2217, opened 2026-02-02
More interop fixes

manodasanW:
Various fixes from debugging unit tests
- Fix vtable ordering
- Add guid attribute to IDIC types
- Update AsmResolver version to address try block ordering issues
- Implement IMarshal interface
- Fix ICollection clear calls to use CallVirt
- Implement clear function for ICollection<keyvaluepair>
- Fix interop assembly names used in unsafe accessors to match what cswinrgen is generating
- Fix IEnumerator for both scenarios where internal .NET types are used and also to support the generic IEnumerator interface as expected by WinUI

### DIFF
--- a/docs/cswinrt3.0-spec.md
+++ b/docs/cswinrt3.0-spec.md
@@ -109,7 +109,8 @@ In CsWinRT 3.0, we worked with the .NET team to complete the proposal for adding
 The API surface of projected runtime classes will be significantly streamlined compared to CsWinRT 2.x. In particular:
 - **`IWinRTObject` and all related infrastructure in each projected runtime class will be removed.** All the shared functionality will be provided by the base `WindowsRuntimeObject` class, which will also handle `IDynamicInterfaceCastable` caching. This also reduces the object size for each runtime class type, as well as improving binary size.
 - **`As<I>()` will be removed.** This functionality is available via `WindowsRuntimeActivationFactory`.
-- **`FromAbi(nint)` will be removed.** This functionality is available via the marshaller type for each runtime class type.
+- **`*.FromAbi(nint)` will be removed.** Native objects can be marshalled to managed objects via `WindowsRuntimeMarshal.ConvertToManaged`.
+- **`*.FromManaged(object)` will be removed.** Managed objects can be marshalled to native objects via `WindowsRuntimeMarshal.ConvertToUnmanaged`. It will be up to callers to perform any additional `QueryInterface` calls that may be needed.
 - **All `IEquatable<T>` support will be removed.** Projected runtime class instances will compare via reference equality, like any other object. Equality support for the underlying native object will be provided as opt-in via the `WindowsRuntimeMarshal` type.
 
 ## Adoption plan


src/WinRT.Runtime2/InteropServices/WindowsRuntimeComWrappers.cs
```diff
@@ -349,7 +349,15 @@ public object GetOrCreateObjectForComInstanceUnsafe(
...
-        IInspectableVftbl.GetRuntimeClassNameUnsafe(interfacePointer, &className).Assert();
+        if (IInspectableVftbl.GetRuntimeClassNameUnsafe(interfacePointer, &className).Failed())
+        {
+            // If we couldn't get the runtime class name, we return it as an IInspectable.
+            WindowsRuntimeObjectReference objectReference = WindowsRuntimeComWrappersMarshal.CreateObjectReference(
+                externalComObject: interfacePointer,
+                iid: in WellKnownWindowsInterfaceIIDs.IID_IInspectable,
+                wrapperFlags: out wrapperFlags);
+            return new WindowsRuntimeInspectable(objectReference) { InspectableObjectReference = objectReference };
+        }
```
Sergio0694, line 354: Mmh wait, this doesn't look right to me. `GetRuntimeClassName` should never fail. What's the scenario?

manodasanW: MemoryBuffer.CreateReference returns E_NOTIMPL for GetRuntimeClassName.

manodasanW: The function returns an interface, so I guess they use some non projected type.  The scenario still works via IDIC.

Sergio0694: Doesn't the WinRT spec say you should just return the interface name then and never fail from there? 😭

src/WinRT.Runtime2/InteropServices/Bindables/BindableIEnumerableAdapter.cs
```diff
@@ -0,0 +1,52 @@
...
+            // implements 'IEnumerator<object>'. If it does, we can directly wrap that in the adapter that
+            // we'll marshal to native. Otherwise, we need an intermediate adapter to adapt the non-generic
+            // managed enumerator that we got. This also means we'll effectively need two nested adapters.
+            IEnumerator<object>? managedObjectEnumerator = managedEnumerator as IEnumerator<object> ?? new BindableIEnumeratorAdapter(managedEnumerator);
+
+            // This is the same adapter we use for 'IEnumerator<T>'. Code paths dealing with it should also
+            // be aware that they might potentially have another nested wrapper. We don't have special logic
+            // in 'cswinrtgen' to track a constructed 'IEnumerator<object>' when we see 'IEnumerable' being
+            // used, however that is not needed given that this interface will always be discovered anyway.
+            // With that, tirmming will then automatically keep code for it if this code path is reachable.
```
manodasanW: trimming

src/WinRT.Runtime2/InteropServices/WindowsRuntimeComWrappers.cs
manodasanW: Well the [doc](https://learn.microsoft.com/en-us/windows/win32/api/inspectable/nf-inspectable-iinspectable-getruntimeclassname) doesn't seem to say you need to.  And it seems there are OS APIs who are doing this today.

Sergio0694: Don't see anything in those docs saying it's optional to implement to be honest 😅

Sergio0694 (approved):

PR #2222, opened 2026-02-03
Make enum arrays blittable

manodasanW:
- Make enum arrays blittable given we only use int and uint as the underlying type and are using fixed expressions for pinning.
- Project the property set parameter in the ABI static methods class as ReadOnlySpan for arrays.

### DIFF
--- a/src/cswinrt/code_writers.h
+++ b/src/cswinrt/code_writers.h
@@ -76,7 +76,7 @@ namespace cswinrt
         return (settings.internal || settings.embedded) ? "internal" : "public";
     }
 
-    bool is_type_blittable(type_semantics const& semantics, bool for_array = false)
+    bool is_type_blittable(type_semantics const& semantics)
     {
         return call(semantics,
             [&](object_type)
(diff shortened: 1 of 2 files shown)


src/cswinrt/code_writers.h
```diff
@@ -6108,9 +6117,9 @@ global::System.Buffers.ArrayPool<%>.Shared.Return(__%_arrayFromPool);
             }
             else
             {
-                m.is_pinnable = is_type_blittable(semantics, true) && !m.is_out();
+                m.is_pinnable = is_type_blittable(semantics) && !m.is_out();
                 m.interop_dll_type = w.write_temp("%", bind<write_interop_dll_type_name>(semantics, typedef_name_type::ArrayMarshaller));
-                m.marshaler_type = is_type_blittable(semantics, true) ? "MarshalBlittable" : "MarshalNonBlittable";
+                m.marshaler_type = is_type_blittable(semantics) ? "MarshalBlittable" : "MarshalNonBlittable";
```
Sergio0694, line 6122: Does this code need updating or is it all leftover that we should eventually cleanup?

manodasanW: Leftover that will get cleaned up at the end.

PR #2230, opened 2026-02-07
Add attributes for authoring support

Sergio0694:
Title.

### DIFF
--- a/src/WinRT.Runtime2/InteropServices/Attributes/WindowsRuntimeComponentAssemblyAttribute.cs
+++ b/src/WinRT.Runtime2/InteropServices/Attributes/WindowsRuntimeComponentAssemblyAttribute.cs
@@ -20,12 +20,12 @@ namespace WindowsRuntime.InteropServices;
     DiagnosticId = WindowsRuntimeConstants.PrivateImplementationDetailObsoleteDiagnosticId,
     UrlFormat = WindowsRuntimeConstants.CsWinRTDiagnosticsUrlFormat)]
 [EditorBrowsable(EditorBrowsableState.Never)]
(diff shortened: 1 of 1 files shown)


src/WinRT.Runtime2/InteropServices/Attributes/WindowsRuntimeAuthoringAssemblyAttribute.cs
```diff
@@ -0,0 +1,31 @@
...
+/// <remarks>
+/// This attribute is emitted by the CsWinRT generator, and it is not meant to be used directly.
+/// </remarks>
+/// <seealso cref="System.Runtime.CompilerServices.ReferenceAssemblyAttribute"/>
+[AttributeUsage(AttributeTargets.Assembly, AllowMultiple = false, Inherited = false)]
+[Obsolete(WindowsRuntimeConstants.PrivateImplementationDetailObsoleteMessage,
+    DiagnosticId = WindowsRuntimeConstants.PrivateImplementationDetailObsoleteDiagnosticId,
+    UrlFormat = WindowsRuntimeConstants.CsWinRTDiagnosticsUrlFormat)]
+[EditorBrowsable(EditorBrowsableState.Never)]
+public sealed class WindowsRuntimeAuthoringAssemblyAttribute : Attribute
```
manodasanW: thoughts on WindowsRuntimeComponentAssemblyAttribute

Sergio0694: Don't feel strongly about it, yeah we can use that. To keep it consistent then is it ok to also use "Component" everywhere else too then? Like, `WinRT.Component.dll`, etc.

PR #2235, opened 2026-02-10
Pass WinRT.* generated .dll-s to 'cswinrtgen'

Sergio0694:
Title. This also allows hiding all implementation detail attributes from the reference assemblies.

### DIFF
--- a/src/WinRT.Interop.Generator/Helpers/SignatureGenerator.Projections.cs
+++ b/src/WinRT.Interop.Generator/Helpers/SignatureGenerator.Projections.cs
@@ -144,7 +144,7 @@ internal partial class SignatureGenerator
         bool useWindowsUIXamlProjections)
     {
         // Just like for generic instantiations, we need to resolve the IID for the type first
-        if (TryGetIIDFromWellKnownInterfaceIIDsOrAttribute(
+        if (!TryGetIIDFromWellKnownInterfaceIIDsOrAttribute(
             type: typeDefinition,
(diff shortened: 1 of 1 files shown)


src/cswinrt/code_writers.h
```diff
@@ -9147,19 +9157,11 @@ return new %(valueReference);
     {
         method_signature signature{ get_delegate_invoke(type) };
         w.write(R"(
-%%%%% delegate % %(%);
+%%%% delegate % %(%);
 )",
             bind<write_winrt_metadata_attribute>(type),
             bind<write_type_custom_attributes>(type, false),
             bind<write_comwrapper_marshaller_attribute>(type),
-            bind([&](writer& w)
```
manodasanW, line 9155: Do we want to emit the guid if it is not a reference projection as it seems this where you are looking right?

Sergio0694: Yeah, because we need it to be in the impl .dll as that's looked up by cswinrtinteropgen. We just don't need it anymore on delegates with this trick, so I'm hiding it there to match the API surface we had before as well 🙂

manodasanW: To clarify, I was saying, do we instead want to flip the check then.  Before the IID was just in reference projections, not impl.  Instead of removing, do we want to flip it to say put it in impl.

Sergio0694: Ooh, yes you're right, missed that one. Fixed it, thank you!

Issue #2171, opened 2026-01-03
Bug: Passing Microsoft.Graphics.Canvas.UI.Xaml.CanvasControl.Invalidate as the callback argument to a DispatcherQueue.TryEnqueue causes an InvalidCastException in the WinRT runtime


HeCoding180:
### Description

When trying to directly pass the `Invalidate` method of a `Microsoft.Graphics.Canvas.UI.Xaml.CanvasControl` instance as the callback argument of the `TryEnqueue` method of a `DispatcherQueue` instance an `InvalidCastException` is thrown.
When passing a lambda expression that just calls the Invalidate method, the exception is **not** thrown!

### Steps To Reproduce

The following should do the trick (I have not verified this outside of my repo):
```c#
DispatcherQueue.TryEnqueue(GraphCanvas.Invalidate);
```
where `GraphCanvas` is a `Microsoft.Graphics.Canvas.UI.Xaml.CanvasControl` instance.
The following code does not throw the exception:
```c#
DispatcherQueue.TryEnqueue(() => GraphCanvas.Invalidate());
```

### Expected Behavior

No `InvalidCastException` is thrown

### Version Info

.NET 9.0 Framework

### Additional Context

The following image shows one such exception with a stack trace:

<img width="1191" height="629" alt="Image" src="https://github.com/user-attachments/assets/2cc17a50-971f-49ea-ad79-733a19e91d2e" />

PR #2241, opened 2026-02-11
Delegate Test on Sealed Types

kythant:
To validate that this issue has been resolved
https://github.com/microsoft/CsWinRT/issues/2171


src/Tests/UnitTest/TestComponentCSharp_Tests.cs
```diff
@@ -93,6 +93,13 @@ public struct Estruct
             E value;
         }
 
+        [Fact]
+        public void TestDelegateCallBackOnSealedType()
+        {
+            // In CSWinRT 2.0, this scenario would throw an exception but this is fixed in 3.0
```
manodasanW, line 99: Consider a comment like passing a projected function as a delegate.

manodasanW (approved):

PR #2249, opened 2026-02-12
Add TypeMarshaling Functional Test to Pipeline with Trimming Fixes

kythant:
…ng scenarios with AOT

### DIFF
--- a/src/Tests/FunctionalTests/TypeMarshaling/Program.cs
+++ b/src/Tests/FunctionalTests/TypeMarshaling/Program.cs
@@ -3,6 +3,7 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Collections.Specialized;
 using System.ComponentModel;
+using System.Diagnostics.CodeAnalysis;
 using System.Numerics;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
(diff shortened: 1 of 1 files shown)


src/Tests/FunctionalTests/TypeMarshaling/Program.cs
```diff
@@ -77,7 +78,7 @@
     // --------------------
     new(typeof(DataErrorsChangedEventArgs), "Microsoft.UI.Xaml.Data.DataErrorsChangedEventArgs", "Metadata", 400),
     new(typeof(INotifyDataErrorInfo),       "Microsoft.UI.Xaml.Data.INotifyDataErrorInfo",       "Metadata", 401),
-    new(typeof(INotifyPropertyChanged),     "Microsoft.UI.Xaml.Data.INotifyPropertyChanged",     "Metadata", 402),
+    new(typeof(INotifyPropertyChanged),          "Microsoft.UI.Xaml.Data.INotifyPropertyChanged",     "Metadata", 402),
```
Sergio0694: Nit: fix extra spacing so it's aligned

src/Tests/FunctionalTests/TypeMarshaling/Program.cs
```diff
@@ -250,6 +251,8 @@
     new(typeof(NotifyCollectionChangedAction?), "Windows.Foundation.IReference`1<Microsoft.UI.Xaml.Interop.NotifyCollectionChangedAction>", "Metadata", 1001),
 ];
 
+UntrimmedTypes();
```
Sergio0694: I don't understand why these changes are needed. Can you just do this:

```csharp
// Add this helper somewhere
[MethodImpl(MethodImplOptions.NoInlining)]
public static void KeepType([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.All)] Type type)
{
}

// Do this before running the tests loop
KeepType(typeof(IServiceProvider))]
KeepType(typeof(INotifyCollectionChanged))]
```

So it's also easier to extend in the future if needed. And then you can run all tests always.

kythant: I tried something similar to what we did in the unit tests but it wasn't working so I was making it use the type for real. I can give this a try.

Sergio0694: That was different, that only allowed cswinrtgen to _see_ the type, but it wouldn't root it. This does 🙂

kythant: That worked! Wow that was so much simpler.

PR #2254, opened 2026-02-16
Packaging changes

manodasanW:
- Publish cswinrtgen, cswinrtimpl, cswinrtprojectiongen to AOT during the build and package them in the NuGet package
- Generalized process by adding `IsBuildTool` property which indicates this is a tool to publish to AOT during build
- Package the task also in the nuget package
- props / target changes to get it working with the .NET SDK and also as part of the cswinrt package itself
- Updated notice file
- Cleanup based on cswinrtgen targets is included by the cswinrt nuget package targets

### DIFF
--- a/src/Tests/FunctionalTests/Directory.Build.props
+++ b/src/Tests/FunctionalTests/Directory.Build.props
@@ -39,7 +39,7 @@
     <ControlFlowGuard>Guard</ControlFlowGuard>
   </PropertyGroup>
 
-  <PropertyGroup Condition="$(IsTargetFrameworkNet8OrGreater) and '$(Platform)' != 'x86' and '$(OutputType)' != 'Library'">
+  <PropertyGroup Condition="$(IsTargetFrameworkNet8OrGreater) and '$(Platform)' != 'x86'">
     <PublishAot>true</PublishAot>
   </PropertyGroup>
  
--- a/src/Directory.Build.targets
+++ b/src/Directory.Build.targets
@@ -19,7 +19,7 @@
     <None Include="$(GeneratedFilesRootDir)**/*.cs" />
     <Compile Condition="'$(GeneratedFilesDir)' != '$(GeneratedFilesRootDir)'" Include="$(GeneratedFilesDir)*.cs" />
   </ItemGroup>
-	
+    
   <ItemGroup>
     <FrameworkReference Remove="Microsoft.Windows.SDK.NET.Ref" />
     <FrameworkReference Remove="Microsoft.Windows.SDK.NET.Ref.Windows" />
@@ -77,5 +77,5 @@
       <Compile Include="$(GeneratedSkipLocalsInitFile)" />
     </ItemGroup>
   </Target>
-	
+    
 </Project>


nuget/NOTICE.txt
```diff
@@ -1,72 +1,152 @@
...
+NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+---------------------------------------------------------
+
+---------------------------------------------------------
+
+System.CommandLine 2.0.0-beta1.21308.1 - MIT
```
Sergio0694: Wait, why `System.CommandLine`? We're using `ConsoleAppFramework`.

nuget/NOTICE.txt
```diff
@@ -1,72 +1,152 @@
...
-THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, 
-EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF 
-MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND 
-NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE 
-LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION 
-OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION 
-WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
-
-
+NOTICES AND INFORMATION
```
Sergio0694, line 1: We should probably also include `CommunityToolkit.HighPerformance` (MIT)?

src/Tests/FunctionalTests/Directory.Build.props
```diff
@@ -39,7 +39,7 @@
     <ControlFlowGuard>Guard</ControlFlowGuard>
   </PropertyGroup>
 
-  <PropertyGroup Condition="$(IsTargetFrameworkNet8OrGreater) and '$(Platform)' != 'x86'">
+  <PropertyGroup Condition="$(IsTargetFrameworkNet8OrGreater) and '$(Platform)' != 'x86' and '$(OutputType)' != 'Library'">
```
Sergio0694: Isn't this imported _before_ the .csproj? If so how can the `OutputType` be set at this point?

nuget/NOTICE.txt
manodasanW: This is from before, probably needs to also be updated.

src/Tests/FunctionalTests/Directory.Build.props
manodasanW: Good point, looks like this didn't help avoid the additional interop compilation for the reason you mention.

PR #2255, opened 2026-02-16
Optimize free-threaded checks when metadata information is available

Sergio0694:
Title. This can skip up to 2 QIs for every single RCW being created.

### DIFF
--- a/src/WinRT.Runtime2/InteropServices/WindowsRuntimeComWrappersMarshal.cs
+++ b/src/WinRT.Runtime2/InteropServices/WindowsRuntimeComWrappersMarshal.cs
@@ -5,7 +5,6 @@ using System;
 using System.ComponentModel;
 using System.Diagnostics.CodeAnalysis;
 using System.Runtime.InteropServices;
-using System.Runtime.Versioning;
 using WindowsRuntime.InteropServices.Marshalling;
 
 #pragma warning disable CS1573
@@ -52,7 +51,6 @@ public static unsafe class WindowsRuntimeComWrappersMarshal
 
     /// <inheritdoc cref="CreateObjectReference(void*, in Guid, out CreatedWrapperFlags)"/>
     /// <param name="marshalingType">The <see cref="CreateObjectReferenceMarshalingType"/> value available in metadata for the type being marshalled.</param>
-    [SupportedOSPlatform("Windows10.0.10240.0")]
     public static WindowsRuntimeObjectReference CreateObjectReference(
         void* externalComObject,
         in Guid iid,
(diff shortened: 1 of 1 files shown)


src/cswinrt/helpers.h
```diff
@@ -1681,6 +1681,26 @@ namespace cswinrt
...
+        {
+            auto sig = marshaling_attr.Value();
+            auto const& fixed_args = sig.FixedArgs();
+            XLANG_ASSERT(fixed_args.size() == 1);
+
+            auto marshaling_type = std::get<int32_t>(std::get<ElemSig::EnumValue>(std::get<ElemSig>(fixed_args[0].value).value).value);
+
+            switch (marshaling_type)
+            {
+            case 2: return "CreateObjectReferenceMarshalingType.Agile"sv;
```
riverar, line 1696: What about types that lie to you? How do devs workaround that?
e.g., `Windows.Networking.PushNotifications.PushNotificationChannel`

Sergio0694: We're going to fix that in the SDK (cc. @jonwis). The point is that for such types, rs-windows is also already broken anyway. The correct thing is to just find the few types that are incorrectly annotated in the SDK (it's just that one as far as we know I think) and just update the annotation.

riverar: Is there a switch to bypass this optimization in cases where folks are working with agile liars and there's no fix available? (Or maybe the workaround is simply that folks manually write those bindings until a fix is available?)

Sergio0694: There isn't currently but I could add a feature switch for it I suppose. @manodasanW thoughts?

jonwis: You probably should add a way to suppress the feature entirely, or on a per-type basis. There are likely custom factories that can't easily be updated in a rational way :\ ... ideally this list comes from an MSDN page or in a kit artifact so we don't have the same patterns as the "tweaks to the headers" that this repo has.

Sergio0694: > "There are likely custom factories that can't easily be updated in a rational way"

Not following, even if that's the case and we can't change them, why can't we fix the annotation? Like, this is not about making factories agile, it's just about making sure that their `[MarshalingBehavior]` annotation is correct.

Sergio0694: Resolving this, added an opt-out switch for this optimization 👍

src/WinRT.Runtime2/InteropServices/WindowsRuntimeComWrappersMarshal.cs
```diff
@@ -37,7 +40,29 @@ public static unsafe class WindowsRuntimeComWrappersMarshal
...
+            marshalingType: CreateObjectReferenceMarshalingType.Unknown);
+
+        wrapperFlags = objectReference.GetReferenceTrackerPtrUnsafe() is null ? CreatedWrapperFlags.None : CreatedWrapperFlags.TrackerObject;
+
+        return objectReference;
+    }
+
+    /// <inheritdoc cref="CreateObjectReference(void*, in Guid, out CreatedWrapperFlags)"/>
+    /// <param name="marshalingType">The <see cref="CreateObjectReferenceMarshalingType"/> value available in metadata for the type being marshalled.</param>
+    [SupportedOSPlatform("Windows10.0.10240.0")]
```
riverar: Nit/Optional: Can delete this. The `SupportedOSPlatform` upstream metadata is all wrong.
https://github.com/microsoft/win32metadata/issues/738

If this didn't come from upstream then ignore!

Sergio0694: Oh good catch, these were actually leftovers and not needed anymore. Updated, thank you! 🙂

PR #2256, opened 2026-02-17
Project renames

manodasanW:
- Removed old WinRT.Runtime and renamed WinRT.Runtime2 to it (keeping folder name as WinRT.Runtime2)
- Renamed cswinrtgen to cswinrtinteropgen
- Renamed RunCsWinRTGeneratorTask.dll to WinRT.Generator.Tasks.dll
- Renamed RunCsWinRTGenerator to RunCsWinRTInteropGenerator
- Remove IIDOptimizer
- Simplify _ResolveCsWinRToolsDirectory target

### DIFF
--- a/build/AzurePipelineTemplates/CsWinRT-Build-Steps.yml
+++ b/build/AzurePipelineTemplates/CsWinRT-Build-Steps.yml
@@ -252,6 +252,7 @@ steps:
     - task: CopyFiles@2 
       displayName: Stage WinRT.Host.Shim
       condition: and(succeeded(), eq(variables['BuildPlatform'], 'x86'), eq(variables['BuildConfiguration'], 'release'))
+      enabled: false
       continueOnError: True
       inputs:
         SourceFolder: $(Build.SourcesDirectory)\src\Authoring\WinRT.Host.Shim\bin\$(BuildConfiguration)\net10.0


nuget/Microsoft.Windows.CsWinRT.CsWinRTGen.targets
```diff
@@ -30,7 +30,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <CsWinRTGenerateInteropAssembly>false</CsWinRTGenerateInteropAssembly>
 
     <!--
-      Default property values for 'cswinrtgen'. Not setting 'CsWinRTToolsDirectory' and
+      Default property values for cswinrt generators. Not setting 'CsWinRTToolsDirectory' and
```
Sergio0694, line 33:
```suggestion
      Default property values for 'cswinrtinteropgen'. Not setting 'CsWinRTToolsDirectory' and
```

manodasanW: That was intentional, some of the properties below, especially the last couple, were common enough to be used by our other generators and are used in some cases rather than duplicating the properties.

Sergio0694: Ah, I see. Then can you put 'cswinrt' in single quotes? Would make it clearer it specifically refers to those.

manodasanW: Sure, probably will address this as part of my source generator fix PR.

PR #2257, opened 2026-02-17
Enable AOT running of UnitTests; Migrate UnitTests from XUnit to MSTest Framework 

kythant:


src/Tests/AuthoringWuxTest/AuthoringWuxTest.csproj
```diff
@@ -16,17 +16,14 @@
   <!-- Temporary workaround, remove when https://github.com/dotnet/sdk/pull/41936 is available in the .NET SDK -->
   <ItemGroup>
     <CompilerVisibleProperty Include="CsWinRTUseWindowsUIXamlProjections" />
-    <RuntimeHostConfigurationOption Include="CSWINRT_USE_WINDOWS_UI_XAML_PROJECTIONS"
-                                    Value="$(CsWinRTUseWindowsUIXamlProjections)"
-                                    Trim="true" />
+    <RuntimeHostConfigurationOption Include="CSWINRT_USE_WINDOWS_UI_XAML_PROJECTIONS" Value="$(CsWinRTUseWindowsUIXamlProjections)" Trim="true" />
   </ItemGroup>
```
Sergio0694, line 20: That should be available now, can we remove this workaround?

src/Tests/UnitTest/ExceptionTests.cs
```diff
@@ -1,23 +1,27 @@
...
-            Assert.False(string.IsNullOrWhiteSpace(exception.Message));
+            Assert.IsNotNull(exception);
+            Assert.IsFalse(string.IsNullOrWhiteSpace(exception.Message));
+
             if (CultureInfo.CurrentUICulture.Name == "en-US")
             {
-                Assert.Equal("The application called an interface that was marshalled for a different thread. (0x8001010E)", exception.Message);
+                Assert.AreEqual(
+                    "The application called an interface that was marshalled for a different thread. (0x8001010E)",
+                    exception.Message);
```
Sergio0694: Super nit: keep it in one line like it was?

src/Tests/UnitTest/TestComponentCSharp_Tests.cs
```diff
@@ -40,51 +42,32 @@
...
-using Xunit.Sdk;
-using static System.Diagnostics.Activity;
+using static UnitTest.UnitTestHelper;
 
 // Test SupportedOSPlatform warnings for APIs targeting 10.0.19041.0:
 [assembly: global::System.Runtime.Versioning.SupportedOSPlatform("Windows10.0.18362.0")]
 
 namespace UnitTest
 {
-    [CLSCompliant(false)]
```
Sergio0694, line 53: Is this no longer needed because the AOT generator from MSTest is causing those types to be visible in the actual code and so cswinrtinteropgen can see them without anything else needed? Or is it something else?

src/Tests/UnitTest/TestComponentCSharp_Tests.cs
```diff
@@ -1081,107 +1079,108 @@ public void TestCustomProjections()
...
-
-            var uriArray = new Uri[] { new Uri("http://aka.ms/cswinrt"), new Uri("http://github.com") };
-            TestObject.ObjectIterableProperty = uriArray;
-            Assert.True(TestObject.ObjectIterableProperty.SequenceEqual(uriArray));
-
-            var objectUriArray = new object[] { new Uri("http://github.com") };
-            TestObject.ObjectIterableProperty = objectUriArray;
-            Assert.True(TestObject.ObjectIterableProperty.SequenceEqual(objectUriArray));
-        }
+        // TODO: Enable once fixed after preview1
```
Sergio0694, line 1113: Just curious, what's missing that causes this test to not work anymore for now?

Sergio0694 (approved):

src/cswinrt.slnx
```diff
@@ -228,7 +200,6 @@
     </Project>
     <Project Path="Tests/UnitTest/UnitTest.csproj">
       <BuildDependency Project="Perf/IIDOptimizer/IIDOptimizer.csproj" />
```
manodasanW: I assume you haven't merged in changes from staging/3.0 yet.

src/Tests/UnitTest/TestComponent_Tests.cs
```diff
@@ -2,393 +2,397 @@
...
     public class TestWinRT
     {
-        public ITests Tests { get; private set; }
+        public ITests Tests { get; private set; } = TestRunner.MakeTests();
 
-        public TestWinRT()
-        {
-            Tests = TestRunner.MakeTests();
-        }
+        // TestInitialize does not work with AOT
```
manodasanW: We can just remove this

src/Tests/UnitTest/TestComponentCSharp_Tests.cs
```diff
@@ -40,51 +42,32 @@
...
     {
-        public Class TestObject { get; private set; }
+        public Class TestObject = new();
 
-        public TestCSharp()
-        {
-            TestObject = new Class();
-        }
+        // TestInitialize does not work with AOT
+        // See https://devblogs.microsoft.com/dotnet/testing-your-native-aot-dotnet-apps/
```
manodasanW: Same here

src/Tests/UnitTest/TestComponentCSharp_Tests.cs
```diff
@@ -112,118 +95,118 @@ public void TestLongClassNameEventSource()
...
             TestObject.EnumStructsProperty = expectedEnumStructs;
-            Assert.Equal(expectedEnumStructs, TestObject.EnumStructsProperty);
+            Assert.IsTrue(expectedEnumStructs.SequenceEqual(TestObject.EnumStructsProperty));
             TestObject.CallForEnumStructs(() => expectedEnumStructs);
-            Assert.Equal(expectedEnumStructs, TestObject.EnumStructsProperty);
+            Assert.IsTrue(expectedEnumStructs.SequenceEqual(TestObject.EnumStructsProperty));
 
             TestObject.EnumStructsProperty = null;
-            Assert.Empty(TestObject.EnumStructsProperty);
+            Assert.IsTrue(TestObject.EnumStructsProperty.SequenceEqual(null));
```
manodasanW: To confirm, does this work as expected?

src/Tests/UnitTest/TestComponentCSharp_Tests.cs
```diff
@@ -535,213 +518,229 @@ public void TestTryGetDataUnsafe()
...
+        [TestMethod]
         public void TestVectorCastConversion()
         {
+
             var vector = TestObject.GetUriVectorAsIInspectableVector();
             var uriVector = vector.Cast<Uri>();
             var first = uriVector.First();
-            Assert.Equal(vector, uriVector);
+            // Assert the sequences are equivalent
+            Assert.IsTrue(vector.Cast<Uri>().SequenceEqual(uriVector)); // [2]
```
manodasanW: Is the comment indicating the size?

src/Tests/UnitTest/TestComponentCSharp_Tests.cs
manodasanW: I assume this is the linq type one for enumerable.

src/Tests/UnitTest/TestComponentCSharp_Tests.cs
```diff
@@ -1811,70 +1810,70 @@ public void RemoveAt(int index)
...
         {
             var expected = new int[] { 0, 1, 2 };
 
             TestObject.BindableIterableProperty = expected;
-            Assert.Equal(expected, TestObject.BindableIterableProperty);
+            Assert.AreEqual(expected, TestObject.BindableIterableProperty);
             TestObject.CallForBindableIterable(() => expected);
             TestObject.BindableIterablePropertyChanged +=
-                (object sender, IEnumerable value) => Assert.Equal(expected, value);
+                (object sender, IEnumerable value) => Assert.IsTrue(expected.SequenceEqual(value.Cast<int>()));
```
manodasanW: For all the Assert.Equal that we converted to Assert.IsTrust(SequenceEqual), have you looked at using [CollectionAssert.AreEqual](https://learn.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.testtools.unittesting.collectionassert.areequal?view=mstest-net-4.0) instead?  Typically the useful thing of Equal functions is it actually indicates what array element is not equal which we miss out with Assert.IsTrue

src/cswinrt.slnx
kythant: Nope

src/Tests/UnitTest/TestComponentCSharp_Tests.cs
kythant: I believe so. TestObject.EnumStructsProperty showed up as {0} so it wasn't completely a nullptr. Is this a mistake on the test code end?

manodasanW: Showing up as {} / empty is expected.

src/Tests/UnitTest/TestComponent_Tests.cs
kythant: Just the comment or the whole commented out method?

src/Tests/UnitTest/TestComponentCSharp_Tests.cs
kythant: Just the comment or the whole commented out method?

manodasanW: Remove it all given we replaced it with an equivalent.

src/Tests/UnitTest/TestComponentCSharp_Tests.cs
kythant: The latter. The DataRow attribute just makes it work. Nothing about AOT

src/Tests/UnitTest/TestComponentCSharp_Tests.cs
kythant: Right but the pointer itself isn't null right? It is pointing to a structure that is empty

src/Tests/UnitTest/TestComponentCSharp_Tests.cs
kythant: Updated to use CollectionAssert.AreEqual

src/Tests/AuthoringWuxTest/AuthoringWuxTest.csproj
kythant: Is this relevant to the PR or should be done in another PR?

src/Tests/UnitTest/TestComponentCSharp_Tests.cs
kythant: Not sure what you mean by this. Do you mean the memory size or the actual value?

manodasanW: the line you added on the right // [2]

kythant: Oh I am not sure why that got added. But I've removed the comments. 
Does the Assert itself look right to you?

manodasanW: was this one of the ones that is going to be converted to CollectionAssert.AreEqual

kythant: Wait the code is outdated. 

It's been updated to 

            CollectionAssert.AreEqual(vector.Cast<Uri>().ToArray(), uriVector.ToArray());

src/Directory.Packages.props
```diff
@@ -52,5 +50,12 @@
     <PackageVersion Include="xunit.runner.console" Version="2.9.3" />
     <PackageVersion Include="xunit.runner.visualstudio" Version="3.1.3" />
     <PackageVersion Include="XunitXml.TestLogger" Version="6.1.0" />
```
Sergio0694, line 52: Are these still used in some test projects we haven't migrated yet, or can we remove them?

kythant: src\Samples\TestEmbedded\UnitTestEmbedded\UnitTestEmbedded.csproj is still using them.

Sergio0694: Probably should be able to delete that whole thing, since we don't support embedded mode anymore. But we can leave that to a separate follow up PR to clean things up then, not here 🙂

kythant: Oh good! I can do that in a separate PR

PR #2262, opened 2026-02-18
Merge master to staging/3.0

manodasanW:


src/Tests/FunctionalTests/DerivedClassAsBaseClass/Program.cs
```diff
@@ -15,4 +17,90 @@
     return 101;
 }
 
+//TODO: weakref rehydration as base class
```
Sergio0694, line 20: How come this is all commented out? Shouldn't this work, just with the resulting RCW after rehydration possibly being less derived than the original one? Perhaps we could keep this test and just update it to reflect that, so it can still be a regression test, and we just leave a note saying in the future once we somehow support this better, the type after rehydration might be more derived?

PR #2275, opened 2026-02-20
Add support for custom overload names to CsWinMD

talagrand:
This brings in support for OverloadAttribute to CsWinMD, so that the generated WinMD file carries that metadata.
Rust does not have support for overloaded methods, so windows-rs based consumers require this attribute to be set for API ergonomics.


Sergio0694:
Ported to 3.0 in #2454. @manodasanW should we just abandon this one or do we still want it?

Sergio0694:
Talked offline, we're not going to take this change for 2.x, and it's fixed in 3.0, so closing this one. Thank you! 🙂

PR #2276, opened 2026-02-20
Add support for Linux cross-builds

talagrand:
This mirrors cppwinrt's cross-builds, and tries to keep cswinrt as similar as possible to cppwinrt in addressing this support (prebuild exe, conditional compilation).


Sergio0694:
As we talked about offline, closing this, since all these C++ tools have been removed in 3.0 anyway 🙂

PR #2281, opened 2026-02-21
Fix various issues from testing with Windows SDK projection build

manodasanW:
- Added logic to find Winmd folder in the Windows SDK targeting pack and use that rather than WinMDs from elsewhere
- Added support for a projection using delay sign.
- Given the projections now can't have additive classes other than what CsWinRT generates, making the DispatcherQueueSynchronizationContext for both WUX and MUX an addition that CsWinRT tool will add
- Fixed issue where NativeObjectReference is null in composable QI scenarios during initialization

### DIFF
--- a/nuget/Microsoft.Windows.CsWinRT.targets
+++ b/nuget/Microsoft.Windows.CsWinRT.targets
@@ -38,7 +38,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       These properties are defined in this property group and not in the 'CsWinRTGenerateProjection'
       target, because they are also needed in authoring scenarios (which will not run that target).
     -->
-    <CsWinRTExeTFM Condition="$([MSBuild]::GetTargetFrameworkVersion('$(TargetFramework)')) >= 8">net8.0</CsWinRTExeTFM>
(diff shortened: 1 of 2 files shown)


nuget/Microsoft.Windows.CsWinRT.targets
```diff
@@ -183,6 +181,18 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     </ItemGroup>
   </Target>
 
+  <Target Name="CsWinRTResolveWindowsMetadata" BeforeTargets="BeforeCompile" AfterTargets="ResolveReferences" Condition="'$(CsWinRTWindowsMetadata)' == ''">
+      <!-- From the Windows SDK projection binary referenced, determine the folder with the WinMDs in the targeting pack. -->
```
Sergio0694, line 185: Super nit:
```suggestion
  <!-- From the Windows SDK projection binary referenced, determine the folder with the WinMDs in the targeting pack. -->
  <Target Name="CsWinRTResolveWindowsMetadata" BeforeTargets="BeforeCompile" AfterTargets="ResolveReferences" Condition="'$(CsWinRTWindowsMetadata)' == ''">
```

PR #2282, opened 2026-02-21
Centralize 'DispatcherQueueSynchronizationContext' logic in 'WinRT.Runtime.dll'

Sergio0694:
Avoids code duplication, makes it easier to review, improve, and maintain.

### DIFF
--- a/src/WinRT.Runtime2/InteropServices/Dispatching/DispatcherQueueProxyHandler.cs
+++ b/src/WinRT.Runtime2/InteropServices/Dispatching/DispatcherQueueProxyHandler.cs
@@ -23,12 +23,12 @@ internal unsafe struct DispatcherQueueProxyHandler
     /// <summary>
     /// The <see cref="GCHandle"/> to the captured <see cref="SendOrPostCallback"/>.
     /// </summary>
-    private GCHandle _callbackHandle;
+    private GCHandle<SendOrPostCallback> _callbackHandle;
 
     /// <summary>
     /// The <see cref="GCHandle"/> to the captured state (if present, or a <see langword="null"/> handle otherwise).
     /// </summary>
-    private GCHandle _stateHandle;
+    private GCHandle<object> _stateHandle;
 
     /// <summary>
     /// The current reference count for the object (from <c>IUnknown</c>).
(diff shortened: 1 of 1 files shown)


src/WinRT.Runtime2/WindowsRuntimeObject.cs
```diff
@@ -590,9 +590,12 @@ static void ThrowArgumentException(Type type)
...
     {
-        // We explicitly don't handle overridable interfaces and 'IInspectable'.
-        // And when 'NativeObjectReference' is null which is during initialization,
-        // those calls are for outer from inner which we don't need to handle either.
+        // We explicitly don't handle overridable interfaces, as well as 'IInspectable' and 'IWeakReference'.
+        // This last one in particular must be ignored to avoid issues when an RCW object is used as a target
+        // for a 'WeakReference' object. Lastly, we also need to not handle this 'QueryInterface' request when
+        // 'NativeObjectReference' is 'null', which will be the case during initialization (because objects are
+        // constructed entirely from this base 'WindowsRuntimeObject' type). In that case, 'QueryInterface' calls
+        // will be coming from the outer instance to the inner one, and they wouldn't need to be handled anyway.
```
manodasanW:
```suggestion
        // will be coming for the outer instance from the inner one where the outer calls `GetInterface` first, and they wouldn't need to be handled anyway.
```

src/WinRT.Runtime2/InteropServices/Dispatching/DispatcherQueueSynchronizationContext.cs
```diff
@@ -0,0 +1,103 @@
...
+
+            RestrictedErrorInfo.ThrowExceptionForHR(hresult);
+
+            // This method will never return 'false', unless 'DispatcherQueueController.ShutdownQueueAsync()'
+            // has been called on the controller for the dispatcher queue in use. In that case, trying to
+            // enqueue a new callback on this dispatcher queue instance is not actually valid. For more info,
+            // see https://docs.microsoft.com/en-us/uwp/api/windows.system.dispatcherqueue.tryenqueue.
+            if (!success)
+            {
+                RestrictedErrorInfo.ThrowExceptionForHR(WellKnownErrorCodes.E_NOT_VALID_STATE);
```
manodasanW, line 81: Any concerns this is going to introduce crashes that aren't happening today such as where another thread is still doing something while the app is shutting down.

Sergio0694: I feel like silently ignoring that failure is also wrong though. For what it's worth, we have this additional check in the Microsoft Store and we never saw crashes because of it. Perhaps it should have a feature switch or something?

manodasanW: Lets keep it for now and monitor for if we see unexpected crashes and then consider if we should remove it.

manodasanW (approved):

PR #2296, opened 2026-02-25
Build Determinism Test for WinRT.Interop.dll

kythant:
Test to see if WinRT.Interop.dll is built deterministically. 

Test builds TestComponent twice with cleaning in between and checks the SHA256 hash of each build.


src/Tests/BuildDeterminismTest/Program.cs
```diff
@@ -0,0 +1,193 @@
...
+
+    /// <summary>
+    /// Deletes bin output, restores packages, builds, and returns the SHA256 hash
+    /// of the output DLL for the given pass (first/second).
+    /// </summary>
+    private string CleanBuildAndHash(string passLabel)
+    {
+        // MSBuild Clean doesn't remove the bin folders, so delete them explicitly.
+        string projectFolder = Path.GetDirectoryName(_projectPath)!;
+        foreach (string dir in new[] { "bin" })
```
Sergio0694: Might as well delete `obj` too to triple check the build is actually clean?

build/AzurePipelineTemplates/CsWinRT-Test-Steps.yml
```diff
@@ -57,6 +57,17 @@ steps:
         dir _build\$(BuildPlatform)\$(BuildConfiguration)\AuthoringWuxConsumptionTest\bin
         _build\$(BuildPlatform)\$(BuildConfiguration)\AuthoringWuxConsumptionTest\bin\AuthoringWuxConsumptionTest.exe --gtest_output=xml:AUTHORINGWUXTEST-$(Build.BuildNumber).xml 
 
+# Run Build Determinism Test
+  - task: CmdLine@2
+    displayName: Run Build Determinism Test
+    condition: and(succeeded(), or(eq(variables['BuildPlatform'], 'x86'), eq(variables['BuildPlatform'], 'x64')))
+    continueOnError: True
```
manodasanW, line 78: lets remove continueOnError, we probably will start removing this for others too.

PR #2307, opened 2026-03-01
Inline event source fields for 'IObservableVector<T>' and 'IObservableMap<K, V>' specialized objects

Sergio0694:
This PR updates the specialized RCWs for those two interfaces to avoid using a `CWT<,>` for the event source objects.

This is done by emitting a new factory type like this:

image

And passing this as type argument to the base specialized RCW types.

### DIFF
--- a/src/cswinrt/code_writers.h
+++ b/src/cswinrt/code_writers.h
@@ -6816,7 +6816,7 @@ remove
                 is_parameter_variable ? "" : ";");
     }
 
-    void write_static_abi_class_members(writer& w, TypeDef const& iface, uint32_t const& abi_methods_start_index = 6)
+    void write_static_abi_class_members(writer& w, TypeDef const& iface, uint32_t const& abi_methods_start_index = 6, bool skip_events = false)
     {
         auto init_call_variables = [&](writer& w)
         {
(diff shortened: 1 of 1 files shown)


src/cswinrt/code_writers.h
```diff
@@ -8565,28 +8716,61 @@ private IObjectReference % => __% ?? Make__%();
...
+        // inlined in the RCW class. This is safe because the interface and its Methods
+        // type are internal, so no other type can reference these event methods.
+        // Only do this for instance interfaces (listed in InterfaceImpl on the class),
+        // not for statics interfaces (referenced via StaticAttribute), whose events
+        // are never inlined and still go through the Methods type.
+        bool skip_exclusive_events = false;
+        if (is_exclusive_to(iface) && !settings.public_exclusiveto)
+        {
+            auto class_type = get_exclusive_to_type(iface);
+            for (auto&& ii : class_type.InterfaceImpl())
```
manodasanW, line 8729: What does this loop achieve?

Sergio0694: It's checking whether the interface is implemented as an instance interface on the runtime class, so it basically filters out static interfaces. The point is we can skip generating events for exclusive instance interfaces, because with this PR they're inlined, so we don't need the CWTs anymore, but not static events. Those still need the CWT and the related code generated in the 'Methods' types. So this loop is making sure we don't just always skip all exclusive interfaces, otherwise the latter case wouldn't compile anymore.

We might be able to optimize this further if we get that support for annotating factories as agile, so we could inline event sources for static events from agile factories. We'll probably need to still keep the CWTs for non-agile factories. But at least that should be very very few at that point.

manodasanW (approved):

PR #2309, opened 2026-03-02
Add TestFailingCompletionHandlerCrashesProcess Test

kythant:


src/Tests/UnitTest/TestComponentCSharp_Tests.cs
```diff
@@ -4092,5 +4092,24 @@ public void TestNontProjectedClassAsBaseClass()
...
+            var instance = new Class();
+            instance.SetFailingCompletedHandler(asyncAction);
+
+            // Wait long enough for the async action to complete and the native
+            // completion handler to fire. If the throwing handler crashes the
+            // process, we will never reach the assertion below.
+            await Task.Delay(TimeSpan.FromSeconds(5));
+
+            // If we get here, the process survived the throwing completion handler.
+            Assert.IsTrue(true, "Process survived the throwing native completion handler.");
```
Sergio0694: So like, if this test passes it means that either:
- The test is somehow wrong
- Something else is wrong 😆

@manodasanW this code should (intentionally) crash the process, right?

manodasanW (approved):

PR #2346, opened 2026-03-18
Use camelCase parameter names in projected struct constructors

Sergio0694:
This PR updates the codegen for projected struct constructors to use **camelCase** parameter names instead of underscore-prefixed field names, producing a cleaner public API surface.

## Motivation

The current codegen for projected structs generates constructor parameters by prepending an underscore to the field name (e.g. `_Stage`, `_BytesSent`). This is not idiomatic C# — the standard convention for parameter names is camelCase.

### `helpers.h`
Added a `to_camel_case()` utility function that:
- Lowercases the first character using invariant ASCII arithmetic (no locale dependency)
- Falls back to prepending an underscore if the first character is not an uppercase ASCII letter (e.g. a field named `_Foo` becomes `__Foo`), to avoid collisions with the original field name

### `code_writers.h`
Updated the struct constructor codegen to use `to_camel_case(field.name)` for parameter names and assignments.

## Example

```csharp
// Before
public HttpProgress(HttpProgressStage _Stage, ulong _BytesSent, ulong? _TotalBytesToSend, ...)
{
    Stage = _Stage;
    BytesSent = _BytesSent;
    ...
}

// After
public HttpProgress(HttpProgressStage stage, ulong bytesSent, ulong? totalBytesToSend, ...)
{
    Stage = stage;
    BytesSent = bytesSent;
    ...
}
```

### DIFF
--- a/src/cswinrt/code_writers.h
+++ b/src/cswinrt/code_writers.h
@@ -9869,7 +9869,7 @@ R"(
             type.TypeName(),
             type.TypeName());
 
-        // ctor (use camelCase parameter names rather than underscore-prefixed names for cleaner public API)
+        // ctor (use camelCase parameter names for cleaner public API)
         w.write("public %(%)\n{\n%\n}\n",
             type.TypeName(),
             bind_list([](writer& w, auto&& field)


src/cswinrt/helpers.h
```diff
@@ -13,6 +13,27 @@ namespace cswinrt
...
+        std::string result(name);
+        if (!result.empty())
+        {
+            if (result[0] >= 'A' && result[0] <= 'Z')
+            {
+                result[0] = result[0] - 'A' + 'a';
+            }
+            else
+            {
+                result.insert(result.begin(), '_');
```
manodasanW: is this needed, can't we just use this.?

Sergio0694: Good point, updated!

src/cswinrt/code_writers.h
```diff
@@ -9863,16 +9869,25 @@ R"(
             type.TypeName(),
             type.TypeName());
 
-        // ctor
+        // ctor (use camelCase parameter names rather than underscore-prefixed names for cleaner public API)
```
manodasanW:
```suggestion
        // ctor (use camelCase parameter names for cleaner public API)
```

PR #2350, opened 2026-03-19
Make WindowsRuntimeDefaultInterface attribute trimmable

Sergio0694:
## Summary

Move the `[WindowsRuntimeDefaultInterface]` attribute from individual projected runtime class types to a centralized `WindowsRuntimeDefaultInterfaces` static class in the `ABI` namespace. This decouples the default interface `Type` references from the class types themselves, allowing the trimmer to remove unused interface types at publish time.

Previously, each projected runtime class carried a `[WindowsRuntimeDefaultInterface(typeof(IDefaultInterface))]` attribute directly on the class. Because attributes are not trimmable, this forced the linker to preserve the interface type even when it was never used at runtime. The attribute is only consumed by the interop generator (`WinRT.Interop.Generator`) to compute WinRT type signatures.

### Attribute definition (`WindowsRuntimeDefaultInterfaceAttribute`)
- Changed from single `Type` parameter (`interfaceType`) to two `Type` parameters (`runtimeClassType`, `interfaceType`)
- Set `AllowMultiple = true` since the centralized type needs multiple attribute instances

### Code generator (`cswinrt.exe`)
- Removed per-class `[WindowsRuntimeDefaultInterface]` emission from `write_class`
- Added `add_default_interface_entry` to accumulate (class, interface) pairs during parallel namespace processing
- Added `write_default_interfaces_class` to emit a single `WindowsRuntimeDefaultInterfaces.cs` file after all namespaces are processed, with entries sorted for deterministic output
- Guarded with `!settings.reference_projection` since the file is only needed for implementation assemblies
- Used `write_begin()` for the file header, consistent with all other generated files

### Interop generator (`WinRT.Interop.Generator`)
- Updated `TryGetDefaultInterfaceFromAttribute` to look up the centralized `ABI.WindowsRuntimeDefaultInterfaces` type instead of searching for the attribute on each individual class
- Added `GetDefaultInterfacesLookup` extension method on `ModuleDefinition` following the same caching pattern as `GetTopLevelTypesLookup`: a `ConditionalWeakTable` holding a `FrozenDictionary<(Namespace, Name), TypeSignature>` for O(1) lookups
- Removed the `InteropReferences` parameter from `TryGetDefaultInterfaceFromAttribute` since attribute type validation is no longer needed (the centralized type only has these attributes)

### Bug fix
- Fixed `write_begin_interface_iids` in `type_writers.h` which was hardcoding version `0.0.0-private.0` instead of using `VERSION_STRING`

## Generated output (before → after)

**Before** (on each class):
```csharp
[WindowsRuntimeDefaultInterface(typeof(IHttpMethod))]
public sealed class HttpMethod : WindowsRuntimeObject, ...
```

**After** (centralized file):
```csharp
// WindowsRuntimeDefaultInterfaces.cs
namespace ABI
{
[WindowsRuntimeDefaultInterface(typeof(global::Windows.Web.Http.HttpMethod), typeof(global::ABI.Windows.Web.Http.IHttpMethod))]
// ... one attribute per runtime class ...
internal static class WindowsRuntimeDefaultInterfaces;
}
```

### DIFF
--- a/src/cswinrt/code_writers.h
+++ b/src/cswinrt/code_writers.h
@@ -401,11 +401,18 @@ namespace cswinrt
 
     void write_type_name(writer& w, type_semantics const& semantics, typedef_name_type const& nameType = typedef_name_type::Projected, bool forceWriteNamespace = false)
     {
+        // When forcing namespace qualification, temporarily clear the current namespace
+        // so that all type references (including generic type arguments) are fully qualified.
+        auto saved_namespace = forceWriteNamespace ? w._current_namespace : std::string_view{};
+        if (forceWriteNamespace) w._current_namespace = {};
+
         for_typedef(w, semantics, [&](auto type)
         {
             write_typedef_name(w, type, nameType, forceWriteNamespace);
             write_type_params(w, type);
         });
+
+        if (forceWriteNamespace) w._current_namespace = saved_namespace;
     }
 
     auto write_type_name_temp(writer& w, type_semantics const& type, char const* format = "%", typedef_name_type const& nameType = typedef_name_type::Projected)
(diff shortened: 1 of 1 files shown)


src/cswinrt/code_writers.h
```diff
@@ -5070,11 +5071,72 @@ R"(
...
+        {
+            return;
+        }
+
+        auto class_name = w.write_temp("global::%.@", type.TypeNamespace(), type.TypeName());
+
+        // Clear the current namespace so that all type references (including generic
+        // type arguments) are fully qualified with 'global::'. This is needed because
+        // the names will be emitted in the 'ABI' namespace, not the original one.
+        auto saved_namespace = w._current_namespace;
```
manodasanW: Doesn't passing true as the last parameter of write_type_name already do this?

PR #2351, opened 2026-03-19
Skip emitting [ContractVersion] and [GCPressure] attributes for implementation assemblies

Sergio0694:
## Summary

When generating an implementation assembly (i.e. not a reference one), the projection generator now skips emitting `[ContractVersion]` and `[GCPressure]` attributes on types. These metadata-only attributes are only relevant for reference assemblies and are unnecessary in the implementation output.

## Changes

In `write_custom_attributes` (`src/cswinrt/code_writers.h`):

- **Removed `GCPressure` from the unconditional early-skip list** (which also covers `Guid`, `Flags`, `ProjectionInternal`), so it now flows into the metadata attribute filter instead of being silently dropped for all builds.
- **Added a dedicated branch for `ContractVersion` and `GCPressure`** in the metadata attribute filter: both are now only emitted when `settings.reference_projection` is true. For implementation assemblies, they are skipped.

The logic is split into two clear, readable branches:
1. If the attribute is `ContractVersion` or `GCPressure` → skip for implementation assemblies, emit for reference assemblies.
2. For all other `Windows.Foundation.Metadata` attributes → only emit if they are one of the always-projected attributes (`DefaultOverload`, `Overload`, `AttributeUsage`, `Experimental`).

### DIFF
--- a/src/cswinrt/code_writers.h
+++ b/src/cswinrt/code_writers.h
@@ -2552,8 +2552,8 @@ remove => %;
         {
             auto [attribute_namespace, attribute_name] = attribute.TypeNamespaceAndName();
             attribute_name = attribute_name.substr(0, attribute_name.length() - "Attribute"sv.length());
-            // Guid, Flags, ProjectionInternal are handled separately
-            if (attribute_name == "Guid" || 
+            // GCPressure, Guid, Flags, ProjectionInternal are handled separately
+            if (attribute_name == "GCPressure" || attribute_name == "Guid" || 
                 attribute_name == "Flags" || attribute_name == "ProjectionInternal") continue;
             auto attribute_full = (attribute_name == "AttributeUsage") ? "System.AttributeUsage" :
                 w.write_temp("%.%", attribute_namespace, attribute_name);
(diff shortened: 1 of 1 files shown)


src/cswinrt/code_writers.h
```diff
@@ -2575,9 +2575,16 @@ remove => %;
                 {
                     allow_multiple = true;
                 }
-                if (attribute_name != "DefaultOverload" && attribute_name != "Overload" && 
-                    attribute_name != "AttributeUsage" && attribute_name != "ContractVersion" &&
-                    attribute_name != "Experimental")
+                // ContractVersion and GCPressure are only emitted for reference assemblies
+                if (attribute_name == "ContractVersion" || attribute_name == "GCPressure")
```
manodasanW: GCPressure I think was never emitted before, but I guess with this change, we are adding it to the reference projection which is fine.

Sergio0694: Oops actually didn't mean to do that, only wanted to skip the other one. Fixed 😄

PR #2362, opened 2026-03-21
Update docs

manodasanW:


docs/interop.md
```diff
@@ -1,223 +1,187 @@
...
+### COM object detection
+
+To check whether an object is a Windows Runtime object:
+
 ```csharp
-var objRef = ObjectReference<T>.Attach(ptr);     // transfers ownership (no AddRef)
-var objRef = ObjectReference<T>.FromAbi(ptr);    // creates a new reference (calls AddRef)
+if (obj is WindowsRuntimeObject)
+{
+    // obj is a Windows Runtime object (either native-backed or managed-authored)
```
Sergio0694: This seems incorrect. Even managed objects (not authored) that simply derive from some composable projected type will pass this check, even though they'd be "normal" C# objects. It's a bit confusing maybe?

docs/structure.md
```diff
@@ -13,15 +13,15 @@ Contains files that assist with publishing to Maestro.
 
 ## [`nuget`](../nuget)
```
Sergio0694, line 14: I feel like this whole document is unnecessary and should just be removed. The copilot instruction docs have the same info and much more in detail, and we also have a skill to auto-update them whenever needed.

docs/usage.md
```diff
@@ -2,56 +2,146 @@
 
 The [C#/WinRT NuGet package](https://www.nuget.org/packages/Microsoft.Windows.CsWinRT/) provides tooling for the following scenarios:
```
Sergio0694, line 3: A bunch of stuff in this doc is also duplicate with what's in the copilot instruction and in the interop generator skill. I think we should keep this strictly just for docs for people on how to actually use the package and generate a projection, and skip repeating how the various tools work. We could just add a link to those two copilot docs instead and say to go read those if more info on the technical details is needed.

docs/interop.md
manodasanW: This was equivalent to the [IWinRTObject check](https://github.com/microsoft/CsWinRT/blob/staging/3.0/docs/interop.md#com-data) that was there before.  But at the same time, given there isn't a scenario where it is used today by others, I can remove it to avoid the confusion from the different scenarios.

docs/structure.md
manodasanW: This doc doesn't need to be in detail but I think could be useful for someone looking at the repo for the first time and I think we should keep both docs for its own purpose rather than having folks read the other.

docs/usage.md
manodasanW: I agree we should keep this doc to what consumers will run into.  I did end up including WinRT.Projection and WinRT.Interop and bit of those details as folks will see it in their bin folder and may wonder what it is and if they should be distributing it in the case of published libraries and unpackaged apps.  And also because folks are potentially going to get errors from those tools and we want them to be able know what they are reporting issues on.

PR #2369, opened 2026-03-26
Move stub exe generation to MSBuild task with multi-stub support

Sergio0694:
## Summary

Move the inline MSVC compilation logic from the `_CsWinRTGenerateStubExe` MSBuild target into a new `GenerateCsWinRTStubExes` MSBuild task, and add support for generating multiple stub `.exe` files with per-item customization.

## Motivation

The existing `_CsWinRTGenerateStubExe` target can only generate a single stub `.exe`. Users who need multiple stubs (e.g. with different subsystems like Windows vs Console) are forced to copy the entire target and maintain it locally, which is error-prone and hard to keep in sync with CsWinRT updates. By moving the logic into a proper MSBuild task that accepts a collection of items, users can declaratively specify all the stubs they need.

## Changes

- **`src/WinRT.Generator.Tasks/GenerateCsWinRTStubExes.cs`**: new MSBuild task that generates one or more stub `.exe` files by invoking MSVC. Takes a `StubExes` item group as input, with optional per-item metadata:
  - `OutputType`: controls PE subsystem (`WinExe` for WINDOWS, else CONSOLE)
  - `Win32Manifest`: path to manifest to embed
  - `AppContainer`: whether to set `/APPCONTAINER`
  - `Platform`: target platform (arm64, x64, x86)
  - `SourceText`: inline C source code for the stub
  - `SourceFile`: path to a custom `.c` source file
- **`nuget/Microsoft.Windows.CsWinRT.targets`**: refactored `_CsWinRTGenerateStubExe` target to use the new task. Adds `UsingTask` registration, documents the `CsWinRTStubExe` item schema with usage examples, and handles `ResolvedFileToPublish` for all generated stubs. When no `CsWinRTStubExe` items are defined, a default item named after the assembly is added, preserving backward compatibility.

### Usage example

```xml
<ItemGroup>
  <CsWinRTStubExe Include="MyApp.Console" OutputType="Exe" />
  <CsWinRTStubExe Include="MyApp.Windowed" OutputType="WinExe" />
  <CsWinRTStubExe Include="MyApp.Custom" SourceFile="src\CustomStub.c" />
</ItemGroup>
```

### DIFF
--- a/src/WinRT.Generator.Tasks/GenerateCsWinRTStubExes.cs
+++ b/src/WinRT.Generator.Tasks/GenerateCsWinRTStubExes.cs
@@ -1,10 +1,7 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-using System;
-using System.Collections.Generic;
 using System.Diagnostics;
-using System.IO;
 using System.Text;
 using Microsoft.Build.Framework;
 using Microsoft.Build.Utilities;
--- a/src/WinRT.Generator.Tasks/GenerateCsWinRTStubExes.cs
+++ b/src/WinRT.Generator.Tasks/GenerateCsWinRTStubExes.cs
@@ -62,6 +62,23 @@ namespace Microsoft.NET.Build.Tasks;
 ///       Defaults to <see cref="DefaultPlatform"/>.
 ///     </description>
 ///   </item>
+///   <item>
+///     <term><c>SourceText</c></term>
+///     <description>
+///       Inline C source code to use for this stub. When set, the text is written
+///       to a <c>.c</c> file in the intermediate directory and compiled. Takes
+///       precedence over <c>SourceFile</c> and the default <see cref="SourceFilePath"/>.
+///     </description>
+///   </item>
+///   <item>
+///     <term><c>SourceFile</c></term>
+///     <description>
+///       Path to a custom <c>.c</c> source file to use for this stub. When set,
+///       the file is copied to the intermediate directory and compiled. Takes
+///       precedence over the default <see cref="SourceFilePath"/>, but is
+///       overridden by <c>SourceText</c> if both are specified.
+///     </description>
+///   </item>
 /// </list>
 /// </remarks>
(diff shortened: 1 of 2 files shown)


src/WinRT.Generator.Tasks/GenerateCsWinRTStubExes.cs
```diff
@@ -0,0 +1,815 @@
...
+            else if (!string.IsNullOrEmpty(sourceFile))
+            {
+                File.Copy(sourceFile, sourceFilePath, overwrite: true);
+            }
+            else
+            {
+                File.Copy(DefaultSourceFilePath!, sourceFilePath, overwrite: true);
+            }
+
+            // Delete any previously-generated broken .exe in the destination (see https://github.com/dotnet/runtime/issues/111313)
```
manodasanW, line 483: Is the mentioned issue somehow related to the delete if exists?

src/WinRT.Generator.Tasks/GenerateCsWinRTStubExes.cs
```diff
@@ -0,0 +1,815 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
manodasanW: I believe once you pull latest, this should be updated to be consistent with others?

manodasanW (approved):

src/WinRT.Generator.Tasks/GenerateCsWinRTStubExes.cs
Sergio0694: Yeah it's about the fact that publishing with `msbuild -t:Publish` makes the .NET SDK still spit out a broken .exe in the publish folder next to the .dll (if you're publishing with `CustomNativeMain), so this code is always just manually deleting the .exe if it already exists, just in case, so it also works around that scenario. Basically if you're producing a stub .exe with a given name and that file somehow already exists, you should just always overwrite it anyway.

src/WinRT.Generator.Tasks/GenerateCsWinRTStubExes.cs
Sergio0694: Updated! 👍

PR #2370, opened 2026-03-26
Fix Source Link

dongle-the-gadget:
Sets `EmbedUntrackedSources` and `ContinuousIntegrationBuild` to `true`, both are required for Source Link to work.


src/Directory.Build.props
```diff
@@ -70,6 +70,7 @@
     <!-- Set remote for source link. -->
     <AdoMirrorBuild Condition="'$(AdoMirrorBuild)' == ''">false</AdoMirrorBuild>
     <GitRepositoryRemoteName Condition="$(AdoMirrorBuild)">github</GitRepositoryRemoteName>
+    <EmbedUntrackedSources>true</EmbedUntrackedSources>
```
Sergio0694, line 73: I wonder if we should use https://github.com/dotnet/reproducible-builds for source link. That would also publish the repository URL, set the repository URL and branch name, and commit name. Right now we're missing all of that information in the package as well. However, @manodasanW mentioned fixing this might require some manual tweaks since we build the package from a separate ADO pipeline, not from here, so we might need to adjust for that.

dongle-the-gadget: Currently CsWinRT doesn't use .NET built-in tooling for NuGet, it handles NuGet manually. I think it might make more sense currently to set it in ADO/build.cmd

Sergio0694: If that's the case, sure. Would you be able to address that too in this PR while you're at it? 🙂

dongle-the-gadget: I've added the repository URL in 56143b0. Branch and commit are a mixed bag right now since I have no ADO experience 😅.

manodasanW: I will kick off a CI run to see what happens with the PR as is now but docs seem to indicate the branch would be BUILD_SOURCEBRANCHNAME and commit would be BUILD_SOURCEVERSION.

https://learn.microsoft.com/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml

manodasanW: I just saw the changes in publish yml, that should work.  Will see what happens.

branch=$(Build.SourceBranchName);commit=$(Build.SourceVersion)

manodasanW: Looks like this got picked up in the nuget.

build/AzurePipelineTemplates/CsWinRT-Build-Steps.yml
```diff
@@ -161,7 +161,7 @@ steps:
       displayName: Build cswinrt.slnx
       inputs:
         solution: $(Build.SourcesDirectory)\src\cswinrt.slnx
-        msbuildArgs: /restore /p:CIBuildReason=CI,VersionNumber=$(VersionNumber),VersionString=$(Build.BuildNumber),AssemblyVersionNumber=$(WinRT.Runtime.AssemblyVersion),GenerateTestProjection=true,AllowedReferenceRelatedFileExtensions=".xml;.pri;.dll.config;.exe.config",PublishBuildTool=true,BuildToolArch=$(CsWinRTBuildToolsArch) /bl:$(Build.SourcesDirectory)\cswinrt.binlog
+        msbuildArgs: /restore /p:CIBuildReason=CI,VersionNumber=$(VersionNumber),VersionString=$(Build.BuildNumber),AssemblyVersionNumber=$(WinRT.Runtime.AssemblyVersion),GenerateTestProjection=true,AllowedReferenceRelatedFileExtensions=".xml;.pri;.dll.config;.exe.config",PublishBuildTool=true,BuildToolArch=$(CsWinRTBuildToolsArch),ContinuousIntegrationBuild=true /bl:$(Build.SourcesDirectory)\cswinrt.binlog
```
manodasanW, line 164: For `ContinuousIntegrationBuild`, thoughts on moving it to the `Directory.Build.props` like below as per [this](https://learn.microsoft.com/dotnet/core/project-sdk/msbuild-props#continuousintegrationbuild).

```
<PropertyGroup Condition="'$(TF_BUILD)' == 'true'">
  <ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
</PropertyGroup>
```
We got multiple VSBuild@1 calls in this yml, and this might avoid repeating setting that to true.

manodasanW: I did another build that triggers the other VSBuilds and I think that needs to be addressed differently due to RunCsWinRTMergedProjectionGenerator generates that, so we can leave it as is.

PR #2376, opened 2026-03-30
User/chriwall/deprecated

Dreynor87:
This finishes the implementation of the deprecated attribute as described here: https://learn.microsoft.com/en-us/uwp/api/windows.foundation.metadata.deprecatedattribute?view=winrt-26100
Along with the DeprecationType enum as described here: https://learn.microsoft.com/en-us/uwp/api/windows.foundation.metadata.deprecationtype?view=winrt-26100


Sergio0694:
Ported to 3.0 in #2457. @manodasanW let me know if you want to just close this or consider taking it too.

PR #2396, opened 2026-04-17
Add reference assembly support for WinRT.Runtime

Sergio0694:
## Summary

Add reference assembly support for `WinRT.Runtime` so the NuGet package ships with a stripped-down public API surface in `ref\net10.0\` and the full implementation in `lib\net10.0\`. Implementation-detail types and members are removed from the reference assembly via a combination of the new `[WindowsRuntimeImplementationOnlyMember]` marker attribute, a `#define WINDOWS_RUNTIME_IMPLEMENTATION_ONLY_FILE` per-file opt-out, wholesale folder exclusions, and `WINDOWS_RUNTIME_REFERENCE_ASSEMBLY` / `WINDOWS_RUNTIME_IMPLEMENTATION_ASSEMBLY` define constants for the few cases that need finer-grained conditional code.

This supersedes #2377, which used a different mechanism (`#if !REFERENCE_ASSEMBLY` guards sprinkled across ~400 files plus `[Obsolete]` / `[EditorBrowsable(Never)]` attributes on the public-but-implementation-only APIs). The new approach is much less invasive in source files and produces a cleaner reference assembly.

## Motivation

`WinRT.Runtime` exposes many public types and members that are "private implementation details" — they have to be public so generated projection code in other assemblies can call them, but they are not part of the versioned API surface and should not be visible to downstream consumers. Previously these were marked with `[Obsolete]` + `[EditorBrowsable(Never)]`, which left them visible in IntelliSense and reflection and prevented us from cleanly evolving the implementation.

With reference assembly support:

1. The normal build produces the implementation assembly (`lib\net10.0\`) with `WINDOWS_RUNTIME_IMPLEMENTATION_ASSEMBLY` defined.
2. A second build with `CsWinRTBuildReferenceAssembly=true` produces the reference assembly (`ref\net10.0\`) with `WINDOWS_RUNTIME_REFERENCE_ASSEMBLY` defined, with all implementation-only sources removed and remaining method bodies stubbed with `throw null`.

The interop generator and other CsWinRT build tools continue to consume the full implementation assembly, so they retain access to all the private implementation detail APIs they need.

## Changes

- **`src/WinRT.Runtime2/Attributes/WindowsRuntimeImplementationOnlyMemberAttribute.cs`**: new internal `[Conditional("WINDOWS_RUNTIME_REFERENCE_ASSEMBLY")]` marker attribute applied at the top of every implementation-only top-level type
- **`src/WinRT.Runtime2/WinRT.Runtime.csproj`**: define `WINDOWS_RUNTIME_IMPLEMENTATION_ASSEMBLY` for normal builds and `WINDOWS_RUNTIME_REFERENCE_ASSEMBLY` for reference assembly builds; add MSBuild target that scans compile items and removes any source containing `[WindowsRuntimeImplementationOnlyMember]` or the `#define WINDOWS_RUNTIME_IMPLEMENTATION_ONLY_FILE` marker; remove implementation-only folders (`ABI\`, `NativeObjects\`, most `InteropServices\` subfolders) wholesale from the reference assembly compile
- **~280 `.cs` files** in `src/WinRT.Runtime2/`: add the `[WindowsRuntimeImplementationOnlyMember]` marker on implementation-only top-level types, or `#define WINDOWS_RUNTIME_IMPLEMENTATION_ONLY_FILE` at the top of files that should be removed wholesale; conditionally include using directives for `Windows.Foundation.Metadata` / `System.Runtime.Versioning` only in the implementation build; stub remaining method bodies in the reference assembly with `throw null`
- **`src/WinRT.Runtime2/WindowsRuntimeObject.cs`** + **`WindowsRuntimeObject.Impl.cs`**: split `WindowsRuntimeObject` into a partial class so the public surface (class declaration + abstract members) lives in the reference assembly and the concrete implementation lives in `WindowsRuntimeObject.Impl.cs`; restore `HasUnwrappableNativeObjectReference` and `IsOverridableInterface` to `abstract` (reverting the temporary workaround from #2371)
- **`build/AzurePipelineTemplates/CsWinRT-Build-Steps.yml`** and **`nuget/Microsoft.Windows.CsWinRT.nuspec`**: build pipeline runs the reference assembly build and packs both the reference and implementation assemblies into the NuGet
- **`docs/diagnostics/cswinrt30001.md`** *(deleted)*: the `CSWINRT3001` diagnostic for the previous `[Obsolete]`-based approach is no longer applicable
- **`src/WinRT.Runtime2/Properties/WindowsRuntimeConstants.cs`** *(deleted)*: file only contained constants for the removed `[Obsolete]` messages
- **`.github/copilot-instructions.md`**, **`.github/skills/interop-generator/SKILL.md`**, **`.github/skills/update-copilot-instructions/SKILL.md`**: document the new reference assembly mechanism (define constants, marker attribute, file-level opt-out, folder exclusions) and how the interop generator continues to use the implementation assembly
- Several smaller fixes: add `[StructLayout(LayoutKind.Sequential)]` to `ComCallData` and `HSTRING_HEADER`, suppress `CS1574` in `WindowsRuntimeFeatureSwitches.cs`, strip `[SupportedOSPlatform]` and `[ContractVersion]` from public types in the impl assembly, remove the unused `UnreachableException` extension method


Sergio0694:
Superseded by #2453.

PR #2399, opened 2026-04-20
Handle COM disconnect errors when cancelling WinRT async operations

Sergio0694:
## Summary

Handle COM disconnect errors gracefully when cancelling a WinRT async action/operation via the cancellation callback in `AsyncInfoToTaskBridge<TResult, TProgress>`, instead of letting the exception propagate and crash the process.

## Motivation

When awaiting a WinRT async API such as:

```csharp
await SomeWinRTAPIAsync().AsTask(token);
```

if the underlying object lives in a COM server that has died (gets disconnected), cancelling `token` would invoke `IAsyncInfo.Cancel()` from inside a `CancellationToken` callback. That call would then fail with `RPC_E_DISCONNECTED` (or related HRESULTs). The exception would propagate out of the cancellation callback and be re-thrown by `CancellationTokenSource.Cancel()` on whichever thread invoked it (often a thread pool thread), bringing down the entire app.

The pre-existing `try`/`catch` around the registration only protected the rare path where the token was already cancelled at registration time, not the typical case where cancellation happens later.

## Changes

- **`src/cswinrt/strings/additions/Windows.Foundation/Windows.Foundation.cs`**: in `AsyncInfoToTaskBridge<TResult, TProgress>`, wrap the `IAsyncInfo.Cancel()` invocation inside the cancellation callback with a `try`/`catch` filter that swallows `COMException` failures with `HRESULT` of `RPC_E_DISCONNECTED`, `RPC_S_SERVER_UNAVAILABLE`, or `JSCRIPT_E_CANTEXECUTE`. The task is still marked as cancelled by the other registration, so awaiters can handle the cancellation as usual without the process being torn down.

### DIFF
--- a/src/cswinrt/strings/additions/Windows.Foundation/Windows.Foundation.cs
+++ b/src/cswinrt/strings/additions/Windows.Foundation/Windows.Foundation.cs
@@ -296,6 +296,7 @@ namespace System
     {
         private const int RPC_E_DISCONNECTED = unchecked((int)0x80010108);
         private const int RPC_S_SERVER_UNAVAILABLE = unchecked((int)0x800706BA);
+        private const int JSCRIPT_E_CANTEXECUTE = unchecked((int)0x89020001);
 
         private readonly CancellationToken _ct;
         private readonly CancellationTokenRegistration _asyncInfoRegistration;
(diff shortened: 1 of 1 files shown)


src/cswinrt/strings/additions/Windows.Foundation/Windows.Foundation.cs
```diff
@@ -294,29 +294,52 @@ internal class VoidReferenceTypeParameter { }
 #endif
     sealed class AsyncInfoToTaskBridge<TResult, TProgress> : TaskCompletionSource<TResult>
     {
+        private const int RPC_E_DISCONNECTED = unchecked((int)0x80010108);
```
manodasanW, line 297: For consistently, maybe also add JSCRIPT_E_CANTEXECUTE  as we had that in the other place too.

https://github.com/microsoft/CsWinRT/blob/03f40ebea962d2bec9d69fada627782c8fe2730e/src/cswinrt/strings/additions/Windows.Foundation/ExceptionDispatchHelper.cs#L27

manodasanW (approved):

PR #2400, opened 2026-04-20
Handle COM disconnect errors when cancelling WinRT async operations (3.0)

Sergio0694:
## Summary

Prevent app crashes when an `IAsyncInfo.Cancel()` call from the cancellation callback in `AsyncInfoTaskCompletionSource` fails because the backing COM server has been disconnected.

## Motivation

When awaiting a WinRT async API such as:

```csharp
await SomeWinRTAPIAsync().AsTask(token);
```

if the underlying object lives in a COM server that has died, cancelling `token` would invoke `IAsyncInfo.Cancel()` from inside a `CancellationToken` callback. That call would then fail with `RPC_E_DISCONNECTED` (or related HRESULTs). The exception would propagate out of the cancellation callback and be re-thrown by `CancellationTokenSource.Cancel()` on whichever thread invoked it (often a thread pool thread), tearing down the entire process.

The pre-existing `try`/`catch` around the cancellation registration only protected the rare path where the token was already cancelled at registration time, since `CancellationToken.Register(...)` only invokes the callback synchronously in that case. In the typical flow, where cancellation happens later, nothing was guarding the `Cancel()` call.

## Changes

- **`src/WinRT.Runtime2/InteropServices/AsyncInfo/TaskCompletionSources/AsyncInfoTaskCompletionSource{TProgress}.cs`**: collapsed the two cancellation registrations (one to mark the task cancelled, one to call `IAsyncInfo.Cancel()`) into a single `OnCancellation` callback. The callback first calls `TrySetCanceled(_cancellationToken)` to preserve the original ordering (so `Canceled` wins as the task state), then calls `IAsyncInfo.Cancel()` inside a `try`/`catch` that swallows any exception. This ensures no exception can escape the callback and crash the process.
- **`src/WinRT.Runtime2/InteropServices/AsyncInfo/TaskCompletionSources/AsyncInfoTaskCompletionSource{TResult, TProgress}.cs`**: same treatment for the generic `TResult` variant.

### DIFF
--- a/src/WinRT.Runtime2/InteropServices/AsyncInfo/TaskCompletionSources/AsyncInfoTaskCompletionSource{TProgress}.cs
+++ b/src/WinRT.Runtime2/InteropServices/AsyncInfo/TaskCompletionSources/AsyncInfoTaskCompletionSource{TProgress}.cs
@@ -4,12 +4,13 @@
 using System;
 using System.Diagnostics;
 using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
 using System.Runtime.Versioning;
 using System.Threading;
 using System.Threading.Tasks;
 using Windows.Foundation;
 
-#pragma warning disable IDE0010
+#pragma warning disable IDE0010, IDE0055
 
 namespace WindowsRuntime.InteropServices;
 
@@ -85,7 +86,10 @@ internal sealed class AsyncInfoTaskCompletionSource<TProgress> : TaskCompletionS
         {
             _asyncInfo.Cancel();
         }
-        catch
+        catch (Exception e) when (e is COMException { HResult:
+            WellKnownErrorCodes.RPC_E_DISCONNECTED or
+            WellKnownErrorCodes.RPC_S_SERVER_UNAVAILABLE or
+            WellKnownErrorCodes.JSCRIPT_E_CANTEXECUTE })
         {
             // The native 'Cancel' call failed (e.g. the COM server hosting the async
             // operation has been disconnected). Swallow the exception here so it doesn't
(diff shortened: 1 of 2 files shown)


src/WinRT.Runtime2/InteropServices/AsyncInfo/TaskCompletionSources/AsyncInfoTaskCompletionSource{TProgress}.cs
```diff
@@ -78,6 +69,34 @@ public AsyncInfoTaskCompletionSource(IAsyncInfo asyncInfo, CancellationToken can
...
+        // Mark the task as canceled first, so that the 'Canceled' status wins over any
+        // failure from the native 'Cancel' call below. This matches the original ordering
+        // (where the cancellation registration ran before the 'IAsyncInfo.Cancel' one).
+        _ = TrySetCanceled(_cancellationToken);
+
+        try
+        {
+            _asyncInfo.Cancel();
+        }
+        catch
```
Sergio0694: @manodasanW I'm wondering if it does make sense to only handle those 3 exceptions, or if we should just always consider calling `Cancel()` a best-effort here. For now I'm just handling anything. I guess the question here is: do we _want_ to crash the whole process if we get some other failure here from `Cancel`..? 🤔

manodasanW: The other side of the issue is we also do not want to hide failures that an app developer maybe wants to notice and fix. Compared to these 3 specific errors which we are saying are error codes from the infrastructure that are expected in these scenarios.

Sergio0694: Makes sense. I'll make the change and also include this as context, so it's clearer for the future 🙂

manodasanW (approved):

PR #2407, opened 2026-04-24
Add authoring analyzers for [ApiContract] and [ContractVersion]

Sergio0694:
## Summary

Adds a set of new Roslyn diagnostic analyzers in the `WinRT.SourceGenerator2` project that validate Windows Runtime component authoring usage of `[ApiContract]`, `[ContractVersion]`, and `[Version]`. Also introduces small reusable extension helpers and refines the public surface of `ContractVersionAttribute`.

All new analyzers are scoped to component-authoring scenarios (`CsWinRTComponent = true`) and are skipped otherwise.

## Motivation

Authoring Windows Runtime components in C# has a number of WinMD-level constraints that are easy to get wrong silently — invalid `[ApiContract]` enum cases, mismatched `[ContractVersion]` constructor usage, missing version metadata, mixed versioning schemes across the public API surface, etc. These analyzers surface those issues at edit/build time with actionable diagnostics, rather than producing a malformed `.winmd` that fails (or worse, succeeds with wrong metadata) downstream.

### New diagnostics (`CSWINRT2010`–`CSWINRT2017`)

- **`CSWINRT2010` — `ValidApiContractEnumTypeAnalyzer`** (Warning): warns when an `[ApiContract]` enum defines enum cases. API contract types are represented as empty struct types in WinRT and any enum cases are ignored when generating the `.winmd`.
- **`CSWINRT2011` / `CSWINRT2012` / `CSWINRT2013` — `ValidContractVersionAttributeAnalyzer`** (Warning): validates `[ContractVersion]` applications:
  - `2011`: the `(uint)` and `(string, uint)` constructors must target an API contract type.
  - `2012`: the `(Type, uint)` constructor must not target an API contract type.
  - `2013`: the `Type` argument of the `(Type, uint)` constructor must itself be a valid API contract type.
- **`CSWINRT2014` — `ApiContractTypeRequiresContractVersionAnalyzer`** (Warning): every `[ApiContract]` enum must declare its contract version using a version-only constructor of `[ContractVersion]`.
- **`CSWINRT2015` — `PublicTypeRequiresVersioningAnalyzer`** (Info): public top-level types in a component should declare their version using either `[ContractVersion(typeof(SomeContract), version)]` (to associate with an API contract) or `[Version(version)]` (to specify a Windows Runtime version). Skips API contract enums (covered by `CSWINRT2014`) and types that already have either attribute applied.
- **`CSWINRT2016` — `PublicTypeRequiresContractVersionAnalyzer`** (Warning): if at least one public type in the component uses `[ContractVersion]`, then all other public types should also use `[ContractVersion]` for consistency. Whole-compilation analysis (collects candidates during symbol analysis, then reports during the compilation end action). Skips API contract enums.
- **`CSWINRT2017` — `PublicTypeMixedVersioningAttributesAnalyzer`** (Warning): public types should not mix `[ContractVersion]` and `[Version]` on the same type — pick one versioning scheme and apply it consistently. Skips API contract enums.

### Files

- **`src/Authoring/WinRT.SourceGenerator2/Diagnostics/Analyzers/`**: new analyzers `ValidApiContractEnumTypeAnalyzer`, `ValidContractVersionAttributeAnalyzer`, `ApiContractTypeRequiresContractVersionAnalyzer`, `PublicTypeRequiresVersioningAnalyzer`, `PublicTypeRequiresContractVersionAnalyzer`, `PublicTypeMixedVersioningAttributesAnalyzer`.
- **`src/Authoring/WinRT.SourceGenerator2/Diagnostics/DiagnosticDescriptors.cs`**: descriptors for `CSWINRT2010`–`CSWINRT2017`.
- **`src/Authoring/WinRT.SourceGenerator2/AnalyzerReleases.Shipped.md`**: release tracking entries for the new diagnostics.
- **`src/Authoring/WinRT.SourceGenerator2/Extensions/AttributeDataExtensions.cs`**: new `AttributeData.GetLocation(...)` and `AttributeData.GetArgumentLocation(int, ...)` extensions, factored out of the analyzer code.
- **`src/Authoring/WinRT.SourceGenerator2/Extensions/ISymbolExtensions.cs`**: new `ISymbol.GetAttributes(INamedTypeSymbol)` overload that yields only attributes of the specified type, eliminating the manual `foreach` + `SymbolEqualityComparer` filter pattern at call sites.
- **`src/WinRT.Runtime2/Windows.Foundation.Metadata/ContractVersionAttribute.cs`**: small API/doc tweaks aligned with the new analyzer expectations.
- **`src/Tests/SourceGenerator2Test/`**:
  - `Test_ValidApiContractEnumTypeAnalyzer.cs`, `Test_ValidContractVersionAttributeAnalyzer.cs`, `Test_ApiContractTypeRequiresContractVersionAnalyzer.cs`, `Test_PublicTypeRequiresVersioningAnalyzer.cs`, `Test_PublicTypeRequiresContractVersionAnalyzer.cs`, `Test_PublicTypeMixedVersioningAttributesAnalyzer.cs`.
  - `Helpers/CSharpAnalyzerTest{TAnalyzer}.cs`: extended `VerifyAnalyzerAsync` with an `isCsWinRTComponent` parameter that injects a `.globalconfig` setting `build_property.CsWinRTComponent = true` so the gated analyzers actually run under test.

### DIFF
--- a/src/Authoring/WinRT.SourceGenerator2/Diagnostics/Analyzers/ApiContractTypeRequiresContractVersionAnalyzer.cs
+++ b/src/Authoring/WinRT.SourceGenerator2/Diagnostics/Analyzers/ApiContractTypeRequiresContractVersionAnalyzer.cs
@@ -58,13 +58,8 @@ public sealed class ApiContractTypeRequiresContractVersionAnalyzer : DiagnosticA
                 }
 
                 // Check whether any '[ContractVersion]' attribute using a version-only constructor is applied
-                foreach (AttributeData attribute in typeSymbol.GetAttributes())
+                foreach (AttributeData attribute in typeSymbol.GetAttributes(contractVersionAttributeType))
                 {
-                    if (!SymbolEqualityComparer.Default.Equals(attribute.AttributeClass, contractVersionAttributeType))
-                    {
-                        continue;
-                    }
-
                     // The version-only constructors are '(uint)' and '(string, uint)'.
                     // The contract-type constructor is '(Type, uint)', which we want to ignore here.
                     if (attribute.AttributeConstructor?.Parameters is [{ Type.SpecialType: SpecialType.System_UInt32 or SpecialType.System_String }, ..])
--- a/src/Authoring/WinRT.SourceGenerator2/Diagnostics/Analyzers/ApiContractTypeRequiresContractVersionAnalyzer.cs
+++ b/src/Authoring/WinRT.SourceGenerator2/Diagnostics/Analyzers/ApiContractTypeRequiresContractVersionAnalyzer.cs
@@ -1,0 +1,83 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Collections.Immutable;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace WindowsRuntime.SourceGenerator.Diagnostics;
+
+/// <summary>
+/// A diagnostic analyzer that validates that <c>[ApiContract]</c> enum types declare their contract version using <c>[ContractVersion]</c>.
+/// </summary>
+[DiagnosticAnalyzer(LanguageNames.CSharp)]
+public sealed class ApiContractTypeRequiresContractVersionAnalyzer : DiagnosticAnalyzer
+{
+    /// <inheritdoc/>
+    public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = [DiagnosticDescriptors.ApiContractTypeMissingContractVersion];
+
+    /// <inheritdoc/>
+    public override void Initialize(AnalysisContext context)
+    {
+        context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+        context.EnableConcurrentExecution();
+
+        context.RegisterCompilationStartAction(static context =>
+        {
+            // This analyzer only applies to Windows Runtime component authoring scenarios
+            if (!context.Options.AnalyzerConfigOptionsProvider.GlobalOptions.GetCsWinRTComponent())
+            {
+                return;
+            }
+
+            // Get the '[ApiContract]' symbol
+            if (context.Compilation.GetTypeByMetadataName("Windows.Foundation.Metadata.ApiContractAttribute") is not { } apiContractAttributeType)
+            {
+                return;
+            }
+
+            // Get the '[ContractVersion]' symbol
+            if (context.Compilation.GetTypeByMetadataName("Windows.Foundation.Metadata.ContractVersionAttribute") is not { } contractVersionAttributeType)
+            {
+                return;
+            }
+
+            context.RegisterSymbolAction(context =>
+            {
+                // Only enum types can be valid API contract types
+                if (context.Symbol is not INamedTypeSymbol { TypeKind: TypeKind.Enum } typeSymbol)
+                {
+                    return;
+                }
+
+                // Immediately bail if the type is not an API contract type
+                if (!typeSymbol.HasAttributeWithType(apiContractAttributeType))
+                {
+                    return;
+                }
+
+                // Check whether any '[ContractVersion]' attribute using a version-only constructor is applied
+                foreach (AttributeData attribute in typeSymbol.GetAttributes())
+                {
+                    if (!SymbolEqualityComparer.Default.Equals(attribute.AttributeClass, contractVersionAttributeType))
+                    {
+                        continue;
+                    }
+
+                    // The version-only constructors are '(uint)' and '(string, uint)'.
+                    // The contract-type constructor is '(Type, uint)', which we want to ignore here.
+                    if (attribute.AttributeConstructor?.Parameters is [{ Type.SpecialType: SpecialType.System_UInt32 or SpecialType.System_String }, ..])
+                    {
+                        return;
+                    }
+                }
+
+                context.ReportDiagnostic(Diagnostic.Create(
+                    DiagnosticDescriptors.ApiContractTypeMissingContractVersion,
+                    typeSymbol.Locations.FirstOrDefault(),
+                    typeSymbol));
+            }, SymbolKind.NamedType);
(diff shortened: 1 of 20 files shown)


src/Authoring/WinRT.SourceGenerator2/Diagnostics/Analyzers/ApiContractTypeRequiresContractVersionAnalyzer.cs
```diff
@@ -0,0 +1,78 @@
...
+            // Get the '[ContractVersion]' symbol
+            if (context.Compilation.GetTypeByMetadataName("Windows.Foundation.Metadata.ContractVersionAttribute") is not { } contractVersionAttributeType)
+            {
+                return;
+            }
+
+            context.RegisterSymbolAction(context =>
+            {
+                // Only enum types can be valid API contract types
+                if (context.Symbol is not INamedTypeSymbol { TypeKind: TypeKind.Enum } typeSymbol)
```
Youssef1313, line 49: nit: Cast instead of type check.

```suggestion
                var typeSymbol = (INamedTypeSymbol)context.Symbol;
                if (typeSymbol.TypeKind != TypeKind.Enum)
```

Sergio0694: I prefer the pattern mostly because it's more compact 😅

Aren't they semantically the same here though? Like, this symbol will always be `INamedTypeSymbol`, meaning both cases would still do a cast and both would always succeed anyway. The only difference is a technically unnecessary `null` check with the pattern, but it shouldn't really matter for perf. Or is there something else?

Youssef1313: Nothing, other than that if you added an additional SymbolKind, the suggestion will cause an invalid cast exception. But otherwise they are equivalent and no perf concern. It's definitely fine if you prefer the pattern.

src/Authoring/WinRT.SourceGenerator2/Diagnostics/Analyzers/PublicTypeRequiresContractVersionAnalyzer.cs
```diff
@@ -0,0 +1,75 @@
...
+
+            // Get the '[ApiContract]' symbol (used to skip API contract types, which are validated by a different analyzer)
+            INamedTypeSymbol? apiContractAttributeType = context.Compilation.GetTypeByMetadataName("Windows.Foundation.Metadata.ApiContractAttribute");
+
+            context.RegisterSymbolAction(context =>
+            {
+                INamedTypeSymbol typeSymbol = (INamedTypeSymbol)context.Symbol;
+
+                // Only consider top-level public types
+                if (typeSymbol is not { DeclaredAccessibility: Accessibility.Public, ContainingType: null })
```
Youssef1313: The pattern isn't applied consistently here though 😄

Sergio0694: Ah, this was an oversight, fixed it! Thank you 😄

src/Authoring/WinRT.SourceGenerator2/Diagnostics/Analyzers/PublicTypeRequiresContractVersionAnalyzer.cs
```diff
@@ -0,0 +1,73 @@
...
+                }
+
+                // Skip if any '[ContractVersion]' attribute is already applied (validity of the
+                // specific constructor used is reported by the other 'ContractVersion' analyzers)
+                if (typeSymbol.HasAttributeWithType(contractVersionAttributeType))
+                {
+                    return;
+                }
+
+                context.ReportDiagnostic(Diagnostic.Create(
```
manodasanW: They might be doing regular versioning rather than contract versioning which we should take into account.

manodasanW: and I believe we default version attribute if not specified.

Sergio0694: Yeah this is why this diagnostic is just info, not a warning. Is that fine or should we still change it?

manodasanW: Given regular versioning and contract versioning have different audiences, I think we basically have a property that someone can set to true if they want to use contract versioning and if that is set, all the diagnostics for it are enabled and enforced.  If it isn't set and it will default to false, then regular versioning is done and the diagnostics are not enforced.  Thoughts?

Sergio0694: Mh having an additional MSBuild property feels a bit overkill perhaps. What about:
- Having an analyzer (info level) that fires if a public type has neither a version nor a contract version
  - I know we add a version by default, this is why the level would just be info, not a warning
- Having an analyzer that warns if there's a public type without a contract version, **only if there's at least another public type with a contract version**. As in: if you do use contract versioning, you should be consistent and not mix.
- All other analyzers are always on, since they're about general correctness things.

This way there's also no property to worry about, things just adapt automatically to your code.
Thoughts?

manodasanW: That could work as long as the diagnostic message for the first one is clear enough to be like you are not using regular or contract versioning, so we will default version to 1 unless you manually specify or so.

Sergio0694: Updated 🙂

manodasanW (approved):

PR #2410, opened 2026-04-27
Port authoring projection support

manodasanW:
- Got authoring projection support working enabling to consume from native app
- Generating a WinRT.Component.dll using the merged projection generator which has all the component projections generated within it.
- Updated JIT hosting support to work with the new projections and simplified it just to do everything in default ALC given we are already merging all the projections
- Initial support for merging multiple components that are consumed by a native app but will be tested later
- Added CsWinRTBuildForNativeConsumer which is what we are using to trigger all the merged projections and so on being created when a regular library is consumed by a native app given it is not a C# exe project
- Added more test cases


src/Authoring/WinRT.SourceGenerator2/AuthoringExportTypesGenerator.Execute.cs
```diff
@@ -281,11 +282,13 @@ namespace ABI.{info.AssemblyName.EscapeIdentifierName()};
...
                 using global::System.Runtime.CompilerServices;
                 using global::System.Runtime.InteropServices;
+                using global::System.Runtime.Versioning;
                 using global::WindowsRuntime.InteropServices;
+                using global::WindowsRuntime.InteropServices.Marshalling;
 
                 using HRESULT = int;
                 using unsafe HSTRING = void*;
-                using IActivationFactory = void;
+                using unsafe IActivationFactory = void*;
```
Sergio0694, line 291: Just curious, any particular reason?

src/WinRT.Interop.Generator/Generation/InteropGenerator.Discover.cs
```diff
@@ -534,15 +617,29 @@ private static string[] GetAssembliesToProcess(InteropGeneratorArgs args)
...
+                fileName.SequenceEqual(InteropNames.WindowsRuntimeComponentDllName))
+            {
+                continue;
+            }
+
+            assembliesToProcess.Add(path);
         }
 
-        return [.. assembliesToProcess];
+        return ([.. assembliesToProcess], [.. assembliesToProcess]);
```
Sergio0694: Is this by accident? The method now returns a tuple but both are the same exact value..?

src/Authoring/WinRT.SourceGenerator2/AuthoringExportTypesGenerator.Execute.cs
manodasanW: when it was just void, that was failing to compile.

src/WinRT.Interop.Generator/Generation/InteropGenerator.Discover.cs
manodasanW: I believe this was from earlier in the porting when I was trying to fix the public key issue a different way by passing 2 set of dlls, but then I ended up just putting the public key without signing it.  I will fix it.

src/WinRT.Runtime2/InteropServices/Activation/IActivationFactory.cs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace WindowsRuntime.InteropServices;
+
+/// <summary>
+/// Enables a class to be activated by the Windows Runtime.
+/// </summary>
+/// <see href="https://learn.microsoft.com/windows/win32/api/activation/nn-activation-iactivationfactory"/>
+[WindowsRuntimeMetadata("Windows.Foundation.FoundationContract")]
+[Guid("00000035-0000-0000-C000-000000000046")]
+[EditorBrowsable(EditorBrowsableState.Never)]
```
Sergio0694, line 16: Why is this hidden? Didn't we want this to be public so users can implement custom factories too?

src/WinRT.Runtime2/InteropServices/ProjectionImpls/IActivationFactoryImpl.cs
```diff
@@ -0,0 +1,78 @@
...
+            return WellKnownErrorCodes.E_POINTER;
+        }
+
+        *instance = null;
+
+        try
+        {
+            IActivationFactory factory = ComInterfaceDispatch.GetInstance<IActivationFactory>((ComInterfaceDispatch*)thisPtr);
+
+            object activated = factory.ActivateInstance();
```
Sergio0694, line 65: If users can implement this, should we validate that `activated` isn't `null`, and return some `E_INVALID***` or something if that's the case?

Sergio0694 (approved):

src/WinRT.Runtime2/InteropServices/Activation/IActivationFactory.cs
manodasanW: Custom factories was next on my list to look at after this is merged. So I might unhide it then.

src/WinRT.Runtime2/InteropServices/ProjectionImpls/IActivationFactoryImpl.cs
manodasanW: yea I can do that when I do custom factories support

PR #2413, opened 2026-05-01
Fix EventSourceCache leak and array marshaller allocation size

manodasanW:
1. EventSourceState.OnDispose: pass local 'thisPtr' (not nulled field '_thisPtr') to EventSourceCache.Remove. The field is nulled on the preceding line, so the cache entry was never actually removed, leaking entries for the lifetime of the application.

2. Array marshallers: allocate using sizeof(TAbi) instead of sizeof(T). The destination buffer is indexed as TAbi*, so the allocation must match the ABI element size.


Sergio0694 (approved):
Oopsies, good finds! I see we didn't have any test failures for those incorrect array marshaller paths allocating the wrong array size, should we add some new unit tests that would exercise that code so we have a regression test as well, just to have a bit more coverage too?

manodasanW:
We do have tests that exercise the code path, its just most of the time the size for both is equal.  And there doesn't seem to be a scenario where we would under allocate, only over allocate with DateTimeOffset.

PR #2417, opened 2026-05-11
Improve hosting support experience

manodasanW:
- Added some native targets for where if CsWinRT package is included by a native app, it will handle making sure the right WinRT.Host.dll gets copied which was an issue in the past when the component was AnyCPU but app was like x64.  It will also handle making the JIT hosting scenarios less complicated in our new model given we need a single-entry point for TypeMap to work.
- To improve the JIT hosting scenarios, WinRT.Host.dll now calls into WinRT.Component.dll which serves as the main entry point (and sets it) and then redirects it to each components activation factory to try to find the type.
- Added sample for AOT scenarios too where we are not doing something similar to JIT but instead providing the developer a way to create a project to merge all the components that they want to use and merge the activation factories for them.  A developer can also choose to publish each component individually which is part of why we are not automating this process.  This is similar to the experience before.
- Added missing binaries in nuget and also cleaned up the included targets.
- There are scenarios where WinRT.Projection.dll may not get generated, so adding handling for that.


nuget/Microsoft.Windows.CsWinRT.targets
```diff
@@ -60,6 +69,10 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <CompilerVisibleProperty Include="CsWinRTMergeReferencedActivationFactories" />
     <CompilerVisibleProperty Include="CsWinRTAotWarningLevel" />
     <CompilerVisibleProperty Include="CsWinRTGenerateManagedDllGetActivationFactory" />
+    <!-- PublishAot is needed by AuthoringExportTypesGenerator.ShouldEmitNativeExports for the
+         merged-aggregator scenario (non-component project with CsWinRTMergeReferencedActivationFactories=true).
+         Authoring.targets exposes the same property but only imports when CsWinRTComponent=true. -->
+    <CompilerVisibleProperty Include="PublishAot" />
```
Sergio0694: Super nit: leave a blank line before the comment and use the same style as the comment above (L37)

src/WinRT.Projection.Generator/Generation/ProjectionGenerator.WinRTComponentSources.cs
```diff
@@ -0,0 +1,278 @@
...
+
+    /// <summary>
+    /// Emits a <c>[ModuleInitializer]</c>-driven <c>SetEntryAssembly</c> in <c>WinRT.Component.dll</c>.
+    /// This designates <c>WinRT.Component.dll</c> as the entry assembly the .NET runtime uses
+    /// to root <c>[TypeMapAssemblyTarget]</c> discovery. Idempotent: skipped if the host has
+    /// already set an entry assembly. Per-component dlls intentionally do not emit a similar
+    /// initializer; consumers using a single component dll directly must call
+    /// <c>Assembly.SetEntryAssembly</c> themselves if they need TypeMap discovery rooted there.
+    /// </summary>
+    private static void WriteProjectionTypesInitializer(ProjectionGeneratorProcessingState processingState)
```
Sergio0694, line 71: I don't really like this one, is there some way we can drop module initializers entirely at least in some scenarios? That is, can we statically determine whether someone else will set the entry assembly already and skip this? For instance, is this needed in all of these scenarios?
- A standalone WinRT component published as NAOT
- An application with also a WinRT component, with .exe/.dll stub merging

Can we figure out a way to perhaps just scope this down to the mixed C++ scenarios that need it?

manodasanW: I thought I scoped this to JIT, but I think I undid that when I changed the approach.  Addressed it by using a property to control when it is emitted in native JIT scenarios rather than always emitting it.  It also won't be needed for JIT once the other .NET change is available.

src/WinRT.Projection.Generator/Generation/ProjectionGenerator.WinRTComponentSources.cs
```diff
@@ -0,0 +1,304 @@
...
+    /// component '.dll' directly must call <see cref="System.Reflection.Assembly.SetEntryAssembly"/>
+    /// themselves if they need interop type map discovery rooted there.
+    /// </summary>
+    private static void WriteProjectionTypesInitializer(ProjectionGeneratorProcessingState processingState)
+    {
+        const string source = """
+            // <auto-generated/>
+            #pragma warning disable
+
+            namespace ABI.WinRT.Component;
```
Sergio0694, line 77: Shouldn't we update all the generated namespaces to be `WindowsRuntime` and not `WinRT`, for consistency with all the other binaries in CsWinRT 3.0? Also keep the various comments in sync if we do this change.

Sergio0694 (approved):

src/WinRT.Projection.Generator/Generation/ProjectionGenerator.WinRTComponentSources.cs
manodasanW: I was just keeping this in sync with the other one for ABI.<ModuleName>.ManagedExports where ModuleName is WinRT.Component.dll rather than having both namespaces.  This would probably be removed later at some point anyways.

PR #2418, opened 2026-05-12
Refactor projection writer: cleanup passes, parallelization, and CI-driven correctness fixes

Sergio0694:
## Summary

Second PR in the CsWinRT projection-writer migration series. Mostly automated cleanup of the C# port that landed in the previous PR: refactor passes, type-semantics consolidation, parallelization, error-discipline alignment, and a stack of CI-driven semantic + cosmetic fixes. The output of the writer now compiles cleanly against the real Windows SDK and matches the original C++ writer's behavior across all 8 regression scenarios.

## Motivation

The previous PR landed the initial C++ -> C# port of `cswinrt` as `WinRT.Projection.Writer`. That port was a faithful 1:1 transcription, but the resulting code carried a lot of structural debt from the C++ shape: long mega-functions, ad-hoc helper conventions inherited from C++ string-template emission, defensive null checks the C++ never had, and a few correctness regressions introduced during automated refactors.

This PR is the cleanup pass: refactor toward C# idioms, share helpers across the writer/factories, parallelize the per-namespace work, and fix every semantic + cosmetic regression we surfaced via CI iteration. The goal is to leave the writer in a state where the only remaining work is hand-polish (next PR) and deleting the legacy C++ tree (final PR).

This is **PR 2 of 4** in the migration series:

1. Initial C# port (already merged)
2. **This PR** -- automated/structural cleanup + correctness fixes
3. Upcoming: manual polishing pass + interpolated-string-handler optimization for zero-allocation emit paths
4. Upcoming: delete the legacy C++ `cswinrt` tree, the test runner, and other temporary scaffolding

### Resolver and helper consolidation

- **`Resolvers/AbiTypeShapeResolver.cs`** -- centralized ABI type classification (BlittablePrimitive, Enum, BlittableStruct, ComplexStruct, NullableT, GenericInstance, RuntimeClassOrInterface, Delegate, Object, MappedAbiValueType, etc.) so emission paths use a single source of truth instead of inline `IsXxx` predicates duplicated across factories. Two helper methods were added: `IsBlittableStruct` (blittable-only) and `IsAnyStruct` (blittable OR complex), and ~36 call sites were updated to use the right one.
- **`Helpers/AbiTypeHelpers.*.cs`** -- split the legacy `AbiTypeHelpers.cs` into focused partials (`AbiTypeNames`, `Blittability`, etc.).
- **`Metadata/TypeSemantics.cs`** -- moved the `TypeSemantics` discriminated union out of helpers and trimmed the trailing-underscore record names.

### Generation pipeline

- **`Generation/ProjectionGenerator.cs` + `Generation/WorkItems/`** -- parallelized projection generation across namespaces, IIDs, and the component-module aggregator via `Parallel.ForEach`. Cross-item shared state moved into `ProjectionGeneratorRunState` with concurrent collections.
- **`Generation/ProjectionGenerator.GeneratedIids.cs`** -- pulled `GeneratedInterfaceIIDs.cs` emission into its own work item; restored proper `IncreaseIndent`/`DecreaseIndent` around the per-IID property emission.
- **`Generation/ProjectionGenerator.Generate.cs`** -- fixed an SDK-mode fallback that forgot to set `hasTypesToProject = true` when no `Microsoft.Windows.SDK.NET` reference assembly was present (the generator silently produced no output).

### Writer infrastructure

- **`Writers/IndentedTextWriter.cs`** -- removed all 4 auto-fixup heuristics (brace-prepend on `{`/`}`, auto-trailing-newline on multilines ending in `;`/`}`/`]`/`#`, brace-prepend in `WriteLine(content)`, always-on idempotent `WriteLine()`). The writer now mirrors the source-generator design: callers are responsible for emitting the right newlines, with `WriteBlock()` providing the canonical brace+indent pattern. Idempotent newline behavior is opt-in via `WriteLine(skipIfPresent: true)`.
- **`Writers/IndentedTextWriterPool.cs`** -- new pool to reduce allocations across parallel work items. Acquired via the disposable `IndentedTextWriterOwner` ref struct.
- **`Helpers/TypedefNameWriter.cs`** -- restored `InAbiNamespace` / `InAbiImplNamespace` tracking that was lost during the refactor; restored 4 missing conditions in the namespace-prefix logic verbatim from the C++ source (`code_writers.h:363-369`) so interface declarations no longer get fully qualified inside their own ABI namespace.

### Error discipline

- **`Errors/WellKnownProjectionWriterExceptions.cs`** -- aligned exception emit with the `WinRT.Interop.Generator` style: error IDs in the 5000+ range, well-known guards in `ProjectionWriter.Run`, per-site `ThrowOrAttach` exception context, no `[SuppressMessage]` workarounds, `Settings` made truly read-only via a one-shot `MakeReadOnly()` pattern.
- **`Generation/ProjectionGenerator.Emit.cs` + `Generation/ProjectionGeneratorArgs.cs`** -- added `MaxDegreesOfParallelism` plumbing through the CLI; better cancellation granularity.

### Code cleanups

- **All factories under `Factories/`** -- 50+ refactor passes (commits prefixed `R2`, `R3`, `F`, `N`, `Pass <N>`) for consistent code style: single-line `if`s expanded with blank-line padding, multi-line `WriteLine` -> raw multiline strings, scratch-writer string-returning overloads, fully-qualified marshaller call sites, etc.
- **`Resources/Additions/Microsoft.UI.Xaml/Microsoft.UI.Xaml.GridLength.cs` + `Resources/Additions/Windows.UI.Xaml/Windows.UI.Xaml.GridLength.cs`** -- fixed a `Pixe)l` typo introduced by the IDE0078 automated refactor (`type is not (GridUnitType.Auto or GridUnitType.Pixe)l && ...` was supposed to be `type is not (GridUnitType.Auto or GridUnitType.Pixel) && ...`).
- **`Resources/Base/ReferenceInterfaceEntries.cs`** -- moved `using static System.Runtime.InteropServices.ComWrappers;` after the normal `using` block per project convention.

### Semantic regressions fixed (surfaced by CI)

The following semantic bugs were introduced during the cleanup passes and caught only by iterating against CI. This is exactly why `validate-compile.ps1` was added (see "Validation infrastructure" below).

- **Nullable<T> getter ABI marshalling** -- `returnIsRefType` was missing `AbiTypeShapeKind.NullableT`, so any `Nullable<T>` getter (e.g. `Wallet.ExpirationDate`) emitted `int __retval` + a direct cast instead of `void* __retval` + `try`/`finally` + `Marshaller.UnboxToManaged`.
- **Enum array (`EnumType[]`) returns** -- `IsBlittablePrimitive` excluded the `Enum` shape, so `returnIsReceiveArray` failed for enum-element arrays. `Span<EnumType>` FillArray emitted `CopyToManaged_X` referencing nonexistent `ABI.Foo.EnumName` types.
- **Complex struct returns** -- `returnIsAnyStruct` covered both `BlittableStruct` AND `ComplexStruct` shapes, so the dispatch never reached the `returnIsComplexStruct` branch. ComplexStruct returns (e.g. `GpioChangeCount`, `GpioChangeRecord`, `AccessListEntry`, `StorePackageUpdateStatus`) emitted the projected struct as `__retval` instead of the ABI struct, with no `ConvertToManaged` wrap.
- **`ProfileUsageMarshaller` missing methods** -- `StructEnumMarshallerFactory.almostBlittable` was wrongly true for complex structs, so `isComplexStruct` became false and the marshaller class skipped emitting `ConvertToUnmanaged` / `ConvertToManaged` / `Dispose`.
- **`source: typeof(Foo)` line split** -- the IDIC `TypeMapAssociation` emitter wrapped `typeof(` with a trailing newline, splitting the typedef name across lines.
- **Comma-on-own-line in vtable args** -- multiline `WriteLine` in the PassArray/Out/Ref/ReceiveArray loops appended a trailing newline so the next iteration's leading `,` landed at start-of-line and got an indent prepended.
- **Interface declaration mangling in component mode** -- TypedefNameWriter dropped 3 conditions from the C++ namespace-prefix logic, causing `internal interface global::ABI.Impl.AuthoringTest.IBasicClassClass` (CS1514/CS1513/CS1001).
- **ABI block type-name resolution (`InAbiNamespace`)** -- the writer was emitting projected-namespace type names without the `global::` prefix when inside the corresponding ABI namespace, causing CS0246 / CS9051 (file-local type misuse) on `LearningModelPixelRange` and similar generic interface types.
- **WriteStruct body indent + WriteFileHeader trailing newline + WriteDefault/ExclusiveToInterfaces patterns** -- 5+ structural emit sites refactored to use the `WriteBlock()` pattern instead of flat raw strings without `IncreaseIndent`.

### Cosmetic regressions fixed

- **R1 (ABI delegate `Invoke` wrapper indent)**, **R2 (XAML CCW marshaller IID accessor)**, **R3 (`struct X{` brace packing in Vftbl)**, **R5 (`GeneratedInterfaceIIDs.cs` member indent)**, **R6 (extra blank line after `namespace X {`)** -- detailed in commit `f6dc5485`. All produce valid C# but were ugly. R4 (long-line wrap from intentional fully-qualified marshaller names) was accepted as the cost of the safety improvement.

### Validation infrastructure

The previous validation harness only diffed PRE writer output against POST. That missed several classes of bug:

- Embedded resource files copied verbatim into the projection (the `Pixe)l` typo had matching PRE/POST diffs because the typo lived in the resource).
- Semantic regressions in marshalling code (the writer happily produces "different but still self-consistent" malformed code).
- Syntax errors that need binding context to surface.

Two additions:

- **`MaxDegreesOfParallelism` plumbing through `cswinrtprojectiongen.exe` CLI** -- so deterministic builds and reduced-parallelism debug runs work the same as the interop generator.
- **Real CI-style compile coverage (kept in session state, not committed)** -- end-to-end build of `WinRT.Sdk.Projection` + `WinRT.Sdk.Xaml.Projection` against the actual Windows 10.0.26100 SDK winmds, plus syntax-only validation across all 8 regression scenarios. Catches CS1xxx parse errors and CS9051 file-local-type errors. Verified by reverting each fix temporarily and confirming the script flags the original CI error.

## Validation

- All 8 regression scenarios produce expected output (244 / 1553 byte-identical vs pre-refactor truth; remaining diffs are the intentional cosmetic improvements: fully-qualified marshaller names, colon-spacing, multi-line `<summary>`, `[FlagsAttribute]` -> `[Flags]`, blank-line consistency).
- `WinRT.Sdk.Projection` and `WinRT.Sdk.Xaml.Projection` compile cleanly against the real Windows SDK.
- `WinRT.Projection.Writer` builds with 0 warnings under `TreatWarningsAsErrors`.
- `WinRT.Projection.Generator` publishes to Native AOT with 0 warnings.
- CI is green.

## Coming next

- **PR 3** (next): manual polishing pass plus a switch to interpolated string handlers in the hot emit paths to eliminate intermediate `StringBuilder`/`string.Format` allocations.
- **PR 4** (final): delete the C++ `cswinrt` tree under `src/cswinrt/`, the `WinRT.Projection.Writer.TestRunner` scaffolding, and any other temporary scaffolding that exists only to bridge the migration.


src/WinRT.Projection.Writer/Models/AbiTypeShapeKind.cs
```diff
@@ -0,0 +1,86 @@
...
+
+    /// <summary>
+    /// A WinRT struct whose fields are all blittable primitives or other blittable structs (no per-field marshalling needed).
+    /// </summary>
+    BlittableStruct,
+
+    /// <summary>
+    /// A WinRT struct with at least one reference-type field (string, generic instance, runtime class, etc.) that needs per-field marshalling via a generated <c>*Marshaller</c>.
+    /// </summary>
+    ComplexStruct,
```
manodasanW, line 35: maybe this should just be NonBlittableStruct?

src/WinRT.Projection.Writer.TestRunner/Program.cs
```diff
@@ -36,10 +36,79 @@ public static int Main(string[] args)
...
+            ProjectionWriter.Run(new ProjectionWriterOptions
+            {
+                InputPaths = inputs, OutputFolder = outputFolder,
+                Include = include, Exclude = exclude,
+                Component = component, Internal = internalMode,
+                ReferenceProjection = refMode, Verbose = false,
+                MaxDegreesOfParallelism = maxDegreesOfParallelism,
+            });
+        }
+        catch (Exception ex) { Console.Error.WriteLine($"ERROR: {ex.Message}"); Console.Error.WriteLine(ex.StackTrace); return 1; }
```
manodasanW, line 107: Consider fixing indenting / formatting throughout this file

Sergio0694: I'll delete this file entirely in a follow-up, this is just for testing 👍

src/WinRT.Projection.Writer/Models/AbiTypeShapeKind.cs
Sergio0694: Agree. I can fix it in the follow-up (to avoid conflicts).

src/WinRT.Projection.Writer/Factories/AbiMethodBodyFactory.DoAbi.cs
```diff
@@ -0,0 +1,902 @@
...
+            writer.Write($"global::ABI.System.TypeMarshaller.ConvertToManaged({pname})");
+        }
+        else if (context.AbiTypeShapeResolver.IsComplexStruct(p.Type))
+        {
+            // Complex struct input (server-side): convert ABI struct to managed via marshaller.
+            writer.Write($"{AbiTypeHelpers.GetMarshallerFullName(writer, context, p.Type)}.ConvertToManaged({pname})");
+        }
+        else if (context.AbiTypeShapeResolver.IsBlittableStruct(p.Type))
+        {
+            // Blittable / almost-blittable struct: pass directly (projected type == ABI type).
```
manodasanW, line 889: almost blittable?

src/WinRT.Projection.Writer/Factories/AbiMethodBodyFactory.DoAbi.cs
```diff
@@ -0,0 +1,902 @@
...
+            writer.Write($"{AbiTypeHelpers.GetMappedMarshallerName(p.Type)}.ConvertToManaged({pname})");
+        }
+        else if (p.Type.IsSystemType())
+        {
+            // System.Type input (server-side): convert ABI Type struct to System.Type.
+            writer.Write($"global::ABI.System.TypeMarshaller.ConvertToManaged({pname})");
+        }
+        else if (context.AbiTypeShapeResolver.IsComplexStruct(p.Type))
+        {
+            // Complex struct input (server-side): convert ABI struct to managed via marshaller.
```
manodasanW, line 884: server side?

src/WinRT.Projection.Writer/Generation/ProjectionGenerator.cs
```diff
@@ -0,0 +1,191 @@
...
+        catch (Exception e) when (!e.IsWellKnown)
+        {
+            throw new UnhandledProjectionWriterException("discovery", e);
+        }
+
+        _token.ThrowIfCancellationRequested();
+
+        ProjectionGeneratorRunState state = new(componentActivatable, componentByModule);
+
+        // Phase 3..6: parallel emission. All file writes happen below; wrap the whole emission
```
manodasanW, line 73: 2?

src/WinRT.Projection.Writer/Generation/Settings.cs
```diff
@@ -0,0 +1,155 @@
...
+
+    /// <summary>
+    /// Gets or sets a value indicating whether projected types are emitted as <c>internal</c> rather than <c>public</c>.
+    /// </summary>
+    public bool Internal { get; init; }
+
+    /// <summary>
+    /// Gets or sets a value indicating whether the projection is embedded into a consuming assembly (forces <c>internal</c> visibility).
+    /// </summary>
+    public bool Embedded { get; init; }
```
manodasanW, line 113: can remove

src/WinRT.Projection.Writer/Factories/ConstructorFactory.FactoryCallbacks.cs
```diff
@@ -0,0 +1,688 @@
...
+                      WindowsRuntimeActivationArgsReference additionalParameters,
+                      WindowsRuntimeObject baseInterface,
+                      out void* innerInterface,
+                      out void* retval)
+                    {
+                """, isMultiline: true);
+        }
+        else
+        {
+            // Sealed Invoke signature is multi-line..
```
manodasanW, line 105:
```suggestion
            // Sealed Invoke signature is multi-line.
```

manodasanW (approved):

src/WinRT.Projection.Writer/Models/AbiTypeShapeKind.cs
Sergio0694: Fixed it in the follow-up 👍

src/WinRT.Projection.Writer/Factories/AbiMethodBodyFactory.DoAbi.cs
Sergio0694: Fixed it in the follow up 👍

src/WinRT.Projection.Writer/Factories/AbiMethodBodyFactory.DoAbi.cs
Sergio0694: Fixed

src/WinRT.Projection.Writer/Generation/Settings.cs
Sergio0694: Done!

PR #2424, opened 2026-05-26
Add CsWinRT 2.x polyfills MSBuild target and wire it into the package

Sergio0694:
## Summary

Adds a new MSBuild target file, `Microsoft.Windows.CsWinRT.CsWinRT2Polyfills.targets`, that generates C# polyfill attributes for CsWinRT 2.x interop (`Microsoft.CodeAnalysis.EmbeddedAttribute` and `WinRT.DynamicWindowsRuntimeCastAttribute`) when `CsWinRTGenerateVersion2PolyfillAttributes` is `true`, and wires it into the NuGet package and build pipeline alongside the existing `CsWinRTGen.targets`.

## Motivation

Libraries that multi-target both CsWinRT 2.x and 3.0 need to be able to reference the 2.x `[DynamicWindowsRuntimeCast]` attribute (used to preserve metadata for dynamic WinRT casts under trimming) from a single shared codebase, without taking a hard reference on the 2.x runtime when compiling against 3.0. Providing a generated polyfill (guarded by the `CSWINRT2_0_POLYFILL_ATTRIBUTES` conditional) lets shared source compile cleanly under 3.0 while still emitting the real attribute when targeting 2.x. The `[Embedded]` attribute is co-generated so the polyfill stays internal even in the presence of `[InternalsVisibleTo]`.

The new target file was added in the previous commit but was not yet packaged or imported; without these changes it would not be picked up by consumers of the CsWinRT NuGet.

## Changes

- **`nuget/Microsoft.Windows.CsWinRT.nuspec`**: include `Microsoft.Windows.CsWinRT.CsWinRT2Polyfills.targets` in the `build` folder of the NuGet package.
- **`nuget/Microsoft.Windows.CsWinRT.targets`**: add an unconditional `<Import>` for the new targets file (the inner `Target` is already gated on `CsWinRTGenerateVersion2PolyfillAttributes`), mirroring the `CsWinRTGen.targets` pattern.
- **`src/cswinrt.slnx`**: list the new file under the `NuGet` solution folder so it shows up in IDEs.
- **`.github/copilot-instructions.md`**: add a row to the `.targets` files table describing the new file's role.


nuget/Microsoft.Windows.CsWinRT.CsWinRT2Polyfills.targets
```diff
@@ -0,0 +1,78 @@
...
+-->
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <!-- Generates polyfills for some CsWinRT 2.x attributes to help with multi-targeting -->  
+  <Target Name="GenerateDynamicWindowsRuntimeCastAttribute"
+          BeforeTargets="BeforeCompile;CoreCompile"
+          DependsOnTargets="PrepareForBuild"
+          Condition="'$(Language)' == 'C#' AND '$(CsWinRTGenerateVersion2PolyfillAttributes)' == 'true'"
+          Inputs="$(MSBuildAllProjects)"
+          Outputs="$(GeneratedDynamicWindowsRuntimeCastAttributeFile)">
```
manodasanW, line 22: Looked like the property is set within the target, I feel like for this to work as expected, it should be outside.

manodasanW (approved):

nuget/Microsoft.Windows.CsWinRT.CsWinRT2Polyfills.targets
Sergio0694: Oh good point, fixed!

PR #2429, opened 2026-06-08
Port WindowsRuntime.Internal.idl to C# in new WinRT.Internal project

Sergio0694:
## Summary

Port the contents of `src/cswinrt/WindowsRuntime.Internal.idl` (the hand-authored Windows Runtime metadata that the legacy C++ `cswinrt` tool used to compile into `WindowsRuntime.Internal.winmd`) to a new C# `WinRT.Internal` project, and wire the existing C# WinMD generator in as the producer of the `.winmd`. The new `WinRT.Internal` project now owns this artifact end-to-end so the CsWinRT toolchain no longer needs the legacy C++ `cswinrt` project to be built at all.

## Motivation

The legacy C++ `cswinrt` project does two things: it produces `cswinrt.exe` (the old projection compiler) **and** it compiles `WindowsRuntime.Internal.idl` into `WindowsRuntime.Internal.winmd`. The first responsibility has already been ported to C# (the `WinRT.Projection.Writer` library + `cswinrtprojectiongen.exe`); the second was still tying the rest of the build to the C++ project.

Moving the `.idl` to a C# source-of-truth + the C# WinMD generator removes that dependency. A follow-up PR can now delete `src/cswinrt` entirely.

### New `WinRT.Internal` C# project

- **`src/WinRT.Internal/WinRT.Internal.csproj`**: targets `net10.0-windows10.0.26100.1` (TFM revision `.1` selects the `cswinrt3` Windows SDK projection reference assemblies which carry `[WindowsRuntimeMetadata]` attributes the WinMD generator reads). Pins `WindowsSdkPackageVersion` so the SDK adds the matching framework reference. Disables all CsWinRT MSBuild processing (`CsWinRTEnabled` / `CsWinRTGenerateProjection` / `CsWinRTGenerateInteropAssembly[2]`) since this project itself feeds back into the CsWinRT toolchain. References the locally built `WinRT.Runtime.dll` directly via a `ProjectReference Private="false"`.
- **`src/WinRT.Internal/HWND.cs`**: struct counterpart of the IDL `HWND` (custom-mapped to `nint` by the projection writer).
- **`src/WinRT.Internal/ProjectionInternalAttribute.cs`**: forward declaration of the `[ProjectionInternal]` marker the projection writer reads to emit `internal` projections.
- **`src/WinRT.Internal/I*Interop.cs`**: one file per interop interface, all 14 of the IDL ones (`IAccountsSettingsPaneInterop`, `IDragDropManagerInterop`, `IInputPaneInterop`, `IPlayToManagerInterop`, `IPrintManagerInterop`, `IRadialControllerInterop`, `IRadialControllerConfigurationInterop`, `IRadialControllerIndependentInputSourceInterop`, `ISpatialInteractionManagerInterop`, `ISystemMediaTransportControlsInterop`, `IUIViewSettingsInterop`, `IUserConsentVerifierInterop`, `IWebAuthenticationCoreManagerInterop`, `IDisplayInformationStaticsInterop`) with their original IIDs, parameter ordering, and reference links to the corresponding Win32 `*interop.idl` documentation pages.

### WinMD generator improvements

Two small fixes to `WinRT.WinMD.Generator` were needed for the output to be functionally equivalent to the C++ `cswinrt`-generated `.winmd`:

- **`src/WinRT.WinMD.Generator/Writers/WinMDWriter.Members.cs`**: `GetWinRTParameterAttributes` now preserves the input parameter's `[In]` / `[Out]` flags rather than unconditionally emitting `[Out]` for any by-reference parameter. When no flag is set on the input, byref params default to `[In]` (matching MIDL's convention for `ref const T`), keeping the projection writer's `ParameterCategoryResolver` classifying `ref Guid riid` as `ParameterCategory.Ref` rather than `Out`.
- **`src/WinRT.WinMD.Generator/Writers/WinMDWriter.cs`** + **`src/WinRT.WinMD.Generator/References/WinMDValues.cs`**: The authored `.winmd` assembly identity is now always stamped `255.255.255.255` (the WinRT "unbound" convention used by every Windows Runtime metadata assembly). The `--assembly-version` argument continues to drive the type-level `[Windows.Foundation.Metadata.Version]` attribute. Verified against the existing `TestComponent` / `TestComponentCSharp` `.winmd` outputs, both of which carry `v255.255.255.255` regardless of the project's `<AssemblyVersion>`.

### MSBuild integration

- **`src/WinRT.Internal/WinRT.Internal.csproj`** (target `GenerateWindowsRuntimeInternalWinMD`): writes a response file to `$(IntermediateOutputPath)` and invokes `cswinrtwinmdgen.exe` directly via `<Exec>` (instead of a `UsingTask`). This avoids the `MSB3027` file-lock contention that happens in Visual Studio when the persistent MSBuild build server keeps a custom-task assembly loaded across incremental builds — MSBuild never needs to load `WinRT.Generator.Tasks.dll` for this project at all.

### Build infrastructure rewiring

- **`src/Directory.Build.props`**: `$(CsWinRTInteropMetadata)` now resolves to `src/WinRT.Internal/bin/$(Configuration)/net10.0-windows10.0.26100.1/WindowsRuntime.Internal.winmd`. `$(AssemblyVersionNumber)` defaults to `3.0.0.0` so `dotnet build` works locally without explicit `/p:AssemblyVersionNumber=...`. Legacy `$(CsWinRTPath)` / `$(CsWinRTExe)` are kept defined as path strings (the `.exe` no longer needs to exist; they will be removed entirely once `src/cswinrt` is dropped).
- **`src/build.cmd`**: `interop_winmd` now points at the WinRT.Internal output instead of `_build/.../cswinrt/bin/`.
- **`build/AzurePipelineTemplates/CsWinRT-Build-Steps.yml`**: the "Stage WindowsRuntime.Internal.winmd" task copies from `src/WinRT.Internal/bin/.../net10.0-windows10.0.26100.1/` rather than the `cswinrt` bin folder; downstream staging into `$(StagingFolder)/native` is unchanged so the nuget pack step still finds it at the same place.
- **`src/cswinrt.slnx`**: `WinRT.Internal` is included in the default solution-level build.

## Verification

- The new `WindowsRuntime.Internal.winmd` contains exactly the same set of public types (`HWND` + 14 interop interfaces + `ProjectionInternalAttribute`) with identical IIDs and method counts as the legacy C++-generated `WindowsRuntime.Internal.winmd`.
- `dotnet build src/WinRT.Sdk.Projection/...` successfully consumes the new `.winmd` and produces a `WinRT.Sdk.Projection.dll` containing the expected `ABI.WindowsRuntime.Internal.IAccountsSettingsPaneInteropMethods` internal projection alongside the regular `Windows.UI.ApplicationSettings.AccountsSettingsPane` runtime class, confirming the full interop pipeline still works.
- Four back-to-back incremental builds (including ones that force rebuilds of `WinRT.Generator.Tasks.dll` and `cswinrtwinmdgen.exe` themselves while the build session is alive) all complete cleanly with `0` errors — no `MSB3027` file-lock issues.

## Follow-up

The legacy `src/cswinrt` directory is still on disk for now (no consumer depends on it being built). It will be deleted entirely in a follow-up PR.

### DIFF
--- a/src/WinRT.Internal/ProjectionInternalAttribute.cs
+++ b/src/WinRT.Internal/ProjectionInternalAttribute.cs
@@ -18,13 +18,13 @@ namespace WindowsRuntime.Internal;
 /// <para>
 /// This attribute is the C# counterpart of the Modern IDL 3.0 declaration:
 /// </para>
-/// <code>
+/// <code language="csharp">
 /// [attributeusage(target_interface)]
 /// [attributename("ProjectionInternal")]
-/// attribute ProjectionInternalAttribute { };
+/// attribute ProjectionInternalAttribute
+/// {
+/// };
 /// </code>
 /// </remarks>
 [AttributeUsage(AttributeTargets.Interface, Inherited = false, AllowMultiple = false)]
-public sealed class ProjectionInternalAttribute : Attribute
-{
-}
+public sealed class ProjectionInternalAttribute : Attribute;
--- a/src/WinRT.Internal/ProjectionInternalAttribute.cs
+++ b/src/WinRT.Internal/ProjectionInternalAttribute.cs
@@ -1,0 +1,30 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+
+namespace WindowsRuntime.Internal;
+
+/// <summary>
+/// Marks a Windows Runtime interface as having an "internal" projection in CsWinRT.
+/// </summary>
+/// <remarks>
+/// <para>
+/// CsWinRT generates an <see langword="internal"/> projection for any interface marked with this
+/// attribute (rather than the default <see langword="public"/> projection). User-friendly wrappers
+/// over the internal projection are exposed via hand-authored extension methods (see e.g.
+/// <c>ComInteropExtensions</c>).
+/// </para>
+/// <para>
+/// This attribute is the C# counterpart of the Modern IDL 3.0 declaration:
+/// </para>
+/// <code>
+/// [attributeusage(target_interface)]
+/// [attributename("ProjectionInternal")]
+/// attribute ProjectionInternalAttribute { };
+/// </code>
+/// </remarks>
+[AttributeUsage(AttributeTargets.Interface, Inherited = false, AllowMultiple = false)]
+public sealed class ProjectionInternalAttribute : Attribute
+{
+}
--- a/src/WinRT.Internal/ProjectionInternalAttribute.cs
+++ b/src/WinRT.Internal/ProjectionInternalAttribute.cs
@@ -10,21 +10,9 @@ namespace WindowsRuntime.Internal;
 /// </summary>
 /// <remarks>
 /// <para>
-/// CsWinRT generates an <see langword="internal"/> projection for any interface marked with this
-/// attribute (rather than the default <see langword="public"/> projection). User-friendly wrappers
-/// over the internal projection are exposed via hand-authored extension methods (see e.g.
-/// <c>ComInteropExtensions</c>).
-/// </para>
-/// <para>
-/// This attribute is the C# counterpart of the Modern IDL 3.0 declaration:
-/// </para>
-/// <code language="csharp">
-/// [attributeusage(target_interface)]
-/// [attributename("ProjectionInternal")]
-/// attribute ProjectionInternalAttribute
-/// {
-/// };
-/// </code>
+/// CsWinRT generates an <see langword="internal"/> projection for any interface marked with this attribute
+/// (rather than the default <see langword="public"/> projection). User-friendly wrappers over the internal
+/// projection are exposed via hand-authored extension methods (see e.g. <c>ComInteropExtensions</c>).
 /// </remarks>
 [AttributeUsage(AttributeTargets.Interface, Inherited = false, AllowMultiple = false)]
 public sealed class ProjectionInternalAttribute : Attribute;
--- a/src/WinRT.Internal/WinRT.Internal.csproj
+++ b/src/WinRT.Internal/WinRT.Internal.csproj
@@ -1,0 +1,9 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net10.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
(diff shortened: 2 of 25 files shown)


src/WinRT.Internal/ProjectionInternalAttribute.cs
```diff
@@ -0,0 +1,30 @@
...
+/// </summary>
+/// <remarks>
+/// <para>
+/// CsWinRT generates an <see langword="internal"/> projection for any interface marked with this
+/// attribute (rather than the default <see langword="public"/> projection). User-friendly wrappers
+/// over the internal projection are exposed via hand-authored extension methods (see e.g.
+/// <c>ComInteropExtensions</c>).
+/// </para>
+/// <para>
+/// This attribute is the C# counterpart of the Modern IDL 3.0 declaration:
```
manodasanW: remove?

src/WinRT.Internal/WinRT.Internal.csproj
```diff
@@ -0,0 +1,168 @@
...
+  <!--
+    ============================================================
+                        GenerateWindowsRuntimeInternalWinMD
+
+    Invokes 'cswinrtwinmdgen.exe' (built from this solution) to produce
+    'WindowsRuntime.Internal.winmd' from the compiled 'WindowsRuntime.Internal.dll'.
+    The '.winmd' is the actual artifact consumed by the CsWinRT toolchain, and
+    the '.dll' itself is just an intermediate build artifact, which is not used.
+
+    The tool is invoked via '<Exec>' (not via the 'RunCsWinRTWinMDGenerator' MSBuild
```
manodasanW, line 94: I feel like we just need to solve this issue in a different way where we build that dll separately only when needed or something as it is a general issue rather than being specific to this.  But this is fine for now.

Sergio0694: My thinking is that once we publish preview 2 with the changes in this PR to the .winmd generator (I mean technically the new .winmd generator isn't in preview 1 at all), we can update this .csproj to just be fully bootstrapped: it'll just be a normal project targeting the latest public version of CsWinRT 3.0 and produce the .winmd. No targets or project references or custom configuration needed. What do you think?

manodasanW: true, that works.

manodasanW (approved):

PR #2455, opened 2026-06-18
Provide a way to suppress AOT partial warnings for certain interfaces

### LINKED ISSUE #1968: Proposal: Address noisy IDisposable vtable warning
### Summary

When CsWinRTAotWarningLevel is set to 2, warnings are enabled for classes not marked partial that implement custom mapped interfaces given we can't generate vtables for them.  Out of all the custom mapped interfaces, IDisposable seems to be a bit noisy as there are classes implementing it but not passed across the WinRT ABI compared to the other ones.  We should find a way to address this.

### Rationale

To allow consumers to be able to set CsWinRTAotWarningLevel to 2 by default and not be impacted by the noise.

### Important Notes

There are a couple approaches:

- If the class just implements IDisposable, maybe we should just ignore it for the warning if not partial.  We will still generate the vtable if marked partial.
- Have a property to suppress just IDisposable related warnings.
- Have a ItemGroup that custom mapped interfaces can be added to and that indicates if a type just implements one of those interfaces, the warning should be suppressed and vtable should be generated if just that interface.

### Open Questions

_No response_

manodasanW:
Certain types like `IDisposable` can lead to noisy warnings to make a class implementing it partial when you are not passing it across the WinRT ABI boundary.  So, this provides an editor config option `cswinrt_aot_warning_suppressed_interfaces` to specify an interface to suppress this warning for.

Fixes #1968


src/Authoring/WinRT.SourceGenerator/Helper.cs
```diff
@@ -248,6 +248,43 @@ public static int GetCsWinRTAotWarningLevel(this AnalyzerConfigOptionsProvider p
...
+        // Returns the set of fully qualified interface names which should be ignored when determining whether
+        // to report the AOT compatibility warning (CsWinRT1028) for non-partial classes.  When a class only
+        // implements WinRT interfaces present in this set (such as the custom mapped System.IDisposable), the
+        // warning is suppressed.  Classes still get vtables generated for these interfaces when marked partial.
+        // The list is configured via the 'cswinrt_aot_warning_suppressed_interfaces' .editorconfig option as a
+        // comma separated list of fully qualified type names.  Note that ';' cannot be used as a separator since
+        // .editorconfig treats it (and '#') as a comment character.  Because it is read from the .editorconfig
+        // options of the syntax tree the type is declared in, it can be scoped to specific folders or files using
+        // .editorconfig sections.
+        public static ImmutableHashSet<string> GetCsWinRTAotWarningSuppressedInterfaces(this AnalyzerConfigOptions options)
```
Sergio0694, line 266: Nit: should be an XML doc, not a single-line comment

Sergio0694 (approved):

Sergio0694 (approved):

jlaanstra (approved):

PR #2456, opened 2026-06-18
Add end-to-end smoke tests for the CsWinRT NuGet package

Sergio0694:
## Summary

Adds minimal, isolated end-to-end smoke tests that consume the real `Microsoft.Windows.CsWinRT` NuGet package — a consumption app and an authoring component — to verify the package works correctly in a real production environment, independent of the repository's own build infrastructure.

## Motivation

The CsWinRT 3.0 build pipeline is complex (reference projections, forwarder assemblies, several post-build generators, and the dual reference/implementation `WinRT.Runtime` assembly added in this branch). The existing unit and functional tests all build against the *local* build infrastructure (`SimulateCsWinRTNugetReference`, project references to the in-repo projections, etc.), so they never exercise the actual packaged `ref`/`lib` assemblies, the packaged generators, or the NuGet build targets the way a real customer consumes them. These smoke tests close that gap: they reference the packed NuGet package exactly like an external app or component author would, validating that the correct assemblies are referenced and that all generators run correctly end-to-end.

## Changes

The smoke tests live under `src/Tests/SmokeTests/` and are intentionally kept out of `cswinrt.slnx` (the package they consume only exists after the build packs it).

- **`src/Tests/SmokeTests/Directory.Build.props`** / **`Directory.Build.targets`**: blank files that isolate the smoke tests from the repository build infrastructure, so they inherit nothing from `src/Directory.Build.*`.
- **`src/Tests/SmokeTests/Directory.Packages.props`**: disables central package management for the smoke tests (and shadows the repository `Directory.Packages.props`), so each project pins the CsWinRT package version directly.
- **`src/Tests/SmokeTests/.gitignore`**: ignores the default `bin`/`obj` output (these projects use the default output folders rather than the repo `_build` redirect).
- **`src/Tests/SmokeTests/Consumption/`**: an `Exe` that calls `JsonObject.Parse(...)` then `Stringify()` from `Windows.Data.Json`, exercising the Windows SDK projection, the interop generator, and the `WinRT.Runtime` ref/impl assemblies.
- **`src/Tests/SmokeTests/Authoring/`**: a `CsWinRTComponent` library exposing a minimal `Greeter` class, exercising WinMD generation, the reference projection, and the forwarder assembly.
- **`src/Tests/SmokeTests/run-smoke-tests.ps1`**: builds and runs the consumption test (asserting a clean exit code), then builds the authoring test and verifies that the generated `Authoring.winmd` is a valid Windows Runtime metadata file defining `Authoring.Greeter`.
- **`src/build.cmd`**: runs the smoke tests after the local `nuget pack` step (x64 only; skippable via `cswinrt_run_smoke_tests=false`).
- **`build/AzurePipelineTemplates/CsWinRT-PublishToNuGet-Steps.yml`**: runs the smoke tests after the CI `NuGet pack` step, against the freshly packed package.

Both projects override `RestoreSources` to restore the package being tested from the local build output (or the CI pack output) and everything else — notably the preview Windows SDK ref pack — from public NuGet. They use the real .NET SDK targeting pack via a pinned `WindowsSdkPackageVersion` and the `.1` Windows TFM revision, mirroring `src/WinRT.Internal`. The package version and source default to the local `build.cmd x64 Release` output and are overridden by the build or CI that produced the package.

## Validation

Verified end-to-end against a real CsWinRT 3.0 package using the committed configuration: the consumption test builds and runs (round-tripping the JSON to `{"a":42}` with a clean exit), and the authoring component's generated `Authoring.winmd` is confirmed to define `Authoring.Greeter`. In CI the same runner executes against the freshly packed package.


nuget/Microsoft.Windows.CsWinRT.targets
```diff
@@ -319,8 +332,10 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       <_CsWinRTRefInputs Include="@(_CsWinRTRefInteropInputs)" />
     </ItemGroup>
```
manodasanW, line 334: I also had similar changes in this file you will need to merge / resolve.

nuget/Microsoft.Windows.CsWinRT.CsWinRTGen.targets
```diff
@@ -171,10 +192,18 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     Outputs="$(_CsWinRTGeneratorInteropAssemblyPath)"
     Condition="'$(CsWinRTGenerateInteropAssembly2)' == 'true'">
 
+    <!-- Pass the 'WinRT.Runtime' implementation assembly (not the reference assembly) to the generator -->
+    <ItemGroup>
```
manodasanW, line 196: Are we swapping just for WinRT.Runtime or for all?

manodasanW: oh ic, it is adding CsWinRTRuntimeImplementationAssemblyPath later which is just the WinRT.Runtime one.

src/WinRT.Projection.Writer/Resources/Additions/Windows.UI.Xaml.Media.Media3D/Windows.UI.Xaml.Media.Media3D.Matrix3D.cs
```diff
@@ -3,8 +3,8 @@ namespace Windows.UI.Xaml.Media.Media3D
 {
     using global::Windows.Foundation;
 
-    [WindowsRuntimeMetadata("Windows.Foundation.UniversalApiContract")]
 #if !CSWINRT_REFERENCE_PROJECTION
+    [WindowsRuntimeMetadata("Windows.Foundation.UniversalApiContract")]
```
manodasanW, line 7: Is the moving of `WindowsRuntimeMetadata` outside of the reference projection to make this only an impl assembly thing?  This attribute is also used for winmd authoring scenarios as we need to make sure we are emitting the reference to the right winmd as we can't tell just from the reference projection.

Sergio0694: Yeah this is needed to make it an impl-detail. This PR just makes it an impl-only type, and then https://github.com/microsoft/CsWinRT/pull/2467 will optimize this to use a lookup and avoid having all types there on the actual types. All tests seem to be passing, including the authoring smoke tests, so we should be fine? I do remember updating the logic to compute the assembly names to account for this in both PRs.

manodasanW (approved):

PR #2467, opened 2026-06-22
Add [WindowsRuntimeType] marker and centralize WinRT metadata mapping

Sergio0694:
## Summary

Splits the dual-purpose `[WindowsRuntimeMetadata("contract")]` attribute into a parameterless `[WindowsRuntimeType]` marker placed on every projected/proxy type, plus a single centralized, fully-trimmable `ABI.WindowsRuntimeMetadataTypes` lookup type that carries the type -> source `.winmd` mapping (one `[WindowsRuntimeMetadata(typeof(T), "stem")]` entry per type). This mirrors how `ABI.WindowsRuntimeDefaultInterfaces` already centralizes default-interface metadata.

## Motivation

The old `[WindowsRuntimeMetadata]` attribute did two unrelated jobs at once: it acted as the "this is a Windows Runtime type" marker, and it also carried the source `.winmd` module name (the "contract"/"stem") as a string on every single projected type. That string is **never** read at runtime — the runtime only ever checks for the *presence* of the attribute. The stem is needed only by build-time tooling (the interop and WinMD generators), and now that reference projections exist, those generators already have to recover that information through a lookup against the implementation projection.

Given that, carrying a per-type string everywhere is pure overhead. Moving to a parameterless marker removes a string from the metadata of every projected type, and relocating the type -> stem mapping onto one dedicated lookup type means the entire mapping is dead-code-eliminated by ILLink/ILC whenever it isn't needed. The net result is smaller metadata with no loss of functionality.

## Changes

- **`src/WinRT.Runtime2/Attributes/`**: add the new parameterless, implementation-only `WindowsRuntimeTypeAttribute` marker; repurpose `WindowsRuntimeMetadataAttribute` to a `(Type, string)`, `class`-targeting, `AllowMultiple` attribute used only on the centralized lookup type; delete the now-redundant `WindowsRuntimeMappedMetadataAttribute`.
- **`src/WinRT.Runtime2/` (manually-projected types + `ABI/` proxy types)**: replace the per-type `[WindowsRuntimeMetadata("...")]` / `[WindowsRuntimeMappedMetadata("...")]` attributes with the parameterless `[WindowsRuntimeType]` marker across all manually-projected types and custom-mapped proxy types.
- **`src/WinRT.Runtime2/InteropServices/TypeMapInfo/`**: update the runtime lookups (`WindowsRuntimeMetadataInfo`, `WindowsRuntimeMarshallingInfo`) to recognize Windows Runtime types via `[WindowsRuntimeType]`, and to distinguish projected vs. proxy types via `[WindowsRuntimeMappedType]`.
- **`src/WinRT.Projection.Writer/`**: emit `[WindowsRuntimeType]` per projected type, and emit the centralized `ABI.WindowsRuntimeMetadataTypes` lookup type (implementation projections only) carrying the `[WindowsRuntimeMetadata(typeof(T), "stem")]` entries, wired through the generator run state, emit context, and finalization step (mirroring the default-interfaces flow).
- **`src/WinRT.Projection.Writer/Resources/Additions/` (XAML structs)**: update the 12 hand-authored full-replacement XAML struct additions (`CornerRadius`, `Duration`, `GridLength`, `KeyTime`, `RepeatBehavior`, `Matrix3D` for both WinUI and UWP XAML) to use `[WindowsRuntimeType]`, placed inside the existing `#if !CSWINRT_REFERENCE_PROJECTION` block alongside the other implementation-only attributes.
- **`src/WinRT.Interop.Generator/`**: recognize projected types via `[WindowsRuntimeType]`; read the source `.winmd` stem from the centralized `ABI.WindowsRuntimeMetadataTypes` lookup on the resolved implementation projection (new `GetWindowsRuntimeMetadataTypesLookup` reader); emit the parameterless marker on generated proxy types.
- **`src/WinRT.WinMD.Generator/`**: recognize Windows Runtime types via `[WindowsRuntimeType]` and recover the assembly/contract name from the centralized lookup type instead of from a per-type attribute.
- **`.github/`**: update `copilot-instructions.md` and the interop generator skill (`SKILL.md` + `references/name-mangling-scheme.md`) to document the new marker and the centralized metadata-types lookup.

## Validation

Built every affected project, and validated end-to-end locally by running the projection writer over the Windows SDK union metadata: the full generated Windows SDK projection (including all XAML namespaces and the new lookup type) compiles cleanly against the new runtime. An AsmResolver probe confirmed the consumption-side lookup correctly reads the emitted entries.


src/WinRT.WinMD.Generator/Extensions/TypeDefinitionExtensions.cs
```diff
@@ -69,33 +70,26 @@ public int? VersionAttributeValue
...
-                }
-
-                // Extract the assembly name from the attribute signature, if possible
-                if (attribute.Signature is { FixedArguments: [{ Element: object assemblyName }] })
-                {
-                    return assemblyName.ToString();
-                }
-
-                return null;
+                return type.DeclaringModule is { } module &&
```
manodasanW, line 89: This will be replaced with my changes.

manodasanW (approved):

PR #2469, opened 2026-06-23
Various fixes from WinUI testing

manodasanW:
- CsWinRTProjectionRefGen seems to be always running even if there is no projection to generate, for now scoping it running to when there are includes
- When there were CsWinRTExcludes but no includes, everything in CsWinRTInputs not excluded was being projected which isn't the behavior from before.  Types only in CsWinRTIncludes should be what is projected in that case.  Updating the TypeFilter logic in general to match what we had before.
- NonBlittable types passed in constructors were ending up failing to compile during merged projection generation as discovered by some Win2D types.  Addressing that and adding tests.

### DIFF
--- a/nuget/Microsoft.Windows.CsWinRT.targets
+++ b/nuget/Microsoft.Windows.CsWinRT.targets
@@ -333,7 +333,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     <Message Condition="'$(_CsWinRTRunProjectionRefGenerator)' != 'true'"
              Text="cswinrtprojectionrefgen: skipped (no namespaces to project)" Importance="Low" />
     <Message Condition="'$(_CsWinRTRunProjectionRefGenerator)' == 'true'"
-             Text="cswinrtprojectionrefgen: @(_CsWinRTRefInputs->Count()) input(s), @(_CsWinRTRefIncludes->Count()) include(s), @(_CsWinRTRefExcludes->Count()) exclude(s)" Importance="$(CsWinRTMessageImportance)" />
+             Text="cswinrtprojectionrefgen: @(_CsWinRTRefInputs->Count()) input(s), @(_CsWinRTRefIncludes->Count()) include(s), @(_CsWinRTRefExcludes->Count()) exclude(s)" Importance="Low" />
     <RunCsWinRTProjectionRefGenerator
         Condition="'$(_CsWinRTRunProjectionRefGenerator)' == 'true'"
         InputWinMDPaths="@(_CsWinRTRefInputs)"
--- a/nuget/Microsoft.Windows.CsWinRT.targets
+++ b/nuget/Microsoft.Windows.CsWinRT.targets
@@ -332,8 +332,6 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
     <Message Condition="'$(_CsWinRTRunProjectionRefGen)' != 'true'"
              Text="cswinrtprojectionrefgen: skipped (no namespaces to project)" Importance="$(CsWinRTMessageImportance)" />
-    <Message Condition="'$(_CsWinRTRunProjectionRefGen)' == 'true'"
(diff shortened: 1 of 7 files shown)


nuget/Microsoft.Windows.CsWinRT.targets
```diff
@@ -319,8 +319,21 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
 
-    <Message Text="cswinrtprojectionrefgen: @(_CsWinRTRefInputs->Count()) input(s), @(_CsWinRTRefIncludes->Count()) include(s), @(_CsWinRTRefExcludes->Count()) exclude(s)" Importance="$(CsWinRTMessageImportance)" />
+    <!--
+      Only run 'cswinrtprojectionrefgen' if there is actually something to project. With no
+      includes, the namespace filter excludes everything, so the generator emits no projection
+      sources (and no base files either, since those are only written alongside a projection).
+      In that case, running the tool is a pure no-op, so we skip launching it entirely.
+    -->
+    <PropertyGroup>
+      <_CsWinRTRunProjectionRefGen>false</_CsWinRTRunProjectionRefGen>
```
Sergio0694: Nit (also update below)
```suggestion
      <_CsWinRTRunProjectionRefGenerator>false</_CsWinRTRunProjectionRefGenerator>
```

nuget/Microsoft.Windows.CsWinRT.targets
```diff
@@ -319,8 +319,21 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+      <_CsWinRTRunProjectionRefGen Condition="'@(_CsWinRTRefIncludes)' != ''">true</_CsWinRTRunProjectionRefGen>
+    </PropertyGroup>
+
+    <Message Condition="'$(_CsWinRTRunProjectionRefGen)' != 'true'"
```
Sergio0694: Don't think we need to show this in the build output, it's just noise. Can we make this verbose?

Sergio0694 (approved):

nuget/Microsoft.Windows.CsWinRT.targets
manodasanW: yea that makes sense.

nuget/Microsoft.Windows.CsWinRT.targets
```diff
@@ -319,8 +319,23 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+    -->
+    <PropertyGroup>
+      <_CsWinRTRunProjectionRefGenerator>false</_CsWinRTRunProjectionRefGenerator>
+      <_CsWinRTRunProjectionRefGenerator Condition="'@(_CsWinRTRefIncludes)' != ''">true</_CsWinRTRunProjectionRefGenerator>
+    </PropertyGroup>
+
+    <Message Condition="'$(_CsWinRTRunProjectionRefGenerator)' != 'true'"
+             Text="cswinrtprojectionrefgen: skipped (no namespaces to project)" Importance="Low" />
+    <Message Condition="'$(_CsWinRTRunProjectionRefGenerator)' == 'true'"
+             Text="cswinrtprojectionrefgen: @(_CsWinRTRefInputs->Count()) input(s), @(_CsWinRTRefIncludes->Count()) include(s), @(_CsWinRTRefExcludes->Count()) exclude(s)" Importance="$(CsWinRTMessageImportance)" />
```
Sergio0694: I don't think we need this? `CsWinRTMessageImportance` defaults to `normal`, which would make this always visible, but the generator itself already always emits a message saying it's running and on how many .winmd-s.

manodasanW: I added this mostly because it was there before and I accidentally deleted it.  But I feel like it can be useful for diagnosing issues in case the generator doesn't run as expected.

Sergio0694: Can we always make it diagnostic at least? It shouldn't be visible by default.

manodasanW: sure

Sergio0694 (approved):

PR #2470, opened 2026-06-23
Stop relying on UB in 'IContextCallback' dispatch logic

Sergio0694:
## Summary

Port the cross-thread access fix from #1865 to the CsWinRT 3.0 runtime (`WinRT.Runtime2`). The context callback dispatch logic no longer stores the managed callback state in a stack local that is then dereferenced from the target thread, which is undefined behavior under the .NET memory model.

## Motivation

When dispatching a callback onto another context via `IContextCallback`, the previous code stored the user-provided state (a managed object) in a local variable on the *original* thread's stack and passed a pointer to it through `ComCallData`. The native `InvokeCallback` stub then dereferenced that pointer while running on the *target* thread. Reading a managed reference that lives on another thread's stack is explicitly undefined behavior, see the [.NET memory model spec on cross-thread access to local variables](https://github.com/dotnet/runtime/blob/main/docs/design/specs/Memory-model.md#cross-thread-access-to-local-variables). This is the same class of issue fixed for CsWinRT 2.x in #1865; this PR brings the fix to the 3.0 runtime.

## Changes

- **`src/WinRT.Runtime2/InteropServices/ObjectReference/ContextCallback.cs`**: rework `CallInContextUnsafe` to stop passing a stack-local managed reference across threads.
  - Add a `[ThreadStatic]` `LocalContextCallbackState` field so the state lives on the managed heap rather than on the calling thread's stack.
  - Write the state into that field, pin it, and pass a pointer (`CallbackData.StatePtr`) to the target context, with a `Thread.MemoryBarrier()` before the cross-context call and a `Volatile.Write(ref LocalContextCallbackState, null)` reset afterwards to avoid extending the state's lifetime.
  - Add a throwaway-array fallback for the recursion case (never observed in the test suite or in Store stress runs), so the path stays correct even if the thread-static slot is already occupied.
  - Change `CallbackData.State` (an `object`) to `CallbackData.StatePtr` (an `object*`).

### DIFF
--- a/src/WinRT.Runtime2/InteropServices/ObjectReference/ContextCallback.cs
+++ b/src/WinRT.Runtime2/InteropServices/ObjectReference/ContextCallback.cs
@@ -105,10 +105,6 @@ internal static unsafe class ContextCallback
             comCallData.dwReserved = 0;
             comCallData.pUserDefined = &callbackData;
 
-            // Add a memory barrier to be extra safe that the target thread will be able to see
-            // the write we just did on 'LocalContextCallbackState' with the state to pass to the callback.
-            Thread.MemoryBarrier();
-
             // Marshal the supplied callback on the target context
             hresult = IContextCallbackVftbl.ContextCallbackUnsafe(
                 thisPtr: contextCallbackPtr,
--- a/src/WinRT.Runtime2/InteropServices/ObjectReference/ContextCallback.cs
+++ b/src/WinRT.Runtime2/InteropServices/ObjectReference/ContextCallback.cs
@@ -68,10 +68,14 @@ internal static unsafe class ContextCallback
         ref object? localContextCallbackState = ref LocalContextCallbackState;
 
         // Store the state object in the thread static to pass to the callback.
-        // We don't need a volatile write here, we have a memory barrier below.
         // A thread local is the most efficient solution for this, given that
         // we need the state to be somewhere on the managed heap to be valid.
         // The GC doesn't allow cross-thread access to managed stack variables.
+        // We're not using a volatile write, as it wouldn't actually help at all.
+        // Volatile writes disallow the write operations to be moved before memory
+        // operations that precede them, but they have nothing to say with respect
+        // to memory operations after them (whereas here we specifically need this
+        // write to not be reordered before following reads from that static field).
         if (localContextCallbackState is null)
         {
             localContextCallbackState = state;
--- a/src/WinRT.Runtime2/InteropServices/ObjectReference/ContextCallback.cs
+++ b/src/WinRT.Runtime2/InteropServices/ObjectReference/ContextCallback.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Runtime.InteropServices;
-using System.Threading;
 
 namespace WindowsRuntime.InteropServices;
 
@@ -116,7 +115,7 @@ internal static unsafe class ContextCallback
         }
 
         // Reset the static field to avoid keeping the state alive for longer
-        Volatile.Write(ref LocalContextCallbackState, null);
+        LocalContextCallbackState = null;
 
         return hresult;
     }
--- a/src/WinRT.Runtime2/InteropServices/ObjectReference/ContextCallback.cs
+++ b/src/WinRT.Runtime2/InteropServices/ObjectReference/ContextCallback.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Runtime.InteropServices;
+using System.Threading;
 
 namespace WindowsRuntime.InteropServices;
 
@@ -11,6 +12,12 @@ namespace WindowsRuntime.InteropServices;
 /// </summary>
 internal static unsafe class ContextCallback
 {
+    /// <summary>
+    /// Storage for the user-provided state for <see cref="CallInContextUnsafe"/>.
+    /// </summary>
+    [ThreadStatic]
+    private static object? LocalContextCallbackState;
+
     /// <summary>
     /// Calls the given callback in the right context, and returns the result of that invocation.
     /// </summary>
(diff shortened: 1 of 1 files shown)


Sergio0694:
@hamarb123 @jkotas I've ported the updated changes to the 3.0 branch, could you take a look when you have time to confirm whether this is the best solution for this and whether the change makes sense? I can also backport to 2.x once we confirm the changes are correct. Thank you! 🙂

src/WinRT.Runtime2/InteropServices/ObjectReference/ContextCallback.cs
```diff
@@ -70,10 +66,49 @@ static int InvokeCallback(ComCallData* comCallData)
...
+        {
+            // In case we recursed on this thread, meaning the local storage already holds a state
+            // for some other caller above us in the stack, we can use a throwaway array to store
+            // the current state. This isn't very efficient, but in practice this case should not
+            // ever happen (it was validated in our entire test suite as well as in stress tests
+            // with the Microsoft Store app, and we never detected recursion on this code path).
+            // However just to be extra safe, we still want to keep this branch functional too.
+            object[] objects = [state];
+
+            localContextCallbackState = ref MemoryMarshal.GetArrayDataReference(objects)!;
```
hamarb123, line 93: I'm pretty sure the JIT can elide the bounds checks here. Or at least, I'd be surprised if it couldn't.

src/WinRT.Runtime2/InteropServices/ObjectReference/ContextCallback.cs
```diff
@@ -70,10 +66,49 @@ static int InvokeCallback(ComCallData* comCallData)
             }
         }
 
+        ref object? localContextCallbackState = ref LocalContextCallbackState;
+
+        // Store the state object in the thread static to pass to the callback.
+        // We don't need a volatile write here, we have a memory barrier below.
```
hamarb123: Assuming that the memory barrier is necessary, a volatile write won't help here as it doesn't disallow the kind of reordering that we want to disallow. Maybe update the comment to reflect this (that volatile write here doesn't actually help at all vs normal write here ito semantics).

In particular volatile write disallows the write to move before memory operations before it, but has nothing to say wrt memory operations after it ito reordering, so it wouldn't help.

hamarb123 (commented):
Lgtm, other than these 2 comments.

src/WinRT.Runtime2/InteropServices/ObjectReference/ContextCallback.cs
hamarb123: Nevermind, it does not seem to produce as good codegen - feels like something the runtime could definitely improve if it knows the exact type & length of the array: https://godbolt.org/z/qvboTn55a

src/WinRT.Runtime2/InteropServices/ObjectReference/ContextCallback.cs
```diff
@@ -70,10 +66,49 @@ static int InvokeCallback(ComCallData* comCallData)
...
+            callbackData.StatePtr = statePtr;
+
+            ComCallData comCallData;
+            comCallData.dwDispid = 0;
+            comCallData.dwReserved = 0;
+            comCallData.pUserDefined = &callbackData;
+
+            // Add a memory barrier to be extra safe that the target thread will be able to see
+            // the write we just did on 'LocalContextCallbackState' with the state to pass to the callback.
+            Thread.MemoryBarrier();
```
jkotas: This should not be needed. The underlying OS operation that is switching the thread and passing the data to it is responsible for issuing the memory barrier.

src/WinRT.Runtime2/InteropServices/ObjectReference/ContextCallback.cs
```diff
@@ -11,6 +12,12 @@ namespace WindowsRuntime.InteropServices;
 /// </summary>
 internal static unsafe class ContextCallback
 {
+    /// <summary>
+    /// Storage for the user-provided state for <see cref="CallInContextUnsafe"/>.
+    /// </summary>
+    [ThreadStatic]
+    private static object? LocalContextCallbackState;
```
jkotas, line 18: Does the name of this thread static match this repo naming convention?

src/WinRT.Runtime2/InteropServices/ObjectReference/ContextCallback.cs
```diff
@@ -84,6 +119,9 @@ static int InvokeCallback(ComCallData* comCallData)
                 pUnk: null);
         }
 
+        // Reset the static field to avoid keeping the state alive for longer
+        Volatile.Write(ref LocalContextCallbackState, null);
```
jkotas: This does not need to be volatile.

src/WinRT.Runtime2/InteropServices/ObjectReference/ContextCallback.cs
hamarb123: Agreed - I checked & I mentioned to sergio on the c# discord during discussing the last PR that it didn't look like it'd probably be necessary to me (for both the memory barrier & the volatile). I think it should be fine without these.

src/WinRT.Runtime2/InteropServices/ObjectReference/ContextCallback.cs
Sergio0694: Updated!

src/WinRT.Runtime2/InteropServices/ObjectReference/ContextCallback.cs
Sergio0694: We're using CamelCase for all static fields in the repo, including private ones.
I'll just keep this as is for now, we can always tweak it later if we do a formatting pass.

src/WinRT.Runtime2/InteropServices/ObjectReference/ContextCallback.cs
```diff
@@ -84,6 +118,9 @@ static int InvokeCallback(ComCallData* comCallData)
                 pUnk: null);
         }
 
+        // Reset the static field to avoid keeping the state alive for longer
+        LocalContextCallbackState = null;
```
Sergio0694, line 122: @hamarb123 @jkotas triple-checking on this, the idea here is that even if we're in the recursive case, we can always unconditionally reset the TLS field here to allow the GC to collect stuff as early as possible, because even if we do have other callers higher in the stack, and this was a nested recursive call, those callers higher up would've already read their context value from the TLS by this point (as each callback does so immediately), so resetting the field wouldn't affect them either way. Does this reasoning seem correct? Thank you! 🙂

manodasanW (approved):

src/WinRT.Runtime2/InteropServices/ObjectReference/ContextCallback.cs
Sergio0694: (clarified offline, that TLS field is only set in the non-recursive case, so this is completely fine)

PR #2474, opened 2026-06-23
Fix constructor marshalling for non-blittable array element types

manodasanW:
Activation-factory constructors taking arrays of non-blittable structs, mapped value types (DateTime/TimeSpan), or HResult/Exception elements hardcoded nint InlineArray storage and void** CopyToUnmanaged/Dispose data pointers, so the array marshaller's UnsafeAccessor failed to bind at runtime (Method not found: ...ArrayMarshaller.Dispose).

Route the constructor's PassArray storage, CopyToUnmanaged, and finally cleanup through the element-kind-aware helpers (GetArrayElementStorageType and typed data-pointer/cast selection), mirroring the RCW caller. Add a TestComponentCSharp constructor and TestNonBlittableArrayConstructor test covering all three element kinds.

### DIFF
--- a/src/WinRT.Projection.Writer/Factories/ConstructorFactory.FactoryCallbacks.cs
+++ b/src/WinRT.Projection.Writer/Factories/ConstructorFactory.FactoryCallbacks.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using AsmResolver.DotNet;
 using AsmResolver.DotNet.Signatures;
@@ -65,6 +66,8 @@ internal static partial class ConstructorFactory
     /// <c>out void* innerInterface</c> params. Iteration over user params is bounded by
     /// <paramref name="userParamCount"/> (defaults to all params).</param>
     /// <param name="userParamCount">If &gt;= 0, only emit the first <paramref name="userParamCount"/> user params (used for composable factories).</param>
+    [SuppressMessage("Style", "IDE0045:Convert to conditional expression",
+        Justification = "if/else if chains over type-class predicates are more readable than nested ternaries.")]
     private static void EmitFactoryCallbackClass(IndentedTextWriter writer, ProjectionEmitContext context, MethodSignatureInfo sig, string callbackName, string argsName, string factoryObjRefName, int factoryMethodIndex, bool isComposable = false, int userParamCount = -1)
     {
         int paramCount = userParamCount >= 0 ? userParamCount : sig.Parameters.Count;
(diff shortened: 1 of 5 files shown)


src/WinRT.Projection.Writer/Factories/ConstructorFactory.FactoryCallbacks.cs
```diff
@@ -677,9 +705,28 @@ private sealed class {{callbackName}} : {{baseClass}}
...
                 ArrayTempNames names = new(raw);
 
+                if (szArr.BaseType.IsHResultException())
+                {
+                    // HResult ABI is just an int; no per-element Dispose, only the pool return.
+                    writer.WriteLine();
+                    writer.WriteLine(isMultiline: true, $$"""
+                        if ({{names.ArrayFromPool}} is not null)
+                        {
+                            global::System.Buffers.ArrayPool<global::ABI.System.Exception>.Shared.Return({{names.ArrayFromPool}});
```
Sergio0694, line 720: We should probably clear the array before returning it to avoid rooting the marshalled exceptions.

src/WinRT.Projection.Writer/Factories/ConstructorFactory.FactoryCallbacks.cs
```diff
@@ -505,14 +506,41 @@ private sealed class {{callbackName}} : {{baseClass}}
...
             {
                 IndentedTextWriterCallback elementProjected = TypedefNameWriter.WriteProjectionType(context, TypeSemanticsFactory.Get(szArr.BaseType));
+
+                // Data pointer type must match the array marshaller's CopyToUnmanaged signature.
+                string dataParamType;
+                string dataCastType;
+
+                if (context.AbiTypeKindResolver.IsMappedAbiValueType(szArr.BaseType))
+                {
+                    dataParamType = AbiTypeHelpers.GetMappedAbiTypeName(szArr.BaseType) + "*";
```
Sergio0694, line 517: Nit: put `GetMappedAbiTypeName` in a local and reuse it rather than calling it twice?

src/WinRT.Projection.Writer/Factories/ConstructorFactory.FactoryCallbacks.cs
```diff
@@ -505,14 +506,41 @@ private sealed class {{callbackName}} : {{baseClass}}
             else
             {
                 IndentedTextWriterCallback elementProjected = TypedefNameWriter.WriteProjectionType(context, TypeSemanticsFactory.Get(szArr.BaseType));
+
+                // Data pointer type must match the array marshaller's CopyToUnmanaged signature.
+                string dataParamType;
+                string dataCastType;
```
Sergio0694: More of a side point but I don't get why `dataCastType` is needed. Seems to me like we could just use `dataParamType` and then put it within parentheses in the template when we need the cast.

src/WinRT.Projection.Writer/Factories/ConstructorFactory.FactoryCallbacks.cs
manodasanW: This is the ABI representation of the exceptions right which is HResult?

Sergio0694: My bad I misread, ignore this 😄

src/WinRT.Projection.Writer/Factories/ConstructorFactory.FactoryCallbacks.cs
manodasanW: Yea not needed, addressed.

Sergio0694 (approved):

PR #2476, opened 2026-06-28
Fix ActivateInstance parameter

### LINKED ISSUE #2477: CsWinRT 3.0 crashes, assumes ActivateInstance instance implements default interface
Originally filed at https://github.com/microsoft/windows-rs/issues/4660 by @dongle-the-gadget and @ahmed605

(Feel free to file a better issue and I'll close this one. Just getting the ball rolling.)

manodasanW:
Fix ActivateInstance parameter to reflect that it is not guaranteed to be the default interface but rather could just be IInspectable

### DIFF
--- a/src/WinRT.Runtime2/WindowsRuntimeObject.cs
+++ b/src/WinRT.Runtime2/WindowsRuntimeObject.cs
@@ -88,24 +88,10 @@ public abstract unsafe class WindowsRuntimeObject :
 
         // This constructor is only meant to be used for sealed types, so there's never a non-delegating (inner) return value.
         // See additional notes in the overload below for more details about how and when that parameter is necessary.
-        // The activation factory returns the 'IInspectable' interface for the activated instance, so we need to do a
-        // 'QueryInterface' for the default interface.
         WindowsRuntimeActivationHelper.ActivateInstanceUnsafe(
             activationFactoryObjectReference: activationFactoryObjectReference,
-            inspectableInterface: out void* inspectableInterface);
-
-        void* defaultInterface;
-
-        try
-        {
-            IUnknownVftbl.QueryInterfaceUnsafe(inspectableInterface, in iid, out defaultInterface).Assert();
-        }
-        finally
-        {
-#pragma warning disable IDE0058 // Return value is discarded, but can't use _ given we used it for the parameter.
-            IUnknownVftbl.ReleaseUnsafe(inspectableInterface);
-#pragma warning restore IDE0058
-        }
+            iid: in iid,
+            defaultInterface: out void* defaultInterface);
 
         // The inner interface pointer isn't used for non-composable types, so we just pass 'null'
         void* innerInterface = null;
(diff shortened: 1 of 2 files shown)


riverar:
Can we get more background (if any) on this and potentially edit the docs https://learn.microsoft.com/windows/win32/api/activation/nf-activation-iactivationfactory-activateinstance

riverar:
@oldnewthing https://devblogs.microsoft.com/oldnewthing/20201028-00/?p=104404 and https://devblogs.microsoft.com/oldnewthing/20200324-00/?p=103586 make claims about object representation "at the ABI" that could use some clarification. (These articles came up in a discussion about this issue.) My suspicion is that you were describing ABI behavior from a C++/WinRT-authored-component perspective, rather than stating overarching rules for all WinRT objects right?

src/WinRT.Runtime2/WindowsRuntimeObject.cs
```diff
@@ -88,9 +88,24 @@ protected WindowsRuntimeObject(
...
         WindowsRuntimeActivationHelper.ActivateInstanceUnsafe(
             activationFactoryObjectReference: activationFactoryObjectReference,
-            defaultInterface: out void* defaultInterface);
+            inspectableInterface: out void* inspectableInterface);
+
+        void* defaultInterface;
+
+        try
+        {
+            IUnknownVftbl.QueryInterfaceUnsafe(inspectableInterface, in iid, out defaultInterface).Assert();
```
Sergio0694: I really dislike paying for this when in practice it's only possibly needed for very esoteric custom WinRT components that do weird things. Can we at least scope this so we only pay this overhead for non-SDK types? We know that all types from the Windows SDK always just return the default interface here.

riverar: How will you determine that without eating into savings?

Sergio0694: Statically, not at runtime.

riverar: > We know that all types from the Windows SDK always just return the default interface here

How do you know that?

riverar: `Windows.Data.Xml.Dom.XmlDocument` activation appears to return a pointer that does not match the QI result for its default interface, heads up.

```csharp

[DllImport("api-ms-win-core-winrt-string-l1-1-0.dll", ExactSpelling = true)]
static extern int WindowsCreateString([MarshalAs(UnmanagedType.LPWStr)] string src, int length, out IntPtr hstring);

[DllImport("api-ms-win-core-winrt-l1-1-0.dll", ExactSpelling = true)]
static extern int RoActivateInstance(IntPtr hstring, out IntPtr instance);

static unsafe int QI(IntPtr p, ref Guid iid, out IntPtr r)
    => ((delegate* unmanaged[Stdcall]<IntPtr, ref Guid, out IntPtr, int>)(*(void***)p)[0])(p, ref iid, out r);

WindowsCreateString("Windows.Data.Xml.Dom.XmlDocument", 32, out nint acid);
RoActivateInstance(acid, out nint inspectable_ptr);
var default_intf_iid = new Guid("F7F3A506-1E87-42D6-BCFB-B8C809FA5494"); // IXmlDocument
Marshal.ThrowExceptionForHR(QI(inspectable_ptr, ref default_intf_iid, out nint default_intf_ptr));
Console.WriteLine(default_intf_ptr == inspectable_ptr ? "Same pointer" : "[!] Different pointer");
```

Sergio0694: Unfortunate. We'll need metadata support for this, I'll work on a spec to bring to API review.

Sergio0694 (approved):

PR #2481, opened 2026-07-03
Add opt-in CCW marshalling for projected Windows Runtime types

Sergio0694:
## Summary

Introduce a new `WindowsRuntimeNativeExposedTypeAttribute` that lets a projected Windows Runtime class be explicitly opted into CCW (COM Callable Wrapper) marshalling code generation, which the interop generator normally skips for projected types. The PR adds interop generator support, three analyzer diagnostics, unit and functional (JIT and Native AOT) test coverage, and improved marshalling failure diagnostics.

## Motivation

The interop generator automatically generates CCW marshalling code for all managed types that implement one or more Windows Runtime interfaces. Projected Windows Runtime types are deliberately skipped, because they are backed by native objects and are marshalled by unwrapping the underlying native object directly, so they never normally require a CCW. There are, however, niche scenarios where a CCW is required for a projected type, and there was previously no way to request one.

The motivating scenario is using a projected collection type as the items source for a XAML control. Consider a `DependencyObjectCollection` assigned as the items source of an `ItemsControl`. The control queries the assigned object (marshalled as an `IInspectable`) for `IBindableIterable` and `IIterable<IInspectable>`. This succeeds for most collection types, but `DependencyObjectCollection` implements neither of those interfaces: it implements `IIterable<DependencyObject>`, and Windows Runtime generic interfaces are not covariant on the native side (each generic instantiation has a completely different IID). To handle this, XAML controls fall back to obtaining a CCW from the source object and querying that instead. The .NET runtime creates an RCW for the native object and uses `ComWrappers` to obtain a CCW proxy over it; because the RCW implements `IEnumerable<DependencyObject>`, which is covariant in C#, the query for `IIterable<IInspectable>` then succeeds through the proxy. This requires a CCW vtable to exist for the projected type, which the interop generator would not normally generate. The new attribute makes it possible to request exactly that.

## Changes

- **`src/WinRT.Runtime2/InteropServices/Attributes/WindowsRuntimeNativeExposedTypeAttribute.cs`**: new public, assembly-level attribute (`AllowMultiple`) that opts a projected Windows Runtime class into CCW marshalling code generation, with detailed documentation of the motivating scenario.
- **`src/WinRT.Interop.Generator/`** (`Discovery/InteropTypeDiscovery.cs`, `Generation/InteropGenerator.Discover.cs`, `References/InteropReferences.cs`, `Errors/WellKnownInteropExceptions.cs`): discover `[assembly: WindowsRuntimeNativeExposedType(typeof(X))]` applications and route the referenced projected types through the user-defined-type CCW pipeline, bypassing the check that normally skips projected types while keeping all other discovery filters.
- **`src/WinRT.Interop.Generator/References/WellKnownInterfaceIIDs.cs`**: add the missing `IStringable` case to `get_IID`. Every projected class implements the manually-projected `IStringable`, so resolving its interface entry is required when CCW code is generated for a projected type; this path was previously unreachable and threw `CSWINRTINTEROPGEN0052`.
- **`src/Authoring/WinRT.SourceGenerator2/Diagnostics/`** (`Analyzers/WindowsRuntimeNativeExposedTypeAnalyzer.cs`, `DiagnosticDescriptors.cs`, `AnalyzerReleases.Shipped.md`): new analyzer emitting `CSWINRT2018` (target type cannot be instantiated), `CSWINRT2019` (target type is not a projected class, so CCW code is already generated automatically), and `CSWINRT2020` (the same type is targeted by more than one application in the assembly). Applications in generated code are excluded, so a duplicate spanning user code and generated code only warns on the user application.
- **Tests** (`src/Tests/SourceGenerator2Test/Test_WindowsRuntimeNativeExposedTypeAnalyzer.cs`, `src/Tests/SourceGenerator2Test/Helpers/CSharpAnalyzerTest{TAnalyzer}.cs`, `src/Tests/UnitTest/NativeExposedTypeTests.cs`, `src/Tests/FunctionalTests/NativeExposedType/`, `src/cswinrt.slnx`, `src/build.cmd`): analyzer unit tests (including a test-harness helper for injecting generated code to validate the generated-code exclusion), plus a self-contained `NativeExposedType` functional test that runs in both JIT and Native AOT and a unit test. Both verify that the interop generator registers a proxy type map association for an opted-in projected type and not for a projected type that was not opted in.
- **`src/WinRT.Runtime2/InteropServices/Attributes/WindowsRuntimeComWrappersMarshallerAttribute.cs`** and **`src/WinRT.Runtime2/InteropServices/TypeMapInfo/WindowsRuntimeMarshallingInfo.cs`**: improve marshalling failure diagnostics. The `NotSupportedException` messages now include the concrete marshaller type, and `ComputeVtables` returns `null` when not overridden so `WindowsRuntimeMarshallingInfo` can throw a detailed exception that names the marshaller and metadata provider types and points to `WindowsRuntimeNativeExposedTypeAttribute` as the remediation.

### DIFF
--- a/src/cswinrt.slnx
+++ b/src/cswinrt.slnx
@@ -347,6 +347,14 @@
       <Platform Solution="*|x86" Project="x86" />
       <Build Solution="*|ARM64" Project="false" />
     </Project>
+    <Project Path="Tests/FunctionalTests/NativeExposedType/NativeExposedType.csproj">
+      <Platform Solution="*|ARM" Project="ARM" />
+      <Platform Solution="*|ARM64" Project="ARM64" />
+      <Platform Solution="*|x64" Project="x64" />
+      <Platform Solution="*|x86" Project="x86" />
+      <Build Solution="*|ARM" Project="false" />
+      <Build Solution="*|ARM64" Project="false" />
+    </Project>
     <Project Path="Tests/FunctionalTests/NonWinRT/NonWinRT.csproj">
       <Platform Solution="*|ARM64" Project="ARM64" />
       <Platform Solution="*|x64" Project="x64" />
--- a/src/cswinrt.slnx
+++ b/src/cswinrt.slnx
@@ -348,11 +348,9 @@
       <Build Solution="*|ARM64" Project="false" />
     </Project>
     <Project Path="Tests/FunctionalTests/NativeExposedType/NativeExposedType.csproj">
-      <Platform Solution="*|ARM" Project="ARM" />
       <Platform Solution="*|ARM64" Project="ARM64" />
       <Platform Solution="*|x64" Project="x64" />
       <Platform Solution="*|x86" Project="x86" />
-      <Build Solution="*|ARM" Project="false" />
       <Build Solution="*|ARM64" Project="false" />
     </Project>
     <Project Path="Tests/FunctionalTests/NonWinRT/NonWinRT.csproj">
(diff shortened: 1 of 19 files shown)


src/cswinrt.slnx
```diff
@@ -420,6 +420,14 @@
       <Build Solution="*|ARM" Project="false" />
       <Build Solution="*|ARM64" Project="false" />
     </Project>
+    <Project Path="Tests/FunctionalTests/NativeExposedType/NativeExposedType.csproj">
+      <Platform Solution="*|ARM" Project="ARM" />
```
manodasanW: once you pull latest, drop ARM.

Sergio0694: Done!

manodasanW (approved):

PR #2484, opened 2026-07-15
Fix SupportedOSPlatform attributes for Desktop extension SDK

manodasanW:
We project types from the Desktop extension SDK, but our lookup for generating SupportedOSPlatform didn't have those in the entries.  In addition, when we did the previous fix related to SupportedOSPlatform to not have an exact matching, it changed some of the attributes in the cases where they were moved from the desktop extension SDK to the main UAP windows SDK.  So this should address those scenarios too.  Given we had already spun a 2.3 build, incrementing version.

### DIFF
--- a/build/AzurePipelineTemplates/CsWinRT-Variables.yml
+++ b/build/AzurePipelineTemplates/CsWinRT-Variables.yml
@@ -4,9 +4,9 @@ variables:
 - name: MinorVersion
   value: 3
 - name: PatchVersion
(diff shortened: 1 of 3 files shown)


manodasanW:
@microsoft-github-policy-service rerun

Sergio0694 (approved):

manodasanW:
@microsoft-github-policy-service rerun

PR #2487, opened 2026-07-20
Enable WinUI tests

manodasanW:
- Enabling the WinUI tests in our repo with workarounds
- The various workarounds I am doing will be addressed as part of WinUI package changes.
- Also fixing a couple issues that were noticed while enabling these tests:
- Fixing incremental builds by having a hashed file which we use to determine if the inputs changed rather than just based on the cs files that are generated.  Previously we had no inputs, so the cs files always got generated and were detected as changed.
- Fixing winmd generation in authoring scenarios to not rely on attributes in the projection to emit the type references for like base types to dependent winmds but rather look at the winmds themselves to determine which is the correct one to emit the type reference.  This allows for the attribute we had previously in the ref projection to be removed.
- Fix issue with ObservableDictionary where the impl was not tracked.


src/WinRT.Projection.Generator/Generation/ProjectionGenerator.Generate.cs
```diff
@@ -115,9 +115,13 @@ private static void BuildWriterOptions(
 
         PathAssemblyResolver resolver = new(resolverPaths);
 
-        bool isWindowsSdkMode = args.WindowsSdkOnly || args.WindowsUIXamlProjection;
         bool isComponentMode = args.AssemblyName == "WinRT.Component";
 
+        // 'WindowsUIXamlProjection' also flows to component mode (for the TypeMapAssemblyTarget union),
```
Sergio0694:
```suggestion
        // 'WindowsUIXamlProjection' also flows to component mode (for the '[TypeMapAssemblyTarget]' union),
```

src/WinRT.WinMD.Generator/Helpers/WindowsRuntimeMetadataNameResolver.cs
```diff
@@ -0,0 +1,128 @@
...
+{
+    /// <summary>
+    /// Builds the <c>(namespace, name)</c> to source <c>.winmd</c> stem lookup from the given inputs.
+    /// </summary>
+    /// <param name="winmdPaths">The input <c>.winmd</c> file or directory paths (third party components and internal metadata).</param>
+    /// <param name="windowsMetadata">The Windows metadata token (path, directory, <c>"local"</c>, <c>"sdk"</c>, <c>"sdk+"</c>, or a version).</param>
+    /// <param name="token">The cancellation token for the operation.</param>
+    /// <returns>The resulting metadata-name lookup.</returns>
+    public static FrozenDictionary<(string? Namespace, string? Name), string> Build(
+        IReadOnlyList<string> winmdPaths,
```
Sergio0694:
```suggestion
        IEnumerable<string> winMDPaths,
```

src/WinRT.WinMD.Generator/WinRT.WinMD.Generator.csproj
```diff
@@ -57,6 +57,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\WinRT.Generator.Core\WinRT.Generator.Core.csproj" />
+    <ProjectReference Include="..\WinRT.Projection.Writer\WinRT.Projection.Writer.csproj" />
```
Sergio0694: Why is this needed?

manodasanW: Using WindowsMetadataExpander to parse the WindowsMetadata property

Sergio0694: Mmh can you move that type to the Core project instead? The WinMD project shouldn't reference the projection writer. If the expander is supposed to be used by multiple generators then being in Core would be the correct thing anyway. Please also adjust the namespace when moving the file to be consistent with the others there.

src/WinRT.WinMD.Generator/Helpers/WindowsRuntimeMetadataNameResolver.cs
```diff
@@ -0,0 +1,128 @@
...
+        }
+    }
+
+    /// <summary>
+    /// Adds all public top-level types from a single <c>.winmd</c> file to the lookup, keyed by their
+    /// namespace and name, mapping each to that <c>.winmd</c> file's stem.
+    /// </summary>
+    /// <param name="builder">The lookup being populated.</param>
+    /// <param name="winmdPath">The <c>.winmd</c> file path.</param>
+    private static void AddWinMD(Dictionary<(string?, string?), string> builder, string winmdPath)
```
Sergio0694: I feel like the name shouldn't be nullable (the code also does check that it isn't). If there's any downstream null checks on it, we can likely cleanup those too, since they shouldn't be needed for this.

src/WinRT.WinMD.Generator/Helpers/WindowsRuntimeMetadataNameResolver.cs
```diff
@@ -0,0 +1,128 @@
...
+        {
+            // Skip the '<Module>' pseudo-type and any non-public types (Windows Runtime types are public)
+            if (!type.IsPublic || type.Name?.Value is not { } name || name.StartsWith('<'))
+            {
+                continue;
+            }
+
+            // First .winmd defining a given (namespace, name) wins; duplicate contract definitions
+            // across metadata files are not expected in practice.
+            _ = builder.TryAdd((type.Namespace?.Value, name), stem);
```
Sergio0694: Isn't it invalid for a WinRT type in a .winmd to have a `null` namespace? Can we check that here, ignore any that are `null` (which should never happen I guess), and make the namespace not-nullable in the tuple too? And like above, cleanup any `null` checks we have on this tuple element elsewhere, if any.

Sergio0694 (approved):

src/WinRT.WinMD.Generator/WinRT.WinMD.Generator.csproj
manodasanW: Done

Sergio0694 (approved):

PR #2489, opened 2026-07-23
Improve repo building experience

manodasanW:
- Previously when you tried to build the slnx in VS, we would often run to WinRT.Generator.Tasks.dll is being used by msbuild and can't be replaced by the build and we would need to kill the processes and then build any changes for it.  In addition, even if we did no changes for it, the builds for it often got triggered due to changing central props / targets file.  To address this, we are using solution filters to filter out the task dll and move that to a separate solution filter project.  This way when changes are done to the task, they can be done and built and then the rest of the projects can be opened and if there are no changes, then other changes don't impact it either.
- Also updating the PlatformToolset defaults across the projects and some dependencies.


Sergio0694:
Nit: seeing this when opening VS still:

> "C:\git\CsWinRT\src\TestWinRT\BenchmarkComponent\BenchmarkComponent.vcxproj : warning  : Platform 'ARM' referenced in the project file 'BenchmarkComponent' cannot be found."

manodasanW:
> Nit: seeing this when opening VS still:
> 
> > "C:\git\CsWinRT\src\TestWinRT\BenchmarkComponent\BenchmarkComponent.vcxproj : warning  : Platform 'ARM' referenced in the project file 'BenchmarkComponent' cannot be found."

That was due to get_testwinrt.cmd was not ran I believe, added it to the docs that someone should run it the first time.

Sergio0694 (approved):

PR #2490, opened 2026-07-24
Put IReferenceArray<object> on vtable for arrays of reference types

manodasanW:
We previously didn't put `IReferenceArray<object>` on the vtable for boxed C# arrays of class types.  We only did it for value types and delegates.  But `IInspectableArray` is a thing that `PropertyValue` recognizes and even has `CreateIInspectableArray`.  So, we are fixing our logic to also project `IReferenceArray<object>` and `IPropertyValue` for C# arrays of class types.  Also double checked with cppwinrt's implementation that they also have `IReferenceArray<object>`.


Sergio0694 (approved):
Does this just work for 3.0 or do we need to make the change there too?

manodasanW:
I think this will need to be ported to 3.0, plan on taking a look after I merge this.

PR #2494, opened 2026-07-29
Only carry over metadata attributes into reference projections

Sergio0694:
## Summary

Implementation projections now carry over no Windows Runtime metadata attributes at all (except `[AttributeUsage]`). Reference projections are completely unchanged: their generated output is byte-identical before and after.

## Motivation

Implementation projections (`WinRT.Sdk.Projection.dll`, `WinRT.Sdk.Xaml.Projection.dll`, `WinRT.Projection.dll`, produced by `cswinrtprojectiongen` at publish time) are only ever *loaded* at runtime. User code, compilers, analyzers and metadata tooling all see the *reference* projection instead. Any Windows Runtime metadata attribute carried into an implementation projection therefore has no consumer.

Worse, attribute blobs cannot be trimmed by ILLink or ILC, so every one of those applications is permanent, unremovable metadata in the shipped application.

`[ContractVersion]`, `[ApiContract]` and the synthesized `[SupportedOSPlatform]` were already reference-projection-only. This generalizes that to every carried-over metadata attribute.

## Changes

- **`src/WinRT.Projection.Writer/Factories/CustomAttributeFactory.cs`**: a new `ShouldCarryOverAttribute` predicate replaces the inline special cases in `WriteCustomAttributes`, so the "reference-only" rule lives in one documented place instead of being spread across the loop. It also drops the previously carried-over attributes from other namespaces (e.g. `Microsoft.UI.Xaml.TemplatePart`, the `MUX*` attributes). Argument formatting now happens *after* the filter, so no work is done for attributes that get dropped (implementation projections are the hot publish-time path). The `[AllowMultiple]` observation and the reference-only `[SupportedOSPlatform]` synthesis are unchanged.
- **`src/WinRT.Projection.Writer/Factories/InterfaceFactory.cs`**: `WriteMethodCustomAttributes` now takes the emit context and returns early for implementation projections. This path was not gated on the projection mode at all, and was the single largest contributor.
- **`src/Tests/ProjectionWriterTest/`**: new test project that drives the projection writer end-to-end through `cswinrtprojectionrefgen` and asserts which attributes survive in each mode.
- **`src/Tests/UnitTest/TestComponentCSharp_Tests.cs`**: `TestRepeatedAttributesAreProjected` inverted (see below).
- **`src/cswinrt.slnx`**, **`build/AzurePipelineTemplates/CsWinRT-Test-Steps.yml`**: wire up the new test project.

`[AttributeUsage]` is deliberately kept in both modes. It is not carried-over Windows Runtime metadata but the .NET modeling of a projected attribute type's own usage contract, and its `AllowMultiple`/`Inherited` arguments drive `Attribute.GetCustomAttributes(inherit: true)` semantics at runtime. It is also negligible in size (36 applications across the entire Windows SDK projection).

## Results

Measured on the real `WinRT.Sdk.Projection.dll` (SDK 26100), at IL metadata level:

| | Before | After |
| --- | --- | --- |
| Assembly size | 23,913,472 bytes | **23,885,824 bytes** (-27,648, -0.116%) |
| Custom attribute applications | 99,279 | **98,543** (-736) |
| `[Overload]` | 720 | **0** |
| `[DefaultOverload]` | 16 | **0** |
| `[WindowsRuntimeType]` | 8,451 | 8,451 (unchanged) |
| `[AttributeUsage]` | 36 | 36 (unchanged) |

Generated sources, running the writer over both metadata sets in both modes:

| Projection | Attribute applications removed | Source bytes removed |
| --- | --- | --- |
| Windows SDK, implementation | **746** (707 `[Overload]`, 27 `[Experimental]`, 12 `[DefaultOverload]`) | 61,554 |
| WinUI, implementation | **411** (146 `[Overload]`, 180 `MUX*`, 73 `ContentProperty`, 6 `[DefaultOverload]`, 4 `InputProperty`, 1 `[Bindable]`, 1 `MUXHasCustomActivationFactory`) | 36,444 |
| Windows SDK, reference | 0 | **0 (byte-identical, 295/295 files)** |
| WinUI, reference | 0 | **0 (byte-identical, 65/65 files)** |

Across every changed implementation-projection file there are **zero added lines**; every removed line is an attribute application. The compiled reference assembly (`ref\Microsoft.Windows.SDK.NET.dll`) is byte-identical too.

## Test coverage

The new `ProjectionWriterTest` project runs `cswinrtprojectionrefgen` as a separate process (exactly as the MSBuild target does) over `Windows.Foundation` in both modes, caching the sources so the whole suite runs in about a second. Its central assertion is that **no** `Windows.Foundation.Metadata` attribute survives into an implementation projection, so any future leak fails the build instead of silently adding permanent metadata. The reference-projection counterparts are asserted too, which keeps the negative assertions from becoming vacuous, along with the `[AllowMultiple]` folding (identical in both modes) and the attributes implementation projections genuinely need. Verified that the suite fails without the writer change and passes with it.

`TestRepeatedAttributesAreProjected` (added in the base PR) asserted the opposite and had to be inverted. Projection assemblies ship a forwarder as their primary output, with the reference assembly under `ref\`, so at runtime those types resolve into the implementation projection generated for the app, which is exactly where these attributes are now dropped. It is replaced by `TestWindowsRuntimeAttributesAreNotProjectedInImplementationProjection`, which asserts the four `[attr_repeatable]` and one `[attr_string]` applications are gone, that no projected Windows Runtime attribute type survives at all, and that `[WindowsRuntimeType]` is still applied so over-stripping is caught too. `TestAllowMultipleIsProjectedOnAttributeUsage` is unchanged and still passes.

One caveat worth flagging: the base PR's assertion that *all four* applications of a repeated `[allowmultiple]` attribute are carried over no longer has runtime coverage, because the runtime only ever sees the implementation projection. There is no type anywhere in the Windows SDK or WinUI metadata with two or more applications of the same attribute, so that scenario is only reachable through `TestComponentCSharp.winmd`, which the new test project does not consume (it uses the `sdk` input token and has no C++/MIDL dependency). Happy to wire the test component's `.winmd` in if that is preferred.

## Out of scope

- The hand-written `[SupportedOSPlatform]` applications in `Resources/Base/ComInteropExtensions.cs` and `Resources/Additions/*` are literal resource text (15 of them survive into the Windows SDK implementation projection). Making those conditional is a separate change.
- `src/Tests/WinMDGeneratorTest` has the same latent `Platform` propagation issue the new test project works around with `UndefineProperties`; left alone as unrelated.

### DIFF
--- a/src/WinRT.Projection.Writer/Factories/CustomAttributeFactory.cs
+++ b/src/WinRT.Projection.Writer/Factories/CustomAttributeFactory.cs
@@ -353,15 +353,20 @@ internal static class CustomAttributeFactory
     }
 
     /// <summary>
-    /// Writes any custom attributes (e.g. <c>[Obsolete]</c>, <c>[Deprecated]</c>,
-    /// <c>[SupportedOSPlatform]</c>) carried over from <paramref name="member"/> to the projection.
+    /// Writes the Windows Runtime metadata custom attributes carried over from <paramref name="member"/>
+    /// to the projection (e.g. <c>[AttributeUsage]</c>, and — in reference projections only —
+    /// <c>[Overload]</c>, <c>[Experimental]</c>, <c>[ContractVersion]</c>, plus the synthesized
+    /// <c>[SupportedOSPlatform]</c>).
     /// </summary>
     /// <remarks>
     /// Attributes are emitted in metadata order, one line per application. Windows Runtime attributes
     /// marked <c>[allowmultiple]</c> can be applied several times to the same member, so every
     /// application is carried over. The <c>[AllowMultiple]</c> metadata attribute itself is not
     /// projected: it is folded into the <c>AllowMultiple</c> named argument of the projected
-    /// <c>[AttributeUsage]</c>, matching how .NET models repeatability.
+    /// <c>[AttributeUsage]</c>, matching how .NET models repeatability. Which applications survive is
+    /// decided by <see cref="ShouldCarryOverAttribute"/>: implementation projections keep only
+    /// <c>[AttributeUsage]</c>, since every other metadata attribute is dead, untrimmable weight in an
+    /// assembly that is only ever loaded at runtime (see that method for the full rationale).
     /// </remarks>
     /// <param name="writer">The writer to emit to.</param>
     /// <param name="context">The active emit context.</param>
@@ -404,11 +409,12 @@ internal static class CustomAttributeFactory
                 continue;
             }
 
-            string fullAttrName = strippedName == "AttributeUsage"
-                ? AttributeUsageName
-                : ns + "." + strippedName;
-
-            List<string> args = WriteCustomAttributeArgs(attr);
+            // '[AllowMultiple]' is never carried over itself: it is observed here (in both projection
+            // modes) and folded into the projected '[AttributeUsage]' after all attributes are seen.
+            if (ns == WindowsFoundationMetadata && strippedName == "AllowMultiple")
+            {
+                allowMultiple = true;
+            }
 
             // '[SupportedOSPlatform]' takes a single platform, so only the first resolved one is emitted
             // (matching the member-level behavior in 'WritePlatformAttributeBody'). A member can carry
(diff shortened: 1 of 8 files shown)


src/WinRT.Projection.Writer/Factories/CustomAttributeFactory.cs
```diff
@@ -353,15 +353,20 @@ public static string GetPlatformAttribute(ProjectionEmitContext context, IHasCus
     }
 
     /// <summary>
-    /// Writes any custom attributes (e.g. <c>[Obsolete]</c>, <c>[Deprecated]</c>,
-    /// <c>[SupportedOSPlatform]</c>) carried over from <paramref name="member"/> to the projection.
+    /// Writes the Windows Runtime metadata custom attributes carried over from <paramref name="member"/>
+    /// to the projection (e.g. <c>[AttributeUsage]</c>, and — in reference projections only —
+    /// <c>[Overload]</c>, <c>[Experimental]</c>, <c>[ContractVersion]</c>, plus the synthesized
```
manodasanW, line 358: Are you saying these attributes are only written for reference projections or the opposite?

manodasanW (approved):

src/WinRT.Projection.Writer/Factories/CustomAttributeFactory.cs
Sergio0694: Only for These are only written for reference projections. For the impl .dll there's a short-circuit path right before this that will just always return false. The only attributes we ever emit for impl .dll-s are the CsWinRT ones we need.

PR #2503, opened 2026-08-01
Project the Windows Runtime [Experimental] attribute as the .NET one

Sergio0694:
## Summary

Custom-map `Windows.Foundation.Metadata.ExperimentalAttribute` to `System.Diagnostics.CodeAnalysis.ExperimentalAttribute`, with the `CSWINRT3005` diagnostic id, in both directions (projection and component authoring). Also adds `docs/attribute-projections.md`, documenting every attribute CsWinRT translates rather than projects verbatim.

> [!NOTE]
> Stacked on top of #2457, which is where the `[Deprecated]` support this builds on comes from. Only the last three commits belong to this PR.

## Motivation

`Windows.Foundation.Metadata.ExperimentalAttribute` was projected as an ordinary Windows Runtime attribute type, and the C# compiler happened to recognize it *by name* and report `CS8305` for its use sites. That is a hard-coded, single, non-actionable warning: it cannot be suppressed for one API without suppressing it for every experimental API in the SDK, it carries no link to any documentation, and it only works because the compiler special-cases that one type name.

C# 12 added [`[Experimental]`](https://learn.microsoft.com/dotnet/csharp/language-reference/proposals/csharp-12.0/experimental-attribute) as the first-class modeling of exactly this concept, and it is strictly richer: it carries a diagnostic id, a help link, and a message, so opting into an experimental API becomes a normal, per-id, greppable suppression. Projecting the Windows Runtime attribute as the .NET one is the same treatment `IClosable`, `IReference<T>` and every other Windows Runtime type with a .NET counterpart already gets.

The behavior around attributes is also currently only discoverable by reading the projection writer and the WinMD generator, which is why this adds a documentation page for it.

### Projection (`.winmd` → C#)

- **`src/WinRT.Projection.Writer/Helpers/MappedTypes.cs`**: custom-maps `Windows.Foundation.Metadata.ExperimentalAttribute` to `System.Diagnostics.CodeAnalysis.ExperimentalAttribute`, so the Windows Runtime attribute type is no longer projected.

- **`src/WinRT.Projection.Writer/Factories/CustomAttributeFactory.cs`**: emits `[Experimental("CSWINRT3005", UrlFormat = ..., Message = ...)]` for every application. Windows Runtime metadata has no per-API diagnostic id (the attribute takes no arguments), so all experimental APIs share one CsWinRT-owned id. The emit loop is extracted into a `WriteAttribute` helper so this and the carried-over path share it.

- **`src/WinRT.Projection.Writer/Factories/InterfaceFactory.cs`**: the same for the interface member level carry-over path.

- **`src/WinRT.Projection.Writer/References/WellKnownDiagnostics.cs`**: new, holds the diagnostic id, message and url format.

- **`src/WinRT.Projection.Writer/Extensions/ProjectionWriterExtensions.cs`**: generated projection files suppress `CSWINRT3005` in their prelude, alongside the `CS0612`/`CS0618` they already suppress. A projection has to name an experimental type in order to project it at all, so the marker is guidance for the consumers of a projection, not for the projection itself. Unlike `CS8305`, this is reported as an error by default, so the suppression is what keeps the Windows SDK projection building.

- **`src/Projections/Directory.Build.props`**: drops the now-redundant `CS8305` `NoWarn` (verified by building the full Windows SDK reference projection without it).

Emission stays reference-projection-only, matching every other metadata attribute: implementation projections are only ever loaded at runtime, and attribute blobs cannot be trimmed by ILLink or ILC.

### Authoring (C# → `.winmd`)

Since the Windows Runtime attribute type is no longer projected, it has no projected form left for a component author to apply. Authored components use the .NET `[Experimental]` instead, and the generator translates it back, so the component looks the same to every other language projection as one authored in MIDL.

- **`src/WinRT.WinMD.Generator/Writers/WinMDWriter.Attributes.cs`**: translates `System.Diagnostics.CodeAnalysis.ExperimentalAttribute` into `Windows.Foundation.Metadata.ExperimentalAttribute`. The diagnostic id, `UrlFormat` and `Message` are dropped, as Windows Runtime metadata has nowhere to carry them.

- **`src/WinRT.WinMD.Generator/Writers/WinMDWriter.Members.cs`**: generalizes the accessor placement established for `[Deprecated]` in #2457, so both attributes land on the accessor method for properties and events (the placement MIDL produces), rather than duplicating it.

### Documentation

- **`docs/attribute-projections.md`**: new, one table per direction covering every attribute CsWinRT translates rather than projects verbatim, including which projection assemblies each result lands in. The two asymmetries easiest to get wrong are called out explicitly: the metadata attribute types that are not projected as types at all, and the fact that `[Obsolete]` on an authored component is *not* translated into `[Deprecated]`, so it is invisible to every other language projection.

- **`docs/diagnostics/cswinrt3005.md`**: new, the diagnostic page, including how to opt into an experimental API and why it is an error rather than a warning.

- **`README.md`**, **`docs/authoring.md`**, **`.github/copilot-instructions.md`**: link the new pages and record the new diagnostic id.

## Testing

- **`src/Tests/ProjectionWriterTest/Test_MappedAttributes.cs`**: new, asserts the exact attribute text the writer emits (so the id, url and message cannot drift from the docs page silently), that it is reference-projection-only, and that neither the Windows Runtime attribute type nor any application of it survives.

- **`src/Tests/WinMDGeneratorTest/Test_ExperimentalAttribute.cs`**: new, asserts the authoring translation for types, methods, properties and events. `WinMDGeneratorRunner` gains the ability to read attributes back out of the generated `.winmd` (via `System.Reflection.Metadata`, with Windows Runtime projections disabled so the raw metadata is asserted), which is what lets these tests assert *placement*, not just success.

- **`src/Tests/UnitTest/TestComponentCSharp_Tests.cs`**: updates the existing `CS8305` suppression to `CSWINRT3005`, and adds a test asserting the attribute does not survive into the implementation projection loaded at runtime.

- **`src/Tests/WinMDGeneratorTest/WinMDGeneratorTest.csproj`**: drops `Platform` from the generator `ProjectReference`'s `UndefineProperties`, matching `ProjectionWriterTest`. Without it, building with an explicit platform puts the tool under `bin\<platform>\` while the tests look for it under `bin\`, so they silently run against a stale tool.

Verified by building the full Windows SDK reference projection (0 warnings, 0 errors, 40 projected `[Experimental]` applications and no leftover Windows Runtime ones) and running both generator test suites.

### DIFF
--- a/src/Tests/SourceGenerator2Test/Test_ExperimentalAttributeTargetAnalyzer.cs
+++ b/src/Tests/SourceGenerator2Test/Test_ExperimentalAttributeTargetAnalyzer.cs
@@ -1,0 +1,287 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Threading.Tasks;
+using WindowsRuntime.SourceGenerator.Diagnostics;
+using WindowsRuntime.SourceGenerator.Tests.Helpers;
+
+namespace WindowsRuntime.SourceGenerator.Tests;
+
+using VerifyCS = CSharpAnalyzerTest<ExperimentalAttributeTargetAnalyzer>;
+
+/// <summary>
+/// Tests for <see cref="ExperimentalAttributeTargetAnalyzer"/>.
+/// </summary>
+[TestClass]
+public sealed class Test_ExperimentalAttributeTargetAnalyzer
+{
+    [TestMethod]
+    [DataRow("public sealed class MyClass;")]
+    [DataRow("public interface IMyInterface;")]
+    [DataRow("public struct MyStruct;")]
+    [DataRow("public enum MyEnum { A }")]
+    [DataRow("public delegate void MyDelegate();")]
+    public async Task ExperimentalOnType_DoesNotWarn(string declaration)
+    {
+        string source = $$"""
+            using System.Diagnostics.CodeAnalysis;
+
+            [Experimental("TEST0001")]
+            {{declaration}}
+            """;
+
+        await VerifyCS.VerifyAnalyzerAsync(source, isCsWinRTComponent: true);
+    }
+
+    [TestMethod]
+    public async Task ExperimentalOnMembers_DoesNotWarn()
+    {
+        const string source = """
+            using System;
+            using System.Diagnostics.CodeAnalysis;
+
+            public sealed class MyClass
+            {
+                [Experimental("TEST0001")]
+                public int MyMethod() => 42;
+
+                [Experimental("TEST0002")]
+                public int MyProperty { get; set; }
+
+                [Experimental("TEST0003")]
+                public event EventHandler<int> MyEvent;
+            }
+
+            public struct MyStruct
+            {
+                [Experimental("TEST0004")]
+                public int MyField;
+            }
+
+            public enum MyEnum
+            {
+                [Experimental("TEST0005")]
+                A
+            }
+            """;
+
+        await VerifyCS.VerifyAnalyzerAsync(source, isCsWinRTComponent: true);
+    }
+
+    [TestMethod]
+    [DataRow("private")]
+    [DataRow("internal")]
+    [DataRow("protected")]
+    public async Task ExperimentalOnNonPublicConstructor_DoesNotWarn(string accessibility)
+    {
+        string source = $$"""
+            using System.Diagnostics.CodeAnalysis;
+
+            public class MyClass
+            {
+                [Experimental("TEST0001")]
+                {{accessibility}} MyClass()
+                {
+                }
+            }
+            """;
+
+        await VerifyCS.VerifyAnalyzerAsync(source, isCsWinRTComponent: true);
+    }
+
+    [TestMethod]
+    public async Task ExperimentalOnStaticConstructor_DoesNotWarn()
+    {
+        const string source = """
+            using System.Diagnostics.CodeAnalysis;
+
+            public sealed class MyClass
+            {
+                [Experimental("TEST0001")]
+                static MyClass()
+                {
+                }
+            }
+            """;
+
+        await VerifyCS.VerifyAnalyzerAsync(source, isCsWinRTComponent: true);
+    }
+
+    [TestMethod]
+    public async Task ExperimentalOnConstructorOfNonVisibleType_DoesNotWarn()
+    {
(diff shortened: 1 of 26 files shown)


src/Tests/WinMDGeneratorTest/Test_ExperimentalAttribute.cs
```diff
@@ -0,0 +1,121 @@
...
+            """);
+
+        AssertIsExperimental(attributes, "IExperimentalComponent");
+        AssertIsExperimental(attributes, "IComponent.ExperimentalMethod");
+
+        AssertIsNotExperimental(attributes, "IComponent");
+        AssertIsNotExperimental(attributes, "IComponent.StableMethod");
+
+        // The .NET attribute type has no Windows Runtime counterpart, so no row may reference it
+        Assert.IsFalse(
```
manodasanW, line 71: Are we saying we don't custom map experimental back into the winmd for authoring scenarios?

Sergio0694: Oh, I might've missed the authoring part, I'll double-check and add that too, good point!

Sergio0694: Fixed!

src/Tests/SourceGenerator2Test/Test_ExperimentalAttributeTargetAnalyzer.cs
```diff
@@ -0,0 +1,287 @@
...
+            using System;
+
+            public sealed class ExperimentalAttribute : Attribute
+            {
+                public ExperimentalAttribute(string diagnosticId)
+                {
+                }
+            }
+
+            public sealed class MyClass
```
manodasanW, line 186: Should we have another version of doesnotwarn for internal constructor or so and internal type with public constructor.

manodasanW (approved):

src/Tests/SourceGenerator2Test/Test_ExperimentalAttributeTargetAnalyzer.cs
Sergio0694: Checked, both are already handled by the existing tests 👍

PR #2505, opened 2026-08-01
Project Windows Runtime struct fields as C# fields

Sergio0694:
## Summary

Projects the fields of Windows Runtime struct types as C# fields instead of `{ readonly get; set; }` properties, across the projection writer, the manually projected types in `WinRT.Runtime`, and the custom-mapped XAML types defined by addition files.

## Motivation

Projecting struct fields as properties turned out to be problematic, and confusing.

The biggest issue is authoring. `cswinrtwinmdgen.exe` maps the public instance **fields** of an authored `struct` to Windows Runtime struct fields, which is the only shape that can round-trip: an author writing a struct against the projected shape (properties) would silently produce an empty struct in the generated `.winmd`. Projecting fields as fields makes consumption and authoring symmetrical: the shape you consume is exactly the shape you author.

Beyond authoring, Windows Runtime structs are plain data — there is no accessor to run any logic behind, so a property only obscures that, and it prevents callers from taking a reference to a member (e.g. to pass it as a `ref` argument, or to mutate an element of an array of structs in place). This also matches what C++/WinRT does, and what CsWinRT 2.x did for generated struct projections.

## Changes

**Projection writer**

- **`src/WinRT.Projection.Writer/Builders/ProjectionFileBuilder.cs`**: `WriteStruct` now emits a public field per Windows Runtime struct field, declared before the constructor so declaration order (which drives the sequential layout of the struct) mirrors the metadata order.
- **`src/WinRT.Projection.Writer/Factories/AbiStructFactory.cs`**, **`src/WinRT.Projection.Writer/Factories/StructEnumMarshallerFactory.cs`**: refresh comments that justified the mapped-struct special cases by claiming their projected public fields don't match the WinMD field layout, which is no longer accurate.

**Manually projected types**

- **`src/WinRT.Runtime2/Windows.Foundation/Point.cs`**, **`Rect.cs`**, **`Size.cs`**, **`src/WinRT.Runtime2/InteropServices/Events/EventRegistrationToken.cs`**: `X`, `Y`, `Width`, `Height` and `Value` become public fields. The negative-value validation that lived in the `Rect`/`Size` setters moves into the constructors that document it (which is also what CsWinRT 2.x did). Managed-only members (`Left`, `Top`, `Right`, `Bottom`, `IsEmpty`, `Empty`, ...) are unchanged.

**Custom-mapped XAML types (addition files, both the `Microsoft.UI.Xaml` and `Windows.UI.Xaml` copies)**

- **`…/Microsoft.UI.Xaml.CornerRadius.cs`**: `TopLeft`, `TopRight`, `BottomRight`, `BottomLeft`
- **`…/Microsoft.UI.Xaml.GridLength.cs`**: `Value`, `GridUnitType`
- **`…/Microsoft.UI.Xaml.Duration.cs`**: `TimeSpan`, `Type`
- **`…/Microsoft.UI.Xaml.Media.Animation.KeyTime.cs`**: `TimeSpan`
- **`…/Microsoft.UI.Xaml.Media.Animation.RepeatBehavior.cs`**: `Count`, `Duration`, `Type`
- **`…/Microsoft.UI.Xaml.Media.Media3D.Matrix3D.cs`**: `M11`-`M34`, `OffsetX`, `OffsetY`, `OffsetZ`, `M44`

These types are defined in full by their addition files, so they kept the property-over-private-backing-field shape inherited from the UWP/WPF managed projections. Leaving them alone would have made them inconsistent with their own siblings, since `Thickness`, `Matrix`, `GeneratorPosition` and `Color` are generated types in CsWinRT 3.0 and therefore now expose fields. Field declaration order (and therefore the sequential layout each of these blittable structs marshals with) is unchanged, and validation that lived in the removed setters is preserved where the constructors already performed it. `GridLength` and `KeyTime` are no longer `readonly struct`-s, since Windows Runtime structs are mutable data. Members that are not Windows Runtime struct fields are left untouched (`GridLength.IsAbsolute`/`IsAuto`/`IsStar`/`Auto`, `Duration.HasTimeSpan`/`Automatic`/`Forever`, `RepeatBehavior.HasCount`/`HasDuration`/`Forever`, `Matrix3D.Identity`/`IsIdentity`/`HasInverse`, and all operators, conversions and formatting helpers).

There is one deliberate behavior change on top of that: the `GridLength` constructors no longer reject negative (or non-finite) values. Negative `GridLength` values turn out to be a valid pattern that real apps rely on (e.g. the Microsoft Store uses them in several custom layout panels), and now that `Value` is a public mutable field the check was trivially bypassable anyway. The `GridUnitType` validation is unchanged, `GridUnitType.Auto` still normalizes the value to the default, and the `IsFinite` helper the removed check was the only user of goes away with it.

**Tests and docs**

- **`src/Tests/ProjectionWriterTest/Test_ProjectedStructs.cs`**: new tests asserting that projected struct fields are emitted as C# fields, and that no projected struct member is emitted as an auto-property, in both projection modes.
- **`docs/cswinrt3.0-spec.md`**: documents the projection rule, and renames the neighbouring `Point`/`Rect`/`Size` section (which was already about their fields rather than properties).
- **`docs/event-infrastructure.md`**: refreshes the `EventRegistrationToken` snippet.

## Validation

- `ProjectionWriterTest` (16), `WinMDGeneratorTest` (17) and `SourceGenerator2Test` (129) all pass.
- The `Windows`, `Windows.UI.Xaml` and `WinAppSDK` projections were regenerated end-to-end and compile cleanly, which covers the emitted structs, their marshallers, and all of the addition files.
- Both `GridLength` addition files were compiled standalone and exercised to confirm negative, `NaN` and infinite values now construct successfully, that `GridUnitType.Auto` still normalizes to the default, and that an out-of-range `GridUnitType` still throws.

### DIFF
--- a/docs/cswinrt3.0-spec.md
+++ b/docs/cswinrt3.0-spec.md
@@ -94,9 +94,15 @@ CsWinRT 3.0 will update the projection of `T[]` parameters to use first-class sp
 
 This means you no longer need to allocate and copy stuff into arrays all over the place (because you also need the size to be exactly right, so you can't even use the array pool). Instead, you'll be able to just use spans normally. Which also means you can now even make it 0-alloc and pass stack-allocated params entirely. This is source compatible thanks to the [first class span](https://learn.microsoft.com/dotnet/csharp/language-reference/proposals/first-class-span-types) types feature of C# 14.
 
-### Fixing `Point`/`Rect`/`Size` properties
+### Fixing `Point`/`Rect`/`Size` fields
 
-These foundational types have historically been projecting their fields as `double`, instead of `float`. This is not ideal for several reasons: it introduces implicit casts when assigning to or reading from them, it doesn't match the WinRT ABI (the backing data is still just floats), and it unnecessary impacts performance when doing lots of heavy calculations with them. In CsWinRT 3.0, we want to try fixing this design aspect and correctly projecting these members as `float`, and monitor what the real impact is on popular projects using WinRT from C#.
+These foundational types have historically been projecting their fields as `double` properties, instead of `float` fields. This is not ideal for several reasons: it introduces implicit casts when assigning to or reading from them, it doesn't match the WinRT ABI (the backing data is still just floats), and it unnecessary impacts performance when doing lots of heavy calculations with them. In CsWinRT 3.0, we want to try fixing this design aspect and correctly projecting these members as `float` fields, and monitor what the real impact is on popular projects using WinRT from C#.
+
+### Projecting struct fields as fields
+
+Windows Runtime structs are plain data: all their members are fields in metadata. CsWinRT 3.0 projects them as C# fields, matching both the metadata and what C++/WinRT does. This keeps the projected shape honest (there is no accessor to run any logic behind), it allows callers to take a reference to a member (e.g. to pass it as a `ref` argument), and it makes authoring work symmetrically: `cswinrtwinmdgen.exe` maps public instance fields of an authored `struct` back to Windows Runtime struct fields, so the shape you consume is exactly the shape you author.
+
+This also applies to the manually projected and custom-mapped struct types, such as `Windows.Foundation.Point` and `Microsoft.UI.Xaml.CornerRadius`. Any additional member those types expose that does not correspond to a Windows Runtime struct field (e.g. `Rect.Left` or `GridLength.IsAuto`) remains a managed-only property.
 
 ### Unifying foundational event handers with .NET
 
--- a/src/WinRT.Projection.Writer/Resources/Additions/Microsoft.UI.Xaml/Microsoft.UI.Xaml.Duration.cs
+++ b/src/WinRT.Projection.Writer/Resources/Additions/Microsoft.UI.Xaml/Microsoft.UI.Xaml.Duration.cs
@@ -8,20 +8,20 @@ namespace Microsoft.UI.Xaml
     [WindowsRuntimeClassName("Windows.Foundation.IReference`1<Microsoft.UI.Xaml.Duration>")]
     [ABI.Microsoft.UI.Xaml.DurationComWrappersMarshaller]
 #endif
-    public readonly struct Duration : IEquatable<Duration>
+    public struct Duration : IEquatable<Duration>
     {
-        private readonly TimeSpan _timeSpan;
-        private readonly DurationType _durationType;
+        public TimeSpan TimeSpan;
+        public DurationType Type;
 
         public Duration(TimeSpan timeSpan)
         {
-            _durationType = DurationType.TimeSpan;
-            _timeSpan = timeSpan;
+            Type = DurationType.TimeSpan;
+            TimeSpan = timeSpan;
         }
 
         private Duration(DurationType durationType)
         {
-            _durationType = durationType;
+            Type = durationType;
         }
 
         public static implicit operator Duration(TimeSpan timeSpan)
@@ -33,9 +33,9 @@ namespace Microsoft.UI.Xaml
         {
             if (t1.HasTimeSpan && t2.HasTimeSpan)
             {
-                return new Duration(t1._timeSpan + t2._timeSpan);
+                return new Duration(t1.TimeSpan + t2.TimeSpan);
             }
-            else if (t1._durationType != DurationType.Automatic && t2._durationType != DurationType.Automatic)
+            else if (t1.Type != DurationType.Automatic && t2.Type != DurationType.Automatic)
             {
                 return Duration.Forever;
             }
@@ -50,9 +50,9 @@ namespace Microsoft.UI.Xaml
         {
             if (t1.HasTimeSpan && t2.HasTimeSpan)
             {
-                return new Duration(t1._timeSpan - t2._timeSpan);
+                return new Duration(t1.TimeSpan - t2.TimeSpan);
             }
-            else if (t1._durationType == DurationType.Forever && t2.HasTimeSpan)
+            else if (t1.Type == DurationType.Forever && t2.HasTimeSpan)
             {
                 return Duration.Forever;
             }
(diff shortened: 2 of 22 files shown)


docs/cswinrt3.0-spec.md
```diff
@@ -94,9 +94,15 @@ CsWinRT 3.0 will update the projection of `T[]` parameters to use first-class sp
...
+### Fixing `Point`/`Rect`/`Size` fields
 
-These foundational types have historically been projecting their fields as `double`, instead of `float`. This is not ideal for several reasons: it introduces implicit casts when assigning to or reading from them, it doesn't match the WinRT ABI (the backing data is still just floats), and it unnecessary impacts performance when doing lots of heavy calculations with them. In CsWinRT 3.0, we want to try fixing this design aspect and correctly projecting these members as `float`, and monitor what the real impact is on popular projects using WinRT from C#.
+These foundational types have historically been projecting their fields as `double` properties, instead of `float` fields. This is not ideal for several reasons: it introduces implicit casts when assigning to or reading from them, it doesn't match the WinRT ABI (the backing data is still just floats), and it unnecessary impacts performance when doing lots of heavy calculations with them. In CsWinRT 3.0, we want to try fixing this design aspect and correctly projecting these members as `float` fields, and monitor what the real impact is on popular projects using WinRT from C#.
```
manodasanW, line 99:
```suggestion
These foundational types have historically been projecting their fields as `double` fields, instead of `float` fields. This is not ideal for several reasons: it introduces implicit casts when assigning to or reading from them, it doesn't match the WinRT ABI (the backing data is still just floats), and it unnecessary impacts performance when doing lots of heavy calculations with them. In CsWinRT 3.0, we want to try fixing this design aspect and correctly projecting these members as `float` fields, and monitor what the real impact is on popular projects using WinRT from C#.
```

src/WinRT.Projection.Writer/Resources/Additions/Microsoft.UI.Xaml/Microsoft.UI.Xaml.Duration.cs
```diff
@@ -8,20 +8,20 @@ namespace Microsoft.UI.Xaml
     [WindowsRuntimeClassName("Windows.Foundation.IReference`1<Microsoft.UI.Xaml.Duration>")]
     [ABI.Microsoft.UI.Xaml.DurationComWrappersMarshaller]
 #endif
-    public readonly struct Duration : IEquatable<Duration>
+    public struct Duration : IEquatable<Duration>
     {
```
manodasanW, line 12: These always had special handling as properties before to enable throwing the exception and so on, do we want to change that?

src/WinRT.Projection.Writer/Resources/Additions/Microsoft.UI.Xaml/Microsoft.UI.Xaml.GridLength.cs
```diff
@@ -8,10 +8,10 @@ namespace Microsoft.UI.Xaml
     [WindowsRuntimeClassName("Windows.Foundation.IReference`1<Microsoft.UI.Xaml.GridLength>")]
     [ABI.Microsoft.UI.Xaml.GridLengthComWrappersMarshaller]
 #endif
-    public readonly struct GridLength : IEquatable<GridLength>
+    public struct GridLength : IEquatable<GridLength>
     {
-        private readonly double _unitValue;
-        private readonly GridUnitType _unitType;
+        public double Value;
```
manodasanW, line 13: This was also kept private before to allow to go through the constructor with validation.

docs/cswinrt3.0-spec.md
Sergio0694: Wait, why though? Those `double` members _were_ properties, otherwise the layout wouldn't be right. The types still had `float` backing fields, and exposed `double` properties.

manodasanW: Oh nvm, I just looked at the top of the old file and missed we also had properties below.

src/WinRT.Projection.Writer/Resources/Additions/Microsoft.UI.Xaml/Microsoft.UI.Xaml.Duration.cs
Sergio0694: Same as below, basically I'm not convinced we should be arbitrarily changing the WinRT API surface here? Like, we can add new helpers, constructors, etc. and stuff that _adds_ to WinRT types, but we have no precedent in 3.0 for just modifying existing APIs. It would also make these types be inconsistent with the other ones: no projected struct is readonly, and all other projected structs always keep the projected fields public. I feel like if people set wrong values we should just let WinUI itself throw an exception appropriately (I also don't remember ever hitting this exception myself, so I'm not even convinced it's that common/useful). And we can still keep the validation in the constructors, since those augment the WinRT types.

What do you think?

Edit: also to note, changing the public fields makes the docs for them incorrect too (since the fields are not there).

manodasanW: The docs do have [this](https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.duration?view=winrt-28000#projection-and-members-of-duration) section today.  Similar section exists for GridLength too and I believe the others that were customized.  It looks like they were customized for providing additional safety, not sure if they gained additional optimizations by being readonly for these particular types.  I can see both sides to this.

Sergio0694: That page would still be valid. We _do_ maintain all of those members, which are just additions. And we'd actually still be matching those fields mentioned in the docs as well, exactly like they're defined in the .winmd too. Basically I'm saying, having helpful additions (including constructors with validation) is completely fine. But we should not be opinionated in _removing_ projected surface areas from arbitrary WinRT structs.

src/WinRT.Projection.Writer/Resources/Additions/Microsoft.UI.Xaml/Microsoft.UI.Xaml.GridLength.cs
Sergio0694: Note that e.g. for this, the docs do mention them publicly ([here](https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.gridlength?view=winrt-28000#fields)) too.

manodasanW (approved):

PR #2506, opened 2026-08-01
Add an analyzer for [Obsolete] on authored APIs without [Deprecated]

Sergio0694:
## Summary

Add `CSWINRT2021`, an analyzer that reports publicly exposed APIs of an authored Windows Runtime component carrying `[Obsolete]` without `[Windows.Foundation.Metadata.Deprecated]`.

## Motivation

The WinMD generator does not translate `[Obsolete]` into `[Deprecated]`: it copies it verbatim, so the `.winmd` ends up with a `System.ObsoleteAttribute` reference that no other language projection (C++/WinRT, `windows-rs`, ...) understands. `[Obsolete]` on its own therefore deprecates an API for nobody but the C# code inside the component itself.

That is not obvious. `[Obsolete]` is *the* way to deprecate an API in C#, it compiles without complaint, and the component builds and works — the deprecation just silently never reaches any consumer. #2503 documented the behavior, but a doc only helps the developer who already suspects there is something to look up.

## Changes

- **`src/Authoring/WinRT.SourceGenerator2/Diagnostics/DiagnosticDescriptors.cs`**, **`AnalyzerReleases.Shipped.md`**: the `CSWINRT2021` descriptor (`Warning`), registered for release tracking.

- **`src/Authoring/WinRT.SourceGenerator2/Diagnostics/Analyzers/ObsoleteWithoutDeprecatedAnalyzer.cs`**: the analyzer. Gated on `CsWinRTComponent`, and silent when both attributes are applied, which is the supported way to deprecate an API for .NET and Windows Runtime consumers alike.

- **`docs/attribute-projections.md`**: links the diagnostic from the `[Obsolete]` note added in #2503.

### Scope

The check covers **types and their members**, not just types: the WinMD generator supports `[Deprecated]` on methods, properties and events too, so `[Obsolete]` on a member is exactly as silently ineffective as it is on a type.

Only what actually reaches the `.winmd` is reported, so that every report has an action the developer can take:

- Types are only reported when public and top level, as Windows Runtime has no nested types.

- Members are only reported when public and declared by such a type, and only for classes and interfaces: a Windows Runtime struct is a plain field aggregate, so the generator drops every member of one other than its public instance fields.

- Accessors are skipped, as they are only exported as part of their property or event (which is reported instead). The generator moves a `[Deprecated]` from the property or event down onto the accessor row, and never reads one written on a C# accessor.

- Constructors are skipped, because `[Deprecated]` does not include `AttributeTargets.Constructor` in its usage and so cannot be applied to one at all. Reporting them would produce a warning whose only resolutions are suppressing it or dropping the `[Obsolete]`.

## Testing

**`src/Tests/SourceGenerator2Test/Test_ObsoleteWithoutDeprecatedAnalyzer.cs`** covers the cases the analyzer must stay quiet for (no attributes, only `[Deprecated]`, both attributes, non-component projects, non-public types and members, nested types, struct members, constructors, accessors) and the ones it must report (every public type kind, and public methods, properties and events of both classes and interfaces).

The accessor test deliberately puts `[Obsolete]` on the accessor rather than on the property: an attribute written on a property is never surfaced on its accessor symbols, so a test relying on that would pass with the guard removed. Verified that this one does fail without it.

Full suite: 148/148 passing.

### DIFF
--- a/docs/attribute-projections.md
+++ b/docs/attribute-projections.md
@@ -52,7 +52,7 @@ What the WinMD generator emits into an authored component's `.winmd` for an attr
 | `[WindowsRuntime.Xaml.GeneratedCustomPropertyProvider]`, `[System.Reflection.DefaultMember]`, and anything under `System.Runtime.CompilerServices` | *nothing* | Either handled by CsWinRT itself or meaningless in Windows Runtime metadata. |
 | Any other public attribute type | itself | Non-public attribute types, and attributes whose signature cannot be read, are skipped. |
 
-> **Note**: `[System.Obsolete]` is **not** translated into `[Windows.Foundation.Metadata.Deprecated]`. It is copied as-is, so it is invisible to every other language projection. Use `[Windows.Foundation.Metadata.Deprecated]` to deprecate an API of an authored component. `[Experimental]` is the exception to that rule only because the Windows Runtime attribute has no projected form to apply.
+> **Note**: `[System.Obsolete]` is **not** translated into `[Windows.Foundation.Metadata.Deprecated]`. It is copied as-is, so it is invisible to every other language projection. Use `[Windows.Foundation.Metadata.Deprecated]` to deprecate an API of an authored component; applying `[Obsolete]` without it is reported as `CSWINRT2021`. `[Experimental]` is the exception to that rule only because the Windows Runtime attribute has no projected form to apply.
 
 ## Related documentation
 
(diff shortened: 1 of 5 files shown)


Sergio0694:
Superseded by #2508. This PR was never actually merged: while #2503 was being fixed, its branch (the base of this PR) briefly contained this PR's commits, and GitHub marks a PR as merged as soon as its head commits become reachable from its base. Those commits were removed from #2503 again two minutes later, so nothing from here landed in \staging/3.0\, #2503 or #2457. A merged PR cannot be reopened, so the work continues in #2508 with byte-identical content.

PR #2508, opened 2026-08-03
Add an analyzer for [Obsolete] on authored APIs without [Deprecated]

Sergio0694:
## Summary

Add `CSWINRT2021`, an analyzer that reports publicly exposed APIs of an authored Windows Runtime component carrying `[Obsolete]` without `[Windows.Foundation.Metadata.Deprecated]`.

> [!IMPORTANT]
> Replaces #2506, which GitHub incorrectly auto-closed as *merged*. Nothing was ever merged: while #2503 was being fixed, its branch (the base of #2506) briefly contained #2506's commits, and GitHub marks a PR merged as soon as its head commits become reachable from its base. The commits were removed from #2503 again minutes later, so these changes are in no branch other than this one. A merged PR cannot be reopened, hence this replacement. The content is byte-identical, rebased onto the current #2503 tip.

> [!NOTE]
> Stacked on top of #2503 (which is itself stacked on #2457). Only the last three commits belong to this PR.

## Motivation

The WinMD generator does not translate `[Obsolete]` into `[Deprecated]`: it copies it verbatim, so the `.winmd` ends up with a `System.ObsoleteAttribute` reference that no other language projection (C++/WinRT, `windows-rs`, ...) understands. `[Obsolete]` on its own therefore deprecates an API for nobody but the C# code inside the component itself.

That is not obvious. `[Obsolete]` is *the* way to deprecate an API in C#, it compiles without complaint, and the component builds and works — the deprecation just silently never reaches any consumer. #2503 documented the behavior, but a doc only helps the developer who already suspects there is something to look up.

## Changes

- **`src/Authoring/WinRT.SourceGenerator2/Diagnostics/DiagnosticDescriptors.cs`**, **`AnalyzerReleases.Shipped.md`**: the `CSWINRT2021` descriptor (`Warning`), registered for release tracking.

- **`src/Authoring/WinRT.SourceGenerator2/Diagnostics/Analyzers/ObsoleteWithoutDeprecatedAnalyzer.cs`**: the analyzer. Gated on `CsWinRTComponent`, and silent when both attributes are applied, which is the supported way to deprecate an API for .NET and Windows Runtime consumers alike.

- **`docs/attribute-projections.md`**: links the diagnostic from the `[Obsolete]` note added in #2503.

### Scope

The check covers **types and their members**, not just types: the WinMD generator supports `[Deprecated]` on methods, properties and events too, so `[Obsolete]` on a member is exactly as silently ineffective as it is on a type.

Only what actually reaches the `.winmd` is reported, so that every report has an action the developer can take:

- Types are only reported when public and top level, as Windows Runtime has no nested types.

- Members are only reported when public and declared by such a type, and only for classes and interfaces: a Windows Runtime struct is a plain field aggregate, so the generator drops every member of one other than its public instance fields.

- Accessors are skipped, as they are only exported as part of their property or event (which is reported instead). The generator moves a `[Deprecated]` from the property or event down onto the accessor row, and never reads one written on a C# accessor.

- Constructors are skipped, because `[Deprecated]` does not include `AttributeTargets.Constructor` in its usage and so cannot be applied to one at all. Reporting them would produce a warning whose only resolutions are suppressing it or dropping the `[Obsolete]`.

## Testing

**`src/Tests/SourceGenerator2Test/Test_ObsoleteWithoutDeprecatedAnalyzer.cs`** covers the cases the analyzer must stay quiet for (no attributes, only `[Deprecated]`, both attributes, non-component projects, non-public types and members, nested types, struct members, constructors, accessors) and the ones it must report (every public type kind, and public methods, properties and events of both classes and interfaces).

The accessor test deliberately puts `[Obsolete]` on the accessor rather than on the property: an attribute written on a property is never surfaced on its accessor symbols, so a test relying on that would pass with the guard removed. Verified that this one does fail without it.

Full suite: 148/148 passing.

### DIFF
--- a/src/Authoring/WinRT.SourceGenerator2/Diagnostics/Analyzers/ObsoleteWithoutDeprecatedAnalyzer.cs
+++ b/src/Authoring/WinRT.SourceGenerator2/Diagnostics/Analyzers/ObsoleteWithoutDeprecatedAnalyzer.cs
@@ -70,7 +70,7 @@ public sealed class ObsoleteWithoutDeprecatedAnalyzer : DiagnosticAnalyzer
                     DiagnosticDescriptors.ObsoleteWithoutDeprecated,
                     context.Symbol.Locations.FirstOrDefault(),
                     context.Symbol));
-            }, SymbolKind.NamedType, SymbolKind.Method, SymbolKind.Property, SymbolKind.Event);
+            }, SymbolKind.NamedType, SymbolKind.Method, SymbolKind.Property, SymbolKind.Event, SymbolKind.Field);
         });
     }
 
@@ -82,8 +82,9 @@ public sealed class ObsoleteWithoutDeprecatedAnalyzer : DiagnosticAnalyzer
     /// <returns>Whether <paramref name="symbol"/> is publicly exposed from the component.</returns>
     private static bool IsPubliclyExposedFromComponent(ISymbol symbol)
     {
-        // Skip symbols with no declaration in source (eg. the implicit parameterless constructor of a
-        // class), which cannot carry an attribute of their own and have nowhere to report a diagnostic
+        // Skip symbols with no declaration in source (eg. the implicit parameterless constructor of a class,
+        // or the 'value__' field of an enum), which cannot carry an attribute of their own and have nowhere
+        // to report a diagnostic
         if (symbol.IsImplicitlyDeclared)
         {
             return false;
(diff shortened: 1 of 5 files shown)


src/Authoring/WinRT.SourceGenerator2/Diagnostics/Analyzers/ObsoleteWithoutDeprecatedAnalyzer.cs
```diff
@@ -0,0 +1,124 @@
...
+/// <c>[Obsolete]</c> attribute applied, but no <c>[Windows.Foundation.Metadata.Deprecated]</c> attribute.
+/// </summary>
+/// <remarks>
+/// <c>[Obsolete]</c> has no Windows Runtime counterpart, so the WinMD generator copies it verbatim rather
+/// than translating it, which leaves the deprecation invisible to every consumer of the component. Having
+/// both attributes applied is the supported way to deprecate an API for .NET and Windows Runtime consumers
+/// alike, so that combination is not reported.
+/// </remarks>
+[DiagnosticAnalyzer(LanguageNames.CSharp)]
+public sealed class ObsoleteWithoutDeprecatedAnalyzer : DiagnosticAnalyzer
```
manodasanW, line 22: Should we just by default map Obsolete to Deprecate and if someone explicitly specifies a Deprecate such as with Remove, that takes precedence rather than our auto generated one.  The Remove one is probably rare compared to Deprecate so the automatic mapping will handle most scenarios.

Sergio0694: I did think about this, but after checking things I'm not convinced we can/should do this. The signature for `[Deprecated]` is fundamentally different, it also takes a version and also possibly a contract name and whatnot. That is all info that I'm not convinced we could/should be somehow generating ourselves. What do you think?

manodasanW: You are right, we don't know the version it would apply to.

manodasanW (approved):

PR #2520, opened 2026-08-13
Reduce exception overhead

manodasanW:
Based on the benchmarks, the native exception saves 24 bytes from cached Boolean metadata and approximately 88 bytes by avoiding intermediate message strings. The restored managed exception saves the 24-byte Boolean box.

### DIFF
--- a/src/WinRT.Runtime2/InteropServices/Exceptions/RestrictedErrorInfo.cs
+++ b/src/WinRT.Runtime2/InteropServices/Exceptions/RestrictedErrorInfo.cs
@@ -124,23 +124,11 @@ public static unsafe class RestrictedErrorInfo
                 // available, we need to concatinate them to produce the final exception message.
                 if (errorCode == restrictedError)
                 {
-                    // Append the description first
-                    if (!string.IsNullOrEmpty(description))
-                    {
-                        errorMessage += description;
-
-                        if (!string.IsNullOrEmpty(restrictedDescription))
-                        {
-                            errorMessage += " ";
-                        }
-                    }
-
-                    // Next append the restricted description. If we have both, we insert a space to separate
-                    // the two. We're not using a newline, as exception messages should always be a single line.
-                    if (!string.IsNullOrEmpty(restrictedDescription))
-                    {
-                        errorMessage += restrictedDescription;
-                    }
+                    errorMessage = string.IsNullOrEmpty(description)
+                        ? restrictedDescription
+                        : string.IsNullOrEmpty(restrictedDescription)
+                            ? description
+                            : string.Concat(description, " ", restrictedDescription);
                 }
             }
             catch (Exception e)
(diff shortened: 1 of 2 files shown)


src/WinRT.Runtime2/InteropServices/Exceptions/RestrictedErrorInfo.cs
```diff
@@ -124,23 +124,11 @@ public static unsafe class RestrictedErrorInfo
...
-                    // the two. We're not using a newline, as exception messages should always be a single line.
-                    if (!string.IsNullOrEmpty(restrictedDescription))
-                    {
-                        errorMessage += restrictedDescription;
-                    }
+                    errorMessage = string.IsNullOrEmpty(description)
+                        ? restrictedDescription
+                        : string.IsNullOrEmpty(restrictedDescription)
+                            ? description
+                            : string.Concat(description, " ", restrictedDescription);
```
Sergio0694, line 131: Nit: keep the comment and some info
```suggestion
                    // Format the exception message. If we have both a description and a restricted description,
                    // we insert a space to separate the two. We're not using a newline, as exception messages
                    // should always be a single line. We're intentionally using this ternary expression with no
                    // interpolation to avoid allocating intermediate buffers/strings, which would add overhead.
                    errorMessage = string.IsNullOrEmpty(description)
                        ? restrictedDescription
                        : string.IsNullOrEmpty(restrictedDescription)
                            ? description
                            : string.Concat(description, " ", restrictedDescription);
```

src/WinRT.Runtime2/InteropServices/Exceptions/RestrictedErrorInfoHelpers.cs
```diff
@@ -16,6 +16,9 @@ namespace WindowsRuntime.InteropServices;
 /// </remarks>
 internal static unsafe class RestrictedErrorInfoHelpers
 {
+    private static readonly object BoxedTrue = true;
+    private static readonly object BoxedFalse = false;
```
Sergio0694, line 23: Nit: we always use XML docs on everything in the runtime

```suggestion
    /// <summary>A cached boxed <see langword="true"/> value, to avoid allocations.</summary>
    private static readonly object BoxedTrue = true;

    /// <summary>A cached boxed <see langword="false"/> value, to avoid allocations.</summary>
    private static readonly object BoxedFalse = false;
```

src/WinRT.Runtime2/InteropServices/Exceptions/RestrictedErrorInfoHelpers.cs
```diff
@@ -127,17 +130,20 @@ public static bool TryGetErrorInfo(
     {
         IDictionary? exceptionData = exception.Data;
 
-        if (exceptionData is not null)
+        if (exceptionData is not null && exceptionData.Count > 0)
```
Sergio0694: Nit: can simplify:
```csharp
if (exception.Data is IDictionary { Count: > 0 } exceptionData)
```

Sergio0694 (approved):

src/WinRT.Runtime2/InteropServices/Exceptions/RestrictedErrorInfo.cs
manodasanW: Restored comment.

PR #2525, opened 2026-08-16
Fix marshaling list collection types used for NotifyCollectionChangedEventArgs

manodasanW:
NotifyCollectionChangedEventArgs exposes collection types which are implemented as .NET internal types.  Even though cswinrtinteropgen does an entire app analysis, it doesn't see these BCL types.  But this is a common scenario someone may pass across the ABI, so we add special handling to add the two types to the proxy map table explicitly.  If they aren't used, they should still get trimmed.


src/WinRT.Interop.Generator/Builders/InteropTypeDefinitionBuilder.UserDefinedType.cs
```diff
@@ -189,19 +190,24 @@ public static void ComWrappersMarshallerAttribute(
...
             ModuleDefinition module,
             bool useWindowsUIXamlProjections,
             out TypeDefinition proxyType)
         {
-            TypeDefinition userDefinedTypeDefinition = userDefinedType.Resolve(module.RuntimeContext);
+            // Almost all user-defined types can be resolved. The exception is the list types used by
+            // 'NotifyCollectionChangedEventArgs', which are registered by name, as they are absent from the reference
+            // assemblies the generator sees (see the discovery of those types). Those carry none of the attributes
+            // checked below, so they just take the same path as any other non-authored type.
+            _ = userDefinedType.TryResolve(module.RuntimeContext, out TypeDefinition? userDefinedTypeDefinition);
```
Sergio0694: I don't like that this will potentially hide legitimate type resolution errors for other types, and also regardless of whether that flag is even set. Can we do this:
- Let's define a centralized place that exposes type signatures for all our special-cased internal types (right now it would just be these two)
- Here call `TryResolve` first. If it fails, check if the type is in that set of special-cased internal types (that class could expose a helper or a set of all of them, so it can also automatically use the right comparer, and here we can just do `Contains`)
- If it matches, continue
- Otherwise, just call `Resolve` to let it throw the original exception as it would've done before

Then we can also reuse that helper (e.g. `InternalInteropReferences` or something) in `DiscoverCollectionChangedListTypes`, by just looping over these two like:

```csharp
foreach (TypeSignature typeSignature in (ReadOnlySpan<TypeSignature>)[
    InternalInteropReferences.SingleItemReadOnlyList,
    InternalInteropReferences.ReadOnlyList])
{
    // Code here
}
```

Something like this, so we have fewer special-cases hardcoded around. What do you think?

Sergio0694 (changes requested):

Sergio0694 (approved):

PR #2529, opened 2026-08-20
Fix CCW discovery for generated bindable property values

### LINKED ISSUE #2458: Bug: Binding issue
### Description

I have try to convert my UWP App to net 10, but i have issue with binding
in the sample that i have attached i use Prism MVVM, this sample have issue for binding Observable collection to ListView. I have try to set also <AllowUnsafeBlocks>true</AllowUnsafeBlocks> or Including CsWinRT 2.2.0.  but no work.

### Steps To Reproduce

[Prism.zip](https://github.com/user-attachments/files/29138976/Prism.zip)

the sample solution is in the folder 

Prism\Sandbox\Windows10\HelloWorld\HelloWorld.sln

Build the solution and start it.

the listview in Xaml  MainPage is in binding with viewModel

```
<ListView ItemsSource="{Binding DisplayItems, Mode=OneWay}" Header="Test list" >
     <ListView.ItemTemplate>
         <DataTemplate>
             <Grid>
                 <TextBlock Text="{Binding}" />
             </Grid>
         </DataTemplate>
     </ListView.ItemTemplate>
</ListView>

```
in the Code behind page:

```
        // uncomment this to bind the list to the UI
        // the App take correct data but from MainPageViewModel 
        // if you comment the App not work, not bind 

        //public ObservableCollection<string> DisplayItems
        //{
        //    get { return _DisplayItems; }
        //}

```
the ListView only works if the property "DisplayItems" is present in the “behind code” of the XAML page; 
if it doesn't find it, the “Binding” doesn't work even if the context is set in the view model. 
The strangest thing is that if it finds the “DisplayItems” property in the behind code, it uses that view model.

### Expected Behavior

Without property in the code Behind the listview not work, even if the “viewModel” is set up correctly

### Version Info

Last version.

### Additional Context

_No response_

tommcdon:
## Description

Fixes #2458.

`GeneratedBindableCustomProperty` generates accessors that return bindable
property values through `object`, but those generated accessors cannot be
analyzed by the same source-generator pass. For getter-only collection
properties, the concrete return type was therefore omitted from
`WinRTGlobalVtableLookup.g.cs` unless another authored or XAML-generated
expression happened to expose it across the WinRT ABI.

This caused UWP XAML bindings such as
`ItemsSource="{Binding DisplayItems}"` to return an
`ObservableCollection<string>` without registering its `IVector<string>`
CCW projection. Applications had to add an unrelated property to the page
code-behind or explicitly use `GeneratedWinRTExposedExternalType` to root
the collection.

## Changes

- Propagate the existing generated-bindable-property discovery flag through
  property initializers, expression-bodied properties, and block-bodied
  getter returns.
- Add source-generator regressions for all three property forms.
- Verify non-bindable properties do not trigger additional lookup generation.
- Verify the generated lookup includes the correct WinRT projections.

## Validation

- SourceGenerator regression tests.
- Validated the repro case from #2458 with the fix.

### DIFF
--- a/src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
+++ b/src/Authoring/WinRT.SourceGenerator/AotOptimizer.cs
@@ -1319,13 +1319,17 @@ namespace Generator
             }
             else if (context.Node is PropertyDeclarationSyntax propertyDeclaration)
             {
+                var isGeneratedBindableCustomPropertyClass =
+                    context.SemanticModel.GetDeclaredSymbol(propertyDeclaration) is IPropertySymbol propertySymbol &&
+                    GeneratorHelper.IsGeneratedBindableCustomPropertyClass(context.SemanticModel.Compilation, propertySymbol.ContainingSymbol);
+
                 // Detect scenarios where the property declaration has an initializer and is to a boxed or cast type during initialization.
                 if (propertyDeclaration.Initializer != null)
                 {
                     var leftSymbol = context.SemanticModel.GetSymbolInfo(propertyDeclaration.Type).Symbol;
                     if (leftSymbol is INamedTypeSymbol namedType)
                     {
-                        AddVtableAttributesForExpression(propertyDeclaration.Initializer.Value, namedType);
+                        AddVtableAttributesForExpression(propertyDeclaration.Initializer.Value, namedType, isGeneratedBindableCustomPropertyClass);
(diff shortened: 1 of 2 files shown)


tommcdon:
@manodasanW @Sergio0694 @jkoritzinsky

manodasanW (approved):

PR #2533, opened 2026-08-24
Pack reference projections automatically

Sergio0694:
## Summary

Reference projections can now be packed without any custom MSBuild in the project: the reference assembly is contributed as `ref/<tfm>`, and the input `.winmd` files as `metadata/`.

## Motivation

Only the forwarder is a build output of a reference projection, so `dotnet pack` produced a package with nothing to compile against in `lib/<tfm>`, and no metadata for the consuming app to generate the merged projection from. The only workaround was to hand-roll a `TargetsForTfmSpecificContentInPackage` target reaching into `$(_CsWinRTRefAssemblyPath)`, an internal property. Since shipping reference projections through NuGet is the point of the CsWinRT 3.0 model, this should just work.

## Changes

- **`nuget/Microsoft.Windows.CsWinRT.CsWinRTGen.targets`**: added `CsWinRTPackReferenceProjection` and `CsWinRTPackWinMDInputs` (both default to `true` for reference projections) and the target that contributes the assets. The `ref/` folder name comes from the SDK's `GetNuGetShortFolderName` task, since NuGet drops a trailing `.0` revision when laying out `lib/` and the two have to line up.

- **`nuget/Microsoft.Windows.CsWinRT.Pack.targets`** (new): suppresses `NU5131` and `NU5100` when those assets are present. Both are false positives: `NU5131` wants a legacy `<references>` node that modern `ref/` + `lib/` packages don't use (this package included), and `NU5100` flags assemblies outside `lib/`, which is what `metadata/` is for.

- **`nuget/Microsoft.Windows.CsWinRT.MultiTargeting.targets`** (new): imports the above into the outer build. NuGet conditions `build/` imports on `$(TargetFramework)`, so they never apply to the outer build of a multi-targeted project, which is where `pack` runs.

- **`nuget/Microsoft.Windows.CsWinRT.nuspec`**, **`nuget/readme.md`**, **`.github/copilot-instructions.md`**: ship the new files and document the new properties.

## Validation

Verified on a project multi-targeting a CsWinRT 2.x leg and a 3.0 reference projection leg: the default packs `ref/` and `metadata/` alongside both `lib/` folders with zero warnings, each opt-out drops just its own assets (still zero warnings), and restore resolves `ref/` for compile and `lib/` for runtime on the 3.0 leg.


src/WinRT.Projection.Writer/Resources/Base/AssemblyAttributes.cs
```diff
@@ -4,11 +4,4 @@
 #if CSWINRT_REFERENCE_PROJECTION
 #pragma warning disable CSWINRT3004 // "Type or member '...' is a private implementation detail"
 [assembly: WindowsRuntime.InteropServices.WindowsRuntimeReferenceAssembly]
```
manodasanW, line 6: Moving this from here to the other isn't going to apply for WinRT.Projection, is it?  Thats generated using rosyln within the generator.

Sergio0694: Reverted!

nuget/Microsoft.Windows.CsWinRT.CsWinRTGen.targets
```diff
@@ -127,6 +127,74 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+
+    <ItemGroup>
+      <TfmSpecificPackageFile Condition="'$(CsWinRTPackReferenceProjection)' == 'true'"
+                              Include="$(_CsWinRTRefAssemblyPath)"
+                              PackagePath="ref\$(_CsWinRTNuGetShortFolderName)\"
+                              CsWinRTPackagedAsset="ref" />
+
+      <!-- The metadata is the same for every target framework, so it is packed once, not per target framework. -->
+      <TfmSpecificPackageFile Condition="'$(CsWinRTPackWinMDInputs)' == 'true'"
+                              Include="@(CsWinRTInputs)"
```
manodasanW: Is CsWinRTInputs here going to end up with dependencies too or no?

Sergio0694: Fixed!

nuget/Microsoft.Windows.CsWinRT.CsWinRTGen.targets
```diff
@@ -127,6 +127,77 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+    That list is explicit rather than derived from '@(CsWinRTInputs)', because the two are not the same set. Inputs are
+    everything the generator reads, which includes dependency metadata the project does not own: 'CsWinRTRemoveWinMDReferences'
+    adds every referenced .winmd to it, and a project may also declare one directly just to resolve types referenced by the
+    .winmd it actually projects. Packing those would redistribute metadata belonging to other packages.
+  -->
+  <PropertyGroup Condition="'$(CsWinRTGenerateReferenceProjection)' == 'true'">
+    <CsWinRTPackReferenceProjection Condition="'$(CsWinRTPackReferenceProjection)' == ''">true</CsWinRTPackReferenceProjection>
+  </PropertyGroup>
+
+  <!-- '@(CsWinRTPackedWinMD)' cannot be tested from an evaluation-time condition (MSB4099), so the target is hooked for
```
manodasanW: not sure how relevant this comment is given we also pack the ref dll and we already gate behind CsWinRTGenerateReferenceProjection.

manodasanW (approved):

PR #2534, opened 2026-08-24
Fix HResult and mapped value type parameter marshalling in generated projections

Sergio0694:
## Summary

The projection writer skipped the marshalling step for `Windows.Foundation.HResult` (and, in a few positions, for `System.Type` and the mapped value types `DateTime` / `TimeSpan`) in several parameter positions, so the generated projection assigned an ABI value where the projected type was expected and failed to compile. This fixes every affected position and adds test coverage for each of them.

## Motivation

Any Windows Runtime interface with an `HRESULT` input parameter produced a projection that fails to build with `CS0029`. The failure happens in the *consuming app's* build, when `cswinrtprojectiongen` generates the merged projection, so it blocks consumption of the component entirely. Found with some Microsoft Store projection .dll-s.

The root cause is that each emission site had its own hand written ladder mapping a type to its ABI form, and those ladders had drifted: `IsMappedAbiValueType` deliberately excludes `HResult` (it is handled explicitly everywhere else), so `HResult` silently fell through to the "already in ABI form" fallback. Investigating that turned up four more positions with the same class of gap, all of which are fixed here.

The bug went unnoticed because CCW stubs are only emitted for standalone interfaces, and every existing `HResult` member in `TestComponentCSharp.idl` lives on `runtimeclass Class`, whose members sit on an `[exclusiveto]` interface that never gets CCW stubs.

## What was broken

| Direction | Position | Affected types |
|---|---|---|
| CCW (`Do_Abi`) | input parameter, property setter, delegate `Invoke` | `HResult` |
| CCW (`Do_Abi`) | non-retval `out` parameter write-back | `HResult`, `System.Type`, `DateTime`, `TimeSpan` |
| RCW | function pointer signature for `out T` | `HResult`, `DateTime`, `TimeSpan` |
| RCW | `out` parameter read-back | `HResult`, `DateTime`, `TimeSpan` |
| RCW | function pointer signature and pinning for `in T` (`ref const`) | `DateTime`, `TimeSpan` |

## Changes

- **`src/WinRT.Projection.Writer/Factories/AbiMethodBodyFactory.DoAbi.cs`**: added the missing `HResult` branch to `EmitDoAbiParamArgConversion` (which covers method parameters, property setters and delegate `Invoke` stubs, since they share the emitter), and the missing `HResult` / `System.Type` / mapped value type branches to the `out` parameter write-back.

- **`src/WinRT.Projection.Writer/Factories/AbiMethodBodyFactory.RcwCaller.cs`**: the function pointer signature builder had its own duplicate ABI type ladder for `out T` and `in T`, and neither copy handled `HResult` or the mapped value types, so both slots were typed `int*` while the matching local was the real ABI type. Both branches now share `GetAbiLocalTypeName`, which is the same helper that types the local, so the two cannot drift apart again. Also added the matching `ConvertToManaged` to the `out` read-back, and routed `in T` parameters whose ABI form differs from the projected one through a marshalled ABI local instead of pinning the projected value.

- **`src/WinRT.Projection.Writer/Resolvers/AbiTypeKindResolver.cs`**: added `RequiresMarshalledAbiInputLocal`, so "this input needs a separate ABI local" is named once instead of being re-derived as "is it a complex struct?" at each of the sites that have to agree on it.

- **`src/Tests/TestComponentCSharp/`**: added `IMarshalledValues`, a standalone interface covering every parameter position where the ABI and projected representations differ, the `HandleResult` delegate for the delegate stub path, and `MarshalledValuesTest`, which implements the interface natively (the RCW direction) and exposes one static per member that invokes it from native code (the CCW direction).

- **`src/Tests/UnitTest/MarshalledValuesTests.cs`**: 16 tests covering each shape in both directions.

## Validation

- Full `UnitTest` suite passes (549 / 550; the one failure is `TestDataTransferManager`, which asserts that `DataTransferManager.GetForWindow(0)` throws and is environment dependent, on a code path this PR does not touch).
- `ProjectionWriterTest` passes (21 / 21).
- The whole Windows SDK projection (`WinRT.Sdk.Projection.dll`) plus the merged projection regenerate and compile cleanly, which exercises the shared ladder refactor across the full SDK surface.
- Reverting the writer changes and rebuilding makes the merged projection fail with 15 `CS0029` / `CS1503` / `CS0266` errors, one for each broken position, so the new coverage is verified to catch the regression rather than merely to pass.

### DIFF
--- a/src/Tests/UnitTest/UnitTestHelper.cs
+++ b/src/Tests/UnitTest/UnitTestHelper.cs
@@ -38,13 +38,13 @@ public static partial class UnitTestHelper
     /// </summary>
     /// <param name="error">The <c>HRESULT</c> to originate the error info for.</param>
     /// <param name="message">The error message to associate with the error info.</param>
-    /// <returns>Whether the error info was originated.</returns>
+    /// <returns>
+    /// Whether the error info was originated. This is <see langword="false"/> only if <paramref name="error"/>
+    /// is a success code. Originating replaces whatever error info the thread already had, so callers don't
+    /// need to clear it first.
+    /// </returns>
     internal static unsafe bool OriginateError(int error, string message)
     {
-        // 'RoOriginateError' is a no-op if the current thread already has originated error
-        // info for the same 'HRESULT', so always start from a clean slate to be deterministic
-        RoClearError();
-
         fixed (char* pMessage = message)
         {
             return RoOriginateErrorW(error, (uint)message.Length, pMessage) != 0;
--- a/src/Tests/UnitTest/UnitTestHelper.cs
+++ b/src/Tests/UnitTest/UnitTestHelper.cs
@@ -20,6 +20,37 @@ public static partial class UnitTestHelper
     [UnmanagedCallConv(CallConvs = [typeof(CallConvStdcall)])]
     internal static unsafe partial int WindowsDeleteString(void* hstring);
 
+    [LibraryImport("api-ms-win-core-winrt-error-l1-1-1.dll")]
+    [UnmanagedCallConv(CallConvs = [typeof(CallConvStdcall)])]
+    private static unsafe partial int RoOriginateErrorW(int error, uint cchMax, char* message);
+
+    /// <summary>
+    /// Clears the <c>IRestrictedErrorInfo</c> object associated with the current thread.
+    /// </summary>
+    [LibraryImport("api-ms-win-core-winrt-error-l1-1-1.dll")]
+    [UnmanagedCallConv(CallConvs = [typeof(CallConvStdcall)])]
+    internal static partial void RoClearError();
+
+    /// <summary>
+    /// Associates an <c>IRestrictedErrorInfo</c> object with the current thread, exactly like a failing
+    /// native call would. That state is ambient and sticky: it outlives the call that produced it, until
+    /// something else originates a new error or clears it. Tests use this to reproduce a "dirty" thread.
+    /// </summary>
+    /// <param name="error">The <c>HRESULT</c> to originate the error info for.</param>
+    /// <param name="message">The error message to associate with the error info.</param>
+    /// <returns>Whether the error info was originated.</returns>
+    internal static unsafe bool OriginateError(int error, string message)
+    {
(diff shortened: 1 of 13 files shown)


manodasanW (approved):

src/Tests/UnitTest/UnitTestHelper.cs
```diff
@@ -20,6 +20,37 @@ public static partial class UnitTestHelper
...
+    /// Associates an <c>IRestrictedErrorInfo</c> object with the current thread, exactly like a failing
+    /// native call would. That state is ambient and sticky: it outlives the call that produced it, until
+    /// something else originates a new error or clears it. Tests use this to reproduce a "dirty" thread.
+    /// </summary>
+    /// <param name="error">The <c>HRESULT</c> to originate the error info for.</param>
+    /// <param name="message">The error message to associate with the error info.</param>
+    /// <returns>Whether the error info was originated.</returns>
+    internal static unsafe bool OriginateError(int error, string message)
+    {
+        // 'RoOriginateError' is a no-op if the current thread already has originated error
```
manodasanW: I don't think that is right?

manodasanW (approved):

src/Tests/UnitTest/UnitTestHelper.cs
Sergio0694: You're right, that comment was wrong. I checked the actual behavior and `RoOriginateError` never no-ops: calling it repeatedly with the same `HRESULT` always succeeds and just replaces the stored error info each time (I verified back-to-back calls and a loop — the last message always wins). The only case it returns `FALSE` is when the `HRESULT` is a success code, and replacement is wholesale, so the previously stored info is gone entirely.

Removed the bogus comment along with the `RoClearError()` call it was justifying, and documented the real semantics on the return value instead. The tests are unaffected: since originating already replaces whatever was there, they're just as deterministic, and the explicit `RoClearError()` calls they make to start from (or return to) a clean thread still do their job. Verified all 8 `ExceptionTests` still pass with the fix, and that the same 2 still fail without it.

PR #2552, opened 2026-09-08
Add throwing buffer data accessors

Sergio0694:
## Summary

Add throwing `GetDataUnsafe` overloads for `IBuffer` and `IMemoryBufferReference`. Move native memory-buffer byte access into the shared internal buffer helpers.

## Motivation

Callers that require a pointer to a buffer's data should not have to repeat `TryGetDataUnsafe` checks and exception handling. Thin throwing wrappers provide that convenience while keeping exception construction in local helpers, and centralizing native access makes the handling of both buffer interfaces consistent.

## Changes

- **`src\WinRT.Runtime2\InteropServices\WindowsRuntimeBufferMarshal.cs`**: Add the two pointer-returning getters as wrappers over the corresponding `TryGetDataUnsafe` methods. Check for null inputs directly in each getter, use parameterless local helpers that only throw `ArgumentException` when retrieval fails, preserve native interop exceptions, and document the new APIs and pointer lifetime requirements.
- **`src\WinRT.Runtime2\InteropServices\Buffers\WindowsRuntimeBufferHelpers.cs`**: Move `IMemoryBufferReference` native interop into a `TryGetNativeData` overload without changing the existing public `TryGetDataUnsafe` behavior.
- **`src\WinRT.Runtime2\Properties\WindowsRuntimeExceptionMessages.cs`**: Add an error message for unsupported memory-buffer references.
- **`src\Tests\UnitTest\BufferMarshalTests.cs`**: Cover native, managed, pinned, and empty buffers; pointer identity, writable data, and capacity results; and null and unsupported input exceptions.

### DIFF
--- a/src/WinRT.Runtime2/InteropServices/Buffers/WindowsRuntimeBufferHelpers.cs
+++ b/src/WinRT.Runtime2/InteropServices/Buffers/WindowsRuntimeBufferHelpers.cs
@@ -3,12 +3,13 @@
 
 using System;
 using System.Diagnostics.CodeAnalysis;
+using Windows.Foundation;
 using Windows.Storage.Streams;
 
 namespace WindowsRuntime.InteropServices;
 
 /// <summary>
-/// Provides internal helpers for working with <see cref="IBuffer"/> objects.
+/// Provides internal helpers for working with Windows Runtime buffer types.
 /// </summary>
 internal static class WindowsRuntimeBufferHelpers
 {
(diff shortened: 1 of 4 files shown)


src/WinRT.Runtime2/InteropServices/Buffers/WindowsRuntimeBufferHelpers.cs
```diff
@@ -174,4 +175,34 @@ public static unsafe bool TryGetNativeData(IBuffer buffer, out byte* data)
...
+    /// <summary>
+    /// Tries to get the underlying data for the specified buffer, only if backed by native memory.
+    /// </summary>
+    /// <param name="buffer">The input <see cref="IMemoryBufferReference"/> instance.</param>
+    /// <param name="data">The underlying data, if retrieved.</param>
+    /// <param name="capacity">The capacity of the buffer, if retrieved.</param>
+    /// <returns>Whether <paramref name="data"/> could be retrieved.</returns>
+    public static unsafe bool TryGetNativeData(IMemoryBufferReference buffer, out byte* data, out uint capacity)
+    {
+        if (buffer is WindowsRuntimeObject { HasUnwrappableNativeObjectReference: true } bufferObject)
```
manodasanW, line 188: so is the only `try` part to this function that we can't unwrap it as an rcw?

Sergio0694: Yeah I just moved this code here for consistency and to keep the public file simpler.

manodasanW: It was more a general question rather than the move, but it is a move, so I will sign off on this specific change.

manodasanW (approved):

src/WinRT.Runtime2/InteropServices/Buffers/WindowsRuntimeBufferHelpers.cs
Sergio0694: Oh, sorry. Yeah the cast is intentionally always throwing because the WinRT spec itself (to my understanding) says that objects implementing `IBuffer` or `IMemoryBufferReference` must also implement the corresponding `xxByteAccess` interfface. So an object failing that QI would be just invalid (as in, not respecting the spec).

PR #2553, opened 2026-09-08
Preserve exact type selection in merged projections

Sergio0694:
## Summary

Preserve the exact type surface of reference projections when generating merged implementation projections. Keep existing namespace-prefix filtering for the Windows SDK, WinUI additions, and reference-projection settings.

## Motivation

The merged projection generator currently passes fully qualified type names from reference assemblies to a prefix filter. An include for `Contoso.User` can therefore match both `Contoso.User2` and the namespace `Contoso.UserProfile`, regenerating types deliberately omitted from the reference projection and introducing unsupported API signatures.

The reference assembly already describes the intended type set. Separating exact identities from intentional prefix rules preserves that surface without requiring new consumer-facing MSBuild properties or adding support for previously excluded APIs.

## Changes

- **`src\WinRT.Projection.Generator\Generation\ProjectionGenerator.Generate.cs`**: populate exact type includes from projection reference assemblies and component metadata, while retaining intentional SDK and WinUI namespace rules.
- **`src\WinRT.Projection.Writer\`**: add the `IncludeTypes` option, preserve exclude precedence and existing prefix semantics, and prevent exact type includes from enabling namespace additions.
- **`src\Tests\ProjectionWriterTest\`**: add synthetic metadata and regression coverage for type-name and namespace collisions, mixed filters, exclude precedence, namespace additions, and required exclusive-to factory infrastructure.

### DIFF
--- a/src/Tests/ProjectionWriterTest/Test_TypeFilter.cs
+++ b/src/Tests/ProjectionWriterTest/Test_TypeFilter.cs
@@ -1,0 +1,113 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using WindowsRuntime.ProjectionWriter.Helpers;
+
+namespace ProjectionWriterTest;
+
+[TestClass]
+public class Test_TypeFilter
+{
+    [TestMethod]
+    [DataRow("Contoso.User", true)]
+    [DataRow("Contoso.User2", false)]
+    [DataRow("Contoso.UserProfile.UserSetupManager", false)]
+    [DataRow("Contoso.User.Profile", false)]
+    [DataRow("Contoso.Users.User", false)]
+    [DataRow("contoso.User", false)]
+    [DataRow("Unrelated.User", false)]
+    public void ExactTypeIncludes_DoNotMatchPrefixes(string name, bool expected)
+    {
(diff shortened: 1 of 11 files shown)


src/Tests/ProjectionWriterTest/Test_TypeFilter.cs
```diff
@@ -0,0 +1,113 @@
...
+[TestClass]
+public class Test_TypeFilter
+{
+    [TestMethod]
+    [DataRow("Contoso.User", true)]
+    [DataRow("Contoso.User2", false)]
+    [DataRow("Contoso.UserProfile.UserSetupManager", false)]
+    [DataRow("Contoso.User.Profile", false)]
+    [DataRow("Contoso.Users.User", false)]
+    [DataRow("contoso.User", false)]
```
manodasanW, line 17: so we are saying here this one is case sensitive right?

manodasanW (approved):

src/Tests/ProjectionWriterTest/Test_TypeFilter.cs
Sergio0694: Yeah it should be, namespaces are case sensitive in .NET

PR #2560, opened 2026-09-11
Fix CCWs for IStringable-only types

Sergio0694:
## Summary

Allow an `IStringable`-only managed type to generate a CCW while preserving validation of zero-additional-interface requests.

## Motivation

An explicit `Windows.Foundation.IStringable` implementation occupies a reserved built-in slot, so it contributes no additional interface entry. Requiring an additional entry caused executable interop generation to fail with `CSWINRTINTEROPGEN0039` for this valid type shape.

## Changes

- **`src\WinRT.Interop.Generator\`**: pass the implemented-interface set to layout lookup, reuse the existing resolver to recognize explicit `IStringable` implementations, and validate every request before accessing the count-keyed cache. Keep structural count validation in the layout factory and leave discovery and other reserved-interface restrictions unchanged.
- **`src\Tests\UnitTest\ComInteropTests.cs`**: add one `IStringable`-only type and one regression that creates its CCW, calls `ToString` through the native vtable, checks the returned string, and releases the string and COM reference. No new test project or CI changes.
- **`.github\skills\interop-generator\SKILL.md`**: document the narrow allowance and cache-validation boundary.

## Validation

- Built the interop generator in Release. The unchanged standalone `.0` control still runs, and the unchanged `.1` repro fails with the original package but builds and runs with the source-built generator.
- Ran the regression's native CCW assertions in an isolated `.1` app on both CoreCLR and Native AOT.
- Full `UnitTest` project execution is blocked locally by a pre-existing `C1083` failure for its long-named C++ fixture, followed by secondary XAML error-handler failures.

### DIFF
--- a/src/Tests/UnitTest/ComInteropTests.cs
+++ b/src/Tests/UnitTest/ComInteropTests.cs
@@ -4,6 +4,7 @@ using System.Runtime.InteropServices.Marshalling;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Windows.ApplicationModel.DataTransfer;
 using Windows.ApplicationModel.DataTransfer.DragDrop.Core;
+using Windows.Foundation;
 using Windows.Graphics.Display;
 using Windows.Graphics.Printing;
 using Windows.Media;
@@ -17,6 +18,8 @@ using Windows.UI.Input.Core;
 using Windows.UI.Input.Spatial;
 using Windows.UI.ViewManagement;
 using TestComponentCSharp;
+using WindowsRuntime.InteropServices;
+using WindowsRuntime.InteropServices.Marshalling;
 
 namespace UnitTest
 {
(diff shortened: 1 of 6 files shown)


src/Tests/UnitTest/ComInteropTests.cs
```diff
@@ -33,6 +36,34 @@ internal partial interface IComInterop
...
+
+                Assert.AreEqual("server test", HStringMarshaller.ConvertToManaged(result));
+            }
+            finally
+            {
+                HStringMarshaller.Free(result);
+            }
+        }
+
+        private sealed class StringableOnly : IStringable
```
manodasanW, line 66: Might make sense to have one more test case with other interfaces and IStringable

Sergio0694: 🤖 Fixed in 1628b58b. The existing CCW test now has two data rows: one for `IStringable` alone and one for `IStringable` plus `IDisposable`. Both create a CCW and invoke `IStringable.ToString` through the native vtable, sharing the same assertions and cleanup.

manodasanW (approved):

PR #2561, opened 2026-09-11
Add ability to register fallback activation handlers

ahmed605:
As the title says, adds ability to register fallback activation handlers.

This can be used as an extension point for activating types that the default activation handlers don't cover, such as [Static WinRT Components](https://github.com/microsoft/cppwinrt/tree/master/vsix/ProjectTemplates/VC/Windows%20Universal/StaticLibrary) for example.

`SetWindowsRuntimeActivationHandler` cannot be used for that since it allows the handler to be set only once, and even if this restriction got removed it still doesn't offer a getter so chaining multiple handlers still won't be possible, it also cannot be used from libraries because that would block the consumer app from changing it.

### DIFF
--- a/src/WinRT.Runtime2/InteropServices/Activation/WindowsRuntimeActivationFactory.cs
+++ b/src/WinRT.Runtime2/InteropServices/Activation/WindowsRuntimeActivationFactory.cs
@@ -68,8 +68,8 @@ public static unsafe class WindowsRuntimeActivationFactory
     {
         ArgumentNullException.ThrowIfNull(fallbackHandler);
 
-        _ = ImmutableInterlocked.Update(ref fallbackHandlers, (arr, it) =>
-            arr.AsSpan().Contains(fallbackHandler) ? arr : arr.Add(it),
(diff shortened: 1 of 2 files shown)


src/WinRT.Runtime2/InteropServices/Activation/WindowsRuntimeActivationFactory.cs
```diff
@@ -140,14 +179,22 @@ private static HRESULT GetActivationFactoryFromAnySourceUnsafe(string runtimeCla
...
+        {
+            return hresult;
+        }
+
         // If manifest free activation is enabled, we must stop here.
         // At this point the 'HRESULT' should be 'REGDB_E_CLASSNOTREG'.
         if (!WindowsRuntimeFeatureSwitches.EnableManifestFreeActivation)
         {
             return hresult;
         }
```
riverar: Is this right? Seems backwards.

manodasanW:
kicked off CI

PR #2568, opened 2026-09-14
Use SDK forwarders for component projection references

Sergio0694:
## Summary

Fix component projection generation for authored components that use Windows SDK types. The production change is confined to reference selection in the component projection target; no generator or runtime code changes are needed.

## Motivation

`ReferencePath` supplies SDK reference projections alongside the generated SDK implementation assemblies, so component compilation sees duplicate SDK definitions (`CS0433`) and incompatible factory parameter identities (`CS1503`). Simply dropping the SDK references is insufficient: authored signatures still reference the original SDK assembly identities, causing `CS0012`. The existing SDK forwarder assemblies preserve those identities while forwarding the types to the generated SDK projections.

## Changes

- **`nuget\Microsoft.Windows.CsWinRT.CsWinRTGen.targets`**: prefer an existing sibling `lib` assembly for packaged `ref` references, following @manodasanW's review suggestion. This selects SDK and other packaged projection forwarders without depending on `RuntimePackAsset`. References without a sibling remain unchanged; `WinRT.Runtime` retains its existing explicit swap flag and implementation-path handling. Leave other generation stages unchanged.
- **`src\Tests\SmokeTests\Authoring\BackgroundTask.cs`**: add an authored SDK-interface implementation with an SDK-typed constructor, guarding against both duplicate definitions and missing original assembly identities.
- **`src\Tests\SmokeTests\MixedConsumption\`**: add a separate mixed SDK/component consumption smoke test. Keep the referenced component managed when publishing the consuming app with Native AOT. The existing `Consumption` project and program are unchanged.
- **Smoke runner and CI**: add the `MixedConsumption` selector, include it in `All`, and run it separately on CoreCLR and Native AOT.

## Verification

Reproduced the reported failure with `3.0.0-preview.260912.1`, then tested a local package using the unchanged published binaries with only the targets file updated. Repeated the checks with the revised sibling-assembly lookup.

- The mixed regression fails against the original package and passes with the fix on CoreCLR and Native AOT.
- The unchanged consumption smoke passes on CoreCLR and Native AOT; standalone authoring Native AOT and the reference-projection smoke also pass.
- A two-component reproduction passes in SDK-only and UWP/XAML modes, including `IBasicVideoEffect`, `IBackgroundTask`, `IValueConverter`, and SDK/XAML-typed constructors. PE inspection confirms canonical SDK references, forwarder destinations, and component default-interface mappings.
- A probe clears `RuntimePackAsset` immediately before component generation: the previous implementation reproduces `CS0433`, while the sibling-assembly lookup passes. The same probe verifies a custom runtime implementation path, exactly one runtime reference, and preservation of framework references without a sibling `lib` assembly.
- The equivalent stable `.0` / CsWinRT `2.3.1` control builds and runs successfully.


nuget/Microsoft.Windows.CsWinRT.CsWinRTGen.targets
```diff
@@ -567,6 +567,11 @@ Copyright (C) Microsoft Corporation. All rights reserved.
       <_ComponentProjectionReferenceAssemblyPaths Include="$(CsWinRTRuntimeImplementationAssemblyPath)" Condition="'$(CsWinRTSwapRuntimeReferenceAssembly)' == 'true'" />
       <_ComponentProjectionReferenceAssemblyPaths Include="@(IntermediateAssembly)" Condition="'$(CsWinRTComponent)' == 'true'" />
 
+      <!-- Use SDK forwarders to preserve authored assembly references without duplicating the generated SDK types. -->
+      <_ComponentProjectionSdkForwarderPaths Include="@(RuntimePackAsset)" Condition="'%(Extension)' == '.dll' and ('%(FileName)' == 'Microsoft.Windows.SDK.NET' or '%(FileName)' == 'Microsoft.Windows.UI.Xaml')" />
```
manodasanW, line 571: I am not sure whether we can rely on RuntimePackAsset being set in all scenarios.  I ran into the same issue in my branch and had this fix:  https://github.com/microsoft/CsWinRT/blob/8162b09175f8a4aa2527560a6067d14aaf1c13a1/nuget/Microsoft.Windows.CsWinRT.CsWinRTGen.targets#L541-L566

Thoughts if that addresses your issue?

Sergio0694: 🤖 Yes, your approach fixes this scenario. Applied it in 1b4203ca, keeping `WinRT.Runtime` on its existing explicit swap path so a configured implementation override is preserved without adding a second runtime reference.

Verified the SDK-only and UWP/XAML two-component repros, including SDK-typed constructors and emitted default-interface identities, plus mixed consumption on CoreCLR/Native AOT and standalone authoring Native AOT. I also cleared `RuntimePackAsset` immediately before component generation: the previous fix reproduces `CS0433`, while the sibling `ref`/`lib` lookup passes. References without a sibling implementation remain unchanged. Thanks for the suggestion!

manodasanW (approved):

PR #2572, opened 2026-09-16
Fix standalone public exclusive interface projections

Sergio0694:
## Summary

Preserve standalone public `[ExclusiveTo]` interfaces when regenerating implementation projections at app build time. Generate their dynamic-cast implementations and ABI helpers even when the owning runtime class belongs to a separate SDK projection, including supplemental projections that need only Windows SDK metadata.

## Motivation

A reference/forwarder projection can expose `Windows.UI.Xaml.IFrameworkElementProtected7` while `FrameworkElement` remains in `WinRT.Sdk.Xaml.Projection.dll`. App-time generation previously lost the public-interface requirements, omitted the standalone ABI helper, and could skip merged projection generation entirely when no package-owned WinMD inputs were present. This left the forwarder without a usable implementation and caused type-loading failures, including `IL1005` under strict Native AOT.

Infer the requirements from each reference assembly's public interface surface instead of enabling global exclusive-interface options or introducing new producer metadata. Existing reference-projection packages therefore do not need to be rebuilt.

## Changes

- **`src\WinRT.Projection.Generator`**: infer public and dynamic-cast requirements per interface and remove stale output when no projection types remain.
- **`src\WinRT.Projection.Writer`**: apply per-type options consistently to interface visibility, ABI implementations, dynamic-cast shims, and type-map associations. Emit standalone helpers and event accessors with interface-local slots while retaining ordinary Fast ABI dispatch through the merged default vtable.
- **`nuget\Microsoft.Windows.CsWinRT.CsWinRTGen.targets`**: let the generator inspect reference assemblies even when a supplemental projection has no package-owned WinMD inputs.
- **`src\Tests\ProjectionWriterTest`**: cover owner-included and owner-excluded interfaces, independent options, Fast ABI slots and events, reference-to-implementation generation, and stale-output removal.
- **`src\Tests\SmokeTests`** and **`build\AzurePipelineTemplates\CsWinRT-PublishToNuGet-Steps.yml`**: add an SDK-only reference/forwarder producer and consumer with CoreCLR and strict Native AOT coverage, assembly-ownership assertions, and an optional real XAML viewport call.
- **`docs\usage.md`** and **`nuget\readme.md`**: document standalone exclusive projections, producer options, and reuse of SDK metadata.

Commits are grouped into implementation fixes, regression coverage and CI, and documentation.


src/WinRT.Projection.Generator/Generation/ProjectionGenerator.Generate.cs
```diff
@@ -288,6 +297,8 @@ private static void BuildWriterOptions(
             OutputFolder = outputFolder,
             Include = includes,
             IncludeTypes = includeTypes,
+            PublicExclusiveToTypes = publicInterfaces,
```
manodasanW, line 300: does this also have non-exclusive ones or no?

Sergio0694: 🤖 Yes. `publicInterfaces` collects all public interfaces from the reference projections being merged (`exportedType.IsPublic && exportedType.IsInterface`), so it includes non-exclusive ones too. Both `PublicExclusiveToTypes` and `IdicExclusiveToTypes` receive that same list.

The writer gates the effects of these options on `type.IsExclusiveTo`—including interface visibility, standalone ABI helpers, and IDIC generation/registration—so the extra non-exclusive entries do not change their projection behavior.

manodasanW (approved):

